PR #370, opened 2023-06-06
✨ zb: Add a from_bytes constructor for Message 

### LINKED ISSUE #326: Add an API similar to `GDBusMessage::from_blob`
In GitLab by @SeaDve on Mar 16, 2023, 15:44

Currently, there is `zbus::Message::from_raw_parts`. However, it is private. 

For the use case, [Bustle](https://github.com/freedesktop/bustle) has DBus messages serialized in a pcap file. There is currently not a convenient way to parse the bytes from each packet from the pcap file into a `zbus::Message`.

Link
* [GDBusMessage.new_from_blob](https://docs.gtk.org/gio/ctor.DBusMessage.new_from_blob.html)

SeaDve:
Closes #326

### DIFF
--- a/zbus/src/message.rs
+++ b/zbus/src/message.rs
@@ -196,6 +196,29 @@ impl Message {
         b.build(body)
     }
 
+    /// Create a message from bytes.
+    ///
+    /// The `fds` parameter is only available on unix. It specifies the file descriptors that
+    /// accompany the message. On the wire, values of the UNIX_FD types store the index of the
+    /// corresponding file descriptor in this vector. Passing an empty vector on a message that
+    /// has UNIX_FD will result in an error.
+    ///
+    /// **Note:** Since the constructed message is not construct by zbus, the receive sequence,
+    /// which can be acquired from [`Message::recv_position`], is not applicable and hence set
+    /// to `0`.
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe as bytes may have an invalid encoding.
+    pub unsafe fn from_bytes(bytes: Vec<u8>, #[cfg(unix)] fds: Vec<OwnedFd>) -> Result<Self> {
+        Self::from_raw_parts(
+            bytes,
+            #[cfg(unix)]
+            fds,
+            0,
+        )
+    }
+
     /// Create a message from its full contents
     pub(crate) fn from_raw_parts(
         bytes: Vec<u8>,


zbus/src/message.rs
```diff
@@ -520,6 +520,19 @@ impl fmt::Display for Message {
...
 }
 
+impl TryFrom<Vec<u8>> for Message {
+    type Error = Error;
+
+    fn try_from(bytes: Vec<u8>) -> Result<Self> {
+        Self::from_raw_parts(
+            bytes,
+            #[cfg(unix)]
+            vec![],
```
zeenix, line 530: Not sure about this being **generally** correct. Not all users will be loading old messages from a file, like is the case for you. I didn't think of this before. Now I'm wondering if the best solution would be to just expose `Message::from_raw_parts`. :thinking:

zeenix: > Now I'm wondering if the best solution would be to just expose `Message::from_raw_parts`. thinking

If we do that, we should document the method better with all arguments also documented so users know what to pass.

SeaDve: I wonder if it would make sense to also use `MessageSequence` as the type of `recv_seq` as that would be self-descriptive.

zeenix: Ah yes, definitely. I missed that it's use u64 as type.

However, now that I think of it, this is only useful for messages received on a `zbus::Connection`. Perhaps it'd be better to provide another wrapper method that doesn't take this argument, `from_bytes`.

BTW, this method should be marked `unsafe` as we can't be sure of `bytes`.

zbus/src/message.rs
```diff
@@ -196,6 +196,17 @@ impl Message {
         b.build(body)
     }
 
+    /// Create a message from bytes.
+    ///
+    /// The created message will have a receive sequence of 0.
```
zeenix, line 201: Kudos for documenting this. I think it would be good to:

1. say why sequence is `0`.
2. document `fds` parameter. @elmarco any suggestions for that?
3. document that the method only has `fds` parameter on unix platforms.

zeenix (changes requested):
LGTM otherwise.

elmarco (changes requested):

zbus/src/message.rs
```diff
@@ -196,6 +196,17 @@ impl Message {
...
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe as bytes may have an invalid encoding.
+    pub unsafe fn from_bytes(bytes: Vec<u8>, #[cfg(unix)] fds: Vec<OwnedFd>) -> Result<Self> {
+        Self::from_raw_parts(bytes, fds, 0)
```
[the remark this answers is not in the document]
SeaDve: Oops, I forgot too add one.

zbus/src/message.rs
zeenix: thanks, that sounds good but i didn't mean just the general description of the paramter but more like what the parameter (could) imply for the user of this API. E.g if they construct a message from old saved message bytes from disk (like is the case for @SeaDve here) then they'll just pass an empty vec here. We need to tell that and when it woudl be empty and when it would be relevant.

elmarco: I think if you pass an empty vec and have UNIX_FD in the message, it will fail to deserialize. We can't really do anything "correct" in this case. It depends how the caller use the API.. but they can always pass /dev/null fds....

zeenix: true. The method is marked `unsafe` for good reasons anyway but we should document this.

SeaDve: > Why sequence is 0.

I'm not sure exactly how to document this

zeenix: >>  Why sequence is 0.
>
> I'm not sure exactly how to document this

Let me do it for you then. :) Add this paragraph:

```rust
/// **Note:** Since the constructed message is not construct by zbus, the [receive sequence](`Message::recv_position`) is not applicable and hence set to `0`.
```

and remove the " with a receive sequence of 0" part from the summary (it's not the most important detail about this method).

P.S. I've forgotte the rustdoc markdown for linking to an identifier so please do check if the above linking works as expected.

SeaDve: Thanks!

zbus/src/message.rs
```diff
@@ -196,6 +196,30 @@ impl Message {
...
+    /// has UNIX_FD will result in an error.
+    ///
+    /// **Note:** Since the constructed message is not construct by zbus, the receive sequence,
+    /// which can be acquired from [`recv_position`], is not applicable and hence set to `0`.
+    ///
+    /// # Safety
+    ///
+    /// This method is unsafe as bytes may have an invalid encoding.
+    ///
+    /// [`recv_position`]: #method.recv_position
```
zeenix, line 213: You don't need to hand-write the link. rustdoc does that for you. See my other comment.

zbus/src/message.rs
```diff
@@ -196,6 +196,30 @@ impl Message {
...
 
+    /// Create a message from bytes.
+    ///
+    /// The `fds` parameter is only available on unix. It specifies the file descriptors that
+    /// accompany the message. On the wire, values of the UNIX_FD types store the index of the
+    /// corresponding file descriptor in this vector. Passing an empty vector on a message that
+    /// has UNIX_FD will result in an error.
+    ///
+    /// **Note:** Since the constructed message is not construct by zbus, the receive sequence,
+    /// which can be acquired from [`recv_position`], is not applicable and hence set to `0`.
```
zeenix, line 207:
```suggestion
    /// which can be acquired from [`Message::recv_position`], is not applicable and hence set to
    ///  `0`.
```

zeenix (commented):
One tiny fix and we're definitely good to merge. :+1: 

Thanks for your patience.

zbus/src/message.rs
SeaDve: Oh, good to know, thanks!

zeenix (approved):
:+1:

SeaDve:
Thanks for the reviews!

PR #372, opened 2023-06-06
Improve `Guid` deserialization

zeenix:
Zeeshan Ali Khan on behalf of MBition GmbH.

[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md#mbition-gmbh).

### DIFF
--- a/zbus/src/fdo.rs
+++ b/zbus/src/fdo.rs
@@ -16,7 +16,9 @@ use zvariant::{
     DeserializeDict, ObjectPath, Optional, OwnedObjectPath, OwnedValue, SerializeDict, Type, Value,
 };
 
-use crate::{dbus_interface, dbus_proxy, DBusError, MessageHeader, ObjectServer, SignalContext};
+use crate::{
+    dbus_interface, dbus_proxy, DBusError, Guid, MessageHeader, ObjectServer, SignalContext,
+};
 
 #[rustfmt::skip]
 macro_rules! gen_introspectable_proxy {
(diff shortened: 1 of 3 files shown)


elmarco (changes requested):
"...verify it it's a" -> "it is"

zbus/src/fdo.rs
```diff
@@ -702,8 +704,13 @@ macro_rules! gen_dbus_proxy {
             fn get_connection_unix_user(&self, bus_name: BusName<'_>) -> Result<u32>;
 
             /// Gets the unique ID of the bus.
+            #[deprecated(since = "3.14.0", note = "Use `get_guid` instead")]
             fn get_id(&self) -> Result<String>;
 
+            /// Gets the unique ID of the bus.
+            #[dbus_proxy(name = "GetId")]
+            fn get_guid(&self) -> Result<Guid>;
```
elmarco, line 712: I think I would prefer to stick with method name, even if it may break user code (this is a rather small change). Or perhaps gate the new API(s) with a feature such as "no-deprecated" ?

zeenix: > I think I would prefer to stick with the original method name, even if it may break user code (this is a rather small change).

If we had a way to know with fair certainty that this will only break things for a small number of users, I'd agree 100% but unfortunately we don't.

Having said that, this seems more like a fix. So maybe it's not a bad option. I'll add `Deref<Target = str>` and `From<Guid> for String` impls to make things easier.

elmarco (approved):

elmarco (commented):
verious -> various

PR #379, opened 2023-06-11
✨ zv: impl Display for Value, Array, Structure, Dict & Maybe

### LINKED ISSUE #366: Add way to display `Value` nicely
I'm not sure if I missed any API, but the current nicest way to display that I know of is through its `Debug` implementation:
```
Structure {
    fields: [
        Str(
            Str(
                Borrowed(
                    ":1.202",
                ),
            ),
        ),
        Str(
            Str(
                Borrowed(
                    "",
                ),
            ),
        ),
        Str(
            Str(
                Borrowed(
                    ":1.202",
                ),
            ),
        ),
    ],
    signature: Signature(
        "(sss)",
    ),
}
```

It looks much more verbose and harder to look at compared to something from `GVariant` through `g_variant_print`:

```
(':1.202', '', ':1.202')
```

I would be willing to submit a pull request for the reimplementation of `GVariant` printing style on `zvariant::Value` if that will be fine.

SeaDve:
Closes #366

~~Marked as a draft since bytestring and str Display implementations are currently broken on edge cases.~~

### DIFF
--- a/zvariant/src/maybe.rs
+++ b/zvariant/src/maybe.rs
@@ -1,8 +1,8 @@
 use serde::ser::{Serialize, Serializer};
 use static_assertions::assert_impl_all;
-use std::convert::TryFrom;
+use std::{convert::TryFrom, fmt::Display};
 
-use crate::{Error, Signature, Type, Value};
+use crate::{value_display_fmt, Error, Signature, Type, Value};
 
 /// A helper type to wrap `Option<T>` (GVariant's Maybe type) in [`Value`].
 ///
@@ -101,6 +101,47 @@ impl<'a> Maybe<'a> {
             signature: self.signature.to_owned(),
         }
     }
+}
+
+impl Display for Maybe<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        maybe_display_fmt(self, f, true)
+    }
+}
+
+pub(crate) fn maybe_display_fmt(
+    maybe: &Maybe<'_>,
+    f: &mut std::fmt::Formatter<'_>,
+    type_annotate: bool,
+) -> std::fmt::Result {
+    if type_annotate {
+        write!(f, "@{} ", maybe.full_signature())?;
+    }
+
+    let (last_inner, depth) = {
+        let mut curr = maybe.inner();
+        let mut depth = 0;
+
+        while let Some(Value::Maybe(child_maybe)) = curr {
+            curr = child_maybe.inner();
+            depth += 1;
+        }
+
+        (curr, depth)
+    };
+
+    if let Some(last_inner) = last_inner {
+        // There are no Nothings, so print out the inner value with no prefixes.
+        value_display_fmt(last_inner, f, false)?;
+    } else {
+        // One of the maybes was Nothing, so print out the right number of justs.
+        for _ in 0..depth {
+            f.write_str("just ")?;
+        }
+        f.write_str("nothing")?;
+    }
+
+    Ok(())
 }
 
 impl<'a, T> From<Option<T>> for Maybe<'a>
--- a/zvariant/src/value.rs
+++ b/zvariant/src/value.rs
@@ -1,5 +1,9 @@
 use core::str;
-use std::{convert::TryFrom, marker::PhantomData};
+use std::{
+    convert::TryFrom,
+    fmt::{Display, Write},
+    marker::PhantomData,
+};
 
 use serde::{
     de::{
@@ -10,12 +14,13 @@ use serde::{
 };
 use static_assertions::assert_impl_all;
 
-#[cfg(feature = "gvariant")]
-use crate::Maybe;
 use crate::{
-    signature_parser::SignatureParser, utils::*, Array, Basic, Dict, DynamicType, ObjectPath,
-    OwnedValue, Signature, Str, Structure, StructureBuilder, Type,
+    array_display_fmt, dict_display_fmt, signature_parser::SignatureParser, structure_display_fmt,
+    utils::*, Array, Basic, Dict, DynamicType, ObjectPath, OwnedValue, Signature, Str, Structure,
+    StructureBuilder, Type,
 };
+#[cfg(feature = "gvariant")]
+use crate::{maybe_display_fmt, Maybe};
 
 #[cfg(unix)]
 use crate::Fd;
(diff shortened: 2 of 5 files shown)


zvariant/src/display.rs
```diff
@@ -0,0 +1,601 @@
...
+#[cfg(test)]
+mod tests {
+    use std::{collections::HashMap, convert::TryFrom};
+
+    use crate::{Fd, ObjectPath, Signature};
+
+    use super::*;
+
+    #[test]
+    fn u8_display() {
```
zeenix, line 317: Really appreciate you adding tests. However, I don't think, the display repr is important enough to warrant so many tests and adding to CI. Just a handul (10-15) random simple and complex cases would be more than enough IMO.

zvariant/src/display.rs
```diff
@@ -0,0 +1,601 @@
+/// Implemented based on https://gitlab.gnome.org/GNOME/glib/-/blob/e1d47f0b0d0893ac9171e24cc7bf635495376546/glib/gvariant.c#L2213
+use std::{
+    ffi::CStr,
+    fmt::{Display, Write},
+};
+
+use crate::{Array, Dict, Structure, Value};
```
zeenix, line 7: I'd rather keep the `Display` impls of each type in their containing file. That's not just a general convention but also consistent with other impls in our crate and repo.

Moroever, I'd ask you to add impl for each type in a separate commit so it'd be easier to add the changes to separate commits if each impl is added to the appropriate files.

SeaDve: > Moroever, I'd ask you to add impl for each type in a separate commit so it'd be easier to add the changes to separate commits if each impl is added to the appropriate files.

I think that would not be easy since the Display impl of `Value` depends on the Display impl of `Maybe` etc. and vice versa, i.e., `value_display_fmt` calls on `maybe_display_fmt` etc. and `maybe_display_fmt` calls on `value_display_fmt`.

SeaDve: I'm also a bit confused about the best way to organize these `*_display_fmt` functions. Should I store them all in `value.rs` and just call them on `maybe.rs` etc. or store them in their respective files (e.g., `maybe.rs`) and just call them in `value.rs`?

SeaDve: > > I think that would not be easy since the Display impl of Value depends on the Display impl of Maybe etc. and vice versa, i.e., value_display_fmt calls on maybe_display_fmt etc., and maybe_display_fmt etc. calls on value_display_fmt.
>
> Is that a problem?

I meant it with splitting the commit into for each file

zvariant/src/value.rs
```diff
@@ -738,3 +891,163 @@ impl<'a> Type for Value<'a> {
...
+            "((true,), (true, false), (true, true, false))"
+        );
+
+        assert_eq!(
+            Value::new((
+                (Some(0_i16), Some(Some(0_i16)), Some(Some(Some(0_i16))),),
+                (None::<i16>, Some(None::<i16>), Some(Some(None::<i16>)),),
+            ))
+            .to_string(),
+            "((@mn 0, @mmn 0, @mmmn 0), (@mn nothing, @mmn just nothing, @mmmn just just nothing))"
```
zeenix, line 1013: * `just just nothing` seems weird. I think it's easier to just (no pun intended) output `nothing`. i-e if the outer maybe is `None` than that's all the value that you need to know about (especially since we're printing the signature).
* wouldn't `0@mn` be better? It'd also be consistent with Rust's way of adding the type at the end. Also, maybe we just use the Rust syntax? This is afterall a Rust API and most users know Rust a lot more than D-Bus. We can leave the signature then to the `Debug` output.

zvariant/src/display.rs
zeenix: > I meant it with splitting the commit into for each file

Ah, sorry I didn't get it. I see the problem. Yeah, that's tough then. I guess this is an atomic commit since all modules need to implement `Display` at the same time to be able to build even. Forgot what I said about splitting commits. Sorry, should have thought about the mutual dependencies.

SeaDve: No worries

zvariant/src/value.rs
SeaDve: > * `just just nothing` seems weird. I think it's easier to just (no pun intended) output `nothing`. i-e if the outer maybe is `None` than that's all the value that you need to know about (especially since we're printing the signature).

I think it would cause a problem if you have something like `Some(None::<Option<i16>>)`. If you only have `nothing` it won't be possible to know where the `nothing` begins unlike in `@mmmn just nothing`.

>  * wouldn't `0@mn` be better? It'd also be consistent with Rust's way of adding the type at the end. Also, maybe we just use the Rust syntax? This is afterall a Rust API and most users know Rust a lot more than D-Bus. We can leave the signature then to the `Debug` output.

I don't have a strong opinion, but I was trying to be consistent with `GVariant` printing.

zeenix: > I think it would cause a problem if you have something like Some(None::<Option<i16>>). If you only have nothing it won't be possible to know where the nothing begins unlike in @mmmn just nothing.

Hmm.. true.

> I don't have a strong opinion, but I was trying to be consistent with GVariant printing.

Ah, I didn't realize there was prior art. In that case, just follow that. The bonus would be that we don't need to hardcode the strings in gvariant format tests but rather compare the glib's (it's a dev dep) output.

BTW, I didn't check but please do make sure that all  gvariant feature requiring code is under a `cfg` (didn't check if it is already or not, just a reminder).

SeaDve: > BTW, I didn't check but please do make sure that all gvariant feature requiring code is under a cfg (didn't check if it is already or not, just a reminder).

I did check by testing it with the `gvariant` feature disabled. Not sure if that was enough and, hopefully, I didn't miss a `cfg`.

zeenix: Cool. As I wrote, I didn't mean to imply that you missed anything. Just came to my head when I was writing my comment so expressed the thought before I forget it completely. :laughing:

zvariant/Cargo.toml
```diff
@@ -24,6 +24,7 @@ ostree-tests = ["gvariant"]
 
 [dependencies]
 byteorder = "1.4.3"
+unicode-general-category = "0.6"
```
zeenix, line 27: We already get a lot of angry users for our deps in zbus, so I'd like to play the devil's advocate here and say that for just one small use, let's not add a new dep?

zvariant/src/maybe.rs
```diff
@@ -103,6 +103,49 @@ impl<'a> Maybe<'a> {
...
+            curr = child_maybe.inner();
+            depth += 1;
+        }
+
+        (curr, depth)
+    };
+
+    if let Some(last_inner) = last_inner {
+        // There are no Nothings, so print out the inner
+        // value with no prefixes.
```
zeenix, line 135: nit: can easily fit in the 100 chars limit. Same for other comments: break only when needed and at 100 chars limit.

zvariant/src/maybe.rs
```diff
@@ -103,6 +103,49 @@ impl<'a> Maybe<'a> {
...
+        value_display_fmt(last_inner, f, false)?;
```
zeenix, line 136: Seems to me, this is the only place where we pass `false` for `type_annotation`. Given that it greatly complicates the code (you don't need separate crate public API if not for this), I wonder if there's a way to avoid this? Perhaps we do always print type annotation? :thinking:

zvariant/src/value.rs
```diff
@@ -738,3 +891,165 @@ impl<'a> Type for Value<'a> {
...
+                -1_i16,
+                65535_u16,
+                -1,
+                1_u32,
+                -9223372036854775808_i64,
+                18446744073709551615_u64,
+                (-1., 1.0, 11000000000., 1.1e-10)
+            ))
+            .to_string(),
+            "(byte 0xff, true, int16 -1, uint16 65535, -1, uint32 1, int64 -9223372036854775808, uint64 18446744073709551615, (-1., 1., 11000000000., 0.00000000011))"
```
zeenix, line 916: Strings literals can be broken over multiple lines using `\` suffix.

zeenix (changes requested):

zvariant/Cargo.toml
SeaDve: It won't be easy though to reimplement the functionality of that library since it does have a lot of hardcoded values: https://docs.rs/unicode-general-category/latest/src/unicode_general_category/tables.rs.html

zvariant/src/maybe.rs
SeaDve: We also disable type annotation on subsequent elements after the first element on dicts and arrays. Aside from being more verbose, we would also break compatibility with `GVariant`.

zvariant/src/value.rs
SeaDve: Thanks! Did not knew of that

zvariant/Cargo.toml
MaxVerevkin: From a quick look it seems that you use this dependency to display special characters inside a string. FYI, `String`'s `Debug` implementation already does this.

SeaDve: Would be indeed nice to exploit that. However, it would have double quotes around them, so we would have to somehow trim that

MaxVerevkin: Is it important to have single quotes?

SeaDve: If we want to consistent with `GVariant`, yeah, otherwise, not

SeaDve: Gave it a try before, but fell back to including this dependency unfortunately because we need more control for this use case, like not escaping double quotes when we enclose the string in single quotes.

MaxVerevkin: Are referring to [`str::escape_debug`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.escape_debug)? This is a method on `char`, so you can just choose not to escape the double quotes.

SeaDve: Ah, my bad, I overlooked that you mentioned char.

I gave it a try. It is consistent with `GVariant` in most cases. However, it differs in more edge cases. I'm not sure with the decision to make here, whether to maintain the consistency or go with the simpler (and one less dependency approach).

zeenix: > Do you mean by this with the scope of zbus as a whole or this `Display` impl? If you mean the latter, it would be a bug if it differs in some places

Not sure I follow. I mean the entire scope of this PR/issue.

zvariant/src/maybe.rs
zeenix: > We also disable type annotation on subsequent elements after the first element on dicts and arrays. Aside from being more verbose, we would also break compatibility with `GVariant`.

hmm.. good point.

zeenix:
Just FYI, related: https://github.com/dbus2/zbus/pull/394

zeenix:
@SeaDve please let me know when this is ready from your side.

SeaDve:
Sorry for taking too long to reply, just finished my finals.

> @SeaDve please let me know when this is ready from your side.

Sure

SeaDve:
I think it is now ready.

The extra dep is now removed, and the debug impl of `str` is now used internally instead. Thus, it uses the double quotes instead of single quotes. Also, changed the impl of `objectpath` and `signature` to match it.

zeenix (approved):
Many thanks for seeing this through. :+1:

PR #381, opened 2023-06-13
🩹 🚸 zv: Equal `Signature`s irrespective of D-Bus marshalled state

luukvanderduim:
This adjusts the `PartialEq` impls to evaluate equality irrespective of
of marshalling of the type.

Two signatures representing the same type can differ as a result of
D-Bus marshalling. D-Bus marshalling will omit outer parentheses from body
singatures.

Consider a `MyType { a: i32, b: i32}`:

`MyType::signature()` returns a "(ii)" `Signature` whereas
`Message::body_signature()` with `MyType` as body, would return
a "ii" `Signature`, which may lead to unexpected inequality of `Signature`s.

Note: `PartialEq` on `Signature` accounts for marshalling only.
One can construct- or deserialize into deeper nested structs or tuples than
than the signature describes but at that point the divergence is expected.

### DIFF
--- a/zvariant/src/signature.rs
+++ b/zvariant/src/signature.rs
@@ -96,7 +96,7 @@ impl<'b> std::ops::Deref for Bytes<'b> {
 ///
 /// [identifies]: https://dbus.freedesktop.org/doc/dbus-specification.html#type-system
 /// [`slice`]: #method.slice
-#[derive(Eq, Hash, Clone)]
+#[derive(Hash, Clone)]
 pub struct Signature<'a> {
     bytes: Bytes<'a>,
     pos: usize,
@@ -446,6 +446,10 @@ impl<'a> PartialEq<&str> for Signature<'a> {
         self.as_bytes() == other.as_bytes()
     }
 }
+
+// According to the docs, `Eq` derive should only be used on structs if all its fields are
+// are `Eq`. Hence the manual implementation.
+impl Eq for Signature<'_> {}
 
 impl<'a> Display for Signature<'a> {
     fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
--- a/zvariant/src/signature.rs
+++ b/zvariant/src/signature.rs
@@ -403,9 +403,35 @@ impl<'a> std::ops::Deref for Signature<'a> {
     }
 }
 
+/// Checks whether the string slice has balanced parentheses.
+fn has_balanced_parentheses(signature_str: &str) -> bool {
+    signature_str.chars().fold(0, |count, ch| match ch {
+        '(' => count + 1,
+        ')' if count != 0 => count - 1,
+        _ => count,
+    }) == 0
+}
+
+/// Determines whether the signature has outer parentheses and if so, return the
+/// string slice without those parentheses.
+fn without_outer_parentheses<'a, 'b>(sig: &'a Signature<'b>) -> &'a str
+where
+    'b: 'a,
+{
+    let sig_str = sig.as_str();
+
+    if let Some(subslice) = sig_str.strip_prefix('(').and_then(|s| s.strip_suffix(')')) {
+        if has_balanced_parentheses(subslice) {
+            return subslice;
+        }
+    }
+    sig_str
+}
+
+/// Evaluate equality of two signatures, ignoring outer parentheses if needed.
 impl<'a, 'b> PartialEq<Signature<'a>> for Signature<'b> {
     fn eq(&self, other: &Signature<'_>) -> bool {
-        self.as_bytes() == other.as_bytes()
+        without_outer_parentheses(self) == without_outer_parentheses(other)
     }
 }
 
--- a/zvariant/src/signature.rs
+++ b/zvariant/src/signature.rs
@@ -556,4 +556,19 @@ mod tests {
         assert_eq!(slice, "t");
         assert_eq!(slice.slice(1..), "");
     }
+
+    #[test]
+    fn signature_equality() {
+        let sig_a = Signature::from_str_unchecked("(asta{sv})");
+        let sig_b = Signature::from_str_unchecked("asta{sv}");
+        assert_eq!(sig_a, sig_b);
+
+        let sig_a = Signature::from_str_unchecked("((so)ii(uu))");
+        let sig_b = Signature::from_str_unchecked("(so)ii(uu)");
+        assert_eq!(sig_a, sig_b);
+
+        let sig_a = Signature::from_str_unchecked("(so)i");
+        let sig_b = Signature::from_str_unchecked("(so)u");
+        assert_ne!(sig_a, sig_b);
+    }
 }


luukvanderduim:
This approach proved to be a too naive an attempt.
The assumption that starting with an opening parenthesis implies one at the end, is false for the wired signatures:
"((ss)ii)" <> "(ss)ii"
We also cannot assume outer parentheses when finding an opening one at the beginning and closing at the end. Coming from the wire they do not need to pertain to the same pair:
"(ii)u(ii)"

A proper test for outer parentheses, is to check for parentheses at beginning and end _and_ then check for parentheses balance in the inner slice: 
"(ii)u(ii)" has outer parentheses, but take subslice "ii)u(ii" and we find invalid balance of parentheses. 
"((ii)u(ii))" has outer parentheses, but take subslice "(ii)u(ii)" and we'll find a valid balance.

luukvanderduim:
DBus omits outer parentheses on structs, where `Type` does not. This is inconvenient and somewhat inconsistent.
The `PartialEq<Signature> for ...` implementations now successfully evaluate equality irrespective of outer parentheses.

`Message::body_signature() == Type::signature()`, also where `Type` is a struct.

The approach relies on `b'('` and `b')'` are emcpded as a single byte in both ASCII and utf8, such that we can byteslice[1..len - 1], when we want the subslice without outher parentheses.

The impls for `str` and `&str` uphold the silent assumption the strings are valid signatures, if they are not, there can be no equality as the `Signature` they are compared to is valid.

zvariant/src/utils.rs
```diff
@@ -316,3 +316,47 @@ where
 {
     input.get(index).ok_or(Error::OutOfBounds)
 }
+
+impl Signature<'_> {
```
zeenix, line 320: Why create another impl in a separate file? I don't see any reason for this.

zvariant/src/utils.rs
```diff
@@ -316,3 +316,47 @@ where
...
+    ///
+    /// The byte wise comparison in this method requires the signature to be
+    /// of an encoding that is equal to utf8 with respect to the encoding of
+    /// ASCII characters.
+    /// D-Bus signature type codes each are a single ASCII character.
+    ///
+    pub(crate) fn without_outer_parentheses(&self) -> Option<&[u8]> {
+        let bytes = self.as_bytes();
+
+        if let [b'(', subslice @ .., b')'] = bytes {
```
zeenix, line 334: we can get rid of the ugly `b` prefixes and make the code prettier/easier to read if we operate on `char`.

zvariant/src/signature.rs
```diff
@@ -354,24 +354,39 @@ impl<'a> std::ops::Deref for Signature<'a> {
...
+// Assumes that `&str` is a valid signature string.
 impl<'a> PartialEq<&str> for Signature<'a> {
     fn eq(&self, other: &&str) -> bool {
-        self.as_bytes() == other.as_bytes()
+        let other = Signature::from_str_unchecked(other);
+        self.eq(&other)
     }
 }
 
+// `Eq` docs: "the derive strategy requires all fields are Eq, which isn’t always desired.".
```
zeenix, line 387: This would be a different logical change and hence it should be split into it's own commit with its own justification.

zvariant/src/utils.rs
```diff
@@ -316,3 +316,47 @@ where
...
+                return Some(subslice);
+            }
+        };
+
+        None
+    }
+}
+
+#[test]
+fn test_without_outer_parentheses() {
```
zeenix, line 350: these cases should be covered by the test you're adding to `signature.rs`.

zvariant/src/signature.rs
```diff
@@ -515,4 +530,19 @@ mod tests {
         assert_eq!(slice, "t");
         assert_eq!(slice.slice(1..), "");
     }
+
+    #[test]
+    fn signature_equality() {
```
zeenix, line 535: Kudos for adding tests. :+1:

zeenix (changes requested):
Looks good otherwise. Also, it would be great to modify zbus code to make use of this in message-handling code.

zvariant/src/utils.rs
luukvanderduim: The reasoning was twofold; that it is a utility to `PartialEq` and I liked the method syntax better than the function.

Would you rather have a function in `util` or move the method to the main impl?

zvariant/src/utils.rs
luukvanderduim: I liked the pattern match `[b'(', subslice @ .., b')'] = bytes ..`, as it was likely efficient, avoids indexing and resulted in the subslice in one go.

But I see that handling chars avoids the need for a safety  notice.

zvariant/src/signature.rs
```diff
@@ -354,24 +354,39 @@ impl<'a> std::ops::Deref for Signature<'a> {
     }
 }
 
-impl<'a, 'b> PartialEq<Signature<'a>> for Signature<'b> {
-    fn eq(&self, other: &Signature<'_>) -> bool {
-        self.as_bytes() == other.as_bytes()
+/// Evaluate equality of two signatures, ignoring outer parentheses.
+impl<'a, 'b> PartialEq<Signature<'b>> for Signature<'a> {
```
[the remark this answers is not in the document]
zeenix: The lifetimes order is still being changed. Please don't resolve threads before ensuring it's been addressed.

zvariant/src/signature.rs
```diff
@@ -354,24 +354,39 @@ impl<'a> std::ops::Deref for Signature<'a> {
...
+    fn eq(&self, other: &Signature<'b>) -> bool {
```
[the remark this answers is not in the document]
zeenix: same here.

zvariant/src/signature.rs
```diff
@@ -354,24 +354,39 @@ impl<'a> std::ops::Deref for Signature<'a> {
...
+            other.without_outer_parentheses(),
+        ) {
+            (Some(lhs_without), None) => lhs_without == other.as_bytes(),
+            (None, Some(rhs_without)) => self.as_bytes() == rhs_without,
+            _ => self.as_bytes() == other.as_bytes(),
+        }
     }
 }
 
+// Assumes that `str` is a valid signature string.
```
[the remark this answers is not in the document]
zeenix: and here.

zvariant/src/signature.rs
```diff
@@ -83,7 +83,7 @@ impl<'b> std::ops::Deref for Bytes<'b> {
 ///
 /// [identifies]: https://dbus.freedesktop.org/doc/dbus-specification.html#type-system
 /// [`slice`]: #method.slice
-#[derive(Hash, Clone)]
+#[derive(Eq, Hash, Clone)]
```
zeenix: I think the rebase got messed up again. You remove this in the first commit, add it again in the second and remove it back in the last commit.

zeenix (changes requested):
I understand you're still grasping around git rebasing etc so I understand completely (I'd make a lot more mistakes in your shoes for sure) but kindly review each line of every commit before (or after) pushing) to catch such mistakes.

luukvanderduim:
Unfortunately I have messed up. I will take your advise to heart, sorry to trouble you with it.

luukvanderduim:
More comfortable with the commits' outline, but shape is just a prerequisite

I remember you mention a preference for string comparisons in favor of the byte slice comparisons.
This is what I ended up with:

```rust
    fn eq(&self, other: &Signature<'_>) -> bool {
        match (
            self.without_outer_parentheses(),
            other.without_outer_parentheses(),
        ) {
            (Some(lhs_without), None) => *lhs_without == **other,
            (None, Some(rhs_without)) => **self == *rhs_without,
            _ => **self == **other,
        }
    }
```
I am a bit torn, as the dereferences do different things and make it less clear what actually happens but on the other hand, this is prettier than having six `.as_bytes()` calls around, and the alternative of having six `.as_str()` about the same. 
I am sure you know what you like better.

zeenix:
> More comfortable with the commits' outline, but shape is just a prerequisite

:+1: 

> I remember you mention a preference for string comparisons in favor of the byte slice comparisons. This is what I ended up with:
> 
> ```rust
> 
> ```

hmm.. 

* why can't `without_outer_parentheses` method always return a slice? If there are no parantheses, it just returns the input as is.
* I wasn't objecting to use of `as_bytes` but rather byte literals because of `b'` suffix. Looks a bit ugly so when there is no need, I'd very much like to avoid it.

luukvanderduim:
>     * why can't `without_outer_parentheses` method always return a slice? If there are no parantheses, it just returns the input as is.

It can - and it does now.
This simplifies `PartialEq` nicely.

Now, I am pondering whether the method `without_outer_parentheses` should be called either `marshall_signature` or just 'marshalled_str' because it is not necessarily removing outer parentheses.

I misunderstood how ` Signature::slice` is not intended to hand out a slice of its bytes.
I see now how ``slice` updates `pos` and `end` and Signature::as_str` does do the slicing.
I presume this is a more widely used pattern I had just not looked at this closely before.

luukvanderduim:
Being curious what the differences would be between the approaches, I rolled [a quick bench crate](https://github.com/luukvanderduim/partialeq_bench)

zeenix:
> > ```
> > * why can't `without_outer_parentheses` method always return a slice? If there are no parantheses, it just returns the input as is.
> > ```
> 
> 
> It can - and it does now.
> This simplifies `PartialEq` nicely.

Nice. :+1: 

> Now, I am pondering whether the method `without_outer_parentheses` should be called either `marshall_signature` or just 'marshalled_str' because it is not necessarily removing outer parentheses.

Nah, `without_outer_parentheses` is fine. Just add doc string to the method, explaining that it doesn't remove parens if there are none.

zeenix:
BTW, not sure if I mentioned here before but this PR should also change zbus code to make use of this new Signature ability.

zeenix:
> You mentioned zbus earlier but I think I read 'could' then.

I wrote `would`. :) But not worries.

> I take it that this refers to `MessageBuilder`?

Yes (at least, see if we do something like that elsewhere too in code please).

zeenix (changes requested):
Actually apart from these minor things, I don't see why we can't merge this work. Am I forgetting something from our conversations on Matrix? :thinking: 

You'll have to rebase on main in any case.

luukvanderduim:
Funny we end up in this situation where you suggest to merge and I have concerns.

On matrix I shared that I had come to realize that `R` (the Rust type), `B` (the `Message` body type), and `D` (the de-serialized Rust type) are not necessarily the same, despite being closely related. They share "a list of zero or more single complete types, each made up of one or more type codes" (in D-Bus spec speak) - but may differ by zero or more outer structs or tuples. 

An "ii" body deserializes to "(((ii)))" just fine, which is understandable, because D-Bus and zvariant don't care about the outer shape, and happily accommodate how the user wants the data. But if some user does this, we are more than one pair of parentheses apart

The question is, should `PartialEq` reflect on equality of the underlying data and be as lenient as going 'R->B->D' can be?, Or should it support the common case of allowing divergence by one pair of parentheses, as this PR? 
Or should it just stick to being a trait to a type and tell it like it is?

The first way offers least chance of surprise and most expensive `PartialEq` , second, this PR, is pragmatic but does not cover all cases, the last seems most principled and is least expensive.

In tests, people probably will not care about more expensive equality.
Users might care if they check signatures in a hot loop?
Could `SignalStream::filter_map` by signature ,ale sense for some people, for example?

Either way, I do think the associated functions on `Signature` make sense.

zeenix:
> Funny we end up in this situation where you suggest to merge and I have concerns.

:laughing: 



> On matrix I shared that I had come to realize that `R` (the Rust type), `B` (the `Message` body type), and `D` (the de-serialized Rust type) are not necessarily the same

Right, I recall now. :) As I tried to explain, I do not care much for theoretical issues. We already need to treat `(x)` the same as `x` and we can't help it, and in theory that's just wrong. So I'm more concerned about consistency here, above all.



> Or should it support the common case of allowing divergence by one pair of parentheses, as this PR?

Yeah, I think this is a good compromise. Hence why I'm thinking this PR is the way to go.

> second, this PR, is pragmatic but does not cover all cases

I don't think we need to cover all cases. You'll notice that `message::Builder` removes only one layer of `()`. So this approach is also consistent with what we've been doing already.

zeenix:
@luukvanderduim You've more than a 100 commits on the PR now. Something went horribly wrong in your rebase. :(

luukvanderduim:
Unfortunate.
Ill go through the motions again.

luukvanderduim:
Sanity seems to have been restored. We are back at three commits, which - looking at the changes - are what we want. 
One commit dropped because the function it modified no longer exists.

zeenix:
> Sanity seems to have been restored. We are back at three commits, which - looking at the changes - are what we want.

Great, thanks again for still not giving up. :)



> One commit dropped because the function it modified no longer exists.

It exists in another form (see commit a0b92da66aa57ec5edb383cc4b8a117602d08262). Please make sure the new code doesn't need these changes that were dropped.

Btw, I think the CI can be fixed by just rebasing on the latest `main` branch.

luukvanderduim:
> > One commit dropped because the function it modified no longer exists.
> 
> It exists in another form (see commit [a0b92da](https://github.com/dbus2/zbus/commit/a0b92da66aa57ec5edb383cc4b8a117602d08262)). Please make sure the new code doesn't need these changes that were dropped.

Looks like it does:

`SignatureParser::validate()` relies on `SignatureParser::from_bytes_unchecked()`
Checks max. length is <= 255.

`SignatureParser::validate()` then relies on `SignatureParser`'s `Iterator` implementation.

`SignatureParser as Iterator::next()` ->
`None` if `self.done()`, otherrwise:

`next_signature()` parses by (start-) character and if it finds an opening parenthesis,
it calls `next_structure_signature()`:

Which errors when it would find an empty struct.
Initializes a 'field parser' and keeps track of fields' lengths.
The `while` loop advances over the `field_parser` slice, calling `parse_next_signature`  
and sums up the lengths of the 'single complete types' / struct field signatures it finds, until
either at premature end or when it finds ')'
- Would Err it `next_char` were not ')' (because it would mean the loop exited prematurely.

     returns `Ok(Signature)` signature of slice of the struct.
  caller `next_signature()` returns `Ok(Signature)`
caller `next()` returns `Some(Ok(Signature))`

`validate()` evaluates these 'single complete type signature' results: `s?`.

Long story short, yes I believe 'balanced parantheses' are checked this way.

Note, there are things the parser does not do, but also is not that important:
- max. braces depth (32), max. total nesting depth (64)
 
> 
> Btw, I think the CI can be fixed by just rebasing on the latest `main` branch.

Reading the log it seems a library is missing in the CI environment. You addressed that?
Will try re-rebasing to main tomorrow.

zeenix:
> Long story short, yes I believe 'balanced parantheses' are checked this way.

:+1: so these changes are good then, good.

> > Btw, I think the CI can be fixed by just rebasing on the latest `main` branch.
> 
> Seems #440 is related.

Right, it should work now. I restarted the failing job. :crossed_fingers: 

> max. braces depth (32), max. total nesting depth (64)

Hmm.. currently these are checked by the serializers and deserializers so it is ultimately checked. However, now I wonder if the checks should be on `Signature` creation itself. The signature is parsed for checking anyway.

luukvanderduim:
> 
> Right, it should work now. I restarted the failing job. crossed_fingers
> 

Rebased to main. 

> > max. braces depth (32), max. total nesting depth (64)
> 
> Hmm.. currently these are checked by the serializers and deserializers so it is ultimately checked. However, now I wonder if the checks should be on `Signature` creation itself. The signature is parsed for checking anyway.

Does moving those checks solve a problem?

zeenix:
> > Hmm.. currently these are checked by the serializers and deserializers so it is ultimately checked. However, now I wonder if the checks should be on `Signature` creation itself. The signature is parsed for checking anyway.
> 
> Does moving those checks solve a problem?

Well ser/de process isn't currently as fast as it should be so maybe it would help there. Signature creation is typically amortized. So it'll definitely optimize the overall code but can't be sure if it'd be significant enough to care.

zvariant/src/signature.rs
```diff
@@ -294,6 +294,29 @@ impl<'a> Signature<'a> {
         clone
     }
 
+    /// Associated function to check whether the signature string slice has
+    /// balanced parentheses.   
+    pub fn has_balanced_parentheses(signature_str: &str) -> bool {
```
zeenix, line 299: * why is this public? It's not even a method of Signature. Its just a private utility function so should be private and separate function.
* I thought we don't need it?

zeenix (changes requested):
Sorry, i still found something not right. :)

zvariant/src/signature.rs
zeenix: ?

zeenix: You addressed the first point but didn't answer my question. IIRC (I've to admit it's hard to remember our last convo on this) we had said we don't need to check for balancing?

zeenix: Looking at the code now and previous comment, I get it. However, I don't think it deserves to be an associated function. Please just move it to be a separate function. Then I believe we're good to merge. :)

luukvanderduim:
Moved the associated functions out. I hope you are happy with where they landed (just above where they are needed, above impl PartialEq<Signature<>> for Signature<> and take note of the lifetime bounds on `without_outer_parentheses` which are new and are necessary to convey the relation between the borrow of the Signature and its internal reference.

zeenix (approved):

PR #385, opened 2023-06-18
zb: add arg0ns to match rule, deprecate arg0namespace

### LINKED ISSUE #382: `arg0namespace` has an overly restrictive type
Currently `-` is not supported in bus names:

> ```
> Err(
>     Variant(
>         Message(
>             "Invalid interface or error name: `-` character not allowed",
>         ),
>     ),
> )
> ```

Even though [the dbus spec](https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus) does discourage the usage of `-`:

> Each element [of a bus name] must only contain the ASCII characters "[A-Z][a-z][0-9]_-", with "-" discouraged in new bus names.

I think it is overly strict and there are still apps that use `-` in their names.

Reported downstream at https://github.com/greshake/i3status-rust/issues/1914.

MaxVerevkin:
arg0namespace has an overly restrictive type (InterfaceName), which disallows valid values (e.g. those containing '-' or single-element names such as 'org').

Further more, a MatchRule with arg0namespace previously would not match a message whose arg0 is not a valid InterfaceName (even if it is a valid BusName).

fixes #382

### DIFF
--- a/zbus/src/match_rule.rs
+++ b/zbus/src/match_rule.rs
@@ -284,8 +284,10 @@ impl<'m> MatchRule<'m> {
         // The arg0 namespace.
         if let Some(arg0_ns) = self.arg0ns() {
             if let Ok(arg0) = msg.body_unchecked::<BusName<'_>>() {
-                if !arg0.starts_with(arg0_ns.as_str()) {
-                    return Ok(false);
+                match arg0.strip_prefix(arg0_ns.as_str()) {
+                    None => return Ok(false),
+                    Some(s) if !s.is_empty() && !s.starts_with('.') => return Ok(false),
+                    _ => (),
                 }
             } else {
                 return Ok(false);
--- a/zbus/src/match_rule.rs
+++ b/zbus/src/match_rule.rs
@@ -93,6 +93,7 @@ pub struct MatchRule<'m> {
     pub(crate) args: Vec<(u8, Str<'m>)>,
     pub(crate) arg_paths: Vec<(u8, ObjectPath<'m>)>,
     pub(crate) arg0namespace: Option<InterfaceName<'m>>,
+    pub(crate) arg0ns: Option<Str<'m>>,
 }
 
 assert_impl_all!(MatchRule<'_>: Send, Sync, Unpin);
@@ -153,6 +154,11 @@ impl<'m> MatchRule<'m> {
         self.arg0namespace.as_ref()
     }
 
+    /// Match messages whose first argument is within the specified namespace.
+    pub fn arg0ns(&self) -> Option<&Str<'m>> {
+        self.arg0ns.as_ref()
+    }
+
     /// Creates an owned clone of `self`.
     pub fn to_owned(&self) -> MatchRule<'static> {
         MatchRule {
@@ -169,6 +175,7 @@ impl<'m> MatchRule<'m> {
                 .map(|(i, p)| (*i, p.to_owned()))
                 .collect(),
             arg0namespace: self.arg0namespace.as_ref().map(|a| a.to_owned()),
+            arg0ns: self.arg0ns.as_ref().map(|a| a.to_owned()),
         }
     }
 
@@ -192,6 +199,7 @@ impl<'m> MatchRule<'m> {
                 .map(|(i, p)| (i, p.into_owned()))
                 .collect(),
             arg0namespace: self.arg0namespace.map(|a| a.into_owned()),
+            arg0ns: self.arg0ns.map(|a| a.into_owned()),
         }
     }
 
@@ -276,8 +284,8 @@ impl<'m> MatchRule<'m> {
         }
 
         // The arg0 namespace.
-        if let Some(arg0_ns) = self.arg0namespace() {
-            if let Ok(arg0) = msg.body_unchecked::<InterfaceName<'_>>() {
+        if let Some(arg0_ns) = self.arg0ns() {
+            if let Ok(arg0) = msg.body_unchecked::<BusName<'_>>() {
                 if !arg0.starts_with(arg0_ns.as_str()) {
                     return Ok(false);
                 }
@@ -362,7 +370,7 @@ impl ToString for MatchRule<'_> {
         for (i, arg_path) in self.arg_paths() {
             add_match_rule_string_component(&mut s, &format!("arg{i}path"), arg_path);
         }
-        if let Some(arg0namespace) = self.arg0namespace() {
+        if let Some(arg0namespace) = self.arg0ns() {
             add_match_rule_string_component(&mut s, "arg0namespace", arg0namespace)
         }
 
@@ -417,7 +425,7 @@ impl<'m> TryFrom<&'m str> for MatchRule<'m> {
                 "path" => builder.path(value)?,
                 "path_namespace" => builder.path_namespace(value)?,
                 "destination" => builder.destination(value)?,
-                "arg0namespace" => builder.arg0namespace(value)?,
+                "arg0namespace" => builder.arg0ns(value)?,
                 key if key.starts_with("arg") => {
                     if let Some(trailing_idx) = key.find("path") {
                         let idx = key[3..trailing_idx]
--- a/zbus/src/match_rule.rs
+++ b/zbus/src/match_rule.rs
@@ -146,10 +146,8 @@ impl<'m> MatchRule<'m> {
 
     /// Match messages whose first argument is within the specified namespace.
     ///
-    /// Note that while the spec allows this to be any string that's a valid bus or interface name
-    /// except that it can lack `.`, we only allow valid interface names. The reason is not only to
-    /// keep things simple and type safe at the same time but also for the fact that use cases of
-    /// only matching on the first component of a bus or interface name are unheard of.
+    /// This function is deprecated because the choice of `InterfaceName` was too restrictive.
+    #[deprecated = "use arg0ns instead"]
     pub fn arg0namespace(&self) -> Option<&InterfaceName<'_>> {
         self.arg0namespace.as_ref()
     }
--- a/zbus/src/match_rule_builder.rs
+++ b/zbus/src/match_rule_builder.rs
@@ -13,6 +13,7 @@ const MAX_ARGS: u8 = 64;
 /// Builder for [`MatchRule`].
 ///
 /// This is created by [`MatchRule::builder`].
+#[derive(Debug)]
 pub struct MatchRuleBuilder<'m>(MatchRule<'m>);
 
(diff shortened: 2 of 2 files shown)


MaxVerevkin:
Is the validation of arg0namespace needed?

zeenix:
> Is the validation of arg0namespace needed?

Yes. We should verify all strings.

zbus/src/match_rule.rs
```diff
@@ -144,15 +145,16 @@ impl<'m> MatchRule<'m> {
     }
 
     /// Match messages whose first argument is within the specified namespace.
-    ///
-    /// Note that while the spec allows this to be any string that's a valid bus or interface name
-    /// except that it can have no `.`, we only allow valid interface names. The reason is not only
-    /// to keep things simple and type safe at the same time but also for the fact that use cases of
-    /// only matching on the first component of a bus or interface name are unheard of.
```
zeenix, line 151: * These docs need to be moved to (maybe modified to fit) the new methods.
* Let's document here why this is deprecated.

zbus/src/match_rule_builder.rs
```diff
@@ -200,12 +200,28 @@ impl<'m> MatchRuleBuilder<'m> {
...
+        let namespace = namespace.try_into().map_err(Into::into)?;
+        self.0.arg0namespace = Some(namespace.clone());
+        self.0.arg0ns = Some(namespace.into());
+
+        Ok(self)
+    }
+
+    /// Set 0th argument's namespace.
+    ///
+    /// The namespace be a valid bus name or a valid element of a bus name.
```
zeenix, line 218: We need a bit more docs here. I know, you're just following my bad example here but don't. :laughing: This is the method that most people will use rather than `MatchRule::arg0ns`.

zbus_names/src/interface_name.rs
```diff
@@ -184,6 +184,12 @@ impl<'name> TryFrom<Cow<'name, str>> for InterfaceName<'name> {
     }
 }
 
+impl<'name> From<InterfaceName<'name>> for Str<'name> {
```
zeenix, line 187: * It'd be strange to only implement this for InterfaceName. Please add for other `Str`-based types too, for consistency.
* Completely separate change from this commit. You know what to do. :)

zeenix (changes requested):

MaxVerevkin:
> > Is the validation of arg0namespace needed?
> 
> Yes. We should verify all strings.

Should I put the validation code in `arg0ns` on move it to a separate function?

zbus/src/match_rule.rs
MaxVerevkin: > These docs need to be moved to (maybe modified to fit) the new methods.

I don't see how this comment applies to the new methods...

zbus/src/match_rule_builder.rs
MaxVerevkin: Should I list the complete list of requirements?

zeenix:
> I suppose I also need to implement `TryFrom<Str> for .*Name` (to set `arg0namespace` in `arg0ns` function for backward compatibility of `TryFrom<&str> for MatchRule`)?

Yeah, users should be able to pass a `&str`, that's actually the main reason for having these generic.

> Unrelated, but may I ask why custom `Str` is used instead of `Cow`?

Hmm.. we should document this better. Cow is very generic and doesn't do special handing of `&'static str` and hence if you do `to_owned` on it, it will do allocation+copying. OTOH `Str` has a special variant for static strings and `to_owned` is free. That's helpful in generic code, where you want to get an owned version, regardless of which variant is it.

zbus/src/match_rule_builder.rs
zeenix: Just copy&paste from the spec and modify a bit?

MaxVerevkin:
I moved `zubs_names` changes to a new PR and rebased this PR, I hope this way it should be easier to review.

zbus/src/match_rule_builder.rs
MaxVerevkin: Perhaps a few examples + a link to the spec would be sufficient?

zeenix: > Perhaps a few examples + a link to the spec would be sufficient?

Sure examples would be even better but we should document the rules anyway. It doesn't have to be long +"Please refer to the D-Bus specification [here](link) for details." at the end would do, yeah.

zeenix:
> Should I put the validation code in `arg0ns` on move it to a separate function?

Oh I missed this one. Given it's the only place this is going to be needed, I think it's fine to keep it all there.

zbus/src/match_rule_builder.rs
MaxVerevkin: For now I left a link to the spec + a few examples of valid and invalid namespaces, similar to all *Name types.

zbus/src/match_rule.rs
zeenix: > I don't see how this comment applies to the new methods...

Not as is, no (hence the "maybe modified to fit" bit). :) Anyway, I think this is covered by the other thread.

Would have been great to address the second bullet here but it's not biggie I guess.

MaxVerevkin: > Would have been great to address the second bullet here but it's not biggie I guess.

But I think I did address it :)

zbus/src/error.rs
```diff
@@ -62,6 +62,8 @@ pub enum Error {
     ///
     /// [MR]: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-routing-match-rules
     InvalidMatchRule,
+    /// Invalid `arg0namespace`
+    InvalidArg0Namespace,
```
zeenix, line 66: Not sure about this one. We use `InvalidMatchRule` for others so let's use that here as well. Otherwise, this type will keep going bigger and bigger. If we want fine-grained errors, we then need to do that for other errors in MatchBuilder as well.

Also, btw this would be a good candidate for a separate (first) commit.

zbus/src/match_rule_builder.rs
```diff
@@ -200,12 +200,73 @@ impl<'m> MatchRuleBuilder<'m> {
     }
 
     /// Set 0th argument's namespace.
+    ///
+    /// This function is deprecated because the choice of `InterfaceName` was too restrictive.
+    #[deprecated = "use arg0ns instead"]
```
zeenix, line 205: nit: Ideally, deprecation should be a separate commit (follow-up). That way if we decide for any reason to revert the deprecation, it'd be a simple `git revert` call.

zbus/src/match_rule_builder.rs
```diff
@@ -200,12 +200,73 @@ impl<'m> MatchRuleBuilder<'m> {
...
+
+        // Rules: https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-bus
+        // minus the requirement to have more than one element.
+
+        if namespace.is_empty() || namespace.len() > 255 {
+            return Err(Error::InvalidArg0Namespace);
+        }
+
+        let (is_unique, namespace_str) = match namespace.strip_prefix(':') {
+            Some(x) => (true, x),
```
zeenix, line 249: `x` -> `s`.

zbus/src/match_rule_builder.rs
```diff
@@ -200,12 +200,73 @@ impl<'m> MatchRuleBuilder<'m> {
...
+    /// - `org`
+    /// - `:org`
+    /// - `:1org`
+    ///
+    /// Eamples of invalid namespaces:
+    /// - `org.`
+    /// - `.org`
+    /// - `1org`
+    /// - `.`
+    /// - `org..freedesktop`
```
zeenix, line 234: Nice! Even better if we convert this into a test case. See the `ObjectPath` docs.

zeenix (changes requested):

zbus/src/match_rule_builder.rs
zeenix: > For now I left a link to the spec + a few examples of valid and invalid namespaces, similar to all *Name types.

Looks good. Examples can be make to work for us as tests too (and hence also we ensure the docs remain in sync with the code). See the [thread](https://github.com/dbus2/zbus/pull/385#discussion_r1239866704).

zeenix: Please resolve this thread once you've read my previous comment.

zbus/src/match_rule_builder.rs
MaxVerevkin: You mean a doctest?

MaxVerevkin: `unwrap_err` requires that `MatchRuleBuilder` implements `Debug`. Is it OK if I add `derive(Debug)` (is a separate commit)?

zeenix: > You mean a doctest?

yes.

> `unwrap_err` requires that `MatchRuleBuilder` implements `Debug`. Is it OK if I add `derive(Debug)` (in a separate commit)?

Sure thing. It's probably a desired change on its own anyway.

zbus/src/match_rule.rs
```diff
@@ -276,8 +284,8 @@ impl<'m> MatchRule<'m> {
         }
 
         // The arg0 namespace.
-        if let Some(arg0_ns) = self.arg0namespace() {
-            if let Ok(arg0) = msg.body_unchecked::<InterfaceName<'_>>() {
+        if let Some(arg0_ns) = self.arg0ns() {
+            if let Ok(arg0) = msg.body_unchecked::<Str<'_>>() {
                 if !arg0.starts_with(arg0_ns.as_str()) {
                     return Ok(false);
                 }
```
MaxVerevkin, line 289: - I believe this implementation was incorrect, because to match `arg0namespace`, the first N elements must match the namespace. For example, `arg0namespace='o'` should not match `org.foo` (just checked in using busctl monitor --match`). I guess it's better to fix in a new commit?
- I'm not sure, should it be `msg.body_unchecked::<BusName<'_>>()` instead?

zeenix: > * I believe this implementation was incorrect, because to match `arg0namespace`, the first N elements must match the namespace. For example, `arg0namespace='o'` should not match `org.foo` (just checked in using `busctl monitor --match`).

Nice catch. 

> I guess it's better to fix in a new commit?

You bet!

> I'm not sure, should it be `msg.body_unchecked::<BusName<'_>>()` instead?

I'm not sure either if we need to be strict. Let's make it stricter and if that's too strict, we can always revert back to `Str`.

zeenix (approved):
LGTM. Thanks!

MaxVerevkin:
Any plans for a new release? :)

zeenix:
> Any plans for a new release? :)

Yes, a few small things on the way.

PR #390, opened 2023-06-25
🍻 Random things (sitting next to Zeeshan)

elmarco:


zbus/tests/xml.rs
```diff
@@ -0,0 +1,11 @@
+#![cfg(feature = "quick-xml")]
+
+use std::convert::TryFrom;
+
+use zbus::quick_xml::Node;
+
+#[test]
+fn invalid_arg_type() {
+    let input = include_str!("data/invalid_arg_type.xml");
+    assert!(Node::try_from(input).is_err());
```
zeenix, line 39: even better if you test for a specific error using `matches`.

zbus/src/address.rs
```diff
@@ -165,6 +165,9 @@ pub enum Address {
     /// type of `stream` is `vsock::VsockStream` with `vsock` feature and `tokio_vsock::VsockStream`
     /// with `tokio-vsock` feature.
     Vsock(VsockAddress),
+    /// A listenable address using a specified path, in which a socket file with a random file name
```
zeenix, line 174: * "using a" -> "using the".
* can we document a bit better, like where does this come from (perhaps link to the appropriate part of the spec as well)?

zbus/src/address.rs
```diff
@@ -168,6 +168,10 @@ pub enum Address {
     /// A listenable address using a specified path, in which a socket file with a random file name
     /// starting with 'dbus-' will be created by the server.
     UnixDir(OsString),
+    /// The same as UnixDir, except that on platforms with abstract sockets, the server may attempt
```
zeenix, line 177: * may? Isn't it that it **will** make the attempt?
* how do clients figure out the full path (same with `UnixDir`)? Probably should document that.

zbus/src/lib.rs
```diff
@@ -41,8 +41,14 @@ pub use dbus_error::*;
...
-mod address;
-pub use address::*;
+pub mod address;
+// re-export mostly for compatibility reasons
+#[cfg(any(
+    all(feature = "vsock", not(feature = "tokio")),
+    feature = "tokio-vsock"
+))]
+pub use address::VsockAddress;
+pub use address::{Address, TcpAddress, TcpAddressFamily};
```
zeenix, line 51: This is half of what I meant. :) 

* Names should be shortened where possible: `address::{Vsock, Tcp, TcpFamily}`.
* We provide type aliases for compatibility but deprecate them to discourage future use.

zbus/src/address.rs
```diff
@@ -674,11 +675,19 @@ impl Display for Address {
     }
 }
 
-impl FromStr for Address {
-    type Err = Error;
+/// A parsed bus address
+#[derive(Debug, PartialEq, Eq)]
+pub struct Parsed<'a> {
```
zeenix, line 680: `Address` is already supposed to be parsed so this feels like a wrong solution. Any other ideas? I'll think about it too..

zeenix (changes requested):
LGTM otherwise.

zvariant/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+#![recursion_limit = "256"]
```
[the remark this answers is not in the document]
elmarco: It's a crate-level attribute. Why would it make it conditional if it's happening within the limits of what dbus supports?

zvariant/src/signature.rs
```diff
@@ -416,7 +416,8 @@ impl<'de> Visitor<'de> for SignatureVisitor {
     }
 }
 
-fn ensure_correct_signature_str(signature: &[u8]) -> Result<()> {
+fn parse_signature_str(signature: &[u8]) -> Result<Vec<Signature<'_>>> {
```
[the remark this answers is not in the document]
elmarco: This is not changing the semantic, nor the complexity either. It's simply accumulating what is already returned by parser.parse_next_signature(). 

We can probably further optimize it by implementing an iterator instead, that would fit the use case well enough.

zvariant/src/signature.rs
```diff
@@ -271,6 +276,15 @@ impl<'a> Signature<'a> {
...
         clone
     }
+
+    /// Returns the number of complete types for the signature.
+    ///
+    /// **Note**: if the signature is invalid, returns 0.
+    pub fn n_types(&self) -> usize {
+        parse_signature_str(self.as_bytes())
+            .map(|s| s.len())
+            .unwrap_or(0)
```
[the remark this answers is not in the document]
elmarco: this is where an iterator would avoid the Vec. Although Rust my by clever enough to inline all this and get rid of the Vec allocations already.

zbus/src/quick_xml.rs
```diff
@@ -371,7 +371,9 @@ mod tests {
 
     #[test]
     fn serde() -> Result<(), Box<dyn Error>> {
-        let node = Node::from_reader(EXAMPLE.as_bytes())?;
+        let node_r = Node::from_reader(EXAMPLE.as_bytes())?;
+        let node: Node<'_> = EXAMPLE.try_into()?;
```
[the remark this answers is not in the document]
elmarco: sure, just need the import.

zbus/src/address.rs
elmarco: It's from the spec. It doesn't have to, so it may.. 

Clients can't figure out the path themself, the server or the user is providing it.

zbus/src/lib.rs
elmarco: Yes, that's also my aim, however! I think the type should still convey enough meaning on it's own though. Tcp or Vsock would be a pretty bad type name, it would be confusing and easily conflict with other related parts.. even in zbus code itself. TcpFamily might be annough, although AddressFamily could be more appropriate and general for ex. So for address:: I don't think we should rename much.

zbus/src/address.rs
elmarco: I thought Address is more "concrete" than Parsed, but I am open to naming suggestions or other proposals :)

zbus/tests/xml.rs
elmarco: ok

zbus/src/quick_xml.rs
```diff
@@ -333,6 +333,19 @@ impl<'a> TryFrom<&'a str> for Node<'a> {
...
 
+fn de_type_valid<'de, D>(deserializer: D) -> Result<String, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let v = String::deserialize(deserializer)?;
+    let sig = zvariant::Signature::try_from(v.as_str())
+        .map_err(|e| serde::de::Error::custom(e.to_string()))?;
+    if sig.n_types() != 1 {
+        return Err(serde::de::Error::custom(format!("Invalid type: {}", v)));
```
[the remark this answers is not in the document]
elmarco: ok

zbus/src/address.rs
elmarco: this is taken almost straight from dbus spec, adding a link

zvariant/src/lib.rs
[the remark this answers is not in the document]
zeenix: > It's a crate-level attribute.

Ughh.. :(

> Why would it make it conditional if it's happening within the limits of what dbus supports?

Because we don't need it at all. The whole reason the test is failing is because we don't allow this level of recursion.

But maybe it's a good thing that we limit recursion at compile-time? We can then remove the checks at serializer end and only check and test the deserialization? Or maybe there is no need for checks at all then as compiler knows the type being deserialized to? Need to look more closely into what this new compiler check is all about..

zbus/src/address.rs
elmarco: It may use non-abstract path. The server typically communicate the address with --print-address (see https://github.com/dbus2/busd/pull/23/commits/f85e20473077f1ed0e009860d6b3a5d6641f699c)

zbus/src/address.rs
elmarco: Hmm? What I wanted here is a way to parse the guid= part, and possibly other options... We would need to break the whole Address API to add fields to every variants. Not my call. This intermdiary type is good enough and flexible, it should fit most cases.

An future alternative may be to completely change the way Address is implemented, to make it opaque and have various accessors. I am not sure this is better either.

zbus/src/address.rs
zeenix: > It may use non-abstract path. The server typically communicate the address with --print-address

But then the actual path is part of the address then. The docs are just confusing in that case.

zbus/src/address.rs
zeenix: > We would need to break the whole Address API to add fields to every variants. Not my call.

No we don't. That's what I'm saying. Because enum is non_exhaustive, we can add new variants w/o breaking any API. Hence my suggestion.

> This intermdiary type is good enough and flexible, it should fit most cases.

As I wrote, it's not a good API. It'll only confuse users (who are unfortunate enough to deal with these types directly).

> An future alternative may be to completely change the way Address is implemented, to make it opaque and have various accessors. I am not sure this is better either.

For sure. Please create an issue for that with `api break` label for that.

elmarco: It's essentially a hard API break since it will no longer returns Address:X(_) but Address::XWithGuid()...

zeenix: > It's essentially a hard API break since it will no longer returns Address:X(_) but Address::XWithGuid()...

It's not. non_exhaustive means we can return any variant, even ones we add later. It's basically an escape hatch. In case of `Error` type, we even have to have it. There is no way, we can predict all error variants in advance.

> and it's bad design anyway, what if we have another option?

Sure but still better than introducing a weird concept. `Address` **is** the parsed type. We should keep in mind that most users don't deal with addresses directly.

zbus/src/lib.rs
zeenix: I realized that perhaps I'm getting a bit too hardcore about modules here. Perhaps you're right and we should still have `Address` in the names etc. However, probably still best to do this separate and for all modules.

zbus/src/address.rs
zeenix: > The more I think about it, the more I think the introduced intermediary "Parsed" type is good design.

You're alone in that. :) However, before we fight each other to death on this, let's first check if guid handling is really that important that it can't wait till the next API break (I'm starting to feel that's not far off at all). The spec says:

> A server may specify a key-value pair with the key guid and the value a hex-encoded 16-byte sequence. [the section called “UUIDs”](https://dbus.freedesktop.org/doc/dbus-specification.html#uuids) describes the format of the guid field. If present, this UUID may be used to distinguish one server address from another.

So this will only affect busd but even then, I'm not convinced it needs to. At least at the moment, it only supports 1 transport per run.

zeenix: If busd can't properly support listening on multiple sockets at the same time until the next zbus API break, I really don't mind at all.

zvariant/src/lib.rs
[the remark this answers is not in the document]
elmarco: I digged a bit: https://github.com/rust-lang/rust/issues/113160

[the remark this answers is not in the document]
zeenix: ah, so a regression in Rust. Even better. :)

zbus/src/address.rs
elmarco: > So this will only affect busd but even then, I'm not convinced it needs to. At least at the moment, it only supports 1 transport per run.

No, the server should generate it. The client should parse it and distinguish different connections.

zvariant/src/lib.rs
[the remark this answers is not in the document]
elmarco: They don't seem to see it as a regression. We should have a solution in zbus. Bumping the default recursion limit shouldn't change anything.

zbus/src/address.rs
elmarco: What is confusing exactly? Can you suggest a different documentation then?

zeenix: > What is confusing exactly?

It's not clear that the actual path is part of the address. Reading the docs, you get the impression that it's just random. Hence my initial question/confusion.

zbus/src/address.rs
zeenix: > No, the server should generate it. The client should parse it and distinguish different connections.

should, yes. Please address my points above, specifically if this really needs to be addressed before 4.0. See #395 for another reason, we might want an API break soon.

zbus/src/lib.rs
elmarco: I don't see much point in doing all modules at the same time. It's a bit easier for case by case instead. But it's easy to move this patch to a separate Draft PR if you prefer.

zbus/src/address.rs
elmarco: Not sure either what to do here. Is this required before 4.0? No. Does this require an API break? No. Can we make it fit in the current Address? Not without breaking every user of Address by having new variants for all existing variants...

You seem to dislike the intermediary address::Parsed, while it offers the required flexibility for extra key=value handling. What to do? If the naming isn't good, let's call it ServerAddress ? we can then rename Address->Transport and alias/deprecate Address. That would fit well with the spec.

zeenix: > Not sure either what to do here.

You answer yourself in the next para. :)

> You seem to dislike the intermediary address::Parsed

Indeed. As I tried to explain, it's totally unnecessary, especially if we can break API. 

> If the naming isn't good

Naming is just a symptom. The real problem is that it's unncessary. It's especially bad because it's exposed to public API. Why does the user need to care? They give you a string and you parse it to an address. Then they use the Address to create a Connection.

zbus/src/lib.rs
zeenix: > I don't see much point in doing all modules at the same time.

consistency.

zbus/src/address.rs
zeenix: > I iterated, are you ok with the current proposal? If not, what do you propose?

I was on vacation and now at GUADEC so didn't get a chance to look at your code yet. If you have addressed this, please resolve the thread. I can always re-open if needed.

zeenix:
@elmarco There are a bunch of things I need to do besides attending talks at GUADEC, so once you've addressed the conversations above, please resolve them and also rebase to resolve conflicts and then I can do another round of review.

zbus/src/lib.rs
elmarco: see #410

zbus/src/address.rs
elmarco: See  [♻️ zv: rename Parsed->ServerAddress and Address->Transport](https://github.com/dbus2/zbus/pull/390/commits/f5b43a9c537a482c6f3d9eacb394145273d37a24)

elmarco:
> @elmarco There are a bunch of things I need to do besides attending talks at GUADEC, so once you've addressed the conversations above, please resolve them and also rebase to resolve conflicts and then I can do another round of review.

If think we should first merge #410 before I rebase. But you could still review the ServerAddress change proposal.

zeenix:
> But you could still review the ServerAddress change proposal.

I already nack'ed that. As I explained above, the naming is not the main issue.

zeenix:
> > I already nack'ed that. As I explained above, the naming is not the main issue.
> 
> yet you don't provide alternative.

I most certainly did: Put the guid in the Address enum. Address should have all the properties of each variant. We do not need generic key/value properties. New properties are rarely (if at all) added but even if we want to be safe, we can add struct for each variant, with properties in private fields with getters. That way, we can always add new properties handling w/o breaking any API.

elmarco:
> > > I already nack'ed that. As I explained above, the naming is not the main issue.
> > 
> > 
> > yet you don't provide alternative.
> 
> I most certainly did: Put the guid in the Address enum. Address should have all the properties of each variant. We do not need API for generic key/value properties. New properties are rarely (if at all) added but even if we want to be safe, we can add struct for each variant, with properties in private fields with getters. That way, we can always add new properties handling w/o breaking any API.

This is wrong design. guid is not transport-related, we shouldn't mix it there and repeat it for each transport type. Furthermore, you break API by doing so (even if you don't seem to agree on that).

zeenix:
> This is wrong design. guid is not transport-related, we shouldn't mix it there and repeat it for each transport type.

Forgot that it applies to all variants. I really don't mind the duplication (as enum will only be one variant at a time and hence the guid string itself will not need to be duplicated) but if you dislike this duplication, we could add an `Address` struct and includes the existing enum (w/ different name) as one of the fields.

> This is wrong design.

Actually we're already doing that for some of the variants (e.g `Tcp`).

> Furthermore, you break API by doing so (even if you don't seem to agree on that).

Let's say that's true but that doesn't matter anymore with main branch now in API break phase.

elmarco:
> > This is wrong design. guid is not transport-related, we shouldn't mix it there and repeat it for each transport type.
> 
> Forgot that it applies to all variants. I really don't mind the duplication (as enum will only be one variant at a time and hence the guid string itself will not need to be duplicated) but if you dislike this duplication, we could add an `Address` struct and include the existing enum (w/ different name) as one of the fields.

That's almost what I propose, ServerAddress has a tranport() -> Transport enum. It is done lazily, so you don't even have to parse it if not needed. The canonical form of a ServerAddress is key=value pairs, following the spec.

What is missing is a way to build a complete ServerAddress (from a transport, an optional guid), and perhaps tweak the fields by accessing the underlying hashmap.

The alternative is to make Transport a simple enum. But then there is no required fields (keys) enforcement at compile time. I would rather keep the current Transport enum.

> 
> > This is wrong design.
> 
> Actually we're already doing that for some of the variants (e.g `Tcp`).
> 

not really, we have a variant for each transport in https://dbus.freedesktop.org/doc/dbus-specification.html#transports

> > Furthermore, you break API by doing so (even if you don't seem to agree on that).
> 
> Let's say that's true but that doesn't matter anymore with main branch now in API break phase.

I don't think we want to bother working users out there with an extra _ to ignore...

zeenix:
> That's almost what I propose, ServerAddress

Huh? There are major diffs:

* That exposes a separation of Server and Client Address. 
* It keeps untyped data and generic key/value pairs.

> > Actually we're already doing that for some of the variants (e.g `Tcp`).
> 
> not really, we have a variant for each transport in https://dbus.freedesktop.org/doc/dbus-specification.html#transports

Not sure I follow. I was talking about our `Address` enum, where `Tcp` variant has a `TcpAddress` struct that can add new information in the future w/o breaking API. I was suggesting we do the same for all variants.

> I don't think we want to bother working users out there with an extra _ to ignore...

As I keep reminding you, the enum is declared `nonexhaustive` already so we're not asking users to add `_` now`, that's [already the case](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute). Talking of which, it seems even adding new fields in a struct variant of a non-exhaustive enum doesn't break existing code.

Moreover, I'm pretty sure direct use of `Address` API is an extremely niche use case and hence I don't mind breaking existing code for this part, if it means having a good simple API.

zeenix:
> > That's almost what I propose, ServerAddress
> 
> Huh? There are major diffs:
> 
>     * It exposes a separation of Server and Client Address.
> 
>     * It keeps untyped data and generic key/value pairs.

I think the key take here is that I would like to go straight from/to string to/from `Address`, no intermediate repr needed and desired.

elmarco:
> > That's almost what I propose, ServerAddress
> 
> Huh? There are major diffs:
> 
>     * It exposes a separation of Server and Client Address.

There is no such thing as "Client Address". See also spec "**Server Addresses**" https://dbus.freedesktop.org/doc/dbus-specification.html#addresses

> 
>     * It keeps untyped data and generic key/value pairs.
 
No, transport() returned a strongly typed and opinionated enum with the required fields, following the spec "**Transports**" https://dbus.freedesktop.org/doc/dbus-specification.html#transports.

> > > Actually we're already doing that for some of the variants (e.g `Tcp`).
> > 
> > 
> > not really, we have a variant for each transport in https://dbus.freedesktop.org/doc/dbus-specification.html#transports
> 
> Not sure I follow. I was talking about our `Address` enum, where `Tcp` variant has a `TcpAddress` struct that can add new information in the future w/o breaking API. I was suggesting we do the same for all variants.
> 
> > I don't think we want to bother working users out there with an extra _ to ignore...
> 
> As I keep reminding you, the enum is declared `nonexhaustive` already so we're not asking users to add `_` now`, that's [already the case](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute). Talking of which, it seems even adding new fields in a struct variant of a non-exhaustive enum doesn't break existing code.
>

I also repeat myself: if we don't use the exisiting variants at all anymore (and duplicate them), it's an even worse break, as existing client code will compile and always run-time fall under the `_` unhandled case.
 
> Moreover, I'm pretty sure direct use of `Address` API is an extremely niche use case and hence I don't mind breaking existing code for this part, if it means having a good simple API.

I am moving that Address API change to a separate PR.

zbus/src/address.rs
elmarco: added

elmarco:
@zeenix should I merge?

zvariant/src/signature.rs
```diff
@@ -433,25 +433,41 @@ impl<'de> Visitor<'de> for SignatureVisitor {
...
 }
 
-fn ensure_correct_signature_str(signature: &[u8]) -> Result<()> {
-    if signature.len() > 255 {
-        return Err(serde::de::Error::invalid_length(
-            signature.len(),
-            &"<= 255 characters",
-        ));
-    }
+struct SignatureIter<'a> {
```
zeenix, line 453: why not just implement `Iterator` for `SigntureParser` itself?

zvariant/src/signature.rs
```diff
@@ -288,6 +293,18 @@ impl<'a> Signature<'a> {
 
         clone
     }
+
+    /// Returns the number of complete types for the signature.
+    ///
+    /// **Note**: if the signature is invalid, returns the first error.
```
zeenix, line 299: Let's keep consistency with Rust's conventions and mention we mean top-level types:

```suggestion
    /// The number of complete top-level signature in `self`.
    ///
    /// # Errors
    /// 
    /// If the signature is invalid, returns the first error.
```

zvariant/src/signature.rs
```diff
@@ -288,6 +293,18 @@ impl<'a> Signature<'a> {
...
+    pub fn n_types(&self) -> Result<usize> {
```
zeenix, line 300: I think `n_signatures` would be more appropriate.

zbus/src/xml.rs
```diff
@@ -183,7 +185,7 @@ pub struct Property<'a> {
     #[serde(rename = "@name", borrow)]
     name: MemberName<'a>,
 
-    #[serde(rename = "@type")]
+    #[serde(deserialize_with = "de_type_valid", rename = "@type")]
```
zeenix, line 188: > We could have used zv:Signature instead of String here, I guess.

Why don't we switch now? Remember, we're in API break and this seems like a good candidate for breakage.

zvariant/src/value.rs
```diff
@@ -985,6 +985,7 @@ mod tests {
                 (@mmn nothing, @mmmn just nothing))"
         );
 
+        #[cfg(unix)]
```
zeenix, line 988: Why didn't our CI pick it up?

elmarco: `cargo test --package zbus --features windows-gdbus` ?

elmarco: `cargo test --features=zbus/windows-gdbus` should solve this

zvariant/src/signature.rs
elmarco: Hmm? you confused me even more. I would assume a Rust "top-level type" is one exposed at the top-level of a namespace... Let's follow the D-Bus spec here. Do not mix Rust types with D-Bus.

zeenix: I'm not confusing them. I'm talking about D-Bus types. If this method returns 2 for `uas`, that means it's only counting the top-level signatures, no?

elmarco: This is not the spec jargon. it's called "complete types"

zvariant/src/signature.rs
zeenix: the spec is not thinking about the context. Here, reader would think you're talking about Rust types.

zvariant/src/signature.rs
elmarco: why not?

zvariant/src/signature.rs
elmarco: No.. it's a dbus signature. Nothing to do with Rust. If you want we refer to the spec from here.

zbus/src/xml.rs
elmarco: done, with the introduction of CompleteType (which I think should be used in many places for correctness)

zvariant/src/signature.rs
elmarco: ok with the current version? would you rename it n_complete_types() instead?

zvariant/src/signature.rs
zeenix: ok. What about rest of the changes suggested for the doc comment?

zvariant/src/signature.rs
zeenix: `n_complete_types` definitely seems better IMO.

zvariant/src/signature_parser.rs
```diff
@@ -30,6 +30,25 @@ impl<'s> SignatureParser<'s> {
...
 
+    pub fn from_bytes(signature: &'s [u8]) -> Result<Self> {
+        if signature.len() > 255 {
+            return Err(serde::de::Error::invalid_length(
+                signature.len(),
+                &"<= 255 characters",
+            ));
+        }
+
+        let signature = unsafe { Signature::from_bytes_unchecked(signature) };
```
zeenix, line 41: `SAFETY` comment would be good here.

zvariant/src/signature_parser.rs
```diff
@@ -30,6 +30,25 @@ impl<'s> SignatureParser<'s> {
         }
     }
 
+    pub fn from_bytes(signature: &'s [u8]) -> Result<Self> {
```
zeenix, line 33: I think the whole method should be marked as unsafe as the safety here depends on the caller. Only thing this method is checking is the length of the signature.

zvariant/src/signature.rs
elmarco: done

zvariant/src/signature_parser.rs
elmarco: ok, actually we don't need to make it pub either (well it's internal anyway, right)

elmarco: ok

zvariant/src/signature.rs
```diff
@@ -288,6 +293,21 @@ impl<'a> Signature<'a> {
 
         clone
     }
+
+    /// Returns the number of complete types for the signature.
```
zeenix, line 298: nit: `Returns` is redundant here and afaik inconsistent with general conventions. Hence why I removed it in my suggestion.

zbus/src/xml.rs
zeenix: You mean you change this code to use the new type in a later commit? Why not move the commit that introduces this type before this one and update this commit to use it already? It would have made review easier too.

zvariant/src/signature.rs
```diff
@@ -361,6 +362,18 @@ impl<'a> TryFrom<&'a str> for Signature<'a> {
     }
 }
 
+/// Try to create a Signature from a `Cow<str>.`
+impl<'a> TryFrom<Cow<'a, str>> for Signature<'a> {
```
zeenix, line 366: I think the new impls here deserve their own commits.

zvariant/src/signature.rs
```diff
@@ -511,6 +511,47 @@ impl std::fmt::Display for OwnedSignature {
...
+    }
+}
+
+impl<'a> Display for CompleteType<'a> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        std::fmt::Display::fmt(&self.0.as_str(), f)
+    }
+}
+
+impl<'a> TryFrom<Signature<'a>> for CompleteType<'a> {
```
zeenix, line 533: would be great to provide:

* `TryFrom<&str>` and `TryFrom<String>`.
* `Deref<Target = Signature>`.
* `From<Signature> for CompleteType`.

zvariant/src/signature.rs
```diff
@@ -511,6 +511,47 @@ impl std::fmt::Display for OwnedSignature {
     }
 }
 
+/// [`Signature`] that identifies a complete type.
```
zeenix, line 514: Let's keep this new type in a separate mod.

zbus/src/address.rs
```diff
@@ -987,7 +987,7 @@ mod tests {
                 port: 2934
             })
             .to_string(),
-            "vsock:cid=98,port=2934,guid=123",
+            "vsock:cid=98,port=2934", // no support for guid= yet..
```
zeenix, line 990: nit: commit log summarizes the benefit, not the change itself. We ask people to [follow these rules](https://wiki.gnome.org/Git/CommitMessages) so we should follow them too.

zbus/src/connection/handshake.rs
```diff
@@ -988,6 +988,7 @@ impl<S: Socket> HandshakeCommon<S> {
         let line = std::str::from_utf8(line_bytes.as_slice())
             .map_err(|e| Error::Handshake(e.to_string()))?;
 
+        trace!("Reading {line}");
```
zeenix, line 991: * Can you append "in handshake code" in the commit log for it to be more specific?
* Just FYI, Lennart was surprised that we don't optimize the handshake by sending over all lines at once, predicting the server's response. I am hoping to look into that soon.

zvariant/src/value.rs
zeenix, line 988: The commit log here "zv: fix build on windows" is also summarizing the benefit and not the change.

.github/workflows/rust.yml
```diff
@@ -101,36 +101,12 @@ jobs:
...
-            c:/bin/*z*
-            c:/bin/*ffi*
-            c:/bin/*intl*
-          key: ${{ runner.os }}-cache
+            c:/bin
+            c:/lib
+            c:/share
+            c:/var
+            c:/ProgramData/chocolatey
+          key: ${{ runner.os }}-cache-2023-08-02.2
```
zeenix, line 109: obviously at least one separate logical change here but commit log doesn't even summarize it.

zeenix (changes requested):
Mostly nits left.

zvariant/src/signature.rs
elmarco: ok

zvariant/src/value.rs
elmarco: done

zbus/src/connection/handshake.rs
elmarco: ok

zbus/src/address.rs
elmarco: ok

zvariant/src/signature.rs
elmarco: done

zvariant/src/signature.rs
elmarco: ok

zvariant/src/signature.rs
elmarco: they are not needed at this point.
"Deref should only be implemented for smart pointers to avoid confusion." iow, we shouldn't abuse it. Perhaps we do.

elmarco: From<Signature> can't be implemented, as it may throw errors. Use TryFrom.

elmarco: I wonder if Rust is not smart enough to figure out you can go try_from &str/String -> Signature -> CompleteType already (is it transitive?).

zeenix: > "Deref should only be implemented for smart pointers to avoid confusion."

Where does it say that?

> we shouldn't abuse it. Perhaps we do.

I don't think so. I think it's good for wrapper types like these.

> I wonder if Rust is not smart enough to figure out you can go try_from &str/String -> Signature -> CompleteType already (is it transitive?).

I doubt it. It's not always desired.

zeenix: > > I wonder if Rust is not smart enough to figure out you can go try_from &str/String -> Signature -> CompleteType already (is it transitive?).
> 
> I doubt it. It's not always desired.

Or maybe. I don't know. :shrug:

zbus/src/address.rs
zeenix: You added details about the error being fixed (which is really nice) but did not address my comment. Something like:

```
zb,test: Don't expect guid in vsock address

<what you have right now>
```

zvariant/src/signature.rs
elmarco: Where does it say that? -> https://doc.rust-lang.org/std/ops/trait.Deref.html#

.github/workflows/rust.yml
zeenix, line 109: "Bump the key manually to force a cache rebuild." doesn't cover the change of caching renaming and also best be split in a separate commit.

zeenix (dismissed):
We're getting close to merge. :)

zvariant/src/value.rs
elmarco: ok

.github/workflows/rust.yml
```diff
@@ -101,36 +101,12 @@ jobs:
...
-            c:/bin/*gresource*
-            c:/bin/*pcre*
-            c:/bin/*z*
-            c:/bin/*ffi*
-            c:/bin/*intl*
-          key: ${{ runner.os }}-cache
+            c:/bin
+            c:/lib
+            c:/share
+            c:/var
```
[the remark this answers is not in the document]
elmarco: Time change isn't noticeable to me. Oh yeah we go from `Cache Size: ~8 MB (8797929 B)` -` Cache Size: ~1930 MB (2023337918 B)`, but at 61.9 MBs/sec that's like 30s.

Where is the cache being managed btw? How often is it purged? What is the limit? How do you inspect it? I wonder if it's chocolatey that takes that much space..

Alternatively, we can build pkgconfig again? no idea...

.github/workflows/rust.yml
elmarco: well it doesn't make much sense to split the change of the cache naming, since I want a new cache with the change, but ok..

.github/workflows/rust.yml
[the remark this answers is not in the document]
elmarco: let see if I can bring that down a bit... not sure it's chocolatey, but 2G for what we built really?

zvariant/src/signature.rs
elmarco: I leave that nice to have out for now.

.github/workflows/rust.yml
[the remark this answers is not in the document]
elmarco: Cache Size: ~10 MB (10788786 B)

Issue #286, opened 2022-11-03
Port file descriptor related API to Rust's safe-io


zeenix:
In GitLab by @msandova on Nov 3, 2022, 19:58

It would be interesting to make use of the newer safe-io APIs in Rust. It in not clear to what extent one can do this without breaking API.

PR #395, opened 2023-07-01
Add safe-io feature

A6GibKm:
Adds a new `safe-io` feature that adds types `zvariant::OwnedFd` and `zvariant::BorrowedFd<'a>`. Introduced a big api break so it is gated behind a feature flag.

Fixes: https://github.com/dbus2/zbus/issues/286

### DIFF
--- a/zvariant/src/fd.rs
+++ b/zvariant/src/fd.rs
@@ -1,14 +1,16 @@
-use serde::{de::Error, Deserialize, Deserializer, Serialize, Serializer};
+use serde::{ser::Error, Deserialize, Deserializer, Serialize, Serializer};
 use static_assertions::assert_impl_all;
-use std::os::unix::io;
+use std::{
+    cmp::PartialEq,
+    os::unix::io::{AsFd, AsRawFd, BorrowedFd, FromRawFd, IntoRawFd, OwnedFd, RawFd},
+    sync::Arc,
+};
 
 use crate::{Basic, EncodingFormat, Signature, Type};
 
-/// A [`RawFd`](https://doc.rust-lang.org/std/os/unix/io/type.RawFd.html) wrapper.
+/// A [`OwnedFd`](https://doc.rust-lang.org/std/os/fd/struct.OwnedFd.html) wrapper.
 ///
-/// See also `OwnedFd` if you need a wrapper that takes ownership of the file.
-///
-/// We wrap the `RawFd` type so that we can implement [`Serialize`] and [`Deserialize`] for it.
+/// We wrap the `OwnedFd` type so that we can implement [`Serialize`] and [`Deserialize`] for it.
 /// File descriptors are serialized in a special way and you need to use specific [serializer] and
 /// [deserializer] API when file descriptors are or could be involved.
 ///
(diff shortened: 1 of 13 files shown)


zeenix:
@A6GibKm Hey, sorry I haven't managed to look into this. Two things:

1. The last commit here is marked WIP. Is that still the case?
2. I'm not sure about feature-gating. I doubt most people will even figure out the new feature so they won't benefit from this. We might want to just do a 4.0 release and break the API. There are other stuff that can use a break.

A6GibKm:
The last commit is wip since i didnt know how to fix it in an elegant way, but removing the feature would fix this.

A6GibKm:
I removed the safe-io feature, and the WIP comment.

zeenix:
@A6GibKm Thanks so much. This needs to be coordindated though with other changes that require breaking changes.

zeenix:
@A6GibKm CI fails. :(

A6GibKm:
whoosp, forgot to format with `+nightly`, it should be fixed now.

A6GibKm:
Following discussions at guadec, I removed Value::OwnedFd but one problem that arises is that it is not possible to do fd.to_owned and get a clone of the reference with static lifetime in a safe way since the lifetime is tied to the original owned descriptor, which once dropped will close the underlying descriptor.

zeenix:
> Following discussions at guadec, I removed Value::OwnedFd but one problem that arises is that it is not possible to do fd.to_owned and get a clone of the reference with static lifetime in a safe way since the lifetime is tied to the original owned descriptor, which once dropped will close the underlying descriptor.

Right, after more thing and offline discussion, I think I'm leaning towards:

* `Value::Fd` keeps our own enum, `Fd`.
* `Fd` enum is like this:
```rust
pub enum Fd<'f> {
   // Need an `Arc` cause we want `Clone` and it be a shallow clone.
   Owned(Arc<OwnedFd>),
   Borrowed(BorrowedFd<'f>),
}
```
* `Value::to_owned` becomes fallible and we document the error cases (from the `dup` manpage).

WDYT?

zeenix:
> WDYT?

@elmarco your input would be greatly appreciated here.

elmarco:
I don't understand the problem. Why is it an issue that Value.to_owned() returns with a clone of `Fd(Arc<OwnedFd>)` ?

zeenix:
> I don't understand the problem. Why is it an issue that Value.to_owned() returns with a clone of `Fd(Arc<OwnedFd>)` ?

`to_owned` issue was only relevant for the case where `Value` can have `BorrowedFd` too. We already moved on from that idea. Sorry, didn't update the PR with subsequent discussions offline and on matrix channel.

The latest issue is `Arc<OwnedFd>` not being useful in APIs like `std::fs::File::from`. The user has to use `Ownedfd::try_clone` (that `dup`s the underlying fd internally) first.

zeenix:
@A6GibKm I know this became complicated but could you finish it soon please?

A6GibKm:
Ok, so I think this requires

- Documentation
- Check whether the structs make sense, there is a thin wrapper for `Arc<Fd>` but might not be needed at all
- We don't expose a wrapper for BorrowedFd since it only makes sense to send/receive a fd together with ownership
- Are macros into_value, try_from_value, ov_try_from, etc correctly used?

A6GibKm:
I don't remember if we discussed that, but I guess one can always deref the Arc and/or clone again the Arc.

zvariant/src/fd.rs
```diff
@@ -1,155 +1,134 @@
...
-impl Serialize for OwnedFd {
+impl Serialize for FdWrapper {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: Serializer,
     {
-        serializer.serialize_i32(self.inner)
+        let dupped = self.0 .0.try_clone().map_err(Error::custom)?;
+        let raw_fd = dupped.into_raw_fd();
+        serializer.serialize_i32(raw_fd)
```
MaxVerevkin, line 99: Looks like this code leaks fds?

Also, I'm not familiar with the way zvariant handles fds, but what is the point of serializing and fd as i32? Seems unnecessary.

zeenix: > Looks like this code leaks fds?
> 
> Also, I'm not familiar with the way zvariant handles fds, but what is the point of serializing and fd as i32? Seems unnecessary.

To send fds across over D-Bus. It's part of the spec.

MaxVerevkin: Ah, I dug into Serializer implementation a little more, it knows if a given i32 is an fd by looking at the signature, nvm :facepalm:

MaxVerevkin: Anyway, it's unfortunate that it is not possible to pass `Arc<Fd>` directly to the serializer without duplicating the fd.

MaxVerevkin: I mean, it is possible to clone the `Arc`, call `into_raw()`, and pass the pointer as `u64`...

zeenix: > I mean, it is possible to clone the `Arc`, call `into_raw()`, and pass the pointer as `u64`...

Yeah. I think we had it all figured out in the end after many discussions at GUADEC.

zeenix:
Superseded by #508.

PR #397, opened 2023-07-03
Monitoring proxy improvements

zeenix:
This proxy was obviously added in a hurry, w/o much thought given but we also didn't have the `MatchRule` type then to use.

Since 2 commits here are breaking API, I will contact the [only](https://github.com/azymohliad/watchmate/blob/3fb3fa89fc76a783f46abdc5f07e9a59ed23ac1f/infinitime/src/freedesktop/notifications.rs#L32) [2](https://github.com/tau-OS/kiri-desktop/blob/97b86e48731e30ba14572b16aec864cdbfa0a171/d5/src/notify.rs#L102) users of this API I could find on Github and see if they'd mind us breaking API for this.

### DIFF
--- a/zbus/src/fdo.rs
+++ b/zbus/src/fdo.rs
@@ -390,7 +390,11 @@ macro_rules! gen_monitoring_proxy {
         trait Monitoring {
             /// Converts the connection into a monitor connection which can be used as a
             /// debugging/monitoring tool.
-            fn become_monitor(&self, n1: &[&str], n2: u32) -> Result<()>;
+            fn become_monitor(
+                &self,
+                match_rules: &[crate::MatchRule<'_>],
+                flags: u32,
+            ) -> Result<()>;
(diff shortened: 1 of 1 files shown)


zeenix:
> Since 2 commits here are breaking API, I will contact the [only](https://github.com/azymohliad/watchmate/blob/3fb3fa89fc76a783f46abdc5f07e9a59ed23ac1f/infinitime/src/freedesktop/notifications.rs#L32) [2](https://github.com/tau-OS/kiri-desktop/blob/97b86e48731e30ba14572b16aec864cdbfa0a171/d5/src/notify.rs#L102) users of this API I could find on Github and see if they'd mind us breaking API for this.

I contacted them both and got [a green light](https://github.com/tau-OS/kiri-desktop/issues/3#issuecomment-1618078971) from them for this change.

PR #404, opened 2023-07-11
WIP: Upgrade syn to 2.0

jplatte:
Didn't convert the huge `macro_rules!` macro and some non-trivial `NestedMeta` usage. I'd rather just convert it to parsing in one step like I did for `match_attribute_with_str_list_value`, but (a) I expected this to be a short contribution and have run out of time for now and (b) I'm not sure this is even what you want, so let's discuss how to proceed from this draft PR.

### DIFF
--- a/zbus_macros/Cargo.toml
+++ b/zbus_macros/Cargo.toml
@@ -21,7 +21,7 @@ proc-macro = true
 
 [dependencies]
 proc-macro2 = "1.0"
-syn = { version = "1.0.103", features = ["extra-traits", "fold", "full"] }
+syn = { version = "2.0.25", features = ["extra-traits", "fold", "full"] }
 quote = "1.0.21"
 proc-macro-crate = "1.2.1"
 regex = "1.6.0"
--- a/zbus_macros/src/proxy.rs
+++ b/zbus_macros/src/proxy.rs
@@ -4,7 +4,7 @@ use quote::{format_ident, quote, quote_spanned, ToTokens};
 use regex::Regex;
 use syn::{
     self, fold::Fold, parse_quote, spanned::Spanned, AttributeArgs, Error, FnArg, Ident, ItemTrait,
-    ReturnType, TraitItemMethod,
+    ReturnType, TraitItemFn,
 };
 use zvariant_utils::{case, def_attrs};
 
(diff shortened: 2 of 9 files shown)


zeenix:
@bilelmoussaoui any chance you can take over?

zeenix:
> I'm not sure this is even what you want,

Can you please elaborate on this?

jplatte:
I'm talking about [this](https://github.com/dbus2/zbus/pull/404/files#diff-fa48bc86b0568c145bf18d728a60539ad74dcc18727284ea8e1a0fb89b7e6011L85-L104) change. It impacts error messages, e.g. supplying `attribute_argument = true` for an `attribute_argument` that expects a string value would previously result in a compilation error saying "invalid literal type for `attribute_argument` attribute", and now will likely say something like "syntax error, expected `"`" (pointing to `true`). Are you okay with error reporting changing to syn's parse errors, from the custom errors based on a parse result?

zeenix:
Thanks for explaining. :+1: 

> Are you okay with error reporting changing to syn's parse errors, from the custom errors based on a parse result?

Yeah. The new error seems more descriptive even. My preference would always been what's most helpful to the user.

zbus_macros/src/iface.rs
```diff
@@ -104,9 +108,27 @@ pub fn expand(args: AttributeArgs, mut input: ItemImpl) -> syn::Result<TokenStre
...
-    for method in &mut input.items {
-        let method = match method {
-            ImplItem::Method(m) => m,
+    for item in &mut input.items {
+        let (method, is_signal) = match item {
+            ImplItem::Fn(m) => (m, false),
+            // Since signals do not have a function body, they don't parse as ImplItemFn…
+            ImplItem::Verbatim(tokens) => {
+                // … thus parse them ourselves and construct an ImplItemFn from that
+                let decl = syn::parse2::<ImplItemSignal>(tokens.clone())?;
```
jplatte, line 117: This was the hardest piece to find out. Apparently syn 1.x just accepted `ImplItemMethod`s without a body, but syn 2.x does not. The error message when re-emitting this `ImplItem::Verbatim` was plenty confusing (it was "expected `impl`" on the first token of the item).

PR #407, opened 2023-07-16
⚡  zvariant, zbus: use borrowed clone if possible

Kijewski:
This PR implements `Signature::as_ref()` to get a borrowed clone of a `Signature`. The code is a copy of `Str::as_ref()`.
  
Even though cloning an `Arc` is cheap, it's not exactly for free. An atomic increment has some inherent overhead, and `Arc::clone()` also `panic!()`s if there are excessively many strong references to an `Arc`. For every `panic!()` some code to unwind the stack has to be generated, which hurts cache locality, and bloats the generated code.


zeenix (approved):
Thanks! Looks good to me and kudos for providing nice atomic commits with details in the log. :+1:

PR #408, opened 2023-07-23
➖ zvariant: remove libc dependency

Kijewski:
This PR implements `OwnedFd` using the built-in `std::os::fd::OwnedFd` without the use of `libc::close()` or `dup()`. `std::os::fd::OwnedFd` is only stable since rust 1.63, but zbus' MSRV is 1.64 now.

Also I don't think there is a need to assume that `RawFd` is an `i32`.

### DIFF
--- a/zvariant/Cargo.toml
+++ b/zvariant/Cargo.toml
@@ -30,7 +30,6 @@ enumflags2 = { version = "0.7.7", features = ["serde"], optional = true }
 zvariant_derive = { version = "=3.15.0", path = "../zvariant_derive" }
 serde_bytes = { version = "0.11", optional = true }
 static_assertions = "1.1.0"
-libc = "0.2.137"
 uuid = { version = "1.2.1", features = ["serde"], optional = true }
 url = { version = "2.3.1", features = ["serde"], optional = true }
 time = { version = "0.3.16", features = ["serde"], optional = true }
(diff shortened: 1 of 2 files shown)


Kijewski:
Hm, [`std::os::fd::OwnedFd`](https://doc.rust-lang.org/stable/std/os/fd/struct.OwnedFd.html) is stable since 1.63, but [`std::os::fd`](https://doc.rust-lang.org/stable/std/os/fd/index.html) is only stable since 1.66? I'm confused.

MaxVerevkin:
> Hm, [`std::os::fd::OwnedFd`](https://doc.rust-lang.org/stable/std/os/fd/struct.OwnedFd.html) is stable since 1.63, but [`std::os::fd`](https://doc.rust-lang.org/stable/std/os/fd/index.html) is only stable since 1.66? I'm confused.

`std::os::fd` is just a new path. You can use `std::os::unix::io::OwnedFd`.

Kijewski:
@MaxVerevkin, ah, thanks! That does the trick. I think the docs could be a little more explicit, though. :)

MaxVerevkin:
> I think the docs could be a little more explicit, though. :)

Yeah. It's also annoying that rust-analyzer autoimports the new path.

zeenix:
@Kijewski Many thanks for this. Before I look into the changes, did you see #395 ?

PR #409, opened 2023-07-23
➖ zbus: use less `async-recursion`

Kijewski:
A recursive depth-first search can be implemented using a deque.


zeenix (approved):
Nice! Wish we could just drop the dep completely. I'm planning on zvariant/zbus 4.0 soon with all the breaking changes we need. I think we can remove this dep completely if we changed the Address to directly parse to the value we need? If you've time and motivation to work on that, please do provide a PR for that. `main` branch will soon be breaking API.

PR #410, opened 2023-07-24
Some code shuffling

elmarco:
Start cleaning up a bit the top namespace. Currently, only top-level address-related types are marked as deprecated. The other types are simply re-exported. This helps already organizing cargo doc into sections.


elmarco:
@zeenix, anything else required here?

zbus/src/lib.rs
```diff
@@ -59,7 +59,7 @@ pub type Address = address::Address;
 mod guid;
 pub use guid::*;
 
-mod message;
+pub mod message;
 pub use message::*;
```
zeenix, line 63: I think we should also provide deprecated export at the top-level for message mod hierarchy.

zbus/src/match_rule/mod.rs
```diff
@@ -8,9 +8,12 @@ use zvariant::Structure;
 use crate::{
     names::{BusName, InterfaceName, MemberName, UniqueName},
     zvariant::{ObjectPath, Str, Type},
-    Error, MatchRuleBuilder, MessageType, Result,
+    Error, MessageType, Result,
 };
 
+mod builder;
+pub use builder::MatchRuleBuilder;
```
zeenix, line 16: The actual builder struct should no longer need the type prefix (e.g this should be just `match_rule::Builder`. For backwards compatibility we should provide an alias with existing name in the root.

zbus/src/blocking/proxy_builder.rs
```diff
@@ -6,7 +6,7 @@ use zvariant::ObjectPath;
 
 use crate::{blocking::Connection, utils::block_on, CacheProperties, Error, Result};
 
-pub use crate::ProxyDefault;
+pub use crate::proxy::ProxyDefault;
```
zeenix, line 9: OT but I was discussing this type and proxy building API in general with someone and I think the current API is a bit weird. I think user should always be able to use the builder pattern, whether they are using the macro-generated proxies or `Proxy`: `Proxy::builder()`. The macros can add the constants into generated proxy impl directly and we can ditch this type (no need to even deprecate since I seriously doubt anyone uses the trait itself) and unify `Builder::{new, new_bare}` as `Builder::new`.

zbus/src/lib.rs
```diff
@@ -86,7 +86,7 @@ pub use match_rule::*;
 
 mod socket_reader;
 
-mod proxy;
+pub mod proxy;
 pub use proxy::*;
```
zeenix, line 84: deprecated export/aliases here too. These changes are supposed to be helping users (not just us devs) and provide them with a Rusty API/patterns.

zbus/src/lib.rs
```diff
@@ -85,6 +83,9 @@ mod socket_reader;
 pub mod proxy;
 pub use proxy::*;
 
+pub mod object_server;
+pub use object_server::*;
```
zeenix, line 87: same here about deprecation in the root.

zbus/src/abstractions/async_drop.rs
```diff
@@ -10,6 +10,7 @@
 /// Hopefully this will be unnecessary [in the future][itf] when Rust gain an `AsyncDrop` itself.
 ///
 /// [itf]: https://rust-lang.github.io/async-fundamentals-initiative/roadmap/async_drop.html
+#[doc(hidden)]
```
zeenix, line 13: This is not supposed to be private at all. Elsewhere in the docs, you'll even find recommendation for using it.

zbus/src/utils.rs
```diff
@@ -17,6 +17,7 @@ pub(crate) fn padding_for_n_bytes(value: usize, align: usize) -> usize {
 /// type alias.  This is currently required because the macros for properties expect a Result
 /// return value, but the macro-generated `receive_` functions need to refer to the actual
 /// type without the associated error.
+#[doc(hidden)]
```
zeenix, line 20: This is not supposed to be private either. Read the docs. :)

zeenix (changes requested):
Really good direction in general and have been on my todo. Thanks for doing it. Need changes though.

zbus/src/lib.rs
```diff
@@ -41,8 +41,20 @@ pub use dbus_error::*;
...
+pub mod address;
+
+#[cfg(any(
+    all(feature = "vsock", not(feature = "tokio")),
+    feature = "tokio-vsock"
+))]
+#[deprecated(note = "Use `address::VsockAddress` instead")]
+pub type VsockAddress = address::VsockAddress;
+#[deprecated(note = "Use `address::TcpAddress` instead")]
+pub type TcpAddress = address::TcpAddress;
```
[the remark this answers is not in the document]
elmarco: I think it's up to us.. not really sure that's necessary. For example, std::net has tcp stuff: https://doc.rust-lang.org/std/net/index.html

zbus/src/lib.rs
elmarco: ok

zbus/src/lib.rs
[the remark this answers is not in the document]
zeenix: True. Not sure, why i remembered wrong. :shrug: Since we're doing this, I think we might as well go all in and take full advantage of the module system.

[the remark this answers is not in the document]
zeenix: but I leave this to you. Please resolve this conversation if you think more strongly about it than me. :)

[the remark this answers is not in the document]
elmarco: also are you sure you want to replace something like Foo with foo::Foo ? (think Address, Message etc). I think the top user facing types could remain exported in the top namespace. Or we could use a prelude::* to help users...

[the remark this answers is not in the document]
zeenix: Hmm.. good point. We could move `Foo` under `foo` but still provide `Foo` on the top-level (w/o deprecation) but not for `foo::Bar`.

> Or we could use a prelude::* to help users...

I always failed to appreciate that pattern. Sure, it makes things easier for users but then it sacrifices readability too much IMO.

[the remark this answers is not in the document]
elmarco: yes, and you never now what is included or not in the prelude... it's just same a having it exported in the top-namespace imho in the end

zbus/src/lib.rs
zeenix: This is not yet resolved. We're exposing `Message*` on top-level afaict.

zbus/src/message/mod.rs
```diff
@@ -18,9 +18,23 @@ use crate::OwnedFd;
...
+};
+
+mod builder;
+pub use builder::MessageBuilder;
+
+mod field;
+pub use field::{MessageField, MessageFieldCode};
+
+mod fields;
+pub use fields::MessageFields;
```
zeenix: * with the new path, I really don't think we should ask user to type `Message` in the names if it's already part of the module. It should be `message::{Message, Builder, Field, FieldCode..}`.
* we need deprecation for top-level for `foo::Bar` (as agreed, I thought?).

elmarco: yes, this is not the latest series...

elmarco:
@zeenix removing the draft status, as I think this should be good enough to merge so we can spread/share the extra work.. the series is already pretty big and something tedious to handle, rebase etc, as you can imagine with mass renaming stuff etc..

zeenix:
> @zeenix removing the draft status,

Somehow it's still in Draft. :thinking:

zeenix (approved):
Since there are way too many commits, I did not do a per-commit review and trust that you ensured all commits are atomic and have appropriate log.

zeenix:
@elmarco so merge?

zbus/src/utils.rs
elmarco: actually, I think we could hide ResultAdapter, it's an implementation detail of the proxy macro. No?

zeenix: hmm.. yeah. so probably you were right to hide it (we've other apis too that are hidden and only `pub` because of macros).

PR #418, opened 2023-07-29
Some connection namespace moves

elmarco:


book/src/connection.md
```diff
@@ -73,7 +73,7 @@ let (client_conn, server_conn) = futures_util::try_join!(
 [`futures::stream::Stream`]: https://docs.rs/futures/latest/futures/stream/trait.Stream.html
 [`futures::sink::Sink`]: https://docs.rs/futures/latest/futures/sink/trait.Sink.html
 [`MessageStream`]: https://docs.rs/zbus/3/zbus/struct.MessageStream.html
-[`ConnectionBuilder::address`]: https://docs.rs/zbus/3/zbus/struct.ConnectionBuilder.html#method.address 
+[`Builder::address`]: https://docs.rs/zbus/3/zbus/struct.ConnectionBuilder.html#method.address
```
zeenix, line 76: The links will also need to be updated (you need to predict it based on a local build). Was this done for the other module changes? :thinking:

zbus/src/object_server/mod.rs
```diff
@@ -18,10 +18,11 @@ use zvariant::{ObjectPath, OwnedObjectPath, OwnedValue, Signature, Type, Value};
 
 use crate::{
     async_lock::{RwLock, RwLockReadGuard, RwLockWriteGuard},
+    connection::WeakConnection,
```
zeenix, line 21: maybe this should just be `connection::Weak`, similar to `std::rc::Weak` and `std::sync::Weak`?

zeenix (changes requested):
Nice job otherwise.

book/src/connection.md
elmarco: I am bad at predicting it though.. is there any way to update it or check it semi-automatically? yes, we will probably need to update all the links for v4 anyway..

zbus/src/object_server/mod.rs
elmarco: yes, I thought about it and forgot, let's do it

elmarco: nah that conflicts with other Weak types.. it's internal anyway, not a big deal.

book/src/connection.md
zeenix: `cargo doc --open` and then find the link?

elmarco: done

zeenix (approved):

PR #420, opened 2023-07-29
Drop serde-xml, rename quick-xml feature

elmarco:
My initial plan was to make `xml` a different module, but I think this is not a good idea as I plan to add some dependencies on zbus.


zeenix:
> My initial plan was to make `xml` a different module,

Do you mean sub-crate?

> as I plan to add some dependencies on zbus

Not sure I follow. :thinking:

elmarco:
> > My initial plan was to make `xml` a different module,
> 
> Do you mean sub-crate?

yeah a different crate

> 
> > as I plan to add some dependencies on zbus
> 
> Not sure I follow. thinking

I would like the introspection tree to provide better typed members. There is InterfaceName, MemberName, but we could have more. Not sure if they are all covered by non-zbus crates. I want to avoid a circular deps... Furthermore, I wonder still if we may want to build a introspection tree instead of just plain-text xml...

elmarco:
In the meantime, we can merge this I suppose.

zeenix:
> I would like the introspection tree to provide better typed members. There is InterfaceName, MemberName, but we could have more. Not sure if they are all covered by non-zbus crates. I want to avoid a circular deps... Furthermore, I wonder still if we may want to build a introspection tree instead of just plain-text xml...

Thanks.

> In the meantime, we can merge this I suppose.

Sure.

zeenix (approved):

PR #421, opened 2023-07-30
Drop deprecated APIs

zeenix:
Time to drop deprecated APIs, excluding the new aliases/re-exports.

### DIFF
--- a/zbus/src/error.rs
+++ b/zbus/src/error.rs
@@ -22,9 +22,6 @@ pub enum Error {
     /// Invalid D-Bus address.
     Address(String),
     /// An I/O error.
-    #[deprecated(note = "Use `Error::InputOutput` instead")]
-    Io(io::Error),
-    /// An I/O error.
     InputOutput(Arc<io::Error>),
     /// Invalid message field.
     InvalidField,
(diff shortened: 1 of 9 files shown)


zbus/src/error.rs
```diff
@@ -22,9 +22,6 @@ pub enum Error {
     /// Invalid D-Bus address.
     Address(String),
     /// An I/O error.
-    #[deprecated(note = "Use `Error::InputOutput` instead")]
-    Io(io::Error),
-    /// An I/O error.
     InputOutput(Arc<io::Error>),
     /// Invalid message field.
```
elmarco, line 26: Rename InputOutput -> Io ?

elmarco (commented):
lgtm otherwise

zbus/src/error.rs
zeenix: > Rename InputOutput -> Io ?

That completely undermines the whole point of deprecating `Io` first: give people a chance to switch to `InputOutput`. Code that switched already will now still break, even though they did what we asked.

I know the naming is very unfortunate but I couldn't come up with something better.

What we could do is keep `Io`, make it the same as `InputOutput`, and deprecate `InputOutput` now. This will still annoy users who switch already though so I don't like this option.

zeenix:
just rebased on latest main.

zbus/src/error.rs
zeenix: > What we could do is keep `Io`, make it the same as `InputOutput`, and deprecate `InputOutput` now. This will still annoy users who switch already though so I don't like this option.

Yeah, I really don't think it's worth it.

PR #424, opened 2023-07-30
Move ::raw under ::connection

elmarco:


zbus/src/lib.rs
```diff
@@ -176,8 +176,9 @@ pub use utils::*;
 #[macro_use]
 pub mod fdo;
 
-mod raw;
-pub use raw::Socket;
+#[deprecated(note = "Use `connection::Socket` instead")]
```
zeenix, line 179: I have yet to encounter direct `Socket` users so I don't think we need deprecation for this one. Moreover, I intend to make `Socket` more high-level (async methods for example) for various reasons so any existing `Socket` using code will break anyway.

zeenix (changes requested):
Good otherwise.

zbus/src/lib.rs
elmarco: it's just moved under connection:: I don't think we want to expose that to the top-level of the crate, especially if there are no users ! :)

zeenix: I'm saying we can just not re-export it here at all.

elmarco: What does it change to have it for a while with the rest of the deprecations?

zbus/src/lib.rs
```diff
@@ -176,8 +176,9 @@ pub use utils::*;
...
+#[doc(hidden)]
+pub use connection::Socket;
```
zeenix, line 181:
```suggestion
```

zbus/src/lib.rs
zeenix: > What does it change to have it for a while with the rest of the deprecations?

I stated 2 reasons above. :) The main one being that if it will be changed and break existing code anyway, deprecation provides nothing IMO.

zeenix: but ok, I will remove it as part of my work.

zeenix (approved):

PR #425, opened 2023-07-31
🚚 zb: Rename Error::QuickXml to Xml

zeenix:
We only have one kind if Xml error now.

### DIFF
--- a/zbus/src/error.rs
+++ b/zbus/src/error.rs
@@ -51,7 +51,7 @@ pub enum Error {
     FDO(Box<fdo::Error>),
     #[cfg(feature = "xml")]
     /// An XML error from quick_xml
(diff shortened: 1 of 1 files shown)


elmarco:
It's still an error from quick_xml though, .. what if we have our own Xml errors then?

zeenix:
> It's still an error from quick_xml though,

Hmm.. yeah. I wasn't too sure about this change tbh. Just that I know users won't care about this fact and `Xml` is simpler to type.

> .. what if we have our own Xml errors then?

Not sure we'll want our own generic Xml error. We'll likely have more specific errors.

elmarco:
I am not acking this, but if you feel strong about it, ok

zeenix:
ok

PR #428, opened 2023-08-04
⚰️ zb: remove unused arg0namespace from match rule

MaxVerevkin:
It was replaced by arg0ns and is now unused.

### DIFF
--- a/zbus/src/match_rule/builder.rs
+++ b/zbus/src/match_rule/builder.rs
@@ -1,4 +1,4 @@
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryInto;
 
 use static_assertions::assert_impl_all;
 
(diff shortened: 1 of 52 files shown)


zbus/src/match_rule/builder.rs
```diff
@@ -1,4 +1,4 @@
-use std::convert::{TryFrom, TryInto};
+use std::convert::TryInto;
```
zeenix, line 1: Now that we've `2021` edition enabled, you can just drop this whole line I think.

zeenix (approved):
Looks good. @elmarco broke the Windows CI so either he'll fix that or I'll revert his commits soon.

zbus/src/match_rule/builder.rs
MaxVerevkin: Awesome

zeenix (approved):
Thanks

PR #430, opened 2023-08-05
Add `addr` module

### LINKED ISSUE #476: Improve D-Bus address API
The current `zbus::Address` enum is suboptimal and assumes that addresses have no properties in common, which isn't correct (e.g `GID`). Let's fix this.

elmarco:
Address the following shortcomings of zbus <=3 Address implementation:
- the Address enum didn't match the spec well, and didn't allow easily more parameter combination. Use a DBusAddr and a Transport structs with field getters instead. Make optional fields and combinations just work without breaking API.  
- didn't handle ;-separated addresses at all
- did not follow the spec quite rigourously, in particular with percent encoded values
- implementation was becoming a mess, mixing parsing, connection handling, os-specifics..

The new implementation doesn't use hashmap (only hashset for optional validation), and try to avoid unnecessary copies.

Fixes #476.


elmarco:
CI is still broken `The pkg-config command could not be found.`, no idea why...

Anyhow, removing the draft status of this PR.

zeenix:
> no idea why...

The action thinks that everything in cached but in reality, it's not and hence the missing pkg-config error.

zbus/src/connection/raw/macos.rs
```diff
@@ -0,0 +1,21 @@
+#![cfg(target_os = "macos")]
+
+use crate::{addr::DBusAddr, process::run, Error, Result};
```
zeenix, line 3: > zb: add/rewrite raw stream connection 

Again, a huge bunch of changes with no details in the commit log of what's going on.

zbus/src/win32.rs
```diff
@@ -26,7 +26,6 @@ use winapi::{
     },
 };
 
-use crate::Address;
```
zeenix, line 29: * The commit log says you're deprecating it but you're in fact dropping it.
* I see potential for further splitting the changes.

zeenix (changes requested):

.github/workflows/rust.yml
```diff
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -D warnings
-      MSRV: 1.64.0
+      MSRV: 1.65.0
```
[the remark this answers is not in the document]
elmarco: 1.65 is ~1.5 months more recent. What's your consideration when bumping rust requirement?

unwrap_or is not a solution. Match introduces extra burden and nesting. Clearly if such a base construct as `let else` made it to stable, there are good reasons.

zbus/src/win32.rs
```diff
@@ -51,7 +51,7 @@ impl Drop for OwnedHandle {
     }
 }
 
-struct Mutex(OwnedHandle);
+pub(crate) struct Mutex(OwnedHandle);
```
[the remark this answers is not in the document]
elmarco: what do you suggest? export internally? :)

zbus/src/addr/address.rs
```diff
@@ -0,0 +1,219 @@
...
+use crate::{Error, Guid, Result};
+
+use super::{
+    percent::{decode_percents, decode_percents_str, Encodable},
+    transport::Transport,
+};
+
+/// A bus address.
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct DBusAddr<'a> {
```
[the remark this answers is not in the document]
elmarco: No relationship, it's a rewrite from scratch.

I use DBus prefix, and Addr, for consistency with rust standard library. Everything address-related uses `addr` consistently, and for types, has a prefix. IpAddr, SocketAddr etc.

zbus/src/connection/raw/macos.rs
elmarco: ok

zbus/src/win32.rs
elmarco: Not dropping, you can still use the legacy Address to parse or manipulate Address. But you can't use it to  connect anymore. I am fine with removing it now if you prefer.

zeenix: > I am fine with removing it now if you prefer.

Yes that's also (what I thought) we decided. I do not believe this new `DBusAddress` is a new API at all. We should just change `Address` to fit our needs. This is not an API typical users use, especially directly.

zbus/src/addr/address.rs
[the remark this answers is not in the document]
elmarco: sigh naming... I have a clear preference for DBusAddr, ToDBusAddr.. over Address, ToAddress. But if you insist, I'll just rename stuff.

zbus/src/win32.rs
elmarco: I am not sure what you mean by "changing Address to fit our needs".. As I see it, DBusAddr is very different.. there are no simple way to "change" the exisiting code, honestly.

elmarco: Also addr could be a new zbus_addr crate. It's fairly standalone (only zbus::Error/Result is reused)

zeenix: Look, we already have a type to represent a dbus address: `Address`. This is good and we should continue with that. Now that type has a few issues and you already have solution for solving that and we can totally break API here. i-e merge DBusAddress into Address.

> Also addr could be a new zbus_addr crate.

Sure but we'll then not want the `Address` API. The user only needs 1 (main) type here and that's what they should have.

.github/workflows/rust.yml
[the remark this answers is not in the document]
zeenix: [We were forced to bump to 1.66](https://github.com/dbus2/zbus/commit/abe3aed324cbcebca11192d62a158090ae006641) so this change can just be removed.

zeenix:
Closing this in favor of #556. One big difference between this PR and #556 is that the latter doesn't make it possible to use `Address` as a reference type. I was half-way making that happen through a bit of a different approach when I realized that in practice that won't give us anything and typical use will have to allocate anyway. See [the commit log](https://github.com/zeenix/zbus/commit/634dab30ae67f124bee3634900e9ca3ef5a7d8b0) of my incomplete attempt.

elmarco:
> Closing this in favor of #556. One big difference between this PR and #556 is that the latter doesn't make it possible to use `Address` as a reference type. I was half-way making that happen through a bit of a different approach when I realized that in practice that won't give us anything and typical use will have to allocate anyway. See [the commit log](https://github.com/zeenix/zbus/commit/634dab30ae67f124bee3634900e9ca3ef5a7d8b0) of my incomplete attempt.

You'd need to give me a few days if you want me to compare our approaches. From a quick glance, it seems you broke the API but didn't solve some of the pending issues, such as address list handling.

I find it sad that your main argument to block my proposal was that we shouldn't break API (I really think it was necessary and weak, because addr API isn't really used outside), but then you just did that in your own way and find it more acceptable now, you should have worked with me from there...

And the main argument to close my work is:
"We've to decode the string in multiple properties involved here and the
input string is readonly so we've to allocate new strings anyway. We
could just keep the original string as is and decode on-demand but given
that most uses of Address will see these properties being read later,
that will only delay alloction. Moreover, it will be worse cause then
we'll have to allocate each time the properties in question are ready."

How is that related to this branch?

Also the split between address parsing handling and raw stream connection ("add/rewrite raw stream connection") is cleaner here. All in all, did you seriously look and compare the existing work and your work?

zeenix:
> You'd need to give me a few days if you want me to compare our approaches.

I gave you several months to address the concerns with this PR and you told me that you won't have the time to address the concerns. So I had to do everything myself cause there was so much in your branch that I disagreed with and I don't like fixing others code. Now that I'm done after 
many days of work, you want to be involved all of a sudden. :cry: 

> it seems you broke the API

As I told you before, we want to break the API. We're breaking API in multiple places already. As long as the typical user has an easy way to port their code, I'm happy. `Address` isn't an API typically used directly by users.

> didn't solve some of the pending issues

It does make it a nice hierarchy, adds guid and makes it possible to add new properties in the future w/o breaking the API. That addresses #476 pretty well.

> such as address list handling.

Pretty sure that can be added on top w/o breaking API. If not, we do a 5.0 at some point.

> And the main argument to close my work is:

The main argument is that you didn't finish your work here, despite several months passing by and 4.0 is super late, as it is. So if I address #476 in a separate PR in my own way, this work becomes obsolete, at least for now.

> How is that related to this branch?

Since it's very difficult to avoid allocations for strings that need % decoding, [your branch is still doing that](https://github.com/dbus2/zbus/pull/430/files#diff-696af3d921a963db5f05808aba41eaa72f849a529e205a4f8c4374ee83f7e879R110) so the issue I describe in the commit log, is relevant.

Having said that, looking again at your PR, I got some ideas on how I can avoid most allocations at least by decoding to buffers on the stack for some properties at least. However, I'll attempt that after merging my PR. I want to release tomorrow at latest cause Mon onward, I've to focus almost 100% on work-work. If I manage to finish before then, good. Otherwise, it'll have to wait for next API break.

zeenix:
> All in all, did you seriously look at this and compare it with the existing work and your work?

Not really. As I told you in my review here, I didn't agree with the overall approach of introducing a secondary address type/hierarchy (which I also said in IM before this PR) so I did at first look into addressing my concerns in this branch but quickly gave up, because I realized I need to start from scratch anyway.

Anyway, if you felt strongly enough about your work getting merged, you had more than sufficient time and reminders to get it done in time. Now it's super late. I'm sorry but that's what will always happen if you don't finish the job.

PR #432, opened 2023-08-08
Add explicit return statement for readability

whot:
A typo fix and an explicit `return` to match the surrounding code

### DIFF
--- a/zbus/src/address.rs
+++ b/zbus/src/address.rs
@@ -422,7 +422,8 @@ impl Address {
                         .unwrap_or_else(|_| format!("/run/user/{}", Uid::effective()));
                     let path = format!("unix:path={runtime_dir}/bus");
 
-                    Self::from_str(&path)
+                    #[allow(clippy::needless_return)]
+                    return Self::from_str(&path);
                 }
 
                 #[cfg(target_os = "macos")]


zeenix (commented):
Nice! thanks so much. CONTRIBUTING didn't convince you to use emoji prefixes? ;)

whot:
> Nice! thanks so much. CONTRIBUTING didn't convince you to use emoji prefixes? ;)

oops. `/me` writes "thou shall not submit PRs from airports" 100 times
as an apology, I've now filed #435 with even more typo fixes :)


Also turns out that clippy was unhappy because it's a needless return so I had to add the `#[allow(...)]` macro to work around that. I'm pretty sure that same clippy error would be triggered on macos for that return there though, so arguably we should just enable that suppression for the whole block.

zeenix:
> Also turns out that clippy was unhappy because it's a needless return so I had to add the `#[allow(...)]` macro to work around that.

Hmm.. in that case, I'd disagree with `Return` usage. I prefer to keep `#[allow..` use to the minimum.

whot:
Yeah, fair enough. Feel free to close this PR though as I said above, I'm pretty sure clippy will fail on macos right now too.

zeenix:
> Yeah, fair enough. Feel free to close this PR though as I said above, I'm pretty sure clippy will fail on macos right now too.

Just to be clear, you mean with this PR, right? It currently seems to work fine on `main`. I ran `cargo clippy --target x86_64-apple-darwin` and I didn't see any warnings.

whot:
Hah, indeed, this appears to be a bug in clippy which doesn't warn about single-line returns but it does in a block. i.e.
```rust
// this code does not trigger a warning
#[cfg(target_os = "macos")]
return Self::from_str("launchd:env=DBUS_LAUNCHD_SESSION_BUS_SOCKET");
```
```rust
// warning: unneeded `return` statement
#[cfg(target_os = "macos")]
{
    return Self::from_str("launchd:env=DBUS_LAUNCHD_SESSION_BUS_SOCKET"); 
}
```
Purely from looking at the code I would've expected either to trigger a warning.

(also thanks, now I learned about `cargo clippy --target` :smile:)

zeenix:
@whot so we can close this then? Please close it, if you agree.

whot:
yep, let's close this

PR #435, opened 2023-08-10
CONTRIBUTING: fix a few typos and a minor rewording

whot:
Please feel free to squash these together, I have no preferences here.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -43,11 +43,11 @@ Same rules apply here as for bug reports and feature requests. Plus:
 
 ### Legal Notice
 
-When contributing to this project, you **implicitly** declare that you:
+When contributing to this project, you **implicitly** declare that:
 
-* have authored 100% of the content.
-* have the necessary rights to the content.
-* agree to providing the content under the [project's license](LICENSE).
+* you have authored 100% of the content,
+* you have the necessary rights to the content, and
+* you agree to providing the content under the [project's license](LICENSE).
 
 ## Running the test suite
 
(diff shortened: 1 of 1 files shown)


zeenix:
> Please feel free to squash these together, I have no preferences here.

Hah, separate logical changes so you did right. :)

CONTRIBUTING.md
```diff
@@ -43,11 +43,11 @@ Same rules apply here as for bug reports and feature requests. Plus:
...
 
-When contributing to this project, you **implicitly** declare that you:
+When contributing to this project, you **implicitly** declare that:
 
-* have authored 100% of the content.
-* have the necessary rights to the content.
-* agree to providing the content under the [project's license](LICENSE).
+* you have authored 100% of the content,
+* you have the necessary rights to the content, and
+* you agree to providing the content under the [project's license](LICENSE).
```
zeenix, line 50: Not at all sure about this change but I don't mind.

zeenix (approved):

zeenix:
Since `busd`'s contributing guide was very much based on this (or was it the other way around? :thinking:), it'd be great to also make these changes [there](https://github.com/dbus2/busd/blob/main/CONTRIBUTING.md).

whot:
Does the "Assert auto traits on item" section apply to busd as well? Or do I leave that out? The rest is a straightforward copy.

zeenix:
> Does the "Assert auto traits on item" section apply to busd as well? Or do I leave that out? The rest is a straightforward copy.

Actually, no. We only do that in zbus to avoid breaking public API. Good catch.

PR #437, opened 2023-08-11
✨ zb: add ProxyImpl trait

### LINKED ISSUE #433: Traits for macro generated Proxies
It would be really nice to have some traits including the common methods of the generated proxies, so that functions which need any of `new`, `builder`, `into_inner`, `inner` could just use those traits as bounds for generic arguments.

I would imagine something like this:
```rs
trait Proxy<'c> {
    // maybe we could even get away with a default impl here?
    pub async fn new(conn: &::zbus::Connection) -> ::zbus::Result<Self> {
        Self::builder(conn).build().await
    }

    pub fn builder(conn: &::zbus::Connection) -> ::zbus::ProxyBuilder<'c, Self>;

    pub fn into_inner(self) -> ::zbus::Proxy<'c>

    pub fn inner(&self) -> &::zbus::Proxy<'c>
}
```

DerZade:
This PR fixes #433 by adding a `ProxyImpl` trait for macro generated proxy-structs to allow narrowing generic arguments to D-Bus proxies.

### DIFF
--- a/zbus/src/blocking/proxy/mod.rs
+++ b/zbus/src/blocking/proxy/mod.rs
@@ -527,3 +527,19 @@ mod tests {
         assert!(signal.args().unwrap().name() == well_known);
     }
 }
+
+/// This trait is implemented by all blocking proxies, which were
+/// generated with the [`dbus_proxy`](zbus::dbus_proxy) macro.
+pub trait ProxyImpl<'c>
+where
+    Self: Sized,
+{
+    /// Returns a customizable builder for this proxy.
+    fn builder(conn: &Connection) -> Builder<'c, Self>;
+
+    /// Consumes `self`, returning the underlying `zbus::Proxy`.
+    fn into_inner(self) -> Proxy<'c>;
+
+    /// The reference to the underlying `zbus::Proxy`.
+    fn inner(&self) -> &Proxy<'c>;
+}
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -1504,3 +1504,19 @@ mod tests {
         Ok(())
     }
 }
+
+/// This trait is implemented by all async proxies, which were
+/// generated with the [`dbus_proxy`](zbus::dbus_proxy) macro.
+pub trait ProxyImpl<'c>
+where
+    Self: Sized,
+{
+    /// Returns a customizable builder for this proxy.
+    fn builder(conn: &Connection) -> Builder<'c, Self>;
+
+    /// Consumes `self`, returning the underlying `zbus::Proxy`.
+    fn into_inner(self) -> Proxy<'c>;
+
+    /// The reference to the underlying `zbus::Proxy`.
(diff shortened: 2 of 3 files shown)


zbus/src/blocking/proxy/mod.rs
```diff
@@ -527,3 +527,19 @@ mod tests {
         assert!(signal.args().unwrap().name() == well_known);
     }
 }
+
+/// This trait is implemented by all blocking proxies, which were
+/// generated with the [`dbus_proxy`](zbus::dbus_proxy) macro.
```
zeenix, line 532: Does this way of linking works? :thinking:

zbus/src/lib.rs
```diff
@@ -231,10 +231,11 @@ mod tests {
     use zvariant::{OwnedObjectPath, OwnedValue, Type};
 
     use crate::{
-        blocking::{self, MessageIterator},
+        blocking::{self, proxy::ProxyImpl as _, MessageIterator},
```
zeenix, line 234: don't think we need these changes here.

zeenix (changes requested):
Looking good otherwise. Also two nits:

* it'd be best to split the commit into two: One to add the new trait and another to implement it in the macros.
* I really appreciate you providing a nice commit log with ref to the issue. :+1:  Just one nit: Please break the lines at 74 chars as per the [guidelines](https://wiki.gnome.org/Git/CommitMessages).

zbus/src/blocking/proxy/mod.rs
DerZade: chrome_sl6iP9qlT1


Yes it does 🙃

zbus/src/fdo.rs
```diff
@@ -17,8 +17,8 @@ use zvariant::{
 };
 
 use crate::{
-    dbus_interface, dbus_proxy, message::Header, object_server::SignalContext, DBusError, Guid,
-    ObjectServer,
+    dbus_interface, dbus_proxy, message::Header, object_server::SignalContext, proxy::ProxyImpl,
+    DBusError, Guid, ObjectServer,
```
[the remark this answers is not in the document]
DerZade: That's actually really smart. I've never seen that pattern.

I actually thought about how I could circumvent existing code from breaking because they have to import the trait, but couldn't come up with a nice solution.

zbus/src/lib.rs
DerZade: Considering your other comment, you are correct.

DerZade:
Thanks a lot for the amazingly fast review.

Sorry, apparently I missed the note about the 75 character limit. That is now corrected along with all your other comments. I also split the commits into two as requested 😇 

Let me know if there is anything else that needs to be addressed.

zbus/src/blocking/proxy/mod.rs
zeenix: Nice. TIL :)

zeenix (approved):
LGTM. Thanks so much.

PR #439, opened 2023-08-11
🔥 message mod API improvements

zeenix:
Many improvements to message module API, improving the ergonomics and performance.

Among other things, it drops/hides a lot of low-level API that IMO we should have never provided. In my defense, I was new to Rust when I wrote this code. :)

---

Zeeshan Ali Khan on behalf of MBition GmbH.

[Provider Information](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md#mbition-gmbh).


zeenix:
For the record, @elmarco ack'ed in a private message. :)

Issue #288, opened 2022-11-08
dbus_proxy: consider switching 'assume_defaults' default value to 'false'


zeenix:
In GitLab by @lucab on Nov 8, 2022, 17:24

This is a followup from https://gitlab.freedesktop.org/dbus/zbus/-/issues/284#note_1624111.

The patch at https://gitlab.freedesktop.org/dbus/zbus/-/merge_requests/587 is introducing a new `assume_defaults` attribute for the `dbus_proxy` macro.
For compatibility reasons, it is initially defaulting to `true` (i.e. the implicit behavior beforehand) and producing a compiler warning when omitted (and not default destination/path are provided).

On the next semver break, it would be good to consider switching the default to `assume_defaults = false` and adjust the compiler warning.

zeenix:
@lucab Looking into this right now. Just one question: 

> and adjust the compiler warning.

what should it become? It doesn't say anything about the default. Should we just remove it now? :thinking:

lucab:
@zeenix I don't remember exactly, but I think it was either about inverting the warning (i.e. warn if it finds an explicit `assume_defaults = false`, which is already the implicit default anyway and can be now dropped) or plainly remove it. As you prefer.

zeenix:
Thanks. Right I don't see the point of warning for turning the implicit into explicit if it's the sane default so I'll just drop the warning. Thanks for the quick action.

PR #445, opened 2023-08-16
A fix & an API change to dbus_proxy

### LINKED ISSUE #444: Support paths in `object*` attributes of `dbus_proxy` macro
As reported [here](https://github.com/dbus2/zbus/issues/392#issuecomment-1680563402):

> If I move the `EntryGroup` to it's own submodule, and change the annotation to `object = "entry_group::EntryGroup"`, I get a different failure:
> 
> ```
> error: custom attribute panicked
>  --> src/mdns.rs:4:1
>   |
> 4 | / #[dbus_proxy(interface = "org.freedesktop.Avahi.Server",
> 5 | |              default_service = "org.freedesktop.Avahi",
> 6 | |              default_path = "/")]
>   | |_________________________________^
>   |
>   = help: message: `"entry_group::EntryGroupProxyBlocking"` is not a valid identifier
> ```

zeenix:
Fixes #444, #288.

### DIFF
--- a/book/src/client.md
+++ b/book/src/client.md
@@ -99,7 +99,10 @@ use std::error::Error;
 
 use zbus::{zvariant::Value, dbus_proxy, Connection};
 
-#[dbus_proxy]
+#[dbus_proxy(
+    default_service = "org.freedesktop.Notifications",
+    default_path = "/org/freedesktop/Notifications"
+)]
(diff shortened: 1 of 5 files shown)


zeenix:
@jhenstridge if you could give it a test, that would be great. :pray:

Issue #307, opened 2023-01-30
Support for awaitable connection


zeenix:
In GitLab by @peelz on Jan 30, 2023, 19:40

It would be nice if connection errors could be reported to the caller by awaiting on the connection, e.g:

```rs
let connection = Connection::session().await?;

if let Err(err) = connection.await {
  error!("Something bad happened: {err}");
}
```

zeenix:
@notpeelz, I'm not sure about this anymore. You can achieve the same already with `MessageStream`:

```rust
let stream = MessageStream::from(Connection::session().await?);

while let Some(res) = stream.next().await {
    if let Err(e) = res {
        error!("Something bad happened: {e}");
    }
}
```

PR #451, opened 2023-08-22
✨ zb: Add callback for socket closed

dreiss:
For the use case of a peer-to-peer D-Bus server, it's convenient to keep a connection alive as long as the peer is connected.  This feature allows a clean-up callback to automatically trigger the release of server-side resources (including dropping the last Connection reference) when the socket is closed by the remote client.

### DIFF
--- a/zbus/src/connection.rs
+++ b/zbus/src/connection.rs
@@ -33,7 +33,7 @@ use crate::{
     socket_reader::SocketReader,
     Authenticated, CacheProperties, ConnectionBuilder, DBusError, Error, Executor, Guid, MatchRule,
     Message, MessageBuilder, MessageFlags, MessageStream, MessageType, ObjectServer,
-    OwnedMatchRule, Result, Task,
+    OwnedMatchRule, Result, Task, CallOnDrop,
 };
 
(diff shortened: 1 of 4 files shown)


zeenix:
Hi, many thanks for the PR. However:

* We generally don't do callbacks. We make use of `event-listener` for such things. See https://docs.rs/zbus/latest/zbus/struct.Connection.html#method.monitor_activity
* This can be achieved through `MessageStream` already by waiting on `None` from the stream. However, maybe we need a more explicit API to avoid unnecessary task wake-ups. 🤔

zeenix:
Also I already had plans for this: https://github.com/dbus2/zbus/issues/307

zeenix:
> Ah. I thought the Connection has to be converted into a MessageStream. I didn't realize that both could exist (and work) at the same time.

Moreover, Connection is also cheaply clonable so you can clone and convert to MessageStream too.

> Listening for None on the MessageStream seems to work just fine (though the one downside is that I think it won't trigger if the socket_reader task panics).

It should never panic. If it does, that's a serious bug.

zeenix:
I'll close this one. If you'd like to take a stab at #307, that'd be great but please create a separate PR.

PR #453, opened 2023-08-23
🐛 zx: Fix declaring proxies for writable properties

mvidner:
The `#[dbus_proxy(property)]` attribute needs to be present also at the `set_foo` function, otherwise the proxy will call a `SetFoo` method instead of `Set`ting the `Foo` property.

Example:
https://github.com/openSUSE/agama/pull/712

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -65,21 +65,22 @@ impl<'i> Display for GenTrait<'i> {
         props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for p in props {
             let name = to_identifier(&to_snakecase(p.name().as_str()));
+            let fn_attribute = if pascal_case(&name) != p.name().as_str() {
+                format!("    #[dbus_proxy(property, name = \"{}\")]", p.name())
+            } else {
+                "    #[dbus_proxy(property)]".to_string()
+            };
 
             writeln!(f)?;
             writeln!(f, "    /// {} property", p.name())?;
-            if pascal_case(&name) != p.name().as_str() {
-                writeln!(f, "    #[dbus_proxy(property, name = \"{}\")]", p.name())?;
-            } else {
-                writeln!(f, "    #[dbus_proxy(property)]")?;
-            }
-
             if p.access().read() {
+                writeln!(f, "{}", fn_attribute)?;
                 let output = to_rust_type(p.ty(), false, false);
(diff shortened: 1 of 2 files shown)


zbus_xmlgen/src/lib.rs
```diff
@@ -65,21 +65,22 @@ impl<'i> Display for GenTrait<'i> {
...
+            };
 
             writeln!(f)?;
             writeln!(f, "    /// {} property", p.name())?;
-            if pascal_case(&name) != p.name().as_str() {
-                writeln!(f, "    #[dbus_proxy(property, name = \"{}\")]", p.name())?;
-            } else {
-                writeln!(f, "    #[dbus_proxy(property)]")?;
-            }
-
```
zeenix, line 76: I'm not sure I follow what is changing here. We were unconditionally add the attribute here already in the removed code.

zeenix (commented):
Super thanks for your contrbution and bonus points for adhering to our contribution guide! However, I've a question.

zbus_xmlgen/src/lib.rs
mvidner: I'm not sure what your angle is so I try to explain my intent. 

Well yes, originally there is one attribute, and you would think that is fine for one property:
```rs
    /// Bar property
    #[dbus_proxy(property)]
    fn bar(&self) -> zbus::Result<u8>;
    fn set_bar(&self, value: u8) -> zbus::Result<()>;
```

but the problem is that the attribute is needed twice, once for the getter and once for the setter. Moving the code to the conditional branch avoids duplicating the attribute in the rare case of a write-only property.

It is documented correctly in the example in https://docs.rs/zbus/3.12.0/zbus/attr.dbus_proxy.html 

```rs
#[dbus_proxy(...)]
trait SomeIface {
    #[dbus_proxy(property)]
    fn a_property(&self) -> fdo::Result<String>;

    #[dbus_proxy(property)]  // HERE!
    fn set_a_property(&self, a_property: &str) -> fdo::Result<()>;
}
```

Sorry if this was already clear to you, were you asking about an implementation aspect instead?

zeenix: Thanks for explaining in detail. However:

> were you asking about an implementation aspect instead?

Indeed. I'm just wondering how your code is doing something different from the existing code. :thinking: I stared at it for a while but I don't get how it's different. :thinking:

zeenix: Ah! I get it now. Sorry, my bad. :)

zeenix (approved):

PR #454, opened 2023-08-23
Rename tests.xml to tests.rs

elmarco:


zeenix (changes requested):
nit: It's more a typo correction:

✏️ zx: Correct name of test code file

PR #455, opened 2023-08-24
💄 zx: support --help and -h

whot:
Best. Feature. Ever!

### DIFF
--- a/zbus_xmlgen/src/main.rs
+++ b/zbus_xmlgen/src/main.rs
@@ -19,6 +19,16 @@ use zbus_xml::{Interface, Node};
 use zbus_xmlgen::GenTrait;
 use zvariant::ObjectPath;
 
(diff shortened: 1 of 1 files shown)


zbus_xmlgen/src/main.rs
```diff
@@ -94,13 +108,7 @@ fn main() -> Result<(), Box<dyn Error>> {
...
         }
         None => {
-            eprintln!(
-                r#"Usage:
-  zbus-xmlgen <interface.xml>
-  zbus-xmlgen --system|--session <service> <object_path>
-  zbus-xmlgen --address <address> <service> <object_path>
-"#
-            );
+            usage();
```
zeenix, line 111: I think we really should just use clap.

zeenix (approved):
Thanks so much.

PR #462, opened 2023-09-03
🚸 Message constructor to return a builder instance

zeenix:
Not only this makes sense anyway, it also makes the method much simpler as we can drop the optional arguments and therefore also the need for user to have to specify the concrete types for these arguments even when they just want to pass `None`.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -471,7 +471,11 @@ impl Connection {
     where
         B: serde::ser::Serialize + zvariant::DynamicType,
     {
-        let m = Message::method_reply(self.unique_name(), call, body)?;
+        let mut b = Message::method_reply(call)?;
+        if let Some(sender) = self.unique_name() {
+            b = b.sender(sender)?;
+        }
+        let m = b.build(body)?;
         self.send_message(m).await
     }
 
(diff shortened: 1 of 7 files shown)


elmarco:
Well, there isn't much point in having ctor then. It is there to make common things easier, and help with spec correctness (the arguments reflect was is required and opional). Why not simply use the Builder directly instead then?

zeenix:
This is the usual pattern. If there is only one way to create the builder, the API is `Type::builder`. If we want we could do the same here and default to a method type but I think that would be needlessly inconvenient. 

This PR does not make things any less convenient. See the updated client code in this PR. The only places where it's**slightly** more inconvenient are methods taking optional arguments. The rest can still just make one call chain in a single statement, which isn't any less inconvenient than a single method call. Actually it's more convenient when you don't need to set optional params.

Having said that, I think we should hide/remove the direct Builder ctors now to have a single consistent way to create messages.

zeenix:
> and help with spec correctness (the arguments reflect was is required and opional)

That is still the case btw. The methods in question, still have the required parameters as arguments.

zeenix:
> but the builder exposes everything regardless of the exact message type. Perhaps there should be specialized builders instead then.

Could do, yes (especially if we hide/drop the Builder ctors methods) but not sure it's worth the effort? The API would be more ergonomic for sure but given that this:

```rust
let msg = Message::method("/", "Hello")?
    .path("/")?
    .build(&())?;
```

doesn't cause any issues at runtime, I'm not sure it's worth splitting and the complexity it'll bring.

Also, keep in mind that currently users can override builder params as many times as they want and even with the split, it will continue to be true.

zeenix:
> Having said that, I think we should hide/remove the direct Builder ctors now to have a single consistent way to create messages.

Done. I went for deprecation.

Issue #403, opened 2023-07-09
Error handling when creating instances via a builder


WhyNotHugo:
I'm bringing this up as a response to the toot:

> If you want changes that require API breakage, now would be a good time to speak out

Something that has low-key bothered me was error handling in builders. In particular, I often have operations that I know are infallible but I need to handle errors due to their API (generally, I just add `.expect("this cannot fail")`, but that's usually an indicator of an API not being quite right.

For example: https://docs.rs/zbus/latest/zbus/struct.ProxyBuilder.html#method.path

If I already have a Path instance, I know that calling this method is infallible, but I still need to handle (ignore?) the error.

### First potential approach

The first one is the more-API-breaking one, and looks something like:

```rust
pub fn path(self, path: ObjectPath) -> Self;

pub fn try_path<P>(self, path: P) -> Result<Self>where
    P: TryInto<ObjectPath<'a>>,
    P::Error: Into<Error>;
```

The major downside of this one is that it requires a lot of work for applications to adapt. 

### Second potential approach

The second approach is less invasive, and to just update the API to:

```rust
pub fn path<P>(self, path: P) -> Self where
    P: TryInto<ObjectPath<'a>>,
```

And then have `ProxyBuilder.build` return an error if the Path was invalid. This approach might get trickier with error types.

### Third potential approach

The third approach is likely the weirdest, but least invasive, and most compatible, by introducing a new function:

```rust
pub fn from_parts(self, path: Path, interface: InterfaceName, destination: BusName) -> T;
```

Where all the "parts" (for lack of a better name) have already been validated, this function cannot fail, and matches what's going on.

zeenix:
Hi @WhyNotHugo,

Thanks for reporting this and also for suggesting solutions.

Given that the main idea behind keeping these generic is convenience, solution 2 seems most interesting at first thought.

> This approach might get trickier with error types.

I guess we can keep the errors generic to (like they're right now)? This will mean us having to keep everything in a Box though but I think the cost of allocation is still too small here to worry about at all and the gain is quite big.

This will still be a breaking change though but super easy to fix. What I'm concerned about though, is that for many it will require  a lot of work (even if it's very trivial changes).  Even though we're breaking API, I'd really prefer to not break most of existing code, at least not in a way that requires a lot of work to port.

So maybe we should consider:

[...]

Issue #321, opened 2023-03-04
ProxyBuilder: add methods that take concrete types instead of generics


zeenix:
In GitLab by @MaxVerevkin on Mar 4, 2023, 17:32

While working on a few projects I noticed that most of the times I already have `ObjectPath` or `BusName`, and when I pass them to `.destination()` / `.path()` I have to unwrap everything.

Personally I think `.path(x.try_into()?)` is better than `.path(x)?` because it is explicit about what can go wrong and it would encourage users of this library to store and pass around `ObjectsPath`s instead of strings.

I do not expect any existing APIs to change, but perhaps concrete-type-variants of `.path()`/`.destination()`/`.interface()` could be added? I don't know how to call them, but maybe someone has any thoughts.

PR #463, opened 2023-09-04
docs: document errors for a few functions

WhyNotHugo:


zbus/src/connection/builder.rs
```diff
@@ -77,11 +77,19 @@ assert_impl_all!(Builder<'_>: Send, Sync, Unpin);
 
 impl<'a> Builder<'a> {
     /// Create a builder for the session/user message bus connection.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error variant if parsing the session bus address fails.
```
zeenix, line 83: nit: We should either remove "variant" or explicitly mention the variant name.

zbus/src/connection/builder.rs
```diff
@@ -266,6 +274,10 @@ impl<'a> Builder<'a> {
     /// interfaces available immediately after the connection is established. Typically, this is
     /// exactly what you'd want. Also in contrast to [`zbus::ObjectServer::at`], this method will
     /// replace any previously added interface with the same name at the same path.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if and only if calling `ObjectPath::try_from(path)` fails.
```
zeenix, line 280: `try_from` is an implementation detail, one that's obvious from the generic constraint. I'd write "if `path` is not a valid object path".

zeenix (changes requested):
Let's be consistent and add these for all failing methods here? Also, don't forget the `blocking::connection::Builder` wrapper.

zbus/src/connection/builder.rs
WhyNotHugo: I don't think this is terribly useful, but I've mentioned the variant explicitly anyway.

zbus/src/connection/builder.rs
WhyNotHugo: Done.

WhyNotHugo:
> Let's be consistent and add these for all failing methods here?

Of course, that's the Rust convention. There's a clippy lint to find all of the missing cases: https://rust-lang.github.io/rust-clippy/master/index.html#/missing_errors_doc

zbus/src/connection/builder.rs
```diff
@@ -77,11 +77,19 @@ assert_impl_all!(Builder<'_>: Send, Sync, Unpin);
 
 impl<'a> Builder<'a> {
     /// Create a builder for the session/user message bus connection.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`Err`](Result::Err) variant if parsing the session bus address fails.
```
zeenix, line 83: Now I understand what you actually meant by "variant" here and why you disagreed with my comment. I thought you mean the "zbus::Error" variant since Result is obvious to even new Rust devs.

My suggest was to either mention the exact `zbus::Error` variant that can be returned from here or just say "error". AFAIK this is the convention too and is useful.

zbus/src/connection/builder.rs
```diff
@@ -266,6 +274,11 @@ impl<'a> Builder<'a> {
     /// interfaces available immediately after the connection is established. Typically, this is
     /// exactly what you'd want. Also in contrast to [`zbus::ObjectServer::at`], this method will
     /// replace any previously added interface with the same name at the same path.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if path cannot be parsed as a valid object path (i.e.: if
+    /// `ObjectPath::try_from(path)` fails).
```
zeenix, line 281: you still kept the internal detail. Please remove the `(..)` part. It's not needed.

zeenix (changes requested):

zeenix:
> > Let's be consistent and add these for all failing methods here?
> 
> Of course,

Cool. Just to be clear, you mean you're going to update the commit to add them?

> that's the Rust convention. There's a clippy lint to find all of the missing cases: https://rust-lang.github.io/rust-clippy/master/index.html#/missing_errors_doc

Nice! I didn't know about this lint.

zbus/src/connection/builder.rs
WhyNotHugo: I think that's more useful. Updated.

zbus/src/connection/builder.rs
WhyNotHugo: The part between the parenthesis is where the actually useful information is contained.

I was writing some code that calls this method, and wasn't quite sure when this could return and error (and therefore, how to handle it). I'm passing an object of type `ObjectPath`. A description saying _"Returns an error if path cannot be parsed as a valid object path"_ doesn't say anything useful, because it's not clear at all _when_ this can fail or _why_.

By specifying that this fails when `ObjectPath::try_from(path)` would fail, then it is clear that this call is likely infallible (because `T::try_from(t: T)` is usually infallible), so I can just use `.expect()` assuming that this does not fail.

---

TBH, taking a `TryInto<T>` as input in a function like this is an anti-pattern. It forces the API to have to represent the error even when it is not applicable, even when it doesn't make sense. It would make much more sense to take an `ObjectPath` as input, and require that the caller do the validation/conversion [if necessary] themselves.

However, this isn't about fixing the API; I'm just trying to make the existing API clearer so it's more obvious when errors need to be handled and how.

WhyNotHugo:
> Cool. Just to be clear, you mean you're going to update the commit to add them?

No, that can be done in a follow-up PR (likely several different ones). There are another 153 occurrences of the same information missing, and each occurrence requires careful inspection to understand when they return an error.

zbus/src/connection/builder.rs
MaxVerevkin: I feel similarly about this API (#321)

zeenix: > because it's not clear at all _when_ this can fail or _why_.

As a user, you don't need to know. If you do, look at the generic constraints. User just needs to know that they need to pass a valid object path. If they don't, they'll get an error.

> TBH, taking a `TryInto<T>` as input in a function like this is an anti-pattern. It forces the API to have to represent the error even when it is not applicable, even when it doesn't make sense. It would make much more sense to take an `ObjectPath` as input, and require that the caller do the validation/conversion [if necessary] themselves.

I disagree and hence whey we've the generic API.

As we discussed before, I'm not against providing API that takes ready objects but the main API should be as easy to use as possible (#321).

zeenix: > > because it's not clear at all _when_ this can fail or _why_.
> 
> As a user, you don't need to know. If you do, look at the generic constraints. User just needs to know that they need to pass a valid object path. If they don't, they'll get an error.

I actually have never seen the "how" being documented in "Error" sections of fallible methods in the Rust world and I think that's the right thing to do.

WhyNotHugo:
I don't know the error type of the other methods. I figured out the ones that I'm calling, and those were somewhat simpler to figure out. Just look at the next method:

https://github.com/dbus2/zbus/blob/6155d22ff1c52d36136488fb5d835a0d37484ae5/zbus/src/connection/builder.rs#L130-L138

The error is cast via `Into::into`, so I've no idea which implementation is called. I tried looking at the `zbus::error::Error` variants, but it's not obvious. Perhaps negative examples would work to figure out what times are returned and then document them?

As I use `zbus`, I come across bits of documentation that are missing and I have to sit down and figure out these nuances. I'm happy to send these back as PRs to save some work to the next person needing to use these function. I don't mind refining on wording so as to make the documentation as clear as possible. But I'm not signing up for fixing all the documentation for stuff that I haven't yet used. I'll be the right person to document those as I actually use them and understand their entire context.

zeenix:
> But I'm not signing up for fixing all the documentation for stuff that I haven't yet used.

I never asked you to fix **all** the documentation. Please read my last comment again.

zeenix:
> I'll be the right person to document those as I actually use them and understand their entire context.

You could at least adjust the wording for `serve_at` docs to include the `iface`?

zbus/src/connection/builder.rs
MaxVerevkin: > User just needs to know that they need to pass a valid object path.

Yeah, and IMHO it can be best described by the type system, by requiring the argument to be of a certain concrete type, not the documentation.

> I'm not against providing API that takes ready objects

I thought #321 was rejected... But good to know. Anyway, this is getting off-topic :)

zeenix: > > User just needs to know that they need to pass a valid object path.
> 
> Yeah, and IMHO it can be best described by the type system, 

It is! The generics say what's the expected type.

> by requiring the argument to be of a certain concrete type, not the documentation.

I'm not the one pushing for docs here but I'm not against them either. We're enforcing compliance whether user has to create the types or we do it for them.

> > I'm not against providing API that takes ready objects
>
> I thought #321 was rejected... But good to know. Anyway, this is getting off-topic :)

Hmm.. actually it was. I confused with #403, where the suggestion is to add additional API.

MaxVerevkin: > where the suggestion is to add additional API.

#321 also suggests to _add_ API (it's in the title :smile:), but #403 is _very_ similar but better written, so I'm happy to see that my idea is not dead.

WhyNotHugo:
> You could at least adjust the wording for serve_at docs to include the iface?

Done, I think that it should be good now :)

zbus/src/connection/builder.rs
```diff
@@ -260,12 +270,16 @@ impl<'a> Builder<'a> {
         self
     }
 
-    /// Register a D-Bus [`Interface`] to be served at a given path.
+    /// Register a D-Bus [`Interface`] `iface` to be served at a given path.
```
zeenix, line 273: Actually this is not what I meant. I would never ask you to add unrelated changes to a commit. :) 

I was mistakenly thinking the `iface` attribute is also using `TryFrom` and is therefore fallible. Hence why I asked for it to be also included in the error docs you're adding. 

So I'm sorry for misguiding you but please remove this change.

WhyNotHugo: Oh, I was a bit confused about this bit, but it has been a few days too :sweat_smile: 

Should be okay now?

zeenix: @WhyNotHugo yes, thanks. :)

zeenix:
> Looks good otherwise but apart from the mistake I made, please add these docs to the blocking wrapper too: `blocking::connection::Builder`.

This is still missing. Just need C&P.

On a related note, it would be nice to make the blocking wrapper methods' docs just point to the underlying async methods' docs so we don't need all this repetition.

WhyNotHugo:
> On a related note, it would be nice to make the blocking wrapper methods' docs just point to the underlying async methods' docs so we don't need all this repetition.

I can add something like this to the blocking wrapper instead:

```
/// Wrapper around `connection::Builder`.
///
/// See its documentation for details (including error conditions).
```

Or I can just copy-paste it, lemme know what you prefer.

zeenix:
> /// Wrapper around `connection::Builder`.
> ///
> /// See its documentation for details (including error conditions).

We need to do a more consistent overall change for this one. If you want to do this, please go ahead but then each method needs to be something like this:

```rust
/// Blocking version of [`connection::Builder::method_name`].
```

as a separate commit that comes before the current one in this PR.

> Or I can just copy-paste it, lemme know what you prefer.

That's fine too. You choose your pain. :)

WhyNotHugo:
This is too much work. Honestly, I'd have preferred if you just rejected my PR and told me to fuck off.

zeenix:
> This is too much work. Honestly, I'd have preferred if you just rejected my PR and told me to fuck off.

I don't know how you arrived at this after I agreed to just c&p:

> Or I can just copy-paste it, lemme know what you prefer.

I suggested that myself.

PR #482, opened 2023-09-28
zbus: Use typed associated constants for ProxyDefault

bilelmoussaoui:
This is a breaking API change, so it makes sense for the upcoming zbus release

### DIFF
--- a/zbus/src/lib.rs
+++ b/zbus/src/lib.rs
@@ -183,6 +183,7 @@ pub mod export {
     pub use async_trait;
     pub use futures_core;
     pub use futures_util;
+    pub use once_cell;
     pub use ordered_stream;
     pub use serde;
     pub use static_assertions;
--- a/zbus/src/proxy/builder.rs
+++ b/zbus/src/proxy/builder.rs
@@ -1,5 +1,6 @@
 use std::{collections::HashSet, marker::PhantomData, sync::Arc};
 
+use once_cell::sync::Lazy;
 use static_assertions::assert_impl_all;
 use zbus_names::{BusName, InterfaceName};
 use zvariant::{ObjectPath, Str};
@@ -151,11 +152,9 @@ where
     pub fn new(conn: &Connection) -> Self {
         Self {
             conn: conn.clone(),
-            destination: T::DESTINATION
-                .map(|d| BusName::from_static_str(d).expect("invalid bus name")),
-            path: T::PATH.map(|p| ObjectPath::from_static_str(p).expect("invalid default path")),
-            interface: T::INTERFACE
-                .map(|i| InterfaceName::from_static_str(i).expect("invalid interface name")),
+            destination: T::DESTINATION.as_ref().cloned(),
+            path: T::PATH.as_ref().cloned(),
+            interface: T::INTERFACE.as_ref().cloned(),
             cache: CacheProperties::default(),
             uncached_properties: None,
             proxy_type: PhantomData,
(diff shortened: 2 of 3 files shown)


bilelmoussaoui:
Actually, my changes are not that great. As clippy complains about `a "const" item should never be interior mutable` and we can't switch to using a static instead of a const as we can't have associated statics so far in rust. The only remaining option is to make  `T::from_static_str` for `ObjectPath`/`InterfaceName`/`BusName` const

zeenix:
> The only remaining option is to make `T::from_static_str` for `ObjectPath`/`InterfaceName`/`BusName` const

That sounds like a good change anyway?

bilelmoussaoui:
It is yes but involves no longer going through TryFrom as that can't be const for obvious reasons

zeenix:
> It is yes but involves no longer going through TryFrom as that can't be const for obvious reasons

That's fine. We just need to ensure the string is checked for validity.

zeenix:
@bilelmoussaoui what's the update on this?

TTWNO:
Since `serde::de::Error:*` are not `const`, you would need to stop using those errors for conversion into `zvariant::Error`. All `serde` errors are put into the `zvariant::Message` variant anyways, so perhaps it isn't that big of a deal to write custom error messages.

zeenix:
> Since `serde::de::Error:*` are not `const`, you would need to stop using those errors for conversion into `zvariant::Error`. All `serde` errors are put into the `zvariant::Message` variant anyways, so perhaps it isn't that big of a deal to write custom error messages.

I'm a bit confused. AFAIK there are no errors involved here. :thinking:

TTWNO:
In order to make `from_static_str` const we must make the checker function const. That checket function currently calls non-const functions from the `serde:::de:Error` trait. But, even if you try to remove that, and wtite the error messages yourself (in this case into `zvariant::Error::Message(...)`, you van't do that either because the `Message` variant accepts a string, which is not possible to create a non-empty one of in const.

This is specifically related to `ObjectPath`, but I suspect it applies elsewhere.

zeenix:
Thanks for explaining. My memory of this PR's details was a bit foggy.

> That checket function currently calls non-const functions from the `serde:::de:Error` trait. But, even if you try to remove that

I think it shouldn't be using `serde::de::Error` at all. We should have our own variants used there in any case (regardless of this PR).

>, and wtite the error messages yourself (in this case into `zvariant::Error::Message(...)`, you van't do that either because the `Message` variant accepts a string, which is not possible to create a non-empty one of in const.

We need dedicated variants then and they can avoid `String`.

TTWNO:
> We need dedicated variants then and they can avoid String.

I understand. I'll give this one a go once I'm done with #496 ... and I promise I'll read the contributors guide first this time ;)

zeenix:
> I understand. I'll give this one a go once I'm done with #496 ... and I promise I'll read the contributors guide first this time ;)

Awesome! I wish most people had your positive attitude towards my perfectionism. :)

MaxVerevkin:
> and then manually convert it into zvariant::Error

Maybe zvariant::Error can have a variant which contains this new enum?

zeenix:
> Are you okay with that?

I guess that's the only option so yeah.

> Maybe zvariant::Error can have a variant which contains this new enum?

Sure. We can do something like `zvariant::Error::MaxDepthExceeded` variant. However, can you please explain a bit why that new struct will be ok to drop in const contexts but not `zvariant::Error` itself? :thinking:

TTWNO:
> However, can you please explain a bit why that new struct will be ok to drop in const contexts but not zvariant::Error itself?

`String`, and `Arc<_>` can both not be dropped in `const` contexts.

zeenix:
@bilelmoussaoui I'm assuming you won't be finishing this anytime soon. @TTWNO if you want to do this, you'll need to create a new PR anyway so I'm closing this.

PR #483, opened 2023-09-29
Bump most dependencies

linkmauve:
Only syn and event-listener are still outdated, as they required quite some more work to fix. Plus syn is already handled in #404.

### DIFF
--- a/zbus/src/connection/socket/unix.rs
+++ b/zbus/src/connection/socket/unix.rs
@@ -1,6 +1,8 @@
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
 use std::io;
+#[cfg(any(target_os = "android", target_os = "linux"))]
+use std::os::unix::io::BorrowedFd;
 #[cfg(unix)]
 use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 #[cfg(all(unix, not(feature = "tokio")))]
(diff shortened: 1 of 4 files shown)


zeenix:
> Plus syn is already handled in #404.

Indeed but it's gone stale though and I don't think the author will have time to finish it. So its up for grabs. :)

zbus/src/connection/socket/unix.rs
```diff
@@ -319,8 +319,11 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<ConnectionCredentials>
     {
         use nix::sys::socket::{getsockopt, sockopt::PeerCredentials};
 
-        let fd = fd.as_raw_fd();
-        getsockopt(fd, PeerCredentials)
+        // TODO: get this BorrowedFd directly from get_unix_peer_creds(), but this requires a
+        // 'static lifetime due to the Task.
```
zeenix, line 323: We need to actually move everything to use Owned and BorrowFd and I've been working on that (to complete #395) this week.

zeenix (commented):
Thanks so much for this contribution. I especially appreciate splitting of commits and details in the commit message. :+1: 

While the changes LGTM and if you like we can merge them, as is but if you want to perfect the job, please split the last commit even further: separate commit for each subcrate and (since only a handful commits are updating) please separate one for each upgrade.

Oh and :cherries: on top, if you could add emoji prefixes on the commits, as per recommendation in [our contributing guide](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). :+1:

zbus/src/connection/socket/unix.rs
linkmauve: Thanks for doing that!  This will nicely improve over this PR. :)

linkmauve:
I have split the last commit, added the emoji, and detailed a bit more the changes at the end.

zeenix (approved):
Thanks so much for splitting. For future ref, kindly also add crate name prefixes (e.g `zbus: ` pr `zb: ` for zbus) to commits after the emoji (see other commits in the repo).

Feel free to change in this PR but I will merge very soon, with or without this change.

zeenix:
@linkmauve oh, seems you've to fix the CI: https://github.com/dbus2/zbus/actions/runs/6354805110/job/17267043161?pr=483 . Since you'll be fixing that, I guess best also add the missing prefixes on the commits. :)

Issue #309, opened 2023-01-31
Support for a shutdown signal


zeenix:
In GitLab by @peelz on Jan 31, 2023, 02:08

zbus could have an API for requesting the connection to be closed.

This ties in #307 

As an example, here's how [hyper](https://crates.io/crates/hyper) handles this:
```rust
let server = Server::bind(&addr)
    .serve(make_service);

// Prepare some signal for when the server should start shutting down...
let (tx, rx) = tokio::sync::oneshot::channel::<()>();
let graceful = server
    .with_graceful_shutdown(async {
        rx.await.ok();
    });

// Await the `server` receiving the signal...
if let Err(e) = graceful.await {
    eprintln!("server error: {}", e);
}

// And later, trigger the signal by calling `tx.send(())`.
let _ = tx.send(());
```

For comparison, the simple "wait forever" implementation with no shutdown signal:

```rs
let server = Server::bind(&addr)
    .serve(make_service);

// Run forever-ish...
if let Err(err) = server.await {
    eprintln!("server error: {}", err);
}
```

notpeelz:
That username isn't mine :p

tacuna:
I'm bumping into this issue, looking at the client example: https://dbus2.github.io/zbus/client.html#signals to handle (multiple) signals. What is the current way of stopping? The only way i could find is calling drop but that is not practical.

Use-case: react to unix signals, for example with the signal-hook crate: https://docs.rs/signal-hook-async-std/latest/signal_hook_async_std/

zeenix:
"bumping" isn't very useful. Someone just needs to implement it.

tacuna:
Apologies, was more looking for confirmation that it is indeed currently not possible and this is the issue to track it

zeenix:
> The only way i could find is calling drop but that is not practical.

How so? You can also use channels (or `event-listener`, which is already a zbus dep) to wait in the main task and signal from another task(s)/threads.

> Use-case: react to unix signals, for example with the signal-hook crate: https://docs.rs/signal-hook-async-std/latest/signal_hook_async_std/

That's no problem.

zeenix:
> Apologies, was more looking for confirmation that it is indeed currently not possible and this is the issue to track it

Actually, it's me who should apologize. I was super tired when I read your comment and missed the "into" in "I'm bumping into this issue". I'm sorry.

Hopefully I'll get around to fixing this soon enough.

zeenix:
[...]

PR #485, opened 2023-10-04
Add cancellation token

Basher4:
There were requests to support graceful shutdown #309 and awaitable connections #307. This change attacks both by adding an option to supply a cancellation token.

Once the token is cancelled, the `socket reader` task is stopped. From what I understand, that in turn lets the object server consume all outstanding messages and once that is done, the dbus interfaces are deregistered.


Basher4:
Doesn't work actually :(

PR #488, opened 2023-10-07
🚸 zb: `Message` now immutable

zeenix:
As it should be.

### DIFF
--- a/zbus/src/blocking/connection/mod.rs
+++ b/zbus/src/blocking/connection/mod.rs
@@ -3,7 +3,7 @@
 use enumflags2::BitFlags;
 use event_listener::EventListener;
 use static_assertions::assert_impl_all;
-use std::{io, num::NonZeroU32, ops::Deref, sync::Arc};
+use std::{io, ops::Deref, sync::Arc};
 use zbus_names::{BusName, ErrorName, InterfaceName, MemberName, OwnedUniqueName, WellKnownName};
 use zvariant::ObjectPath;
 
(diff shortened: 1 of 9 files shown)


elmarco:
I really don't have your expertise, but this change makes me feel uneasy.. Fortunately, it's quite low-level, and if we need to add back "unsafe" hatch  later, it should be possible...

Since the serial_num is assigned once when the message is created, you can't reuse it if you need to broadcast it to several peers, or again later. Acktually it wasn't possible, right? And it won't guarantee ordering of sent message serials anymore.. but that should not be an issue either, as long as they are unique.

 And for FDs, that means that you will now *have* to `dup()` them to get them, which may create extra issues that were possible to avoid previously. Other than that, if you are confident that's the way to go, then ok.

zeenix:
> I really don't have your expertise, but this change makes me feel uneasy.. Fortunately, it's quite low-level, and if we need to add back "unsafe" hatch later, it should be possible...

We can (if needed) bring unsafe hack back, yes. Adding new API is never problematic. :)

> Since the serial_num is assigned once when the message is created, you can't reuse it if you need to broadcast it to several peers or again later.

serial number is only really relevant to method calls and I don't know of any possible use case for that. Don't even think it's possible with bus involved.

OTOH signals are what you broadcast but there serial is irrelevant anyway and with bus (typical case), you still send it out once and the bus broadcasts for you.

> And it won't guarantee ordering of sent message serials anymore.. but that should not be an issue either, as long as they are unique

Exactly, that's not an issue or at least it should not be, as per the spec^1.

> And for FDs, that means that you will now _have_ to `dup()` them to get them,

Only if you want to own/keep them yourself. At least currently, you can still deserialize to unowned `Fd` type, where there is no duplication involved.

I say "currently" because this might have to change for completing #395. I will try by best to still support deserializing to `BorrowedFd` but I can't promise it. In any case, that's a different future PR/discussion. :)

> which may create extra issues that were possible to avoid previously.

I can't think of any issues that can't be easily resolved with existing API after this PR. Do you have a concrete example?

---

^1 Admittedly, this will likely make things worse for xdg-dbus-proxy (see https://github.com/flatpak/xdg-dbus-proxy/issues/46) but that's on them to make assumptions not backed by the specification.

PR #499, opened 2023-10-28
✨ zx: add `--split-interfaces` option

FineFindus:
Adds an option to split interfaces into individual files, simplifying maintenance and eliminating the need to extract them from a single large output file. This is especially helpful when dealing with files containing many different interfaces, such as the [org.freedesktop.UDisks2](https://github.com/storaged-project/udisks/blob/a76eda89a4c747f12dc05670f376d95d8ec4cd45/data/org.freedesktop.UDisks2.xml) one.
 The resulting files are stored in the current directory, with names based on the interface name.


zbus_xmlgen/src/main.rs
```diff
@@ -113,29 +114,55 @@ fn main() -> Result<(), Box<dyn Error>> {
...
     Ok(())
 }
 
 fn write_interfaces(
     interfaces: &[&Interface<'_>],
     standard_interfaces: &[&Interface<'_>],
-    service: Option<BusName<'_>>,
-    path: Option<ObjectPath<'_>>,
+    service: &Option<BusName<'_>>,
+    path: &Option<ObjectPath<'_>>,
```
zeenix, line 165: Never seen Option being passed as ref and seems unrelated to the commit. The name structs are free (no runtime cost) to clone so I'd not bother.

zeenix (changes requested):
LGTM aside from 1 nitpick.

zeenix:
@FineFindus oh and the CI is failing.

zeenix:
> @FineFindus oh and the CI is failing.

Oh, looks like that's unrelated to your commits. Please rebase on current main and it should be fine. Windows one might need a  rerun though. For some reason, the pkg-config download from sourceforge fails often. :shrug:

zbus_xmlgen/src/main.rs
```diff
@@ -161,8 +161,8 @@ fn main() -> Result<(), Box<dyn Error>> {
 fn write_interfaces(
     interfaces: &[&Interface<'_>],
     standard_interfaces: &[&Interface<'_>],
-    service: &Option<BusName<'_>>,
-    path: &Option<ObjectPath<'_>>,
+    service: Option<BusName<'_>>,
+    path: Option<ObjectPath<'_>>,
```
zeenix, line 165: Please squash this fixup commit into the relevant commit and force push. :pray:

zeenix (commented):
Are the double shortlogs in the last commit on purpose?

```
 ✨ zx: add split interfaces option

♻️ zx: clone service and path
```

FineFindus:
No real reason, just git deciding to add it automatically when squashing and me not noticing :). Fixed it.

zeenix:
> No real reason, just git deciding to add it automatically when squashing and me not noticing :). Fixed it.

Thanks! but could you add some description there? Feel free to c&p the PR description.

zbus_xmlgen/src/main.rs
```diff
@@ -25,6 +25,7 @@ fn usage() {
   zbus-xmlgen <interface.xml>
   zbus-xmlgen --system|--session <service> <object_path>
   zbus-xmlgen --address <address> <service> <object_path>
+  zbus-xmlgen <interface> --split-interfaces
```
zeenix, line 28: * This should be `<interface.xml>`, shouldn't it? 
* Does `--split-interfaces` not work with `--system|--session|--address` options? :thinking: This message gives the impression that it doesn't.

zeenix (changes requested):
I also now notice an inconsistency here. Normally the generated code is just thrown on the console, except (now) when `--split-interfaces` is specified. Perhaps, it'd be better and we make `--split-interfaces` implied by default? We can add a different option `--output` (short form: `-o`) to output to a specific file. If user specifies `-o -`, we output everything to the stdout. This will also be consistent with how other tools (e.g gdbus) work I think. WDYT?

zbus_xmlgen/src/main.rs
FineFindus: The option works with all arguments, which is why I set it to `<interface>`. I didn't want to add it to the end of the other options, in fear someone would interpret it as required :)

FineFindus:
Frankly, I'm not a big fan of creating a bunch of files without explicit user permission. This is especially the case when you accidentally run the command (e.g. to try something out) and now have to clean up all the newly created files. 
So using `--split-interfaces` without `-o` would just print the interfaces to stdout as it does now, and only create files if `-o` is specified?

zbus_xmlgen/src/main.rs
zeenix: `<interface>` is opaque and doesn't tell the user what it means. My first guess as a user would be that you have to specify an interface name, which is wrong.

zeenix:
> This is especially the case when you accidentally run the command (e.g. to try something out) and now have to clean up all the newly created files.

That's not our problem to solve really. They could run the command accidently with `--split-interfaces` as well.

> So using `--split-interfaces` without `-o` would just print the interfaces to stdout as it does now,

What's `--split-interfaces` doing then? It's been just ignored and that's very unexpected. Even if we keep the current behavior, this should result in an error. I really think we should change the default behavior as that would mean everything is consistent.

zbus_xmlgen/src/main.rs
```diff
@@ -25,10 +25,17 @@ fn usage() {
   zbus-xmlgen <interface.xml>
   zbus-xmlgen --system|--session <service> <object_path>
   zbus-xmlgen --address <address> <service> <object_path>
+  zbus-xmlgen <interface.xml> -o <output-file>
```
zeenix, line 28: this doesn't need to be listed separately. `-o` is supported in all cases. So I'd just append `[-o|--output <file_path>]` to the 3 lines above and remove this line.

zeenix (changes requested):
The default behavior is changing here (which is a pretty big change and I'd like to see it hilighted) and the commit log summary doesn't capture it. It would be best to split the behavior change into a separate commit that comes before this one.

zeenix (approved):
As I wrote in the PM:

> you did things the other way around. :(
> the change of behavior still doesn't get hilighted in the commit history
> I suggested that you first change the behavior so output is no longer sent to stdout but to separate interface files, and then add the -o so user can modify (including - for getting the old behavior)
> but oh well
> I'll just have to try hard and remember when writing the release notes

PR #500, opened 2023-10-31
zv: store entries of Dict as BTreeSet to keep it ordered

### LINKED ISSUE #484: zvariant: Implement `PartialEq` for `Dict` that ignores the order of `entries`
The implemented `PartialEq` trait for a `HashMap` in `std` does not take the order into account but `zvariant::Dict` does.

Example:

```rust
let mut map1: HashMap<i32, &str> = HashMap::new();
map1.insert(1, "one");
map1.insert(2, "two");
map1.insert(3, "three");

let mut map2: HashMap<i32, &str> = HashMap::new();
map2.insert(3, "three");
map2.insert(1, "one");
map2.insert(2, "two");

assert_eq!(map1, map2); // does not panic

let dict1 = zbus::zvariant::Dict::from(map1);
let dict2 = zbus::zvariant::Dict::from(map2);

assert_eq!(dict1, dict2); // panics
```

This stems from `Dict` not implementing `PartialEq` but [using derive](https://github.com/dbus2/zbus/blob/acc34f5cbbe3e385f6e1ed2281b8f77edb23e3fa/zvariant/src/dict.rs#L18) and hence `self.entries` is compared against `other.entries` and since both of them are `Vec`, the order suddenly matters.

While this is not a problem in flat structures where `Dict` can be converted into a `HashMap` for such a comparison, it becomes one in nested structures like `HashMap<String, OwnedValue>`.

When two instances of such a `HashMap` are compared with some values being `Dict`s the equality of these `HashMap` now depends on the order in which the `Dict`s have been constructed which cannot be controlled in most cases, especially in de-serialization.

Example:

```rust
let mut nested_map1: HashMap<&str, OwnedValue> = HashMap::new();
nested_map1.insert("one", OwnedValue::from(dict1));
nested_map1.insert("two", OwnedValue::from(Value::from("foo")));

let mut nested_map2: HashMap<&str, OwnedValue> = HashMap::new();
nested_map2.insert("one", OwnedValue::from(dict2));
nested_map2.insert("two", OwnedValue::from(Value::from("foo")))
assert_eq!(nested_map1, nested_map2); // panics
```

mguentner:
adds a test to assert the desired behavior described in #484.

Credits to @sivizius for reviewing and contributing.

closes #484

### DIFF
--- a/zvariant/src/dict.rs
+++ b/zvariant/src/dict.rs
@@ -15,7 +15,7 @@ use crate::{value_display_fmt, Basic, DynamicType, Error, Signature, Type, Value
 ///
 /// [`Value`]: enum.Value.html#variant.Dict
 /// [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
-#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord)]
+#[derive(Debug, Clone, Hash, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Dict<'k, 'v> {
     entries: BTreeSet<DictEntry<'k, 'v>>,
     key_signature: Signature<'k>,
@@ -261,7 +261,7 @@ where
 
 // TODO: Conversion of Dict from/to BTreeMap
 
-#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
+#[derive(Debug, Clone, Hash, PartialOrd, Ord, PartialEq, Eq)]
 struct DictEntry<'k, 'v> {
     key: Value<'k>,
     value: Value<'v>,
--- a/zvariant/src/dict.rs
+++ b/zvariant/src/dict.rs
@@ -15,7 +15,7 @@ use crate::{value_display_fmt, Basic, DynamicType, Error, Signature, Type, Value
 ///
 /// [`Value`]: enum.Value.html#variant.Dict
 /// [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Dict<'k, 'v> {
     entries: Vec<DictEntry<'k, 'v>>,
     key_signature: Signature<'k>,
@@ -261,7 +261,7 @@ where
 
 // TODO: Conversion of Dict from/to BTreeMap
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
 struct DictEntry<'k, 'v> {
     key: Value<'k>,
     value: Value<'v>,
--- a/zvariant/src/dict.rs
+++ b/zvariant/src/dict.rs
@@ -1,5 +1,5 @@
 use std::{
-    collections::HashMap,
+    collections::{BTreeSet, HashMap},
     fmt::{Display, Write},
     hash::BuildHasher,
 };
@@ -17,7 +17,7 @@ use crate::{value_display_fmt, Basic, DynamicType, Error, Signature, Type, Value
 /// [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
 #[derive(Debug, Clone, PartialEq, PartialOrd, Eq, Ord)]
 pub struct Dict<'k, 'v> {
-    entries: Vec<DictEntry<'k, 'v>>,
+    entries: BTreeSet<DictEntry<'k, 'v>>,
     key_signature: Signature<'k>,
     value_signature: Signature<'v>,
     // should use a separate lifetime or everything should use the same but API break.
@@ -32,7 +32,7 @@ impl<'k, 'v> Dict<'k, 'v> {
         let signature = create_signature(&key_signature, &value_signature);
 
         Self {
-            entries: vec![],
+            entries: BTreeSet::new(),
             key_signature,
             value_signature,
             signature,
@@ -56,7 +56,7 @@ impl<'k, 'v> Dict<'k, 'v> {
         check_child_value_signature!(self.key_signature, key.value_signature(), "key");
         check_child_value_signature!(self.value_signature, value.value_signature(), "value");
 
-        self.entries.push(DictEntry { key, value });
+        self.entries.insert(DictEntry { key, value });
 
         Ok(())
     }
@@ -70,7 +70,7 @@ impl<'k, 'v> Dict<'k, 'v> {
         check_child_value_signature!(self.key_signature, K::signature(), "key");
         check_child_value_signature!(self.value_signature, value.dynamic_signature(), "value");
 
-        self.entries.push(DictEntry {
+        self.entries.insert(DictEntry {
             key: Value::new(key),
             value: Value::new(value),
         });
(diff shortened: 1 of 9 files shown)


zvariant/src/lib.rs
```diff
@@ -1177,7 +1177,7 @@ mod tests {
         let v: Value<'_> = dict.into();
         assert_eq!(v.value_signature(), "a{sv}");
         let encoded = to_bytes(ctxt, &v).unwrap();
-        assert_eq!(dbg!(encoded.len()), 68);
+        assert_eq!(dbg!(encoded.len()), 66);
```
mguentner, line 1180: In the original code where the order of the `Dict` entries mattered, the internal entries looked like this:
```
       [ { key: "hello", value: "there"}, { key:"bye", "now"} ]
```
with the change to `BTreeSet` they are reordered to
```
       [  { key:"bye", "now"}, { key: "hello", value: "there"} ]
```

The serializer (zvariant/src/dbus/ser.rs:134) is adding various paddings for alignment. I am quite certain that this is the reason the len of the encoded bytes goes down but the content remains correct (see following lines of the test).

zeenix:
> adds a test to assert the desired behavior described in #484.

Cool, thanks for doing this.

> Credits to @sivizius for reviewing and contributing.

We don't see the review so will have to take your word for it. :)

zvariant/src/signature.rs
```diff
@@ -20,29 +22,13 @@ use crate::{signature_parser::SignatureParser, Basic, EncodingFormat, Error, Res
...
-            (Bytes::Borrowed(b), Bytes::Owned(o)) => **b == *o.as_ref(),
-            (Bytes::Static(s), Bytes::Owned(o)) => **s == *o.as_ref(),
-            (Bytes::Borrowed(b1), Bytes::Borrowed(b2)) => **b1 == **b2,
-            (Bytes::Owned(b1), Bytes::Owned(b2)) => *b1.as_ref() == *b2.as_ref(),
-            (Bytes::Owned(b1), Bytes::Static(s2)) => *b1.as_ref() == **s2,
-            (Bytes::Static(s1), Bytes::Static(s2)) => **s1 == **s2
-        }
-    }
-}
-
```
zeenix: This is a fixup commit. please squash it into the relevant previous commit and update the log.

zeenix (changes requested):
In general looking good. Apart from the inline comments in the commits:

* It looks like @sivizius added 2 commits after to improve your previous commits. They should be squashed back into your commits so we've a nice linear history.
* Great work on keeping the commits atomic and adding rational to each commit. :+1: Could I also convince you to add emojis as per the recommendation in our contribution guide? :) It just keeps the history very consistent.

zvariant/src/dict.rs
```diff
@@ -259,9 +259,10 @@ where
     }
 }
 
+
 // TODO: Conversion of Dict from/to BTreeMap
 
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialOrd, Ord, PartialEq, Eq)]
```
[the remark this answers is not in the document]
zeenix: A commit can have a dependency on a previous commit, so if I understand correctly, yes.

mguentner:
formating issues should be fixed now

sivizius:
> > Credits to @sivizius for reviewing and contributing.
> 
> We don't see the review so will have to take your word for it. :)

Here is some context: https://github.com/mguentner/zbus/commit/db9474a1889dedd17b26996f9a6ff7f6652e411d. 

We tried to conform to [C-COMMON-TRAITS](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits)](https://rust-lang.github.io/api-guidelines/interoperability.html#types-eagerly-implement-common-traits-c-common-traits), so even though `impl Hash for Value<'_>` is not relevant to solve #484, it was provided for completeness. 

`BTreeSet` was chosen over `BTreeMap`, so dictionary-entries with the same key but different values are still allowed, which is the current behaviour. We like to avoid any breakage, even though, we are not sure if there is any use case for this.

The custom comparison for `Bytes` was implemented similar to [`std::borrow::Cow`](https://doc.rust-lang.org/src/alloc/borrow.rs.html#345-380), because, as mentioned in the commit message, the ownership state of the data should not impact the result of any comparison of the data itself.

There is an open PR https://github.com/mguentner/zbus/pull/4 by myself regarding comparison of floating point values using [`f64::total_cmp`](https://doc.rust-lang.org/std/primitive.f64.html#method.total_cmp), which is probably what we want in case of NaN-values. This methods implements the `totalOrder`-predicate [doi:10.1109/IEEESTD.2008.4610935](https:/doi.org/10.1109/IEEESTD.2008.4610935), section 5.10. Due to the `partial_cmp` before, only paragraph d is relevant.

zeenix (approved):
Great work!

zeenix:
> There is an open PR [mguentner#4](https://github.com/mguentner/zbus/pull/4) 

Thanks. I will have a look at it soon. Somehow I didn't get an email notification for that. :shrug: 

> This methods implements the `totalOrder`-predicate [doi:10.1109/IEEESTD.2008.4610935](https:/doi.org/10.1109/IEEESTD.2008.4610935), section 5.10. Due to the `partial_cmp` before, only paragraph d is relevant.

That link is a 404 for btw. Is it restricted?

sivizius:
> Thanks. I will have a look at it soon. Somehow I didn't get an email notification for that. 🤷

It is an PR in the repository with the branch of this PR. Maybe that is why?

> That link is a 404 for btw. Is it restricted?

Copy-Paste-Fail, I have fixed it.

zeenix:
> > Thanks. I will have a look at it soon. Somehow I didn't get an email notification for that. 🤷
> 
> It is an PR in the repository with the branch of this PR. Maybe that is why?

You mean that branch is based on the branch of this PR? I don't think that matters at all. I think there may be an issue with my email alias' forwarding.

> > That link is a 404 for btw. Is it restricted?
> 
> Copy-Paste-Fail, I have fixed it.

Thanks. Happens to us all. :)

zeenix:
> @mguentner please fix the formatting issue (make sure to squash int the relevant commit) and we should be ready to merge. Thanks so much for this.

@mguentner Please don't give up this close to the finish line. :)

mguentner:
On it. Won't stall it.

mguentner:
@zeenix Should be good to go. Let's see the workflow result :slot_machine:

zvariant/src/value.rs
```diff
@@ -100,6 +101,22 @@ pub enum Value<'a> {
...
+
+impl Ord for Value<'_> {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.partial_cmp(other)
+            .unwrap_or_else(|| match (self, other) {
+                (Self::F64(lhs), Self::F64(rhs)) => lhs.total_cmp(rhs),
+                // `partial_cmp` is returns `Some(_)` if either the discriminants are different
+                //   or if both the left hand side and right hand side is `Self::F64(_)`,
+                //   because `f64` is the only type in this enum, that does not implement `Ord`.
+                // This `match`-arm is therefore unreachable.
```
zeenix, line 144: Nice catch! I find it strange that the std API is making this a dynamic problem. In the case of `PartialEq` it's decided at compile time, what can and can't be compared. Anyway:

* A typo here: the first `is` can be dropped.
* while you're removing that anyway, might as well remove the extra space on line 112 and 113.

zeenix (changes requested):
LGTM otherwise.

zvariant/src/value.rs
mguentner: :heavy_check_mark:

PR #502, opened 2023-11-03
zv,zn: impl `Basic` for applicable types

EHfive:
Similar to `String` is `Basic`, `OwnedObjectPath` should also be `Basic`. 
This allows to convert `OwnedValue` into `HashMap` with `OwnedObjectPath` key type.

### DIFF
--- a/zbus_names/src/bus_name.rs
+++ b/zbus_names/src/bus_name.rs
@@ -5,7 +5,10 @@ use core::{
 };
 use std::{borrow::Cow, sync::Arc};
 
-use crate::{Error, OwnedUniqueName, OwnedWellKnownName, Result, UniqueName, WellKnownName};
+use crate::{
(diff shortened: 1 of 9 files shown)


zeenix (approved):
LGTM. Could I ask you to also do the same for other Owned* types based on `Str` in the same PR (could be the same commit as well)?

EHfive:
Seems there is only `OwnedSignature` left in zvariant, should I also implement `Basic` for name types that wrapping `Str` in `zbus_names`?

zeenix:
> Seems there is only `OwnedSignature` left in zvariant, should I also implement `Basic` for name types that wrapping `Str` in `zbus_names`?

Yes please! Although best to put that in a separate commit.

zeenix (approved):
Nice, many thanks.

PR #504, opened 2023-11-05
🐛 zb: use the correct traversal order for node tree introspection

EHfive:
The previous implementation is just incomplete and would cause all child nodes to be concated one by one instead of being enclosed.

Also updated the e2e tests to test this case.

Fixes 916b4cce00a0b85cb6e9a19695f9f1387832682e

---
**The bug** introduced by 916b4cce00a0b85cb6e9a19695f9f1387832682e

```
$ busctl introspect --user com.example.Service /
Invalid introspection data: duplicate method 'Get' on interface 'org.freedesktop.DBus.Properties'.

$ busctl tree --user com.example.Service
Failed to introspect object /example of service com.example.Service: Unknown object '/example'
Failed to introspect object /Service of service com.example.Service: Unknown object '/Service'
Failed to introspect object /com/Service of service com.example.Service: Unknown object '/com/Service'
└─ /com
  └─ /com/example
    └─ /com/example/Service
```

### DIFF
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -3,7 +3,7 @@
 use event_listener::{Event, EventListener};
 use serde::Serialize;
 use std::{
-    collections::{hash_map::Entry, HashMap, VecDeque},
+    collections::{hash_map::Entry, HashMap},
     fmt::Write,
     marker::PhantomData,
     ops::{Deref, DerefMut},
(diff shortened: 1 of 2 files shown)


zbus/tests/e2e.rs
```diff
@@ -473,6 +473,30 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+    let root_introspect_proxy = zbus::fdo::IntrospectableProxy::builder(&conn)
+        .destination("org.freedesktop.MyService")?
+        .path("/")?
+        .build()
+        .await?;
+    debug!("Created: {:?}", root_introspect_proxy);
+
+    let root_xml = root_introspect_proxy.introspect().await?;
+    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes())
+        .map_err(|e| Error::Failure(e.to_string()))?;
```
zeenix, line 485: Seems we're missing a `impl From<zbus_xml::Error> for zbus::Error`, which would make this `map_err` call redundant. Could you please add that instead (in a separate commit that comes before this one)?

zbus/src/object_server/mod.rs
```diff
@@ -298,38 +298,68 @@ impl Node {
...
+        let mut stack = Vec::new();
+        stack.push(Fragment::Node {
+            name: "",
+            node: self,
+            level: 0,
+        });
+
+        // This can be seen as traversing the fragment tree in pre-order DFS with formatted XML
+        // fragment, splitted `Fragment::Node`s and `Fragment::End` being current node, left
+        // subtree and right leaf respectively.
```
zeenix, line 321: kudos for adding nice comment to help read the code easily.

zbus/tests/e2e.rs
```diff
@@ -473,6 +473,30 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+            .nodes()
+            .iter()
+            .find(|&n| n.name().is_some_and(|n| n == name))
+            .expect("Child node not exist");
+    }
+    let iface = node
+        .interfaces()
+        .iter()
+        .find(|i| i.name() == "org.freedesktop.MyIface");
+    assert!(iface.is_some());
```
zeenix, line 498: nit: I think it's best to use `any` instead of `find` here.

zeenix (changes requested):
Looks good in general. Thanks for fixing this regression.

zbus/tests/e2e.rs
EHfive: Not sure where to add it, zbus_xml does not use zbus, and zbus only use zbus_xml for testing.. Seems a feature gate might be needed for either case.

zeenix: > zbus only use zbus_xml for testing

Right i forgot zbus doesn't depend on zbus_xml.

zbus/tests/e2e.rs
```diff
@@ -473,6 +473,30 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+            .nodes()
+            .iter()
+            .find(|&n| n.name().is_some_and(|n| n == name))
+            .expect("Child node not exist");
+    }
+    let has_my_iface = node
+        .interfaces()
+        .iter()
+        .any(|i| i.name() == "org.freedesktop.MyIface");
+    assert!(has_my_iface);
```
zeenix, line 498: could have just wrapped the whole call in `assert!` but this is ok too.

zeenix (approved):

PR #505, opened 2023-11-05
zx: use clap for arg parsing

FineFindus:
Update `zbus-xmlgen` to use [`clap`](https://crates.io/crates/clap). This change converts the `--session`, `--system`, `--address`, and `file` arguments into subcommands.

Builds on https://github.com/dbus2/zbus/pull/499 to also add the `--split-interfaces` options.

### Possible future improvements
- Improve help/documentation by adding descriptions/comments to arguments.
- Generate completions and a man page

### DIFF
--- a/zbus_xmlgen/Cargo.toml
+++ b/zbus_xmlgen/Cargo.toml
@@ -29,6 +29,7 @@ zbus = { path = "../zbus", version = "4.0.0" }
 zbus_xml = { path = "../zbus_xml", version = "4.0.0" }
 zvariant = { path = "../zvariant", version = "4" }
 snakecase = "0.1.0"
+clap = { version = "4.4", features = ["derive", "wrap_help"] }
 
 [dev-dependencies]
 pretty_assertions = "1.3"
--- a/zbus_xmlgen/src/cli.rs
+++ b/zbus_xmlgen/src/cli.rs
@@ -1,0 +1,45 @@
+use std::path::PathBuf;
+
+use clap::Parser;
+
+#[derive(Parser, Debug)]
+#[command(author, version, about, long_about = None)]
+pub struct Args {
+    #[clap(subcommand)]
+    pub command: Command,
+
+    /// Specify the destination for saving the output. If no argument is provided, the parsed
+    /// interfaces will be stored in separate files. If a filename is provided, the output will
+    /// be saved to that file. Use '-' to print the output to stdout.
+    #[clap(short, long, allow_hyphen_values = true)]
+    pub output: Option<String>,
+}
+
+#[derive(Parser, Debug, Clone)]
+pub enum Command {
+    /// Generate code for interfaces in the specified file.
+    #[clap()]
+    File { path: PathBuf },
+
+    /// Generate code for interfaces from the specified system service.
+    #[clap()]
+    System {
+        service: String,
+        object_path: String,
+    },
+
+    /// Generate code for interfaces from the current users session.
+    #[clap()]
+    Session {
+        service: String,
+        object_path: String,
+    },
+
+    /// Generate code for interfaces from the specified address.
+    #[clap()]
+    Address {
+        address: String,
+        service: String,
+        object_path: String,
+    },
+}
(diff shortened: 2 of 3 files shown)


zbus_xmlgen/src/main.rs
```diff
@@ -19,112 +18,162 @@ use zbus_xml::{Interface, Node};
...
+    service: String,
+    object_path: String,
+) -> Result<
+    (
+        Node<'a>,
+        Option<BusName<'a>>,
+        Option<ObjectPath<'a>>,
+        String,
+    ),
+    Box<dyn Error>,
```
FineFindus, line 104: This results in a [`type_complexity` clippy warning](https://rust-lang.github.io/rust-clippy/master/index.html#/type_complexity). Not sure if it makes sense to create a type just for this, since the returned tuple is immediately destructured.

zeenix: creating a type is so trivial that I always heed to this warning. :) Coming up with names is the issue. :)

zeenix:
@FineFindus hey. You think you'll be able to finish this soon?

FineFindus:
It's basically done, the only two things missing are #499 being merged and the MSRV CI failing. I'm still not quite sure why, for some reason it tries to build a newer clap version than specified. Maybe a caching problem? We could try a re-run, hoping that the cache is now old enough not to be an issue-.

FineFindus:
Rebased the pr to the latest `main`. I had a bit of trouble rebasing with the additional changes in https://github.com/dbus2/zbus/pull/499, so I ported the code in a single commit. I also found some minor issues with the previous version along the way, see the following commits.

zbus_xmlgen/Cargo.toml
```diff
@@ -29,7 +29,7 @@ zbus = { path = "../zbus", version = "4.0.0" }
 zbus_xml = { path = "../zbus_xml", version = "4.0.0" }
 zvariant = { path = "../zvariant", version = "4" }
 snakecase = "0.1.0"
-clap = { version = "4.4", features = ["derive"] }
+clap = { version = "4.4", features = ["derive", "wrap_help"] }
```
zeenix, line 32: This should be in the previous commit.

zbus_xmlgen/src/cli.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// be saved to that file. Use '-' to print the output to stdout.
+    #[clap(short, long, allow_hyphen_values = true)]
+    pub output: Option<String>,
+}
+
+#[derive(Parser, Debug, Clone)]
+pub enum Command {
+    /// Generate code for interfaces in the specified file.
+    #[clap()]
+    File { path: PathBuf },
```
zeenix, line 22: Having these as subcommands mean the interface is changing already? If so, this should be mentioned in the commit log.

zbus_xmlgen/src/main.rs
```diff
@@ -140,6 +140,9 @@ fn write_interfaces(
     let mut process = match Command::new("rustfmt")
         .stdin(Stdio::piped())
         .stdout(Stdio::piped())
+        // rustfmt may post warnings about features not being enables on non-nightly rust
```
zeenix, line 149: * typo: enabled -> enabled.
* "stable" might be a better way to say "non-nightly". :wink:

zbus_xmlgen/src/cli.rs
```diff
@@ -11,7 +11,7 @@ pub struct Args {
     /// Specify the destination for saving the output. If no argument is provided, the parsed
     /// interfaces will be stored in separate files. If a filename is provided, the output will
     /// be saved to that file. Use '-' to print the output to stdout.
-    #[clap(short, long, allow_hyphen_values = true)]
+    #[clap(short, long, allow_hyphen_values = true, global = true)]
```
zeenix, line 14: A bit of explanation in the commit log about what "global" means here would be great.

zeenix (changes requested):
Thanks for updating. Some questions and concerns still. Also the last commit feels very much like a fixup of the second commit. Is it? If so, please squash it into that. If not, please add some description to the commit.

zeenix (approved):
LGTM. There were issues in the testing but as we discussed in the PMs, they're not caused by your changes here.

Issue #257, opened 2022-03-31
xmlgen: Incorrect codegen with more complex type structures


zeenix:
In GitLab by @PolyMeilex on Mar 31, 2022, 12:20

Hi, I have a method like this one:
```xml
    <method name="GetAbcd">
      <arg type="u" name="a" direction="out" />
      <arg type="a((ssss)a(siiddada{sv})a{sv})" name="b" direction="out" />
      <arg type="a(iiduba(ssss)a{sv})" name="c" direction="out" />
      <arg type="a{sv}" name="d" direction="out" />
    </method>
```
After piping it into xmlgen I get this type as return value:
```rs
     (
        u32,
        Vec<(String, String, String, String)>,
        Vec<(
            i32,
            i32,
            f64,
            u32,
            bool,
            Vec<(String, String, String, String)>,
        )>,
        std::collections::HashMap<String, zbus::zvariant::OwnedValue>,
    )

```
There is several fields missing in this definition, here is the same type manually written by me to match what xml defines:
```rs
(
    u32,
    Vec<(
        (String, String, String, String),
        Vec<(
            String,
            i32,
            i32,
            f64,
            f64,
            Vec<f64>,
            HashMap<String, zbus::zvariant::OwnedValue>,
        )>,
        HashMap<String, zbus::zvariant::OwnedValue>,
    )>,
    Vec<(
        i32,
        i32,
        f64,
        u32,
        bool,
        Vec<(String, String, String, String)>,
        HashMap<String, zbus::zvariant::OwnedValue>,
    )>,
    HashMap<String, zbus::zvariant::OwnedValue>,
)
```

nagatavit:
Hi, I stumbled on this problem today, and dig a little to find out if it was easy to fix, and it was! I've opened the PR #506, however, I never contributed to an open source project, so I'm kinda lost on what I should do, next.

Can someone give me some pointers for the next steps? :thinking:

zeenix:
> Hi, I stumbled on this problem today, and dig a little to find out if it was easy to fix, and it was! I've opened the PR #506,

Thanks so much. :pray: 

> however, I never contributed to an open source project, so I'm kinda lost on what I should do next.

Really? You did great and even followed the contributors guide to the letter. Most folks don't do it so well.

> Can someone give me some pointers on how to proceed? 🤔

[...]

PR #506, opened 2023-11-07
🐛 zbus_xmlgen: Fix nested signature parser

nagatavit:
Fixes: #257 

When the xml signature had a nested element (such as a nested dict), the parser wasn't cleaning up the closing character after peeking the iterator. Fixed by consuming the next iterator before proceeding.

### DIFF
--- a/zbus_xmlgen/tests/data/sample_object0.rs
+++ b/zbus_xmlgen/tests/data/sample_object0.rs
@@ -1,6 +1,9 @@
 #[dbus_proxy(interface = "com.example.SampleInterface0", assume_defaults = true)]
 trait SampleInterface0 {
 
+    /// BarplexSig method
+    fn barplex_sig(&self, rule: &(&[i32], i32, std::collections::HashMap<&str, &str>, i32, &[i32], i32, &[&str], i32, bool)) -> zbus::Result<Vec<(String, zbus::zvariant::OwnedObjectPath)>>;
+
     /// Bazify method
     fn bazify(&self, bar: &(i32, i32, u32)) -> zbus::Result<zbus::zvariant::OwnedValue>;
 
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -228,7 +228,11 @@ fn to_rust_type(ty: &CompleteType, input: bool, as_ref: bool) -> String {
                 loop {
                     let c = it.peek().unwrap();
                     match **c as char {
-                        STRUCT_SIG_END_CHAR | DICT_ENTRY_SIG_END_CHAR => break,
+                        STRUCT_SIG_END_CHAR | DICT_ENTRY_SIG_END_CHAR => {
+                            // consume the closing character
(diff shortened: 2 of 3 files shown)


zeenix:
Many thanks for this contribution. Before I look closer, could you see what's the issue with the filname? There is a warning about it from Github:

Screenshot from 2023-11-07 13-22-08

zeenix:
> Many thanks for this contribution. Before I look closer, could you see what's the issue with the filname? There is a warning about it from Github:

Oh, I think I know what's happening. You added an emoji to the branch name as well, which git is fine with but Github is not. It's ok. Just keep in mind for any future contributions to not do that. Emoji prefixes in commit log and PR title are more than enough. :)

zbus_xmlgen/tests/data/sample_object0.rs
```diff
@@ -1,6 +1,9 @@
 #[dbus_proxy(interface = "com.example.SampleInterface0", assume_defaults = true)]
 trait SampleInterface0 {
 
+    /// BarplexSig method
+    fn barplex_sig(&self, rule: &(&[i32], i32, std::collections::HashMap<&str, &str>, i32, &[i32], i32, &[&str], i32, bool)) -> zbus::Result<Vec<(String, zbus::zvariant::OwnedObjectPath)>>;
```
zeenix, line 5: Too long a line. The lint part of the CI will fail on this. Please run `cargo +nightly fmt` on you machine to ensure that all code is correctly-formatted.

zeenix (changes requested):
LGTM otherwise.

zbus_xmlgen/tests/data/sample_object0.rs
zeenix: Actually,[ I was wrong](https://github.com/dbus2/zbus/actions/runs/6779306008/job/18441228520?pr=506). :laughing: This file is not part of the build so `cargo fmt` will not see it. You need to run `rustfmt zbus_xmlgen/tests/data/sample_object0.rs` manually. It would be great if we could somehow check the formatting for this file in the CI.

nagatavit: Hi, checking the comments now!

This test file I actually had to disable the auto formatter from my editor :grimacing:, because the test was failing with the generated code and I didn't wanted to tweak with the test (only to add a new case).

Do you think it is necessary to format it? It would be necessary to adapt the test macro I think.

zeenix: > This test file I actually had to disable the auto formatter from my editor 😬, because the test was failing with the generated code and I didn't wanted to tweak with the test (only to add a new case).

You can always have other changes in the same branch/PR, just commit them separately (if you use commandline, `git add -p` is your friend).

> Do you think it is necessary to format it? It would be necessary to adapt the test macro I think.

Right. That would be great. However, you can do that in a separate PR afterwards.

zeenix (approved):

Issue #323, opened 2023-03-06
`emits_changed_signal` for `dbus_interface` properties


zeenix:
While !479 added support of `org.freedesktop.DBus.Property.EmitsChangedSignal` for proxies, it would be good to also support that in interfaces. This is likely a lot less work than it was for properties.

mmasque:
Hi, I'm interested in working on this issue. I have been thinking a bit about an approach in `zbus_macros`. 
I have one design question. Suppose I have the following interface
```rust
#[dbus_interface(name = "org.zbus.MyGreeter1")]
impl Greeter {
    #[dbus_interface(property)]
    async fn foo(&self) -> u64 {
        0
    }
    #[dbus_interface(property)]
    async fn set_foo(&mut self, value: u64) -> zbus::fdo::Result<()> {
        self.foo = value;
        Ok(())
    }
```

- I think it makes sense to allow `emits_changed_signal` to be put in the `foo` header, and use that value to generate the signal function `foo_changed` when `emits_changed_signal` is `true` or `invalidate`. When set to `false` or `const`, the function could either a) still be generated and do nothing or b) not be generated at all.
- If we choose a) above, `set_foo` could always call `foo_changed`, and thus we wouldn't need to pass `emits_changed_signal` to the `set_foo` header. This isn't a very desirable approach, because in the default case we end up with a stub public function the user can call. 
- If we choose b) above, the implementation gets more complicated, because we need to only call the function `foo_changed` when it exists, and the macro code for `set_foo` needs a way to know whether the function exists or not. So there would be some refactoring to do to make sure that `foo` gets processed before `set_foo` and that we store whether the `foo_changed` function was generated.

Would like to hear your thoughts before I PR!

zeenix:
> Would like to hear your thoughts before I PR!

Sure, sounds good to me. About your open question:

> * If we choose b) above, the implementation gets more complicated, because we need to only call the function `foo_changed` when it exists, and the macro code for `set_foo` needs a way to know whether the function exists or not. So there would be some refactoring to do to make sure that `foo` gets processed before `set_foo` and that we store whether the `foo_changed` function was generated.

That shouldn't be difficult and we have a lot of code in our macros that does similar things. It's actually what makes the macro code hard to modify for even small changes. :( If you refactor/clean up the code as part of your PR, that's even better. If you could refactor first in a separate commit before the real change (same PR), that would be great.

[...]

PR #514, opened 2023-11-28
#323 enable `emits_changed_signal` on `dbus_interface`

marcelmasqueprodrive:
Implements  #323: enables interface property macros on getters to be annotated with `emits_changed_signal`.

### DIFF
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -123,6 +123,36 @@ trait MyIface {
 
     #[dbus_proxy(allow_interactive_auth)]
     fn test_interactive_auth(&self) -> zbus::Result<()>;
+
+    #[dbus_proxy(property)]
+    fn emits_changed_default(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_emits_changed_default(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property(emits_changed_signal = "true"))]
+    fn emits_changed_true(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_emits_changed_true(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property(emits_changed_signal = "invalidates"))]
+    fn emits_changed_invalidates(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_emits_changed_invalidates(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property(emits_changed_signal = "const"))]
+    fn emits_changed_const(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_emits_changed_const(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property(emits_changed_signal = "false"))]
+    fn emits_changed_false(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_emits_changed_false(&self, count: u32) -> zbus::Result<()>;
 }
 
 #[derive(Debug, Clone)]
@@ -136,11 +166,24 @@ enum NextAction {
 struct MyIfaceImpl {
     next_tx: Sender<NextAction>,
     count: u32,
+    emits_changed_default: u32,
+    emits_changed_true: u32,
+    emits_changed_invalidates: u32,
+    emits_changed_const: u32,
+    emits_changed_false: u32,
 }
 
 impl MyIfaceImpl {
     fn new(next_tx: Sender<NextAction>) -> Self {
-        Self { next_tx, count: 0 }
+        Self {
+            next_tx,
+            count: 0,
+            emits_changed_default: 0,
+            emits_changed_true: 0,
+            emits_changed_invalidates: 0,
+            emits_changed_const: 0,
+            emits_changed_false: 0,
+        }
     }
 }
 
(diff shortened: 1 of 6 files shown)


zeenix:
Nice! I really liked how you did a great job at splitting your work into atomic commits and added details to commit logs. :+1: It would be awesome if you could also add subcrate prefixes (after the emoji to commit shortlog ( think `zm: ` in your case), then they'll be perfect and consistent with rest of the history. :)

zbus_macros/src/iface.rs
```diff
@@ -62,17 +63,21 @@ impl<'a> Property<'a> {
...
             read: false,
             write: false,
+            emits_changed_signal: PropertyEmitsChangedSignal::True,
             ty: None,
             doc_comments: quote!(),
         }
     }
 }
 
+#[derive(PartialEq)]
```
zeenix, line 73: This change went into the wrong commit. This should be part of the commit adding `MethodType` enum.

zbus_macros/src/iface.rs
```diff
@@ -62,17 +63,21 @@ impl<'a> Property<'a> {
...
 enum MethodType {
     Signal,
     Property(PropertyType),
     Other,
 }
+
+#[derive(PartialEq)]
```
zeenix, line 80: Same here. This change belongs in the previous commit.

zbus_macros/src/utils.rs
```diff
@@ -55,7 +55,7 @@ pub fn is_blank(s: &str) -> bool {
 /// Standard annotation `org.freedesktop.DBus.Property.EmitsChangedSignal`.
 ///
 /// See <https://dbus.freedesktop.org/doc/dbus-specification.html#introspection-format>.
-#[derive(Debug, Default)]
+#[derive(Debug, Default, Clone)]
```
zeenix, line 58: I wouldn't have pointed this one out but since you're going to be fixing the commits, I think best to split this into a separate one too that comes before this ("✨ Adds iface support for EmitsChangedSignal flag.") commit.

zbus_macros/tests/tests.rs
```diff
@@ -220,13 +220,17 @@ fn test_interface() {
     <arg name="arg" type="y"/>
     <arg name="other" type="s"/>
   </signal>
-  <property name="MyCustomProperty" type="u" access="readwrite"/>
+  <property name="MyCustomProperty" type="u" access="readwrite">
+    <annotation name="org.freedesktop.DBus.Property.EmitsChangedSignal" value="true"/>
```
zeenix, line 224: I think `true` is the default? If so, we should skip the annotation when it's `true`. I think other implementations also skip it as otherwise, it will mean a lot of noise in many introspection files.

zbus_macros/src/iface.rs
```diff
@@ -397,8 +395,8 @@ pub fn expand(args: AttributeArgs, mut input: ItemImpl) -> syn::Result<TokenStre
                     // * For argument type with lifetimes, we convert from `Value` (so
                     //   `TryFrom<Value<'_>>` is required).
                     //
-                    // * For all other arg types, we convert the passed value to `OwnedValue` first and
-                    //   then pass it as `Value` (so `TryFrom<OwnedValue>` is required).
+                    // * For all other arg types, we convert the passed value to `OwnedValue` first
+                    //   and then pass it as `Value` (so `TryFrom<OwnedValue>` is required).
```
zeenix, line 399: Is this change needed? If so, it doesn't belong in this commit.

zbus_macros/src/iface.rs
```diff
@@ -185,16 +151,13 @@ pub fn expand(args: AttributeArgs, mut input: ItemImpl) -> syn::Result<TokenStre
         } else {
             None
         };
-
```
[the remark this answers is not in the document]
zeenix: You ended up adding a new empty line now. :laughing:

zbus_macros/src/iface.rs
```diff
@@ -208,7 +171,6 @@ pub fn expand(args: AttributeArgs, mut input: ItemImpl) -> syn::Result<TokenStre
         } else {
             quote!(c.reply(m, &reply).await)
         };
-
```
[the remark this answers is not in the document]
zeenix: Same here. Now you're adding an empty line instead of removing.

zbus_macros/src/iface.rs
```diff
@@ -527,37 +572,75 @@ pub fn expand(args: AttributeArgs, mut input: ItemImpl) -> syn::Result<TokenStre
...
+                            )
                         }
-                    );
+                        PropertyEmitsChangedSignal::Invalidates => {
+                            quote!(
+                                #zbus::fdo::Properties::properties_changed(
+                                    signal_context,
+                                    #zbus::names::InterfaceName::from_static_str_unchecked(#iface_name),
+                                    &::std::collections::HashMap::new(),
+                                    &[],
```
zeenix, line 593: Didn't catch this the last time. The last argument here is supposed to be the list of all invalidated properties. So it think this should be `&[#member_name)],`. It would be great to have a test case for this. Perhaps add the case to both interface and proxy side in `zbus/tests/e2e.rs`?

zeenix (changes requested):
Looking really good otherwise. The request for a testcase is just that: a request. It's ok if you don't have the time to do that, just let me know.

zbus_macros/src/iface.rs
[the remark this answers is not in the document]
zeenix: [looks like the CI wouldn't have let you get away with this anyway](https://github.com/dbus2/zbus/actions/runs/7099835568/job/19327305216?pr=514). :laughing:

zbus_macros/tests/tests.rs
zeenix: Did you remove the test all together? :thinking:

zbus_macros/src/iface.rs
marcelmasqueprodrive: This code goes in the `properties_changed_method`, so I do not think that we should be reporting that the property is getting invalidated when we call the `changed` method. However, the `prop_invalidate_method` code should have `&[#member_name]`, this is a few lines below this comment

zeenix: > This code goes in the `properties_changed_method`, so I do not think that we should be reporting that the property is getting invalidated when we call the `changed` method. However

Not sure I follow. This is as per the behavior defined by [the spec](url):

```
If one or more properties change on an object, the org.freedesktop.DBus.Properties.PropertiesChanged signal may be emitted (this signal was added in 0.14):

              org.freedesktop.DBus.Properties.PropertiesChanged (STRING interface_name,
                                                                 ARRAY of DICT_ENTRY<STRING,VARIANT> changed_properties,
                                                                 ARRAY<STRING> invalidated_properties);
        

where changed_properties is a dictionary containing the changed properties with the new values and invalidated_properties is an array of properties that changed but the value is not conveyed. 
```

Otherwise, the client gets a signal w/o any information at all and hence not useful.

> However, the `prop_invalidate_method` code should have `&[#member_name]`, this is a few lines below this comment

Good that you mentioned. I also missed that you are adding a new separate method for invalidating. I think this is not needed. The method should still be `_changed`. The behavior is then defined by the type of property declared by the user.

zbus/tests/e2e.rs
```diff
@@ -430,6 +463,105 @@ impl MyIfaceImpl {
...
+            modifiable_true: 0,
+            modifiable_invalidates: 0,
+            modifiable_const: 0,
+            modifiable_false: 0,
+        }
+    }
+}
+
+#[dbus_interface(interface = "org.freedesktop.MyEmitsChangedSignalIface")]
+impl MyEmitsChangedSignalIfaceImpl {
```
zeenix, line 488: Thanks for adding this. However, I don't think we need to add another interface just for this. Why not add new properties to the existing interface and proxy? You'll basically just need 4-5 new props.

zeenix (changes requested):
Thanks. Sorry to ask you for another round but I think it'll be done soon.

zbus/tests/e2e.rs
```diff
@@ -123,6 +123,36 @@ trait MyIface {
...
+    #[dbus_proxy(property)]
+    fn set_modifiable_default(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property)]
+    fn modifiable_true(&self) -> zbus::Result<u32>;
+
+    #[dbus_proxy(property)]
+    fn set_modifiable_true(&self, count: u32) -> zbus::Result<()>;
+
+    #[dbus_proxy(property)]
```
zeenix, line 139: why isn't the proxy side making use of the (existing) `emits_changed_signal` attributes? Same for other methods.

zbus/tests/e2e.rs
```diff
@@ -699,6 +817,102 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+            let expected_property_key = "ModifiableInvalidates";
+            let args = changed.args()?;
+            assert_eq!(
+                args.invalidated_properties(),
+                &vec![expected_property_key.to_string()]
+            );
+
+            assert_eq!(args.changed_properties().keys().len(), 0);
+        }
+        None => panic!(""),
```
zeenix, line 882: Might as well, just `unwrap` then and reduce code complexity? Same for other matches you're adding here.

zeenix (changes requested):
So sorry, still 2 small things.

zeenix (approved):
Great job!

zeenix:
> Implements #323

btw, for future reference please use the keywords that GH can pick up on and automatically close the issues (e.g here it a `Fixes #323.` line at the end, would have done the trick.

zeenix:
> Implements #323: enables interface property macros on getters to be annotated with `emits_changed_signal`.

@mmasque For future reference, using the GH keywords (e.g `Fixes #323` here) closes the issue automatically for us.

PR #515, opened 2023-11-28
Remove stack overflow main

sivizius:
In version 3.15.0 of zvariant, this results in a stack overflow:

```rust
#[derive(Deserialize, Type)]
#[zvariant(signature = "a{sv}")]
struct Outer {
    foo: OwnedValue,
}

#[test_log::test]
fn convert() {
    let ctxt = EncodingContext::<LE>::new_dbus(0);
    let value = <HashMap<String, OwnedValue>>::from([
        ("foo".into(), 23.into()),
        ("bar".into(), 42.into()),
    ]);
    let data = to_bytes(ctxt, &value).unwrap();
    eprintln!("{data:02x?}");
    let good: Outer = from_slice(&data, ctxt).unwrap();
    eprintln!("{:?}", good.foo);
}
```

This is because if serde encounters an unknown field, it tries to parse it as `Content`
in : https://docs.rs/serde/latest/src/serde/private/de.rs.html#301. This calls https://docs.rs/serde/latest/src/serde/de/mod.rs.html#1231 and then on the deserializer (here `zvariant::dbus::Deserializer`) `deserialize_any`: https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#96. After https://docs.rs/zvariant/latest/src/zvariant/de.rs.html#589, `deserilaize_seq` of the deserializer is called: https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#276, which calls `visit_seq` on the visitor (`ContentVisitor`) with `ValueDeserializer` (this implements `SeqAccess`): https://docs.rs/serde/latest/src/serde/private/de.rs.html#493. There, `next_element` of `SeqAccess = ValueDeserializer` is called: https://docs.rs/serde/latest/src/serde/de/mod.rs.html#1728. This then calls https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#589 with `PhantomData<Content>`. The only thing this function can do, besides returning `Ok(None)` and `Err(…)`, is calling `deserialize` on `seed`, which is `PhantomData`, which now calls `deserialize` of `Content`: https://docs.rs/serde/latest/src/serde/de/mod.rs.html#794. Ah, shit, here we go again.

This PR is more of an initial draft, gvariant is basically unaddressed and untested, but this might be of interest nevertheless. 

Related: https://gitlab.freedesktop.org/dbus/zbus/-/issues/164, https://gitlab.freedesktop.org/dbus/zbus/-/issues/312

### DIFF
--- a/zbus/tests/serde.rs
+++ b/zbus/tests/serde.rs
@@ -1,0 +1,21 @@
+use byteorder::LE;
+use serde::Deserialize;
+use std::collections::HashMap;
+use zbus::zvariant::{serialized::Context, to_bytes, OwnedValue, Type};
+
+#[derive(Deserialize, Type)]
+#[zvariant(signature = "a{sv}")]
+struct Outer {
+    foo: OwnedValue,
+}
+
+#[test_log::test]
+fn convert() {
+    let ctxt = Context::<LE>::new_dbus(0);
+    let value =
+        <HashMap<String, OwnedValue>>::from([("foo".into(), 23.into()), ("bar".into(), 42.into())]);
+    let data = to_bytes(ctxt, &value).unwrap();
+    eprintln!("{data:02x?}");
+    let good = data.deserialize::<Outer>().unwrap().0;
+    eprintln!("{:?}", good.foo);
+}
--- a/zvariant/src/dbus/de.rs
+++ b/zvariant/src/dbus/de.rs
@@ -1,4 +1,5 @@
-use serde::de::{self, DeserializeSeed, EnumAccess, MapAccess, SeqAccess, Visitor};
+use core::convert::TryFrom;
+use serde::de::{self, Deserialize, DeserializeSeed, EnumAccess, MapAccess, SeqAccess, Visitor};
 use static_assertions::assert_impl_all;
 
 use std::{marker::PhantomData, str};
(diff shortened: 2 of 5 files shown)


zeenix:
@sivizius thanks so much for providing this fix and also explaining in detail what the issue was. However, could you please provide the fix based on the main (3.x serious) branch and remove redundant commits? Once 4.x is out, that will be the only supported series. I ported the test case for you already:

```rust
   #[test]
    fn issue_515() {
        use crate::OwnedValue;

        #[derive(Deserialize, Type)]
        #[zvariant(signature = "a{sv}")]
        struct Outer {
            #[allow(unused)]
            foo: OwnedValue,
        }

        let ctxt = Context::<LE>::new_dbus(0);
        let value = <HashMap<String, OwnedValue>>::from([
            ("foo".into(), 23.into()),
            ("bar".into(), 42.into()),
        ]);
        let data = to_bytes(ctxt, &value).unwrap();
        let _outer: (Outer, _) = data.deserialize().unwrap();
    }
```

sivizius:
Converted to proper Draft (TIL: GitHub has draft PRs…but less prominent than GitLab.)

PR #516, opened 2023-11-29
bug zbus_xmlgen: Fix structs with a single element

pauloserrafh:
When the xml element had a struct containing only a single type, the parser was returning the element itself. This was wrong since a D-Bus struct is expected to be translated into a tuple, regardless of how many types it has (except for 0 elements, which is not accepted according to D-Bus spec). Returning a single-element tuple type instead.

Note: As defined on Rust syntax, the single-element tuple type has a trailing comma inside parenthesis, not to be confused with compiler lint "unused_parens". Examples:
(i32,) --> OK. Single-element tuple with i32 element. (i32) --> ERROR. Wrong use of parenthesis that clutters readability.

About the change itself on _zbus_xmlgen/src/lib.rs_, it was noticed that we could somewhat simplify by removing the else-if condition and doing it all in the same format. However, IMHO the approach on the PR is more readable. 
Please see the other possible approach and let me know what would be the preferred approach.
```
                if dict {
                    vec.join(", ")
                } else {
                    format!(
                        "{}({}{})",
                        if as_ref { "&" } else { "" },
                        vec.join(", "),
                        if vec.len() == 1 { "," } else { "" }
                    )
                }
```

First time contributor here, sorry if I missed something on the contributing guidelines, please let me know and I'll fix ASAP. Thanks @nagatavit for the support on the investigation IRL.

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -241,7 +241,7 @@ fn to_rust_type(ty: &CompleteType, input: bool, as_ref: bool) -> String {
                 } else if vec.len() > 1 {
                     format!("{}({})", if as_ref { "&" } else { "" }, vec.join(", "))
                 } else {
-                    vec[0].to_string()
+                    format!("{}({},)", if as_ref { "&" } else { "" }, vec[0])
                 }
             }
             _ => unimplemented!(),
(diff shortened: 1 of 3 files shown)


zeenix (approved):
Thanks so much for this fix. LGTM!

Issue #329, opened 2023-04-13
Use std's abstract unix socket API


zeenix:
Apparently abstract unix socket support [was added to std a while ago](https://github.com/rust-lang/rust/pull/85379). We should make use of that instead of hacking the address etc.

Issue #105, opened 2020-11-20
A Connection/Proxy that can be shared between threads


zeenix:
In GitLab by @Spider on Nov 20, 2020, 18:34

In Async-std, when the executor recognizes that a task is blocking on something, it will automatically spawn a new thread that takes over the asyncronous executor mainloop function.

This means that the current thread may change quietly without intent, which can make using zbus a bit tricky. Therefore, it would be useful for me if Connection and/or Proxy'd objects that own a connection could implement Send in a safe manner.

PR #517, opened 2023-11-30
Actual smol rs update

### LINKED ISSUE #479: Bump to smol-rs subcrates' 3.0
Hopefully it's not a lot of work.

TTWNO:
Slight modification of #494 , fixes #479 

Depends on getting [rust_uds_windows/#9](https://github.com/haraldh/rust_uds_windows/pull/9) merged upstream.

For now I have added my person Github entry as the location for `uds_windows`. This should let us test of the tests pass now.

Once I get this merged upstream, I'll commit, squash, add emojis, etc. here.

### DIFF
--- a/book/src/blocking.md
+++ b/book/src/blocking.md
@@ -185,13 +185,13 @@ fn main() -> Result<(), Box<dyn Error>> {
         name: "GreeterName".to_string(),
         done: event_listener::Event::new(),
     };
-    let done_listener = greeter.done.listen();
+    let mut done_listener = greeter.done.listen();
     let _handle = connection::Builder::session()?
         .name("org.zbus.MyGreeter")?
         .serve_at("/org/zbus/MyGreeter", greeter)?
         .build()?;
 
-    done_listener.wait();
+    done_listener.as_mut().wait();
 
     Ok(())
 }
--- a/book/src/server.md
+++ b/book/src/server.md
@@ -262,14 +262,14 @@ async fn main() -> Result<()> {
         name: "GreeterName".to_string(),
         done: event_listener::Event::new(),
     };
-    let done_listener = greeter.done.listen();
+    let mut done_listener = greeter.done.listen();
     let _connection = Builder::session()?
         .name("org.zbus.MyGreeter")?
         .serve_at("/org/zbus/MyGreeter", greeter)?
         .build()
         .await?;
 
-    done_listener.wait();
+    done_listener.as_mut().wait();
 
     Ok(())
 }
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -107,7 +107,7 @@ nix = { version = "0.27", default-features = false, features = [
 [target.'cfg(target_os = "macos")'.dependencies]
 # FIXME: This should only be enabled if async-io feature is enabled but currently
 # Cargo doesn't provide a way to do that for only specific target OS: https://github.com/rust-lang/cargo/issues/1197.
-async-process = "1.7.0"
+async-process = "2.0.0"
 
 [target.'cfg(any(target_os = "macos", windows))'.dependencies]
 async-recursion = "1.0.0"
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -53,7 +53,7 @@ futures-util = { version = "0.3.25", default-features = false, features = [
   "std",
 ] }
 async-lock = { version = "2.6.0", optional = true }
-async-broadcast = "0.5.0"
+async-broadcast = "0.6.0"
 async-executor = { version = "1.5.0", optional = true }
 blocking = { version = "1.0.2", optional = true }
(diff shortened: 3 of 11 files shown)


zeenix:
@TTWNO I converted to a `Draft` to avoid accidental merge.

TTWNO:
`asynx-io` on a -windows-` platform does use it: https://docs.rs/async-io/latest/x86_64-pc-windows-msvc/async_io/struct.Async.html#method.new

That's what was causing the compile etror previously, since `Async::new(...)` requires an argument that implements `AsSocket`

zeenix:
> `asynx-io` on a `-windows-` platform does use it: https://docs.rs/async-io/latest/x86_64-pc-windows-msvc/async_io/struct.Async.html#method.new

That's very strange. I saw that it's a dev dep only: https://github.com/smol-rs/async-io/blob/master/Cargo.toml#L53

> That's what was causing the compile etror previously, since `Async::new(...)` requires an argument that implements `AsSocket`

hmm.. so the latest async-io is broken on windows, essentially?

TTWNO:
> hmm.. so the latest async-io is broken on windows, essentially?

No. But `uds_windows::UnixStream` doesn't implement `AsSocket`, which is now required by `async-io` (as of 2.x).

That's the problem I'm trying to fix upstream. Looks like the only tests failing now are for Linux.

TTWNO:
I can't reproduce the Linux build error. Based on the "OS error 22", it's an "invalid parameter" problem; now to recreate and fix it....

TTWNO:
I understand what you mean, but that seems extremely weird since these tests work on my local machine. I'm wondering if it's an "Ubuntu is too old" problem.

TTWNO:
I see the branch I based this off of is quite old, and didn't include the `safe-io2` branch. I tried merging that in to see if that fixes the issue. It looks like we were playing with `RawFd`s in `zbus/src/connection.rs`, even though `async-io` now expects to deal with `Borrowed/OwnedFd`s now. Perhaps that was the mix up?

:crossed_fingers:

TTWNO:
Just confirmed with `act`, which can run Github CI locally using Docker... No bueno. Looks like whatever broke it is still there.

notgull:
This problem seems to be `dbus`-specific: I was able to use `Async::connect()` to connect to `seatd` just fine. See here: https://github.com/smol-rs/async-io/actions/runs/7073336022/job/19253127954#step:9:24

zeenix:
@notgull Many thanks for looking into this! :+1: 

> I ran the tests with `strace` and I didn't see `EINVAL` anywhere, which makes me skeptical that this is a failed system call. Could this be [bytecodealliance/rustix#869](https://github.com/bytecodealliance/rustix/issues/869) ?

Oh? That's very much possible. Does this mean rustix doesn't support/has problems with abstract unix sockets? I wonder if #329 would help with this. :thinking:

zeenix:
@TTWNO async-io 2.2.2 was just released and it has the fix for the issue. I restarted the failing Linux job and hopeful that it'll now pass because it should pick up the latest release. However, it'd be best to change async-io version in the PR to ensure we only build against the fixed version.

TTWNO:
Wicked! Looking forward to getting any cleanups that are necessary :)

TTWNO:
Should `async-io` have a `2.2.0` minimum version, then?

zeenix:
> Should `async-io` have a `2.2.0` minimum version, then?

As I wrote, let's require `2.2.2`.

TTWNO:
Alright, got those dependencies updated and have removed the link to my git release. As long as the tests pass, I think this should be ready to rock!

zeenix:
> Alright, got those dependencies updated and have removed the link to my git release. As long as the tests pass, I think this should be ready to rock!

Awesome! However, there is still one concern though: It looks like event-listener might need another API break to effectively remove some footguns/unsoundness issues. Need to figure how to solve that.

In the meantime, could you please clean up the commits?

TTWNO:
Updated to remove specific issue listed, should this wait until [event-listener #105](https://github.com/smol-rs/event-listener/pull/105)?

notgull:
It might take a while to make sure everything works. So I'd say to go ahead with this first

zeenix:
> It might take a while to make sure everything works. So I'd say to go ahead with this first

Hmm.. but the event-listener update would then be a breaking change since we return `Pin<Box<EventListener>>` to the user and I was hoping not to make any more breaks any time soon.

Thinking more about this, I guess  we can add our own abstraction and expose that or just return `impl Future<Output = ()>` since all you can do with returned listener is wait on its Future impl.

zeenix:
> Thinking more about this, I guess we can add our own abstraction and expose that or just return `impl Future<Output = ()>` since all you can do with returned listener is wait on its Future impl.

Yeah, actually let's do that but I'll do it on a separate PR. This one has waited long enough.

Issue #512, opened 2023-11-26
xmlgen: hide clippy lints for interface functions


FineFindus:
Generated zbus functions often have many arguments, which causes [clippy to complain about having a function that returns too many arguments](https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments), or complex types, which causes the [type complexity lint](https://rust-lang.github.io/rust-clippy/master/index.html#/type_complexity).
Since the user has no control over the interface design, I think it makes sense to automatically hide the lints when generating the interface code.

PR #518, opened 2023-11-30
✨ zx: hide clippy lints for generated functions

FineFindus:
This hides the [`clippy::too_many_arguments`](https://rust-lang.github.io/rust-clippy/master/index.html#/too_many_arguments) and [`clippy::type_complexity`](https://rust-lang.github.io/rust-clippy/master/index.html#/type_complexity) lints for generated functions.

Note that type complexity is only an estimate, so future adjustments may be necessary. I tested it with the interface files for [`org.freedesktop.UDisks2.xml`](https://github.com/storaged-project/udisks/blob/a76eda89a4c747f12dc05670f376d95d8ec4cd45/data/org.freedesktop.UDisks2.xml) and confirmed that it is working correctly.

Closes https://github.com/dbus2/zbus/issues/512.

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -77,6 +77,7 @@ impl<'i> Display for GenTrait<'i> {
             if p.access().read() {
                 writeln!(f, "{}", fn_attribute)?;
                 let output = to_rust_type(p.ty(), false, false);
+                hide_clippy_type_complexity_lint(f, p.ty().signature())?;
                 writeln!(f, "    fn {name}(&self) -> zbus::Result<{output}>;",)?;
             }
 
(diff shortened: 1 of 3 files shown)


zbus_xmlgen/src/lib.rs
```diff
@@ -294,3 +314,54 @@ pub fn pascal_case(s: &str) -> String {
...
+                        break;
+                    }
+                    _ => score += 5 * estimate_type_complexity(it),
+                }
+            }
+        }
+        _ => {}
+    };
+    score
+}
```
zeenix, line 367: This code is a bit complicated so it would be great to have tests for this. Just add the cases to the existing test.

zeenix:
Oh and btw, the max chars on a line in the commit log description part should be 74. No biggie though.

PR #531, opened 2023-12-30
Stop using the async-std crate

sophie-h:
The async-std crate didn't had a release in 1 1/2 years and is pulling in a lot of outdated and unnecessary dependencies.

Therefore I replaced all uses with actively maintained crates.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -844,8 +844,8 @@ impl Connection {
     /// # // https://gitlab.freedesktop.org/zeenix/zbus/-/jobs/34023494
     /// # #[cfg(all(not(feature = "tokio"), not(target_os = "windows")))]
     /// # {
+    /// use async_global_executor::{block_on, spawn};
     /// use zbus::connection::Builder;
-    /// use async_std::task::{block_on, spawn};
     ///
     /// # struct SomeIface;
     /// #
@@ -869,7 +869,7 @@ impl Connection {
     ///            loop {
     ///                conn.executor().tick().await;
     ///            }
-    ///        });
+    ///        }).detach();
     ///     }
     ///
     ///     // All your other async code goes here.
--- a/README.md
+++ b/README.md
@@ -50,20 +50,21 @@ impl Greeter {
     }
 }
 
-// Although we use `async-std` here, you can use any async runtime of choice.
-#[async_std::main]
-async fn main() -> Result<(), Box<dyn Error>> {
-    let greeter = Greeter { count: 0 };
-    let _conn = connection::Builder::session()?
-        .name("org.zbus.MyGreeter")?
-        .serve_at("/org/zbus/MyGreeter", greeter)?
-        .build()
-        .await?;
+// Although we use `smol` here, you can use any async runtime of choice.
+smol_macros::main! {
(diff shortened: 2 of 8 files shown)


zbus/src/connection/mod.rs
```diff
@@ -844,8 +844,8 @@ impl Connection {
     /// # // https://gitlab.freedesktop.org/zeenix/zbus/-/jobs/34023494
     /// # #[cfg(all(not(feature = "tokio"), not(target_os = "windows")))]
     /// # {
+    /// use async_global_executor::{block_on, spawn};
```
sophie-h, line 847: This comment needs to be adjusted

```
    /// Here is how one would typically run the zbus executor through async-std's single-threaded
    /// scheduler:
```

We could also use `smol` instead of `async-global-executor`. But that might pull more dependencies.

zeenix:
Thanks but async-std is only a dev dep so it dragging in obsolete deps isn't a very compelling argument, I'm afraid. I still want to ensure that zbus continues to work with it.

We primarily use smol-rs crates. Dev deps are irrelevant to our users. Developers shouldn't need to care about async-std being under-mai tainted. 

Having said that, if any of the deps brought in by async-std has any security issues, it'd be definitely a compelling argument as we wouldn't want contributors to compromise their dev env.

sophie-h:
The problem with dev dependencies is that distros have to package them as well if they want to run tests. And since the deps of async-std are outdated there will be several versions of the same crate that have to be packaged.

zeenix:
> The problem with dev dependencies is that distros have to package them as well if they want to run tests.

Oh, I didn't think/know of this aspect. This changes things for sure. :)

I think it'd be best to just use tokio then, since it is the most popular runtime. Could you kindly provide a PR for that?

sophie-h:
Cool. I will hopefully get back to it later this month.

zeenix:
>  I will hopefully get back to it later this month.

Actually, looking at the repo, I still see sufficient recent [activity](https://github.com/async-rs/async-std/activity) on the project and one year is not that long for a release of a stable project. We've ourselves haven't had a release for more than 6 months now. 😞

Has this already presented an issue for a distro?

zeenix:
> Has this already presented an issue for a distro?

@sophie-h ?

sophie-h:
Well, the general situation with distros packaging rust apps is not great. One reason that has been reported is the amount of dependencies needed, yes. If this specifically is the most important part, idk.

Right now, zbus is pulling a bunch of older versions of crates that cause us to depend on multiple versions. Maybe that would be more important to tackle first https://sophie-h.pages.gitlab.gnome.org/cargo-lock-analyzer/

zeenix:
> One reason that has been reported is the amount of dependencies needed, yes

I think this wouldn't be relevant here if the distros didn't insist on dev deps being packaged as well.

> zbus is pulling a bunch of older versions of crates that cause us to depend on multiple versions

I'd more more than happy to update deps where possible. I'm also pushing for `event-listener` crate situation to be resolved soon so we don't end up with multiple versions of that in zbus deps.

Anyway, please do update the PR as discussed before. It's almost been [one month now](https://github.com/dbus2/zbus/pull/531#issuecomment-1872930834). :laughing:

PR #532, opened 2023-12-31
demo for a bug triggered by objectManager sending a signal before hello

caramelhooves:
This code snippet fails with error:

called `Result::unwrap()` on an `Err` value: InputOutput(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })

while dbus complains

dbus-broker: Peer :1.505 is being disconnected as it sent a message before calling Hello().


zeenix:
Closing for #533

PR #535, opened 2024-01-01
🚸 zb: Connection::monitor_activity now returns Activity type

zeenix:
The `event-listener` crate will be breaking API again soon so it's best to abstract away the small part that we expose to the user. This will allow us to update to the new API without breaking our own API.


zeenix:
cc @TTWNO

PR #538, opened 2024-01-03
♻️ zb/win32: Refactor raw pointer handling

eNV25:
- Use `ptr.cast()` to convert type keeping `const`ness, instead of `as`
- Use `ptr::addr_of_mut!()` to get mut pointer, intead of `&mut`

### DIFF
--- a/zbus/src/win32.rs
+++ b/zbus/src/win32.rs
@@ -4,6 +4,7 @@ use std::{
     net::SocketAddr,
     os::windows::prelude::OsStrExt,
     ptr,
+    mem,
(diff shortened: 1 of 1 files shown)


zeenix:
@eNV25 Thanks. Just some linting to be fixed to satisfy the CI. Don't worry about the warnings, they'll hopefully be resolved by https://github.com/dbus2/zbus/pull/539.

zeenix:
oh auto-merged even though lint job failed. :( I need to update the settings to require the lint job too. :)

PR #540, opened 2024-01-06
✨ zd,zv: `Value` derive improvements for enums

katyo:
- Support enums without explicit values
- Support string representation for enums

### DIFF
--- a/zvariant/src/lib.rs
+++ b/zvariant/src/lib.rs
@@ -1447,10 +1447,10 @@ mod tests {
         let decoded: NoReprEnum = encoded.deserialize().unwrap().0;
         assert_eq!(decoded, NoReprEnum::Variant2);
 
-        #[derive(Deserialize, Serialize, Type, Debug, PartialEq)]
-        #[zvariant(signature = "s")]
+        #[derive(Deserialize, Serialize, Type, Value, OwnedValue, Debug, PartialEq)]
+        #[zvariant(signature = "s", rename_all = "snake_case")]
         enum StrEnum {
-            Variant1,
+            VariantOne,
             Variant2,
             Variant3,
         }
@@ -1460,6 +1460,20 @@ mod tests {
         assert_eq!(encoded.len(), 13);
         let decoded: StrEnum = encoded.deserialize().unwrap().0;
         assert_eq!(decoded, StrEnum::Variant2);
+
+        assert_eq!(
+            StrEnum::try_from(Value::Str("variant_one".into())),
+            Ok(StrEnum::VariantOne)
+        );
+        assert_eq!(
+            StrEnum::try_from(Value::Str("variant2".into())),
+            Ok(StrEnum::Variant2)
+        );
+        assert_eq!(
+            StrEnum::try_from(Value::Str("variant4".into())),
+            Err(Error::IncorrectType)
+        );
+        assert_eq!(StrEnum::try_from(Value::U32(0)), Err(Error::IncorrectType));
 
         #[derive(Deserialize, Serialize, Type)]
         enum NewType {
--- a/zvariant/src/lib.rs
+++ b/zvariant/src/lib.rs
@@ -128,8 +128,8 @@ mod tests {
     use crate::Fd;
     use crate::{
         serialized::{Context, Format},
-        Array, Basic, DeserializeDict, DeserializeValue, Dict, Error, ObjectPath, Result,
-        SerializeDict, SerializeValue, Str, Structure, Type, Value, BE, LE, NATIVE_ENDIAN,
+        Array, Basic, DeserializeDict, DeserializeValue, Dict, Error, ObjectPath, OwnedValue,
+        Result, SerializeDict, SerializeValue, Str, Structure, Type, Value, BE, LE, NATIVE_ENDIAN,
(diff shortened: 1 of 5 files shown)


katyo:
As a future improvement: what about using external crate for ident case conversion?

zvariant_derive/tests/tests.rs
```diff
@@ -59,6 +59,22 @@ fn derive_enum_no_repr() {
     assert_eq!(TestEnum::try_from(Value::U32(3)), Err(Error::IncorrectType));
     assert_eq!(TestEnum::try_from(Value::Str("Foo".into())), Err(Error::IncorrectType));
 }
+
+#[test]
+fn derive_enum_string() {
```
zeenix, line 64: Kudos for adding tests. :+1: However, I'd prefer you modify the existing tests in `zvariant/src/lib.rs`. The `derive` test case there already covers enums as strings. You can just add the Value bits to the same enum(s).

zvariant_derive/tests/tests.rs
```diff
@@ -28,15 +28,37 @@ fn derive_struct() {
...
-    assert_eq!(RequestNameFlags::signature(), "u")
+    assert_eq!(RequestNameFlags::signature(), "q");
+    assert_eq!(Value::from(RequestNameFlags::AllowReplacement), Value::U16(0x01));
+    assert_eq!(RequestNameFlags::try_from(Value::U16(0x04)), Ok(RequestNameFlags::DoNotQueue));
+    assert_eq!(RequestNameFlags::try_from(Value::U16(0x03)), Err(Error::IncorrectType));
+}
+
+#[test]
+fn derive_enum_no_repr() {
+    #[derive(Debug, Clone, Copy, PartialEq, Type, Value, OwnedValue)]
```
zeenix, line 47: * same here: try and modify existing tests (preferably in `zvariant` as people often copy code from tests and I want to direct use of macro crates these days).
* Please only add what is needed. E.g do we need `Clone` and `Copy` here? :thinking:

zeenix (changes requested):
Thanks for this contribution! Some nits to fix. Talking of which, could you kindly add emoji prefixes as per [our contribution guidelines](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests)?

zeenix:
> As a future improvement: what about using external crate for ident case conversion?

I'm not sure what you mean but I'd like to avoid adding a dep unless there is a compelling rationale presented. :)

zeenix:
@katyo Hi there. Do you think you'll have time to finish this soon? No hurry, just a polite ping. :)

katyo:
@zeenix I'm a little bit busy lately, but I try fix this pr asap.

zvariant/src/lib.rs
```diff
@@ -1535,10 +1535,10 @@ mod tests {
         let decoded: NoReprEnum = encoded.deserialize().unwrap().0;
         assert_eq!(decoded, NoReprEnum::Variant2);
 
-        #[derive(Deserialize, Serialize, Type, Debug, PartialEq)]
-        #[zvariant(signature = "s")]
+        #[derive(Deserialize, Serialize, Type, Value, OwnedValue, Debug, PartialEq)]
+        #[zvariant(signature = "s", rename_all = "snake_case")]
```
zeenix, line 1539: Would be nice to document these additions in the macro docs.

zvariant_derive/src/value.rs
```diff
@@ -269,34 +297,76 @@ fn impl_enum(
...
-                )
+                #from_val
             }
         }
 
         #into_value
     })
 }
+
+fn enum_name_for_variant(
```
zeenix, line 349: `SerializeDict` and `DeserializeDict` already support these renaming attributes so I'm sure there is existing code you can use.

zeenix (changes requested):
Looks great in general. Thanks for updating. BTW, you forgot to add emojis [still](https://github.com/dbus2/zbus/pull/540#pullrequestreview-1807484939). ;)

zvariant_derive/src/value.rs
katyo: Of course, but currently `dict_name_for_field` expect `syn::Field` as an argument, not something like `Literal` or `String`. I would like to avoid modifying existing code which is not related to this PR.

zeenix: It **is** related to this PR since there is no good reason to duplicate the code. AFAIK, it's a pretty trivial change so why not just do it? It will need to be renamed and moved to utils mod btw.

katyo: Ok

zeenix:
> @zeenix I'm a little bit busy lately, but I try fix this pr asap.

Just a gentle reminder, in hope that you're not super busy and can finish this PR. 🙏

zeenix:
> Just a gentle reminder, in hope that you're not super busy and can finish this PR. 🙏

I guess not. 😞

zeenix:
I rebased this branch and since it took some effort, I already added myself as the co-author of the commits. Next I will try to address the concerns/comments raised.

zeenix (approved):

PR #544, opened 2024-01-14
📝 zb: update documentation for Guid::generate

alyssais:
Connection::new_unix_server no longer exists, so I've replaced it with the appropriate current method.

---

There's also a reference to this method in book-1.0, but I wasn't sure whether that should be updated.

### DIFF
--- a/zbus/src/guid.rs
+++ b/zbus/src/guid.rs
@@ -25,9 +25,8 @@ pub struct Guid(String);
 assert_impl_all!(Guid: Send, Sync, Unpin);
 
 impl Guid {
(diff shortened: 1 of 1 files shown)


zeenix (approved):
Thanks so much.

zeenix:
> There's also a reference to this method in book-1.0, but I wasn't sure whether that should be updated.

book-1.0 is about 1.0 API so it should have that ref. I'll be removing the book-1.0 anyway as part of the upcoming 4.0 release.

zeenix:
@alyssais don't sweat the CI failure, it's caused by https://github.com/rust-lang/rust-clippy/issues/12133 which was fixed 2 days ago so as soon as clippy nightly has it, the CI will go green and we can merge.

Issue #542, opened 2024-01-08
dbus_proxy with DynamicType not working (requires Type to be implemented)


alexg-axis:
I'm having issues using the `dbus_proxy` macro when trying to use `DynamicType`. The trait is documented as being a solution to when a type is not known ahead of type (i.e. cannot implement `Type`). 

The error I get is the following.

```
The trait bound `Value: zvariant::Type` is not satisfied.
Required by a bound in `zbus::blocking::Proxy::<'a>::call`
```

Here's a small example of what I'm trying to do.

```rust
#[dbus_proxy(
    default_service = "com.example.Service",
    interface = "com.example.Service",
    default_path = "/com/example/Service",
    gen_async = false,
    gen_blocking = true
)]
trait Some {
    /// SomeCall.
    fn some_call(
        &self,
        some_types: HashMap<&str, Type>,
        some_values: HashMap<&str, Value>,
    ) -> zbus::Result<()>;
}
```

```rust
#[derive(Serialize, Deserialize, zvariant::Type)]
#[zvariant(signature = "s")]
enum Type {
    #[serde(rename = "VALUE_TYPE_STRING")]
    String,
    #[serde(rename = "VALUE_TYPE_INTEGER")]
    Integer,
    #[serde(rename = "VALUE_TYPE_BOOLEAN")]
    Boolean,
    #[serde(rename = "VALUE_TYPE_DOUBLE")]
    Double,
}

#[derive(Serialize, Deserialize)]
enum Value {
    String(String),
    Integer(i32),
    Boolean(bool),
    Double(f64),
}

impl DynamicType for Value {
    fn dynamic_signature(&self) -> Signature<'_> {
        match self {
            Self::String(_) => Signature::try_from("s").unwrap(),
            Self::Integer(_) => Signature::try_from("i").unwrap(),
            Self::Boolean(_) => Signature::try_from("b").unwrap(),
            Self::Double(_) => Signature::try_from("d").unwrap(),
        }
    }
}
```

The example might look a bit weird from a dbus API perspective, but it makes more sense in context, without any fields stripped out.

Without knowing too much of the internals, from an API point of view, there doesn't seem to be any reason to require `Type` over `DynamicType` in `Call`. Especially as it seems to do with deserialization, which is not used in this case.

Is there any way to facilitate such a use of zbus?

zeenix:
[...]

PR #547, opened 2024-01-16
Add SerializeTuple and use it for code generated by #[dbus_proxy]

danieldg:
Fixes: https://github.com/dbus2/zbus/issues/542

### DIFF
--- a/zbus/src/blocking/proxy/mod.rs
+++ b/zbus/src/blocking/proxy/mod.rs
@@ -227,7 +227,7 @@ impl<'a> Proxy<'a> {
         M: TryInto<MemberName<'m>>,
         M::Error: Into<Error>,
         B: serde::ser::Serialize + zvariant::DynamicType,
-        R: serde::de::DeserializeOwned + zvariant::Type,
+        R: for<'d> zvariant::DynamicDeserialize<'d>,
     {
         block_on(self.inner().call(method_name, body))
     }
@@ -251,7 +251,7 @@ impl<'a> Proxy<'a> {
         M: TryInto<MemberName<'m>>,
         M::Error: Into<Error>,
         B: serde::ser::Serialize + zvariant::DynamicType,
-        R: serde::de::DeserializeOwned + zvariant::Type,
+        R: for<'d> zvariant::DynamicDeserialize<'d>,
     {
         block_on(self.inner().call_with_flags(method_name, flags, body))
     }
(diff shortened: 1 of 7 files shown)


zvariant/src/type.rs
```diff
@@ -140,6 +143,26 @@ where
     }
 }
 
+/// A helper type to serialize a tuple whose elements implement [DynamicType] but not [Type].
+///
+/// This is required because tuples already have an implementation of [DynamicType] via the blanket
+/// implementation of [DynamicType] where `T: Type`, but that results in a bound of [Type] on each
+/// element, which is stronger than needed for serializing.
+#[derive(Debug, Copy, Clone)]
```
zeenix, line 151: Why `Copy`? We shouldn't assume `T` to be `Copy`.

zvariant/src/type.rs
```diff
@@ -140,6 +143,26 @@ where
...
+pub struct SerializeTuple<T>(pub T);
+
+impl Type for SerializeTuple<()> {
+    fn signature() -> Signature<'static> {
+        Signature::from_static_str_unchecked("")
+    }
+}
+
+impl<T: Serialize> Serialize for SerializeTuple<T> {
```
zeenix, line 160: I know it's not needed for #542 but would be good to add `Deserialize` impl as well, while you're at it.

zeenix (changes requested):
LGTM otherwise. Thanks for looking into this. :+1:

zvariant/src/type.rs
danieldg: It's a `#[derive]` so it doesn't require that `T` is `Copy`.  Technically it might be better to derive all the traits (`Eq` and so forth) but the rest are less useful.

zeenix: TIL that derives don't add the impl if the generics don't satisfy the the underlying trait.

zeenix (changes requested):
Other than that some nits on the commit logs:

* " zb: Relax type bound on call/call_with_flags" doesn't say what restriction is being lifted.
* " zv: add OwnedStructure" could use a very short rationale.

zeenix:
@danieldg this got off my radar. All review comments addressed?

danieldg:
I believe so.

zeenix:
> " zv: add OwnedStructure" could use a very short rationale.

Still the case but ok, no biggie.

I forgot to mention, we've started doing [emoji prefixes in git commits](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). Could you please add those? Should be pretty easy with my [gimoji](https://github.com/zeenix/gimoji) tool. :)

Then I can merge.

zeenix (approved):

PR #520, opened 2023-12-08
🐛 zv: remove stack overflow main

sivizius:
In version 3.15.0 of zvariant, this results in a stack overflow:

```rust
#[derive(Deserialize, Type)]
#[zvariant(signature = "a{sv}")]
struct Outer {
    foo: OwnedValue,
}

#[test_log::test]
fn convert() {
    let ctxt = EncodingContext::<LE>::new_dbus(0);
    let value = <HashMap<String, OwnedValue>>::from([
        ("foo".into(), 23.into()),
        ("bar".into(), 42.into()),
    ]);
    let data = to_bytes(ctxt, &value).unwrap();
    eprintln!("{data:02x?}");
    let good: Outer = from_slice(&data, ctxt).unwrap();
    eprintln!("{:?}", good.foo);
}
```

This is because if serde encounters an unknown field, it tries to parse it as `Content`
in : https://docs.rs/serde/latest/src/serde/private/de.rs.html#301. This calls https://docs.rs/serde/latest/src/serde/de/mod.rs.html#1231 and then on the deserializer (here `zvariant::dbus::Deserializer`) `deserialize_any`: https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#96. After https://docs.rs/zvariant/latest/src/zvariant/de.rs.html#589, `deserilaize_seq` of the deserializer is called: https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#276, which calls `visit_seq` on the visitor (`ContentVisitor`) with `ValueDeserializer` (this implements `SeqAccess`): https://docs.rs/serde/latest/src/serde/private/de.rs.html#493. There, `next_element` of `SeqAccess = ValueDeserializer` is called: https://docs.rs/serde/latest/src/serde/de/mod.rs.html#1728. This then calls https://docs.rs/zvariant/latest/src/zvariant/dbus/de.rs.html#589 with `PhantomData<Content>`. The only thing this function can do, besides returning `Ok(None)` and `Err(…)`, is calling `deserialize` on `seed`, which is `PhantomData`, which now calls `deserialize` of `Content`: https://docs.rs/serde/latest/src/serde/de/mod.rs.html#794. Ah, shit, here we go again.

This PR is more of an initial draft, gvariant is basically unaddressed and untested, but this might be of interest nevertheless. 

Related: https://gitlab.freedesktop.org/dbus/zbus/-/issues/164, https://gitlab.freedesktop.org/dbus/zbus/-/issues/312

[...]

PR #552, opened 2024-01-23
✨ zv: do not ignore field-names of a{sv}-structs

sivizius:
If `#[derive(Serialize)]` is used on a `struct`, serde will generate some code using `serialize_struct` to construct a `SerializeStruct`-object and calls its method `serialize_field` to serialise the fields. 

If a signature is set with `#[derive(Type)] #[zvariant(signature = "a{sv}")]` and e.g. the dbus-serialiser is used, its implementation of the trait-method `serialize_struct` will use the first character of the signature, which is `a`, and will return a `StructSeqSerializer::Seq`, for which the traits `SerializeStruct` and `SerializeStructVariant` are implemented. But these implementations implement `serialize_field` in a way, that will ignore the field names.

This would be fine for e.g. `av`, where structs are (de)serialised like tuples, but `a{sv}` should be (de)serialised as dictionaries.

### DIFF
--- a/zvariant/src/dbus/ser.rs
+++ b/zvariant/src/dbus/ser.rs
@@ -687,7 +687,14 @@ macro_rules! serialize_struct_named_fields {
             {
                 match self {
                     StructSeqSerializer::Struct(ser) => ser.serialize_field(key, value),
-                    StructSeqSerializer::Seq(ser) => ser.serialize_element(value),
+                    StructSeqSerializer::Seq(ser) => match ser.ser.0.sig_parser.next_char()? {
+                        DICT_ENTRY_SIG_START_CHAR => {
+                            use ser::SerializeMap as _;
+                            ser.serialize_key(key)?;
+                            ser.serialize_value(value)
+                        }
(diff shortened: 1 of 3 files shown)


sivizius:
This is somewhat related to #520, as both are first steps, that might obsolete `SerializeDict`/`DeserializeDict`.

zeenix:
> This is somewhat related to #520,

Oh, I had missed that #520 was about use of `serde::Deserialize`. I thought it was just a bug fix.

> as both are first steps, that might obsolete `SerializeDict`/`DeserializeDict`.

That would be great. I'm trying my best to make zvariant and zbus 4.0 happen by end of this week. If you could finish this work before then, that would be super great since it would be best to deprecate those macros (or even remove them) in a breaking release.

If that's not posible, it's not a big deal though so don't stress it. :)

sivizius:
Sadly, I do not share your optimism about deprecating them till end of week, but I will try my best. Especially testing will take some time.

zeenix:
I guess @sivizius doesn't plan to finish this.

PR #553, opened 2024-01-23
🔨: introduce shell.nix and flake.nix

sivizius:
`shell.nix` (for `nix-shell`) and `flake.nix` (e.g. for `nix develop`) makes developing under nix/NixOS easier as it automagically enforces a development-environment with every dependency included.

A simple `nix-shell` or `nix develop` is necessary before e.g. `cargo test` does not complain about missing dependencies or wrong version of rustc. In future PRs, [pre-commit-hooks.nix ](https://github.com/cachix/pre-commit-hooks.nix) could be introduced to add pre-commit-hooks e.g. for formatting, linting, spell-checking, etc., and more.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,101 @@
+{
+  "nodes": {
+    "fenix": {
+      "inputs": {
+        "nixpkgs": [
+          "nixpkgs"
+        ],
+        "rust-analyzer-src": "rust-analyzer-src"
+      },
+      "locked": {
+        "lastModified": 1705991032,
+        "narHash": "sha256-C+ePPXLMOxkJfJMqihUC1XbNr9Xy2F6BpWrmGDsgzXk=",
+        "owner": "nix-community",
+        "repo": "fenix",
+        "rev": "2189a3d994aaee6f83d3fc92deb13c458dd03dbd",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-community",
+        "repo": "fenix",
+        "type": "github"
+      }
+    },
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1705309234,
+        "narHash": "sha256-uNRRNRKmJyCRC/8y1RqBkqWBLM034y4qN7EprSdmgyA=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "1ef2e671c3b0c19053962c07dbda38332dcebf26",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "ref": "main",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1706035044,
+        "narHash": "sha256-Mjk20bVbCc0GyjHNXeWHgx7vvLnak7tyV9fEs9iWpo0=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "edcff388a08f24b62dd09a27e86eb0c4313eb798",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "master",
+        "repo": "nixpkgs",
+        "type": "github"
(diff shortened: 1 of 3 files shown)


zeenix:
Thanks. Not sure though. Ideally, we shouldn't have any distro-specific stuff in upstream repo but I'd be willing to make an exception if other rust projects are doing the same?

Oh and I'm not sure I'll be able to review the files myself as I don't know anything about nix. I asked a friend who does, to help review but they're currently sick. Fortunately, there's no big hurry on this. 😊

MaxVerevkin:
> Ideally, we shouldn't have any distro-specific stuff in upstream repo

I also think this is out of place, especially in cases when core maintainers aren't using nix.

sivizius:
I use NixOS and wrote this to build and test these crates and like to share this setup, because it might be useful for everyone who uses NixOS/nix and wants to contribute. However, for everyone else, this is basically useless. Nevertheless, NixOS/nix seems to be quite common these days.

MaxVerevkin:
Why isn't there a central place to share these configs? Like https://github.com/NixOS/nixpkgs/ but for development instead of packages.

zeenix:
> Nevertheless, NixOS/nix seems to be quite common these days.

Of course but so are many other dev setups and distros etc. Hence why I asked for examples of other upstream Rust projects doing this. I think that would be the only way for me to be sure enough that we should do this.

> for everyone else, this is basically useless.

That's the problem. I hate dead/useless code/information with a passion. It adds to cognitive load as one sees information they may never ever need.

gila:
@zeenix [Hyprland](https://github.com/hyprwm/Hyprland/blob/main/flake.nix) is a rust example -- however one difference IMHO is that it produces an executable as an artifact. Zbus (correct if im wrong) does not produce binaries and should be used as a crate. 

So, the value added here is minor. The reason is that zbus is light on dependencies to get it to build. Having that said, the PR itself looks good to me. 

(neovim and bpftrace are other examples that have included nix files in the repo)

sivizius:
> Zbus (correct if im wrong) does not produce binaries and should be used as a crate.

Tests, examples and benchmarks are binaries. But no *installables* (you usually do not install and use these on your system…I mean…you could, but probably don’t).

sivizius:
> I also think this is out of place, especially in cases when core maintainers aren't using nix.

This does not really work with different branches, as different versions of the code might require different (versions of) dependencies. However, with flakes, `nix develop github:sivizius/zbus/zbus4-flake-nix` kinda works.

> other rust projects are doing the same
> I also think this is out of place, especially in cases when core maintainers aren't using nix.

This does not really work with different branches, as different versions of the code might require different (versions of) dependencies. However, with flakes, `nix develop github:sivizius/zbus/zbus4-flake-nix` kinda works.

zeenix:
@gila Hey, thanks for chiming in. @sivizius Can these files be hidden (`.` prefix)? I'd feel much better merging if there could be one hidden dir that holds all the files.

Kloenk:
> Hey, thanks for chiming in. @sivizius Can these files be hidden (`.` prefix)? I'd feel much better merging if there could be one hidden dir that holds all the files.

Sadly only the shell.nix can. flake.nix and flake.lock need to be in the root folder as it's relativ and nix expects it like that

zeenix:
> Sadly only the shell.nix can. flake.nix and flake.lock need to be in the root folder as it's relativ and nix expects it like that

Then I'm afraid, I can't accept flake.*. I hope `shell.nix` helps already. @sivizius Please update the PR accordingly.

Kloenk:
the shell.nix is something that is imported by flake.nix so it can be move but still needs the flake.nix.

I never tried to move the flake.nix in a subfolder, maybe it could work with `src = ../.;`?

zeenix:
@sivizius do you think you'll have time in the near future to finish your PRs? I'll be happy to wait for the other PRs as they are applicable to everyone but not this one, sorry.

zeenix:
> @sivizius do you think you'll have time in the near future to finish your PRs? I'll be happy to wait for the other PRs as they are applicable to everyone but not this one, sorry.

I guess not. :(

PR #554, opened 2024-01-23
✨ zv: impl different enum-variant tag-types

sivizius:
Previously, `enums` can only be serialised as `(u…)`, except for unit-variants. With this PR, enum-variants can be serialised as `_(…)` and `(_…)`, where `_` is one of `s` (variant name), `u` (as before), `b`, `y`, … (all other numeric types). Except for `s`, `u`, `t` and `x`, this might be fallible, e.g. an `enum` with >256 Variants cannot be encoded as `u8` (`y`).

TODO: tests, deserialisation


zeenix:
:disappointed:

PR #557, opened 2024-01-27
zb: Replace once_cell usage

bilelmoussaoui:
Replace the usage of OnceCell/Lazy from the once_cell crate with std types equivalents. The types were added before the current MSRV of zbus


zeenix (approved):
Nice!

Issue #174, opened 2021-07-10
Consider using dbus-server-address-parser


zeenix:
The author of the [dbus-server-address-parser](https://crates.io/crates/dbus-server-address-parser) crate contacted me and suggested we consider using it instead of our own parsing code. While at the moment our parsing code isn't much but I think in future it'll get more complex so worth having a look for sure.

PR #562, opened 2024-02-01
Use dbus-addr crate

elmarco:
I published https://crates.io/crates/dbus-addr based on my work from #430.

zbus can rely on it instead of providing its own Address implementation.

This allows to cleanup some code (mixing code parsing and stream connection etc)..

The other advantage of this series is the support for ;-separated address.

This change is mostly unnoticeable from a typical zbus API user, who rely on session() most of the time, or DBus addresses as string.


zeenix:
> I published https://crates.io/crates/dbus-addr based on my work from #430.

So I reject your work for an alternative impl and you think putting it in a separate crate under your maintenance would convince me to be more open to those changes? :laughing: This is a pretty big change so close to a major release that's been in the making for months. Even if I'd be fine with the API of the new crate (I haven't looked but as you said it's based on your work and I didn't agree with your approaches there), I think we need to postpone this to 5.0.

> zbus can rely on it instead of providing its own Address implementation.

What does it really achieve? We can put many things in separate crate but it's only useful if people need to parse dbus addresses separately from zbus but I seriously doubt there are sufficient use cases of that. Typically people don't even directly use our `Address` API even.

Besides we already [considered use of an external crate](https://github.com/dbus2/zbus/issues/174) for address handling but in the end, [I decided](https://github.com/dbus2/zbus/issues/174#issuecomment-1543544782) (w/o objections fro you) that it's not worth it. I think the rationale to close that issue still applies here.

> This allows to cleanup some code (mixing code parsing and stream connection etc).
 
I'll have to look closer to know what that implies and if I agree with it but neither of the advantages you specified requires splitting into a separate crate.

> This change is mostly unnoticeable from a typical zbus API user, who rely on session() most of the time, or DBus addresses as string.

Exactly and hence no compelling reason to split.

----

P.S. I see some fixes that are independent of this PR here. Please split them into another PR so they can already go in.

elmarco:
I didn't remember someone suggested we use an existing crate. There are many differences between dbus-addr and  dbus-server-address-parser. dbus-addr is rooted in zbus, and uses Cow everywhere it possibly can, and parses things strictly.

Your Address changes are also very recent (more recent than mine!), so this argument doesn't hold much.

Imho splitting code in separate crate is a good practice, even if there is only a single user so far (the crate was released yesterday...)

zeenix:
> Your Address changes are also very recent (more recent than mine!), so this argument doesn't hold much.

Those changes are based on what we discussed and agreed many months ago (although they don't include all the changes we agreed on). Also, I was doing almost all the work for 4.0 so they were not disruptive to my planning.

Also, I didn't introduce a completely new crate at the last minute.

> Imho splitting code in separate crate is a good practice, even if there is only a single user so far (the crate was released yesterday...)

Let's be a bit specific. Is it useful for anyone not using zbus, in practice (i can see this being true in theory)? If not, I don't see why we need to do this now. We can do it if/when the use case presents itself.

To be very honest, this split seems like your attempt to workaround my objections/concerns to your PR. That is not a good rationale for me to accept this, sorry.

elmarco:
"completely new" is not correct, it's rooted on zbus code, the same tests are used (with a few more tests).

The fact that there are already crates parsing dbus addresses shows that there is interest in having independent crate for it. And again, if code can be split in a separate crate, it is generally a good idea to do it. Many rust crates are neatly divided, even if they belong and are used only by to a top project.

I recall your main objection was that you didn't want to break API, but now you did.

zeenix:
> The fact that there are already crates parsing dbus addresses shows that there is interest

We can't assume that. Are those crates being actually used by other dbus crates to be useful? Please point to specific use cases this will satisfy.

> I recall your main objection was that you didn't want to break API, but now you did.

That was not my objection at all. [I in fact told you the exact opposite](https://github.com/dbus2/zbus/pull/430#discussion_r1293319363). :) Before #430, we agreed on Address becoming a struct and that is a clear API-break. So I don't know how you got this very wrong impression. Please see the logs and my comments your old PR again.

---

Bottom line, I'm not objecting to split per say but:

1. It will have to be a split of our own API (mostly just moving). That would also make it possible to split w/o breaking API as we can provide re-exports.
2. I don't see any compelling reason to do this now. If we can do it w/o breaking API, it can happen soon. If not, we'll wait for 5.0.

elmarco:
> > The fact that there are already crates parsing dbus addresses shows that there is interest
> 
> We can't assume that. Are those crates being actually used by other dbus crates to be useful? Please point to specific use cases this will satisfy.

Address parsing without connection? That would be for example configuration file handling. It could also be testing tools that just want to connect to a dbus server without the need for dbus implementations (fuzzers etc). Or simple proxy code for ex.

> Bottom line, I'm not objecting to split per say but:
> 
>     1. It will have to be a split of our own API (mostly just moving). That would also make it possible to split w/o breaking API as we can provide re-exports.

"split of our own API" ? what does this mean?

>     2. I don't see any compelling reason to do this now. If we can do it w/o breaking API, it can happen soon. If not, we'll wait for 5.0.

The main compelling reasons are:
- ;-addr list handling
- remove the burden from zbus itself, sharing responsability
- zbus code cleanups
- various code cleanups and improvements in dbus-addr too

zeenix:
> > > The fact that there are already crates parsing dbus addresses shows that there is interest
> > 
> > 
> > We can't assume that. Are those crates being actually used by other dbus crates to be useful? Please point to specific use cases this will satisfy.
> 
> Address parsing without connection? That would be for example configuration file handling. It could also be testing tools that just want to connect to a dbus server without the need for dbus implementations (fuzzers etc). Or simple proxy code for ex.

Can you show me one example in any language of this? If you're going to connect to dbus, why would you not just use the full D-Bus API?

You continue to just come up with random theoretical possibilities as the rationale, Please be specific and give practical examples.

> "split of our own API" ? what does this mean?

A simple (mostly) verbatim move. This will allow us to be able to do that even in a stable release, using re-exports. So we finalize the API in zbus first and then split it.

>     * ;-addr list handling

Given that you addressed this in your PR, you know that this is not relevant to splitting so why present it as an argument for that?

>     * remove the burden from zbus itself, sharing responsability

My question was, what's the compelling reason to do this **now**?

>     * zbus code cleanups
>     * various code cleanups and improvements in dbus-addr too

Both are not relevant to splitting. The same could be achieved w/o the split.

zeenix:
I spent enough time on this already and you didn't succeed in convincing me that this is something we need to do right now. I'm closing this with the bottom line [above](https://github.com/dbus2/zbus/pull/562#issuecomment-1921449788). Sorry! I'll also now unsubscribe from this issue. Feel free to provide PRs for improvements to the Address API for 4.0 (I will hold it off till Sun for you) and create PRs for address split after 4.0.

Oh and I would want to be a co-maintainer of such a split and it should remain in zbus umbrella and repo (i-e `zbus_address`).

elmarco:
> You continue to just come up with random theoretical possibilities as the rationale, Please be specific and give practical examples.

Those are not random, they are practical. Haven't you ever seen address parsing or url or email without actually implementing a browser or a mail client. You are not considering seriously my argument. And even then, I don't agree with you that it *needs* different users to have code split in a different crate.

> 
> > "split of our own API" ? what does this mean?
> 
> A simple (mostly) verbatim move. This will allow us to be able to do that even in a stable release, using re-exports. So we finalize the API in zbus first and then split it.

It's not a move, I wrote the code from scratch before your implementation. I am not going to justify every choice in implementation I made, (you didn't have to justify yours btw... since nobody reviewed)

> 
> > ```
> > * ;-addr list handling
> > ```
> 
> Given that you addressed this in your PR, you know that this is not relevant to splitting so why present it as an argument for that?

Because dbus-addr handles it with this PR, and this is compelling to be compliant with dbus address handling.

> 
> > ```
> > * remove the burden from zbus itself, sharing responsability
> > ```
> 
> My question was, what's the compelling reason to do this **now**?
> 

Actually, I proposed to do it in the original PR. And you pressed me to release my code somewhere to not just have to trash it...

> > ```
> > * zbus code cleanups
> > * various code cleanups and improvements in dbus-addr too
> > ```
> 
> Both are not relevant to splitting. The same could be achieved w/o the split.

Well it's done in dbus-addr and with this PR. But you don't want to consider it seriously.

PR #565, opened 2024-02-01
🩹 fix impl (Try)From for owning values

anphir:
When trying to implement the builder pattern working with `zvariant::Value` arguments, I encountered a strange compiler error about lifetimes when using generic functions.

The following simplified code shows the problem, with comments explaining the error encountered (functions `fail1`, `fail2`).

```rust
use zbus::zvariant;

/// A simplified builder to focus on the compile problem.
pub struct Values<'a> {
    values: Vec<zvariant::Value<'a>>,
}

impl<'a> Values<'a> {
    pub fn new() -> Self {
        Self { values: Vec::new() }
    }

    /// Adds a value.
    /// 
    /// Accepts anything than can be converted into a `zvariant::Value` with
    /// a sufficient lifetime.
    pub fn value<V>(mut self, value: V) -> Self
    where
        V: Into<zvariant::Value<'a>>,
    {
        self.values.push(value.into());
        self
    }
}

// Using explicit lifetimes to show eliding is not a problem.

// As expected, this compiles just fine.
pub fn ok1<'a>(value1: &'a str) -> Values<'a> {
    Values::new().value(value1)
}

// Surprisingly this fails to compile.
// The compiler claims that "`'a` must outlive `'static`".
pub fn fail1<'a>(value1: &'a str) -> Values<'a> {
    Values::new().value(value1).value("owned value".to_string())
}

// This fails to compile too.
pub fn fail2<'a>(value1: &'a str) -> Values<'a> {
    let value2: String = "owned value".to_string();
    Values::new().value(value1).value(value2)
}

// Explicitly converting into a `zvariant::Value` works, though.
pub fn ok2<'a>(value1: &'a str) -> Values<'a> {
    let value2: zvariant::Value<'a> = "owned value".to_string().into();
    Values::new().value(value1).value(value2)
}

// It works also when using a `static` lifetime.
pub fn ok3<'a>(value1: &'a str) -> Values<'a> {
    let value2: zvariant::Value<'static> = "owned value".to_string().into();
    Values::new().value(value1).value(value2)
}

fn main() {
    println!("DONE");
}

```

Looking at zbus source code I noticed that `From<String>` is implemented for static lifetime only for `zvariant::Value`,
in contrast to e.g. `zvariant::Str` where it is implemented for all lifetimes.
When relying on the compiler to derive types (as in function `fail1`), this seems to make it think lifetimes need to be static.
Implementing `From<String>` for all lifetimes (as already is the case for `zvariant::Str`) fixes this problem.

I noticed that other owning types have also implementations of `From` and `TryFrom` limited to static lifetime, but wanted to keep the changes small.
If you think they are correct, I'm willing to look for other `impl`s needing an update.

### DIFF
--- a/zbus/src/guid.rs
+++ b/zbus/src/guid.rs
@@ -92,7 +92,7 @@ impl<'g> TryFrom<Str<'g>> for Guid<'g> {
     }
 }
 
-impl TryFrom<String> for Guid<'static> {
+impl TryFrom<String> for Guid<'_> {
     type Error = crate::Error;
 
     fn try_from(value: String) -> std::result::Result<Self, Self::Error> {
@@ -191,7 +191,7 @@ impl Borrow<str> for OwnedGuid {
     }
 }
 
-impl From<OwnedGuid> for Guid<'static> {
+impl From<OwnedGuid> for Guid<'_> {
     fn from(o: OwnedGuid) -> Self {
         o.0
     }
@@ -209,7 +209,7 @@ impl From<Guid<'_>> for OwnedGuid {
     }
 }
 
-impl From<OwnedGuid> for Str<'static> {
+impl From<OwnedGuid> for Str<'_> {
     fn from(value: OwnedGuid) -> Self {
         value.0 .0
     }
--- a/zbus/src/match_rule/mod.rs
+++ b/zbus/src/match_rule/mod.rs
@@ -511,7 +511,7 @@ impl Deref for OwnedMatchRule {
     }
 }
 
-impl From<OwnedMatchRule> for MatchRule<'static> {
+impl From<OwnedMatchRule> for MatchRule<'_> {
     fn from(o: OwnedMatchRule) -> Self {
         o.into_inner()
     }
--- a/zbus_names/src/bus_name.rs
+++ b/zbus_names/src/bus_name.rs
@@ -306,7 +306,7 @@ impl<'name> From<&BusName<'name>> for BusName<'name> {
     }
 }
 
-impl TryFrom<OwnedValue> for BusName<'static> {
+impl TryFrom<OwnedValue> for BusName<'_> {
     type Error = Error;
 
     fn try_from(value: OwnedValue) -> Result<Self> {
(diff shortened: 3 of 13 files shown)


zeenix:
@anphir Thanks so much for the PR. I'll have a look soon. I'm sorry for keeping you waitng but I'm currently at FOSDEM conf and busy with trying to roll out 4.0.

zeenix:
@alyssais These changes look good to me. I should have taken a look earlier as they're pretty straight-forward. :) 

>  If you think they are correct, I'm willing to look for other impls needing an update.

Yes please! Also, appreciate you looked at our contribution guide before submission but please add name/abbreviation of each subcrate you modify in the commit log summary part (e,g `zb,zb,..`). Best would be to keep each subcrate changes in separate commits.

anphir:
> @alyssais These changes look good to me ...

I don't know if @alyssais is willing to work on this, but I still am. ;-)

> Yes please! Also, appreciate you looked at our contribution guide before submission but please add name/abbreviation of each subcrate you modify in the commit log summary part (e,g `zb,zb,..`). Best would be to keep each subcrate changes in separate commits.

So far I noticed only zvariant needing more changes, but will check all code.
Anyway I will keep _this_ commit to include changes to zvariant only.

zeenix:
> > @alyssais These changes look good to me ...
> 
> I don't know if @alyssais is willing to work on this, but I still am. ;-)

Sorry, for some reason GH recommended that when I typed `@a` here. :)

> Anyway I will keep _this_ commit to include changes to zvariant only.

Great. Just to be clear, I only mean commit. It can and should go in the same PR.

zeenix:
Converting to draft for now to avoid accidental merges.

anphir:
> Just to be clear, I only mean commit. It can and should go in the same PR.

OK, thanks for clarifying.

When checking the source code of all packages, I found some others besides zvariant needing changes.
So if I understood correctly, we should have multiple commits with appropriate prefix (e.g. zb for zbus) within this single pull request.

zeenix:
> So if I understood correctly, we should have multiple commits with appropriate prefix (e.g. zb for zbus) within this single pull request.

Correct. :)

zeenix:
@anphir Thanks so much!

PR #568, opened 2024-02-04
zb: Add missing nix feature

bilelmoussaoui:
The nix::unistd::Uid API requires the user feature, per https://docs.rs/nix/0.26.4/nix/unistd/struct.Uid.html

Somehow it worked out by luck as some other dependency was enabling that feature...

### DIFF
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -80,7 +80,7 @@ winapi = { version = "0.3", features = [
 uds_windows = "1.0.2"
 
 [target.'cfg(unix)'.dependencies]
-nix = { version = "0.26.0", default-features = false, features = ["socket", "uio"] }
+nix = { version = "0.26.0", default-features = false, features = ["socket", "uio", "user"] }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 # FIXME: This should only be enabled if async-io feature is enabled but currently


rjmac:
The other dependency was xdg-home.  It released 1.1.0 this morning, in which it stopped depending on nix itself.  If your build was broken by this, pinning xdg-home to 1.0.0 until zbus is updated should fix it.

ogoffart:
It worked before because the feature was was bought up by the xdg_home crate.
But the new release of the xdg_home create a few hours ago doesn't depend on the nix crate anymore and therefore broke the zbus crate.
And now zbus and therefore most crates that depends on it are no longer building

A new 3.x release of the zbus crate with this patch would solve the problem and would be much appreciated

bilelmoussaoui:
> It worked before because the feature was was bought up by the xdg_home crate. But the new release of the xdg_home create a few hours ago doesn't depend on the nix crate anymore and therefore broke the zbus crate. And now zbus and therefore most crates that depends on it are no longer building
> 
> A new 3.x release of the zbus crate with this patch would solve the problem and would be much appreciated

I see, thanks for the investigation.

zeenix:
I'm so sorry. I didn't realise releasing a package that removes a dependency could break everything. 🤦‍♂️

I'll cherry-pick on 3.x branch soon and roll out a release.

zeenix:
> I'll cherry-pick on 3.x branch soon and roll out a release.

oh this **was** for the 3.x branch. We don't need it on main (4) branch? :thinking:

bilelmoussaoui:
> > I'll cherry-pick on 3.x branch soon and roll out a release.
> 
> oh this **was** for the 3.x branch. We don't need it on main (4) branch? 🤔

You already have it in the main branch.

PR #571, opened 2024-02-05
📝book: update zbus-xmlgen commands

FineFindus:
Update the `zbus-xmlgen` commands in the book to the new cli introduced in https://github.com/dbus2/zbus/pull/505


zeenix (approved):
Thanks.

PR #586, opened 2024-02-11
🚑️ zb: Fix a pointer access on Windows

zeenix:
Details at: https://github.com/microsoft/windows-rs/issues/2842

### DIFF
--- a/zbus/src/win32.rs
+++ b/zbus/src/win32.rs
@@ -214,7 +214,7 @@ pub fn socket_addr_get_pid(addr: &SocketAddr) -> Result<u32, Error> {
     let tcp_table = tcp_table.as_mut_ptr().cast::<MIB_TCPTABLE2>();
     let num_entries = unsafe { (*tcp_table).dwNumEntries };
     for i in 0..num_entries {
-        let entry = unsafe { (*tcp_table).table.get_unchecked(i as usize) };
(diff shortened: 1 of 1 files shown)


zbus/src/win32.rs
```diff
@@ -214,7 +214,7 @@ pub fn socket_addr_get_pid(addr: &SocketAddr) -> Result<u32, Error> {
     let tcp_table = tcp_table.as_mut_ptr().cast::<MIB_TCPTABLE2>();
     let num_entries = unsafe { (*tcp_table).dwNumEntries };
     for i in 0..num_entries {
-        let entry = unsafe { (*tcp_table).table.get_unchecked(i as usize) };
+        let entry = unsafe { *(*tcp_table).table.as_ptr().add(i as usize) };
```
Nerixyz, line 217: @zeenix I think it would also be possible to use [std::slice::from_raw_parts](https://doc.rust-lang.org/std/slice/fn.from_raw_parts.html#), which might give the compiler a better understanding of what you're doing, and you could use a regular loop.

So you could have something like

```rs
let entries = unsafe { std::slice::from_raw_parts((*tcp_table).table.as_ptr(), (*tcp_table).dwNumEntries as usize) };
```

zeenix: Oh [that works](https://github.com/dbus2/zbus/pull/587) and you're right, it makes things much clearer. Thanks!

PR #588, opened 2024-02-12
zb: Implement ProxyDefault for blocking::Proxy

bilelmoussaoui:


zbus/src/blocking/proxy/mod.rs
```diff
@@ -355,6 +359,12 @@ impl<'a> Proxy<'a> {
     }
 }
 
+impl<'a> ProxyDefault for Proxy<'a> {
```
zeenix, line 362: nit: you can elide the lifetimes here:

```rust
impl ProxyDefault for Proxy<'_>
```

zeenix (changes requested):
LGTM otherwise. Thanks!

zeenix (approved):

PR #595, opened 2024-02-16
🐛zx: correctly check if fdo ifaces is empty

FineFindus:
Fixes an issue where instead of checking if `fdo_standard_ifaces` was empty, it was accidentally checked if the prefix string was empty. Found by (and thanks to) @Cobinja in https://github.com/dbus2/zbus/commit/35c15c8d9e8e730dee6bdbfa6997cd7d83a8f738#r138622943.

I totally blame clippy for not having a lint against this :)


zeenix:
> I totally blame clippy for not having a lint against this :)

They'd be happy to receive a report on this. :wink:

FineFindus:
[Done](https://github.com/rust-lang/rust-clippy/issues/12307) :)

PR #648, opened 2024-02-22
Reproduce and resolve #647

### LINKED ISSUE #647: zbus 4.0.1 property read never returns, or hangs
I'm having some difficulty chasing this one down. Just did an update to 4.0.1. Everythign so far is okay except property reads (writes are okay),

Server:
```rust
use zbus::fdo::Error as ZbErr;

    /// Return the current LED brightness
    #[zbus(property)]
    async fn brightness(&self) -> Result<LedBrightness, ZbErr> {
        let ctrl = self.0.lock().await;
        dbg!("I was called"); // yes this prints, the call was made
        Ok(ctrl.sysfs_node.get_brightness().map(|n| n.into())?) // but never returns???
    }

    /// Set the keyboard brightness level (0-3)
    #[zbus(property)]
    async fn set_brightness(&mut self, brightness: LedBrightness) -> Result<(), ZbErr> {
        let ctrl = self.0.lock().await;
        Ok(ctrl.sysfs_node.set_brightness(brightness.into())?)
    }
```

Interface example:
```rust
    /// Brightness property
    #[zbus(property)]
    fn brightness(&self) -> zbus::Result<LedBrightness>;
    #[zbus(property)]
    fn set_brightness(&self, value: LedBrightness) -> zbus::Result<()>;
```

What xml gen gives (and I update, this also doesn't read):
```rust
    /// Brightness property
    #[zbus(property)]
    fn brightness(&self) -> zbus::Result<u32>;
    #[zbus(property)]
    fn set_brightness(&self, value: u32) -> zbus::Result<()>;
```

I've used D-Feet to read all properties fine. So the server at least is fully functional. It is only the client side that is an issue, and only with reads.

The absolute minimum in an example main():
```rust
    let conn = zbus::blocking::Connection::system().unwrap();
    let a = rog_dbus::zbus_aura::AuraProxyBlocking::new(&conn).unwrap(); // from xml gen
    dbg!(a.set_brightness(rog_aura::LedBrightness::Low).unwrap());
    dbg!(a.brightness().unwrap()); // hung here
```

I'm not sure what other info I can give here.

zeenix:


zeenix:
See https://github.com/dbus2/zbus/issues/647#issuecomment-1960382750

PR #651, opened 2024-02-22
zvariant: Fix compilation of gvariant feature on non-unix platforms

felinira:
This would fail compilation on eg. windows because the FD generic does not actually exist there.

Reproduce via `cargo build --features="gvariant" --target=x86_64-pc-windows-gnu`


zeenix:
Thanks so much @felinira . We should enable gvariant testing on Windows CI so these things don't happen.

zeenix:
@felinira btw, for future contributions, kindly add emoji prefixes to the commits and make use of abbreviated crate name, as per [our contributing guidelines](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md).

PR #656, opened 2024-02-23
⬆️  🚧 zb: Bump async-executor to latest (WIP)

zeenix:
For now we're just testing the fix for the latest release of async-executor, that was yanked because of a deadlock.


zeenix:
This was superseded by #698.

PR #668, opened 2024-03-05
zb: ObjectServer should handle methods sequentially

elmarco:
Messages should be dispatched in the order they are received, otherwise, the client can't do message batching etc..
    
Spawning a task is best left to the actual implementation.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -996,13 +996,16 @@ impl Connection {
                                     // destination doesn't matter if no name has been registered
                                     // (probably means name it's registered through external means).
                                     if !names.is_empty() && !names.contains_key(dest) {
-                                        trace!("Got a method call for a different destination: {}", dest);
+                                        trace!(
+                                            "Got a method call for a different destination: {}",
+                                            dest
+                                        );
 
                                         continue;
                                     }
                                 }
                             }
-                            let member = match hdr.member() {
+                            match hdr.member() {
                                 Some(member) => member,
                                 None => {
                                     warn!("Got a method call with no `MEMBER` field: {}", msg);
@@ -1010,27 +1013,16 @@ impl Connection {
                                     continue;
                                 }
                             };
-                            trace!("Got `{}`. Will spawn a task for dispatch..", msg);
-                            let executor = conn.inner.executor.clone();
-                            let task_name = format!("`{member}` method dispatcher");
-                            executor
-                                .spawn(
-                                    async move {
-                                        trace!("spawned a task to dispatch `{}`.", msg);
-                                        let server = conn.object_server();
-                                        if let Err(e) = server.dispatch_message(&msg).await {
-                                            debug!(
-                                                "Error dispatching message. Message: {:?}, error: {:?}",
-                                                msg, e
-                                            );
-                                        }
-                                    }
-                                    .instrument(trace_span!("{}", task_name)),
-                                    &task_name,
-                                )
-                                .detach();
+                            let server = conn.object_server();
+                            if let Err(e) = server.dispatch_message(&msg).await {
+                                debug!(
+                                    "Error dispatching message. Message: {:?}, error: {:?}",
+                                    msg, e
+                                );
+                            }
                         } else {
-                            // If connection is completely gone, no reason to keep running the task anymore.
+                            // If connection is completely gone, no reason to keep running the task
+                            // anymore.
                             trace!("Connection is gone, stopping associated object server task");
                             break;
                         }
--- a/zbus/src/message/field.rs
+++ b/zbus/src/message/field.rs
@@ -32,7 +32,7 @@ pub(super) enum FieldCode {
     ErrorName = 4,
     /// Code for [`Field::ReplySerial`](enum.Field.html#variant.ReplySerial)
     ReplySerial = 5,
-    /// Code for [`Field::Destinatione`](enum.Field.html#variant.Destination)
+    /// Code for [`Field::Destination`](enum.Field.html#variant.Destination)
     Destination = 6,
     /// Code for [`Field::Sender`](enum.Field.html#variant.Sender)
     Sender = 7,
--- a/zbus/src/message/mod.rs
+++ b/zbus/src/message/mod.rs
@@ -295,6 +295,7 @@ impl fmt::Debug for Message {
         let mut msg = f.debug_struct("Msg");
         let h = self.header();
         msg.field("type", &h.message_type());
+        msg.field("serial", &self.primary_header().serial_num());
         if let Some(sender) = h.sender() {
             msg.field("sender", &sender);
         }
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -699,7 +699,6 @@ impl ObjectServer {
         })
     }
 
-    #[instrument(skip(self, connection))]
     async fn dispatch_method_call_try(
         &self,
         connection: &Connection,
@@ -766,7 +765,6 @@ impl ObjectServer {
         )))
     }
 
-    #[instrument(skip(self, connection))]
     async fn dispatch_method_call(&self, connection: &Connection, msg: &Message) -> Result<()> {
         match self.dispatch_method_call_try(connection, msg).await {
             Err(e) => {


zbus/src/connection/mod.rs
```diff
@@ -1011,26 +1014,16 @@ impl Connection {
...
+                            if let Err(e) = server.dispatch_message(&msg).await {
+                                debug!(
+                                    "Error dispatching message. Message: {:?}, error: {:?}",
+                                    msg, e
+                                );
+                            }
                         } else {
-                            // If connection is completely gone, no reason to keep running the task anymore.
+                            // If connection is completely gone, no reason to keep running the task
+                            // anymore.
```
zeenix, line 1026: Irrelevant change to the commit?

zeenix (changes requested):
> Messages should be dispatched in the order they are received, otherwise,
the client can't do message batching etc..

I'm not against this change but it's pretty big behavior change so let me play devil's advocate here:

1. what exactly is message batching and how does the current behavior breaks this use case?
2. since nobody really complained about this before, I'm willing to bet the order is only important in niche use cases. Given that and also that this is a behavior change, I think we should provide an option for the new behavior and document this better.

Also to keep in mind that spawning a task doesn't necessarily imply methods will be handled out of order. With some runtimes, it may always be handled in order (especially if the impls are synchronous and runtime is single-threaded).

> zb: fix out of order message dispatching

nit: This makes it sound like a pure bug fix rather than a behavior that's intended. Let's summarize the change in the shortlog, not the effect/benefit as per our guidelines. Also it's not clear that you're talking about method call dispatches only.

zeenix:
> Also to keep in mind that spawning a task doesn't necessarily imply methods will be handled out of order. With some runtimes, it may always be handled in order (especially if the impls are synchronous and runtime is single-threaded).

Thinking more about this, the intention was also that object server can handle multiple method calls at once and I think that's a good default behavior to have. So I'm leaning more and more towards the idea of providing an option (perhaps ObjectServer setter method or connection::Builder method) to explicitly enable serializing of method dispatch.

zbus/src/connection/mod.rs
elmarco: yep.. but CI cargo fmt complains.

elmarco:
> what exactly is message batching and how does the current behavior breaks this use case?

Sending together "DoThisFirst() ThenDoThat()" without having to wait for a reply for DoThisFirst().

Reordering signals can also have bad consequences. For example, with position tracking etc..

> since nobody really complained about this before, I'm willing to bet the order is only important in niche use cases. Given that and also that this is a behavior change, I think we should provide an option for the new behavior and document this better.

I think they didn't complained yet because it's happening under stressed environment. Or because figuring out where the problems comes from is hard. Or because they work around it, using serial etc.. I am quite confident that none of the dbus libraries out there do this kind of out-of-order dispatching by default.

> nit: This makes it sound like a pure bug fix

Given then rationale above, it's a bug fix to me. 

> the intention was also that object server can handle multiple method calls

I understand, but as I said, you better leave that to the implementation discretion. Typically, good IPC handlers should be quick anyway, and have second-half handlers for long running tasks or other mechanisms.

Imho having an object server option to spawn tasks on each message isn't necessary or desirable.

elmarco:
commit 2da0ecd98f449aa2fcae2ff7f656fe91a13d8b9a

@danieldg 

"dbus clients that depend on call ordering will wait for the first call to return before making the second" that's not true when you do batching.

Is the deadlock you are talking about reentering zbus dispatch? Could you provide an example/test?

"Interfaces already use locking to ensure they only handle one mutating call at a time" so that kind of deadlock is still possible?

zbus/src/connection/mod.rs
zeenix: > but CI cargo fmt complains.

that doesn't mean it should go in irrelevant commit. :)

zeenix:
> > what exactly is message batching and how does the current behavior breaks this use case?
> 
> Sending together "DoThisFirst() ThenDoThat()" without having to wait for a reply for DoThisFirst().

That's not a use case. What's the use case for not waiting for responses? If the order of execution is important, then why would not wait for replies? In case of Rust, there are APIs for running futures in parallel and sequentially in a batch. You can make use of that.

> Reordering signals can also have bad consequences. For example, with position tracking etc..

That's not relevant here. This is specifically about processing of method calls by ObjectServer.

> > ince nobody really complained about this before, I'm willing to bet the order is only important in niche use cases. Given that and also that this is a behavior change, I think we should provide an option for the new behavior and document this better.
> 
> I think they didn't complained yet because it's only happening under stressed environment. Or because figuring out where the problems comes from is hard. Or because they work around it, using serial etc.. I am quite confident that none of the dbus libraries out there do this kind of out-of-order dispatching by default.

It's a lot more likely that people want to wait for replies of a method before firing other methods, especially if the ordering of method calls is important. Where the order is not important then there is no problem.

Clients relying on how a service handle requests and timing etc, is a bad idea as that's internal logic of the service. As you say yourself, services can spawn tasks themselves (although not true, see below) so when they do, can client really rely on the timing then? I don't see any diff for the client, whether service does it itself or zbus does it for them.

> > nit: This makes it sound like a pure bug fix
> 
> Given then rationale above, it's a bug fix to me.

You didn't read what a wrote carefully. Stress on the word "pure". Most changes are bug fixes in the end but there is a difference between fixing a mistake and changing an intentional behavior to fix a bug.

> I understand, but as I said, you better leave that to the implementation discretion.

How do they do that? ObjectServer will still await for a response so no matter how many tasks the service spawns, the methods will still be processed sequentially. That means, services won't be able to handle multiple requests at the same time.

> Typically, good IPC handlers should be quick anyway, and have second-half handlers for long running tasks or other mechanisms.

Yeah but we can't rely on that and it depends on so many factors.

> "dbus clients that depend on call ordering will wait for the first call to return before making the second" that's not true when you do batching.

Can you show me examples of D-Bus code out there that does this batch processing while the ordering is important to it? I am very doubtful that this combo exists (except for your code) and if it does, it's super niche and relies on assumptions about service implementation internals that is not backed by any spec requirement or even recommendation.

danieldg:
This change, as I understand it, would prevent *any* ObjectServer methods from running in parallel. It makes the distinction between `&self` and `&mut self` on `Interface` implementers much less useful. But after reading the discussion, I now think my initial parallelization assumptions were a bit too aggressive.

> Is the deadlock you are talking about reentering zbus dispatch? Could you provide an example/test?
>
> "Interfaces already use locking to ensure they only handle one mutating call at a time" so that kind of deadlock is still possible?

The deadlock this was trying to fix was if, for example, one binary wanted to provide an implementation of both  `org.freedesktop.portal.Desktop` and `org.freedesktop.Notifications`, but not seize ownership of the names (allowing others to `request_name`).  The `Desktop` portal object methods are proxies that call the `Notifications` protocol to do the actual work, so they would be making a call to the object server - but to a different name and path, so it wouldn't deadlock even with `&mut self` calls.  I don't have an example, however, and the ones that I can think of are often better served by having multiple dbus `Connection` objects.  Multiple `Connection` objects is currently used to work around coarse granularity in dbus filtering (https://github.com/flatpak/flatpak/issues/4077), so perhaps it's a good solution to work around recursion too.

A partial solution: have each `Interface` have a task for handling its incoming message queue (this task could be lazily started and exit if the queue is empty, if we want to avoid having a ton of idle tasks and queues taking up memory).  That only ensures on a *per-object* basis that ordering is stable, however; I could see inter-object ordering stability being desirable too.  No proper batched call ordering on objects not known to be served by the same client is possible, of course, but I think the `systemd` dbus API creates sub-objects in ways that could end up being well-ordered enough to batch calls.

Regarding the "spawn your own task" solution: the macro-provided API also doesn't provide an easy way to launch a task to handle a method - the result of the call must be present before you return from the `async fn`.  Even the raw `Interface` API doesn't make it easy, as you only have `&'call self`, which means that you need to take and use `InterfaceRef` just to get access to your data.

Maybe we should extend the interface macros to add an annotation to make it easier to "detach" a known long-running call from the object server - something like `#[detached]` that produces the current behavior, but only for that method.

Alternatively, we could add a `#[zbus(result_sender)]` parameter that adds an object with one `submit_result<R>(self, result: R)` method - which would be movable into a detached task.  The actual function would have return type `()`, and could do things like choose to handle the task immediately or spawn it off.

elmarco:
See also my question on dbus spec:
https://gitlab.freedesktop.org/dbus/dbus/-/issues/179#note_2317797

I will try to update my PR based on the various comments and concerns.

elmarco:
It's not necessary a break, but rather a clarification I am looking for. My example with the bus implementation should already give a hint of what is implicitly expected by users, and most existing implementations respect the delivery ordering for that reason.

> Before you do all the work, can you give a summary of your intended changes? I wouldn't want you to waste your time.

I am leaning towards making task spawning an option on object server, defaulting to true, with an opt-in/out flag on the method macro.

elmarco:
> It's not necessary a break, but rather a clarification I am looking for. My example with the bus implementation should already give a hint of what is implicitly expected by users, and most existing implementations respect the delivery ordering for that reason. (btw, parallel handling is not incompatible with ordering..)
 
> > Before you do all the work, can you give a summary of your intended changes? I wouldn't want you to waste your time.
> 
> I am leaning towards making task spawning an option on object server, defaulting to true, with an opt-in/out flag on the method macro.

zeenix:
> It's not necessary a break, but rather a clarification I am looking for.

If a requirement was never specified or even hinted to in the spec, making it explicit will definitely be a break. As I pointed out in the dbus issue, we know at least one set of services will become non-complaint: zbus-based ones.



> I am leaning towards making task spawning an option on object server, defaulting to true, with an opt-in/out flag on the method macro.

Sure, that sounds good to me, as long as the implementation isn't too complicated.

elmarco:
closing in favor of optional no_spawn in #675

elmarco:
closing in favor of optional no_spawn in https://github.com/dbus2/zbus/pull/675

Issue #674, opened 2024-03-13
A better object_server::Interface


elmarco:
`object_server::Interface` could use a better design (https://docs.rs/zbus/latest/zbus/object_server/trait.Interface.html)

- The whole trait shouldn't need to be async. But only the "object methods".
- We shouldn't have to do double lookups for Ref and Mut versions.
- Properties and Methods should be treated more equal. Only getters can use a special tratement to make sure they return Values and allow simpler code generation or building GetAll().
- Properties shouldn't need to return OwnedValues.
- member lookups could use perfect hash, and static/const versions.
- the XML interface could use a static XML AST instead of a dynamically built string.
- the Interface types should be private to the proxy macro (similar to other macros crates, like serde etc), so we can evolve this without API break.

We should be able to introduce an experimental Interface2 in parallel with the existing Interface, by opt-in feature or macro flag, if API breaking isn't an option. Or we could adapt the code by gradually moving/replacing Interface trait methods.

After some brainstorming, this is what I would aim for atm:
```rust
// we can't store "async fn", so we return a future instead
type FutureFn<'a> = Pin<Box<dyn Future<Output = ()> + Send + 'a>>;

// this will need extra context arguments: conn, msg etc..
enum Callable {
    Fn(for<'a> fn(&'a dyn Any) -> Result<FutureFn<'a>, String>),
    FnMut(for<'a> fn(&'a mut dyn Any) -> Result<FutureFn<'a>, String>),
}

struct Method {
    name: MemberName<'static>,
    callable: Callable,
}

// TODO: use future like Callable
// even getters can be mut, why not
enum Getter {
    Fn(fn(&dyn Any) -> fdo::Result<Value>),
    FnMut(fn(&mut dyn Any) -> fdo::Result<Value>),
}

struct Property {
    name: PropertyName<'static>,
    getter: Option<Getter>,
    setter: Option<Callable>,
}

trait Interface {
    const NAME: InterfaceName<'static>;

    const METHODS: phf::Map<&'static str, Method>;

    const PROPERTIES: phf::Map<&'static str, Property>;

    // ideally we would build a static AST, but zbus_xml needs changes
    // const XML: xml::Interface<'static>;
}
```

zeenix:
> > > The whole trait shouldn't need to be async. But only the "object methods".
> > 
> > 
[...]

PR #675, opened 2024-03-15
Make task spawning for method call optional

elmarco:
Clients can expect calls to be handled in order they are sent. Although the specification doesn't explicitly state so  (https://gitlab.freedesktop.org/dbus/dbus/-/issues/179), bus implementations have respected message ordering. This can also be  expected when the bus and client are fully managed, as in p2p.

Add an option to avoid the default task spawning when dispatching methods on a specific interface.

This currently prevents parallel handling of other incoming calls which will have to wait for the method to return (current pending calls are still running in parallel).

Supersedes: #668

### DIFF
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -300,6 +300,9 @@ impl<'a> Builder<'a> {
     /// interfaces available immediately after the connection is established. Typically, this is
     /// exactly what you'd want. Also in contrast to [`zbus::ObjectServer::at`], this method will
     /// replace any previously added interface with the same name at the same path.
+    ///
+    /// Standard interfaces (Peer, Introspectable, Properties) are added on your behalf. If you
+    /// attempt to add yours, [`Builder::build()`] will fail.
     pub fn serve_at<P, I>(mut self, path: P, iface: I) -> Result<Self>
     where
         I: Interface,
@@ -448,13 +451,13 @@ impl<'a> Builder<'a> {
             for (path, interfaces) in self.interfaces {
                 for (name, iface) in interfaces {
                     let iface = iface.clone();
-                    let future =
-                        object_server
-                            .inner()
-                            .at_ready(path.to_owned(), name.clone(), || iface);
-                    let added = future.await?;
-                    // Duplicates shouldn't happen.
-                    assert!(added);
+                    let added = object_server
+                        .inner()
+                        .at_ready(path.to_owned(), name.clone(), || iface)
+                        .await?;
+                    if !added {
+                        return Err(Error::InterfaceExists(name.clone(), path.to_owned()));
+                    }
                 }
             }
 
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -450,10 +450,9 @@ impl<'a> Builder<'a> {
             let object_server = conn.sync_object_server(false, None);
             for (path, interfaces) in self.interfaces {
                 for (name, iface) in interfaces {
-                    let iface = iface.clone();
                     let added = object_server
                         .inner()
-                        .at_ready(path.to_owned(), name.clone(), || iface)
+                        .add_arc_interface(path.clone(), name.clone(), iface.clone())
                         .await?;
                     if !added {
                         return Err(Error::InterfaceExists(name.clone(), path.to_owned()));
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -2,14 +2,11 @@
 use async_io::Async;
 use event_listener::Event;
 use static_assertions::assert_impl_all;
+use std::collections::{HashMap, HashSet, VecDeque};
 #[cfg(not(feature = "tokio"))]
 use std::net::TcpStream;
 #[cfg(all(unix, not(feature = "tokio")))]
 use std::os::unix::net::UnixStream;
-use std::{
-    collections::{HashMap, HashSet, VecDeque},
-    sync::Arc,
-};
 #[cfg(feature = "tokio")]
 use tokio::net::TcpStream;
 #[cfg(all(unix, feature = "tokio"))]
@@ -25,7 +22,6 @@ use zvariant::{ObjectPath, Str};
 
 use crate::{
     address::{self, Address},
-    async_lock::RwLock,
     names::{InterfaceName, WellKnownName},
     object_server::{ArcInterface, Interface},
     Connection, Error, Executor, Guid, OwnedGuid, Result,
@@ -312,8 +308,7 @@ impl<'a> Builder<'a> {
     {
         let path = path.try_into().map_err(Into::into)?;
         let entry = self.interfaces.entry(path).or_default();
-        entry.insert(I::name(), ArcInterface(Arc::new(RwLock::new(iface))));
-
+        entry.insert(I::name(), ArcInterface::new(iface));
         Ok(self)
     }
 
@@ -452,9 +447,11 @@ impl<'a> Builder<'a> {
             let object_server = conn.sync_object_server(false, None);
             for (path, interfaces) in self.interfaces {
                 for (name, iface) in interfaces {
-                    let future = object_server
-                        .inner()
-                        .at_ready(path.to_owned(), name, || iface.0);
+                    let iface = iface.clone();
+                    let future =
+                        object_server
+                            .inner()
+                            .at_ready(path.to_owned(), name.clone(), || iface);
                     let added = future.await?;
                     // Duplicates shouldn't happen.
                     assert!(added);
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -24,7 +24,7 @@ use crate::{
     async_lock::Mutex,
     blocking,
     fdo::{self, ConnectionCredentials, RequestNameFlags, RequestNameReply},
-    message::{Flags, Header, Message, Type},
+    message::{Flags, Message, Type},
     proxy::CacheProperties,
     DBusError, Error, Executor, MatchRule, MessageStream, ObjectServer, OwnedGuid, OwnedMatchRule,
     Result, Task,
@@ -980,35 +980,13 @@ impl Connection {
                                     }
                                 }
                             }
-                            let member = match hdr.member() {
-                                Some(member) => member,
-                                None => {
-                                    warn!("Got a method call with no `MEMBER` field: {}", msg);
-
-                                    continue;
-                                }
-                            };
-                            trace!("Got `{}`. Will spawn a task for dispatch..", msg);
-                            let executor = conn.inner.executor.clone();
-                            let task_name = format!("`{member}` method dispatcher");
-                            // SAFETY: msg becomes 'static when moved to the task, and dropped after
-                            let hdr = unsafe { std::mem::transmute::<_, Header<'static>>(hdr) };
-                            executor
-                                .spawn(
-                                    async move {
-                                        trace!("spawned a task to dispatch `{}`.", msg);
-                                        let server = conn.object_server();
-                                        if let Err(e) = server.dispatch_call(&msg, &hdr).await {
-                                            debug!(
-                                                "Error dispatching message. Message: {:?}, error: {:?}",
-                                                msg, e
-                                            );
-                                        }
-                                    }
-                                    .instrument(trace_span!("{}", task_name)),
-                                    &task_name,
-                                )
-                                .detach();
+                            let server = conn.object_server();
+                            if let Err(e) = server.dispatch_call(&msg, &hdr).await {
+                                debug!(
+                                    "Error dispatching message. Message: {:?}, error: {:?}",
+                                    msg, e
+                                );
+                            }
                         } else {
                             // If connection is completely gone, no reason to keep running the task
                             // anymore.
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -996,7 +996,7 @@ impl Connection {
                                     async move {
                                         trace!("spawned a task to dispatch `{}`.", msg);
                                         let server = conn.object_server();
-                                        if let Err(e) = server.dispatch_message(&msg).await {
+                                        if let Err(e) = server.dispatch_call(&msg).await {
                                             debug!(
                                                 "Error dispatching message. Message: {:?}, error: {:?}",
                                                 msg, e
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -24,7 +24,7 @@ use crate::{
     async_lock::Mutex,
     blocking,
     fdo::{self, ConnectionCredentials, RequestNameFlags, RequestNameReply},
-    message::{Flags, Message, Type},
+    message::{Flags, Header, Message, Type},
     proxy::CacheProperties,
     DBusError, Error, Executor, MatchRule, MessageStream, ObjectServer, OwnedGuid, OwnedMatchRule,
     Result, Task,
@@ -991,12 +991,14 @@ impl Connection {
                             trace!("Got `{}`. Will spawn a task for dispatch..", msg);
                             let executor = conn.inner.executor.clone();
                             let task_name = format!("`{member}` method dispatcher");
+                            // SAFETY: msg becomes 'static when moved to the task, and dropped after
+                            let hdr = unsafe { std::mem::transmute::<_, Header<'static>>(hdr) };
                             executor
                                 .spawn(
                                     async move {
                                         trace!("spawned a task to dispatch `{}`.", msg);
                                         let server = conn.object_server();
-                                        if let Err(e) = server.dispatch_call(&msg).await {
+                                        if let Err(e) = server.dispatch_call(&msg, &hdr).await {
                                             debug!(
                                                 "Error dispatching message. Message: {:?}, error: {:?}",
                                                 msg, e
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -971,7 +971,10 @@ impl Connection {
                                     // destination doesn't matter if no name has been registered
                                     // (probably means name it's registered through external means).
                                     if !names.is_empty() && !names.contains_key(dest) {
-                                        trace!("Got a method call for a different destination: {}", dest);
+                                        trace!(
+                                            "Got a method call for a different destination: {}",
+                                            dest
+                                        );
 
                                         continue;
                                     }
@@ -1005,7 +1008,8 @@ impl Connection {
                                 )
                                 .detach();
                         } else {
-                            // If connection is completely gone, no reason to keep running the task anymore.
+                            // If connection is completely gone, no reason to keep running the task
+                            // anymore.
                             trace!("Connection is gone, stopping associated object server task");
                             break;
                         }
--- a/zbus/src/object_server/interface.rs
+++ b/zbus/src/object_server/interface.rs
@@ -142,9 +142,26 @@ pub trait Interface: Any + Send + Sync {
     fn introspect_to_writer(&self, writer: &mut dyn Write, level: usize);
 }
 
-/// A newtype for a reference counted Interface trait-object, with a manual Debug impl.
+/// A type for a reference counted Interface trait-object, with associated run-time details and a
+/// manual Debug impl.
 #[derive(Clone)]
-pub(crate) struct ArcInterface(pub(crate) Arc<RwLock<dyn Interface>>);
+pub(crate) struct ArcInterface {
+    pub instance: Arc<RwLock<dyn Interface>>,
+    pub spawn_tasks_for_methods: bool,
+}
+
+impl ArcInterface {
+    pub fn new<I>(iface: I) -> Self
+    where
+        I: Interface,
+    {
+        let spawn_tasks_for_methods = iface.spawn_tasks_for_methods();
+        Self {
+            instance: Arc::new(RwLock::new(iface)),
+            spawn_tasks_for_methods,
+        }
+    }
+}
 
 impl fmt::Debug for ArcInterface {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
--- a/zbus/src/object_server/interface.rs
+++ b/zbus/src/object_server/interface.rs
@@ -72,6 +72,14 @@ pub trait Interface: Any + Send + Sync {
     where
         Self: Sized;
 
+    /// Whether each method call will be handled from a different spawned task.
+    ///
+    /// Note: When methods are called from separate tasks, they may not be run in the order in which
+    /// they were called.
+    fn spawn_tasks_for_methods(&self) -> bool {
+        true
+    }
+
     /// Get a property value. Returns `None` if the property doesn't exist.
     async fn get(&self, property_name: &str) -> Option<fdo::Result<OwnedValue>>;
 
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -294,20 +294,6 @@ impl Node {
         self.add_arc_interface(I::name(), ArcInterface::new(iface))
     }
 
-    // Takes a closure so caller can avoid having to create an Arc & RwLock in case interface was
-    // already added.
-    fn at<F>(&mut self, name: InterfaceName<'static>, iface_creator: F) -> bool
-    where
-        F: FnOnce() -> ArcInterface,
-    {
-        match self.interfaces.entry(name) {
-            Entry::Vacant(e) => e.insert(iface_creator()),
-            Entry::Occupied(_) => return false,
-        };
-
-        true
-    }
-
     async fn introspect_to_writer<W: Write + Send>(&self, writer: &mut W) {
         enum Fragment<'a> {
             /// Represent an unclosed node tree, could be further splitted into sub-`Fragment`s
@@ -516,28 +502,25 @@ impl ObjectServer {
         P: TryInto<ObjectPath<'p>>,
         P::Error: Into<Error>,
     {
-        self.at_ready(path, I::name(), move || ArcInterface::new(iface))
+        self.add_arc_interface(path, I::name(), ArcInterface::new(iface))
             .await
     }
 
-    /// Same as `at` but expects an interface already in `Arc<RwLock<dyn Interface>>` form.
-    // FIXME: Better name?
-    pub(crate) async fn at_ready<'node, 'p, P, F>(
-        &'node self,
+    pub(crate) async fn add_arc_interface<'p, P>(
+        &self,
         path: P,
         name: InterfaceName<'static>,
-        iface_creator: F,
+        arc_iface: ArcInterface,
     ) -> Result<bool>
     where
         P: TryInto<ObjectPath<'p>>,
         P::Error: Into<Error>,
-        F: FnOnce() -> ArcInterface,
     {
         let path = path.try_into().map_err(Into::into)?;
         let mut root = self.root().write().await;
         let (node, manager_path) = root.get_child_mut(&path, true);
         let node = node.unwrap();
-        let added = node.at(name.clone(), iface_creator);
+        let added = node.add_arc_interface(name.clone(), arc_iface);
         if added {
             if name == ObjectManager::name() {
                 // Just added an object manager. Need to signal all managed objects under it.
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -731,18 +731,6 @@ impl ObjectServer {
         )))
     }
 
-    async fn dispatch_method_call(&self, connection: &Connection, msg: &Message) -> Result<()> {
-        match self.dispatch_method_call_try(connection, msg).await {
-            Err(e) => {
-                let hdr = msg.header();
-                debug!("Returning error: {}", e);
-                connection.reply_dbus_error(&hdr, e).await?;
-                Ok(())
-            }
-            Ok(r) => r,
-        }
-    }
-
     /// Dispatch an incoming message to a registered interface.
     ///
     /// The object server will handle the message by:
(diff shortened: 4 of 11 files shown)


zeenix:
> Clients can expect calls to be handled in order they are sent. Although the specification doesn't explicitly state so (https://gitlab.freedesktop.org/dbus/dbus/-/issues/179), 

As explained multiple times, the spec does not say it at all in any way or form, not even implicitly or remotely hinting towards it. Please stop insisting that it's the case, unless/until you get a confirmation from any of the spec maintainers/authors on your assertion here.

> bus implementations have respected message ordering.

Order of messages for the bus is important because bus needs to replicate the behavior of the service. Very strange things can happen under very normal circumstances if bus reorders messages (no batching or strange assumptions from clients needed). Hence I find this very irrelevant.

> this can also be expected when the bus and client are fully managed, as in p2p.

This is the only valid use case (even though it's even more niche since its only clients that batch method calls with inter-dependencies) here. I has to ask you to modify the description of the PR and the relevant commit message to focus solely on this use case.

I'll review once you've done that.

zbus_macros/src/lib.rs
```diff
@@ -215,6 +215,15 @@ pub fn dbus_proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
 /// `for T` on your behalf, to handle the message dispatching and introspection support.
 ///
+/// The trait accepts the `interface` attributes:
+///
+/// * `name` - the D-Bus interface name
+///
+/// * `no_spawn` - Do not spawn async tasks to call the methods. By default, zbus will spawn a task
+///   for each method call. This may result in methods being handled out of order and may not be the
+///   expected or desired behaviour. Note that D-Bus has currently not stated explicitely message
+///   ordering guarantees (see also <https://gitlab.freedesktop.org/dbus/dbus/-/issues/179>).
```
zeenix, line 225: * Remove "currently" and update to reflect that we're talking about method handling order. Message ordering ordering and this linked issue is very much irrelevant.
* I don't think this should be decided on per interfaces bases. I just want a simple boolean setting on either ConnectionBuilder or ObjectServer.

danieldg:
Another possible use case that you could document is inter-operating with an existing client that assumes this type of ordering (including when it does it unintentionally, or by using `NoReplyExpected`).  Accidental dependencies on things like this (which turn in to race conditions) are unfortunately common and hard to debug, so identifying this as a "make buggy clients happy" knob could help - both in making things work and in pointing out that it is in fact a bug in the client, not zbus.

zbus_macros/src/lib.rs
zeenix: @elmarco actually I've a better suggestion that you'll probably like: we can make non-spawning the default for p2p. Since separate clients will be on separate connections anyway and hence have their ObjectServer running in a separate task. 

Also we don't need to provide an option to change this behaviour unless someone asks for it. I don't see much point in parallelising method calls from the same client. We should document this behaviour though.

I'd be fine with doing the same for the bus case but that would be adding a lot of complexity to track all the peers and managing tasks for them.

elmarco: > @elmarco actually I've a better suggestion that you'll probably like: we can make non-spawning the default for p2p. Since separate clients will be on separate connections anyway and hence have their ObjectServer running in a separate task.

We could discuss the various use cases having an option to set the spawning behaviour at either Method, Interface, ObjectServer, Connection. They are *all* valid to me.  

But I think the sweet spot is at the Interface level, because it is a set of related methods and implementation, and you can document this expected behaviour better than at the ObjectServer or Connection level. At the ObjectServer or Connection level, you enforce a behaviour on other implementations that may not require it.

There is no fundamental difference from a p2p connection and bus connection, and both could serve a single interface or multiple interface/multiplexed communication. Imho, we shouldn't behave differently because of that.

> Also we don't need to provide an option to change this behaviour unless someone asks for it. I don't see much point in parallelising method calls from the same client. We should document this behaviour though.

There is as much reasons to want or no want parallelism with p2p or bus connection.

zeenix:
BTW, if you can break the cleanup/fixes commits into a separate PR, we can already merge those.

zbus_macros/src/lib.rs
elmarco: > > There is no fundamental difference from a p2p connection and bus connection, and both could serve a single interface or multiple interface/multiplexed communication.
> 
> It's not about the number of interfaces served. It's about the relationship between client and connection 

The same p2p connection can be shared for different usages & interfaces, just like a bus connection.

> 
> > Imho, we shouldn't behave differently because of that.
> 
> The only use-case you presented was about p2p and if we can satisfy that use case w/o needing new API (at least temporarily), I'd prefer that.

It's not specific to p2p. org.qemu.Display1.Keyboard has Press and Release methods for example. There is no reason a client would have to wait for a reply for Press before sending Release. This interface is exposed on the default bus by default. The implementation is currently C/GDBus, but you should get my point.

zeenix: > > > There is no fundamental difference from a p2p connection and bus connection, and both could serve a single interface or multiple interface/multiplexed communication.
> > 
> > 
> > It's not about the number of interfaces served. It's about the relationship between client and connection
> 
> The same p2p connection can be shared for different usages & interfaces, just like a bus connection

As I wrote, it's not about interfaces (only if we take the approach in this PR, which I'm objecting to for other reasons I mentioned). 

> It's not specific to p2p. org.qemu.Display1.Keyboard has Press and Release methods for example. 

Again, dependencies between methods is **not** an argument alone. You can find countless examples of methods depending on each other out in the wild.

> There is no reason a client would have to wait for a reply for Press before sending Release.

If it's from the same client, yes they shouldn't need to. Ideally both in case of p2p and bus, but as I wrote, it'd be hard to implement it for the bus but in case of p2p, there is only one client per connection so it's easy to make it happen.

elmarco: > > The same p2p connection can be shared for different usages & interfaces, just like a bus connection
> 
> As I wrote, it's not about interfaces (only if we take the approach in this PR, which I'm objecting to for other reasons I mentioned).

I disagree, typically a set of related methods on an interface must be handled in order. If you are lucky enough to have an interface that doesn't need ordering, you can keep zbus default behaviour.

p2p or bus connection, it doesn't matter, they can serve several unrelated objects & interfaces with different constraints. Your point about making several p2p connections doesn't hold, they are usually sharing the same connection to uniquely identify a client and avoid pointless extra resources usage and connection setups.

> 
> > It's not specific to p2p. org.qemu.Display1.Keyboard has Press and Release methods for example.
> 
> Again, dependencies between methods is **not** an argument alone. You can find countless examples of methods depending on each other out in the wild.

Good you agree they are countless existing examples where ordering or methods matters!

zeenix: > Good you agree they are countless existing examples where ordering of methods matters! (the norm, not the exception to me)

I never disagreed with that at all. If you think I did, you misunderstood. A method call in general is incomplete w/o a return. If you had 2 async methods with interdependency, you won't do something like this in Rust:

```rust
// `method_b` depends on the side effects of `method_a` here.
let batch = [method_a(), method_b()];
...
for method in batch {
    tokio::task::spawn(method);
}
```

---

Since it's clear to me that we won't be able to convince each other after having spent a lot of our time on this already, I am no longer interested in discussing it further. I hate to do this but I'm pulling the maintainer card and ask you to either take one of the 2 options I suggested (either make non-spawning default only for the p2p and document this behavior, or add the ConnectionBuilder/ObjectServer method to disable task spawning) or close the PR.

elmarco: With task spawning, it's quite explicit that there are no dependencies. The trouble is the client is simply doing `send([methodA,methodB])` and can reasonably expect that to be handled in order, unless the implementation clearly state otherwise.

elmarco: Setting a no-spawn option _globally_ on Connection/ObjectServer brings back the potential dead-lock issue https://github.com/dbus2/zbus/commit/2da0ecd98f449aa2fcae2ff7f656fe91a13d8b9a. By setting `no_spawn` on the Interface, the _specific_ implementation states that it doesn't rely on reentering (no blocking calls etc).

zeenix: > Setting a no-spawn option _globally_ on Connection/ObjectServer brings back the potential dead-lock issue [2da0ecd](https://github.com/dbus2/zbus/commit/2da0ecd98f449aa2fcae2ff7f656fe91a13d8b9a)

I don't think so. The commit was made before ObjectServer started using its own MessageStream (i-e queue), which is separate from the stream used by method replies so the deadlock mentioned in the commit log is not possible anymore. Almost everything uses its own stream now in zbus even and this was done to avoid all such deadlocks. 

But maybe I'm missing something. Can describe where and how this can deadlock?

danieldg: The stream is still bounded (by `DEFAULT_MAX_QUEUED`). Since it's bounded, you could get this deadlock if you wait on any dbus recv operation inside your handler (the usual cause of such a wait would be "making a dbus call").  And even if it were unbounded (or the bound is not hit), you can still get a deadlock if you have a circular dependency of dbus calls that cycles back to this ObjectServer.  The circular dependency is a bit easier to just document as "don't risk doing that if you set no-spawn".

elmarco: @danieldg +1, correct

zeenix: > Since it's bounded, you could get this deadlock if you wait on any dbus recv operation inside your handler (the usual cause of such a wait would be "making a dbus call").

True. However, although after all the debates on this topic, I still remain unconvinced that anyone other than @elmarco needs this anyway but just in case I'm wrong and there are other people/projects out there "batching method calls with inter-dependencies", I am open to a connection option and I think it's more than sufficient to warn about this possible deadlock in the docs.

> you can still get a deadlock if you have a circular dependency of dbus calls that cycles back to this ObjectServer.

D-Bus calls are always more expensive and have a higher latency so using it for intra-process communication sounds like a bad idea in general. There are much better ways of doing that, e.g channels. So as you said, we shouldn't need to care about this case at all. Just document it, if really needed.

zeenix: > > Since it's bounded, you could get this deadlock if you wait on any dbus recv operation inside your handler (the usual cause of such a wait would be "making a dbus call").

Actually looking back at the code, I'm not sure that's true. The ObjectServer's receiver is [for `METHOD_CALL` messages only](https://github.com/dbus2/zbus/blob/f45718536c7fba6b88c3db071bb90ccc2efa4522/zbus/src/connection/mod.rs#L957) so [any `METHDO_RETURN` messages won't be broadcasted to it](https://github.com/dbus2/zbus/blob/6a78d8ef191391c69c4fffe530741db67a314bac/zbus/src/connection/socket_reader.rs#L63) and hence socket reader task won't deadlock on broadcasting any method call returns while the method dispatch is waiting on a method call.

Did I miss something?

danieldg: Yes: if parties on the bus have sent us `DEFAULT_MAX_QUEUED + 1` method calls, then `receive_msg` will block on the `broadcast_direct` call.  That prevents any new messages, regardless of type, from being processed.

zeenix: > Yes: if parties on the bus have sent us `DEFAULT_MAX_QUEUED + 1` method calls, then `receive_msg` will block on the `broadcast_direct` call.

But why? The queues are separate. Perhaps it would help me understand if you could describe an example scenario.

danieldg: 1. Client A makes the AccidentalDeadlockFunction method call; this ends in zbus's queue.
2. Client B sends `DEFAULT_MAX_QUEUED + 1` method calls to `GetFoo` (say, a badly designed API where you can get only one property at a time; this doesn't matter exactly, just the number of messages).
3. https://github.com/dbus2/zbus/blob/f45718536c7fba6b88c3db071bb90ccc2efa4522/zbus/src/connection/socket_reader.rs#L75 starts sending to the queue.  After `DEFAULT_MAX_QUEUED`, it blocks (2 messages left to send).
4. The ObjectServer dequeues one method call (ADF) and calls it. It makes a call to `DBusProxy::GetConnectionUnixProcessId`; this gets sent out.
5. The dbus server sends the reply on the socket.
6. `receive_msg` runs again, and is able to send one more message via `broadcast_direct`.  There's one `GetFoo` left (parsed but not sent), and just behind it is the reply from `GetConnectionUnixProcessId`.
7. Deadlock.

zeenix: Thanks for explaining! Yeah, now I see it.

However, does having the option on the interface level eliminate this deadlock possibility? I think it only reduces the likelihood, assuming that developer disabling spawning, designs their interfaces and code carefully. If so, I don't see this as a good rationale for putting it on the interface level.

danieldg: To avoid the deadlock, you could have the documentation declare: if you ever wait on this `Connection` object inside your `ObjectServer` (by making a method call without NoReply, etc), you must declare your interface to be spawning (or if the default is spawning, must not declare it no-spawn).  The `GetFoo` calls in this example could be declared no-spawn without any problems.

Putting it at the interface level is "closer" to the problem (you could even put it at the method level) and so it's more likely to be noticed - though you could argue that either way.  If you have *multiple* interfaces, making the entire connection no-spawn puts restrictions on all of them - which is more of a burden to check.

zeenix: > Putting it at the interface level is "closer" to the problem (you could even put it at the method level)

Well, I think if we do this in a more granular way, method-level seems like the best option. If you hit this issue on an interface-level, there may not be something you could do about it (e.g if you're implementing a standardized interface, that you can not change).

@elmarco can you modify the PR to move the option to methods instead? I know this will require breaking changes on `Interface` but [as I wrote](https://github.com/dbus2/zbus/issues/674#issuecomment-2020310501) in #674, I think we should treat this an unstable API. We just need to document it more explicitly.

If so, please address other comments too and rebase the PR.

zeenix: > I think we should treat this an unstable API. We just need to document it more explicitly.

Having said that, we should still try to make the change a bit more future-proof:

```rust

pub struct CallSettings {
    spawn_task: bool,
}

impl CallSettings {
    pub fn new() {
        MethodCallSettings::default()
    }

    pub fn set_spawn_task(mut self, spawn_task: bool) -> Self {
         self.spawn_task = spawn_task;

         self
    }
}
```

and then add an `Option<CallSettings>` to `Interface::{call, call_mut}`.

elmarco: @zeenix if we are breaking Interface, I would rather implement part of my proposal from #674. Property and Method can have additional details, and there is no call & call_mut API duplication.

To me, the method granularity isn't the most appropriate. Usually, you have a related methods in an interface that you want to handle in order. Annotating each interface method is cumbersome, but that's fine to me if you believe it's better.

zeenix:
@elmarco rebase please.

zbus/src/object_server/interface.rs
```diff
@@ -70,6 +70,11 @@ pub trait Interface: Any + Send + Sync {
     where
         Self: Sized;
 
+    /// Whether methods will be called with a spawned task (and potentially out of order)
```
zeenix, line 73: * "with" -> "from".
* I think "out of order" isn't correct. The task are spawned in order, just that they may not be scheduled or the dispatched methods may not respond in order. Perhaps better add a separate note instead:

```
Note: When methods are called from separate tasks, they may not return in the order in which they were called. Typically, this is not a problem and therefore, the default implementation returns `true`.
```

That way, we also document the default value.

zbus/src/object_server/mod.rs
```diff
@@ -782,8 +782,35 @@ impl ObjectServer {
             })?
         };
 
-        self.dispatch_call_from_iface(iface, connection, msg, hdr)
-            .await
+        trace!("acquiring read lock on interface `{}`", iface_name);
+        let read_lock = iface.read().await;
```
zeenix, line 786: * `read_lock` here is not a lock but a guard.
* I prefer naming the guards by what they contain (rather than naming it after their smart pointer container), so:

```rust
let with_spawn = {
     ...
     let iface = iface.read().await;
     ...
     iface.with_spawn()
};
```

This also gets rid of explicit `drop`. Call me a weirdo but I prefer using scopes over manual drops. :)

zbus/src/connection/mod.rs
```diff
@@ -1005,35 +1005,13 @@ impl Connection {
...
-                                            debug!(
-                                                "Error dispatching message. Message: {:?}, error: {:?}",
-                                                msg, e
-                                            );
-                                        }
-                                    }
-                                    .instrument(trace_span!("{}", task_name)),
-                                    &task_name,
-                                )
-                                .detach();
```
zeenix, line 1031: This is very confusing. So until the commit "♻️ zb: drop the mandatory task spawning", we were essentially spawning tasks twice for each method? So if we only reverted this commit, we'll be in a weird state.

Why not first do the needed refactor so that you don't need this additional commit?

zeenix (changes requested):
LGTM in general.

zbus/src/connection/mod.rs
elmarco: Well, if we drop spawning before introducing it, we are in a weird state (worse since it regresses). What's wrong with spawning twice? Alternatively, just squash all changes, but that's harder to review/bisect.

zbus_macros/src/lib.rs
```diff
@@ -215,6 +215,18 @@ pub fn dbus_proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// properties or signal depending on the item attributes. It will implement the [`Interface`] trait
 /// `for T` on your behalf, to handle the message dispatching and introspection support.
 ///
+/// The trait accepts the `interface` attributes:
+///
+/// * `name` - the D-Bus interface name
+///
+/// * `spawn` - By default, zbus will spawn a task for each method call. This may result in methods
+///   being handled out of order and may not be the expected or desired behaviour.
```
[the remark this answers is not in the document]
elmarco: The default could be problematic. What do you suggest?

zbus_macros/src/lib.rs
```diff
@@ -215,6 +215,18 @@ pub fn dbus_proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///
+///   Be careful if you set this attribute to `false`: the object server will block until the
+///   functions returns. In this context, no other methods are handled and the message queues can
+///   become full, blocking the whole connection. You must make sure the implementation doesn't
+///   create such a dead-lock situation, by not blocking/awaiting if possible.
```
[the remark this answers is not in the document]
elmarco: Well, I don't like vague detailslike "in certain situations" either. but fine.

zbus/src/object_server/interface.rs
elmarco: It's not simply about "return in the order", it's also execution. I disagree about "Typically, this is not a problem" too.

zbus/src/object_server/mod.rs
elmarco: This is based on existing code, I just followed the same style. The dispatch usage requires different variables with explicit drop because of the read/write locking.

It can't be done the way you suggest.

zbus/src/connection/mod.rs
```diff
@@ -1016,12 +1016,14 @@ impl Connection {
                             trace!("Got `{}`. Will spawn a task for dispatch..", msg);
                             let executor = conn.inner.executor.clone();
                             let task_name = format!("`{member}` method dispatcher");
+                            // SAFETY: msg becomes 'static when moved to the task, and dropped after
+                            let hdr = unsafe { std::mem::transmute::<_, Header<'static>>(hdr) };
```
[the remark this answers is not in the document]
elmarco: If you remove the lifetime cast, you get this compile error: ```argument requires that `msg` is borrowed for `'static```

The borrow-checker isn't able to correlate the lifetimes of 'a msg & hdr<'a>. They must become 'static to be given to spawn. 

Anyway, this is a temporary hack, as explained in the commit message.

zbus_macros/src/lib.rs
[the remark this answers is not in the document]
zeenix: It's less about what I or you like, it's more about what's useful to the user. They don't need to know what internals are involved, just what they have to do, to avoid deadlocks.

zbus_macros/src/lib.rs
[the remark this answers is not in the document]
zeenix: > The default could be problematic.

Only in some very rare specific use cases that assume unspecified requirements, so no need to scare most users who will never have any issues with it.

> What do you suggest?

"In certain situation, where clients assume unspecified requirements on services, this behavior could be problematic. For example, <your use case described here>. You can set this to `false` to enable compatibility with such clients."

zbus/src/connection/mod.rs
zeenix: whatever, I don't have the energy for debate this one. :)

zbus/src/object_server/mod.rs
zeenix: > The dispatch usage requires different variables with explicit drop because of the read/write locking.

Not sure, I follow. why can't my suggestion work? scoping is a very usual way to drop locks as soon as they're no longer needed.

zeenix: > This is based on existing code, I just followed the same style.

Pretty sure, the existing code doesn't call this boolean returning method, since you added it in this PR. :)

elmarco: ok, I added a scope to drop the explicit drop

zbus_macros/src/lib.rs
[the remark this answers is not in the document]
elmarco: see new version

zbus/src/connection/mod.rs
```diff
@@ -996,41 +996,25 @@ impl Connection {
...
                                 }
                             }
-                            let member = match hdr.member() {
-                                Some(member) => member,
-                                None => {
-                                    warn!("Got a method call with no `MEMBER` field: {}", msg);
-
-                                    continue;
-                                }
-                            };
```
zeenix, line 1012: Seems this is now checked only **after** the spawn? This could mean that a misbehaving client can get us to spawn lots of tasks needlessly. Let's avoid that and check the message integrity etc before spawning the task?

zeenix (changes requested):
LGTM otherwise. If you could, please do check if these changes somehow doesn't end up breaking busd (at some point we should just add busd to zbus' CI to ensure this).

zbus/src/connection/mod.rs
zeenix: * Member name is mandatory for method call messages so "unknown member" means an invalid message here.
* Regardless of this distinction, the point is that we should ensure we've all essentials in the method call before spawning a task for its dispatch.

elmarco: Because the way Interface::call{_mut} is designed, the return value needs to hold the lock. (in contrast with my proposal from https://github.com/dbus2/zbus/issues/674).

So if we want to avoid multiple redundant lookups, we need to do the Method resolution in the spawned task, we can't easily gather all the details before.

elmarco: (Msg and Header can become 'static,  but not 'a in Header<'a>)

zeenix: > But it fails the borrow-checker that it isn't able to match the lifetime of Msg and Header<'_> to 'static when moving those to a future.

Right, I guess because of `spawn` requirements. Still, as I said, I'd rather pay the small[1] penalty of header getter.

---

1. [As documented](https://docs.rs/zbus/latest/zbus/message/struct.Message.html#method.header), this method is no longer costly, just allocates a vec for the fields.

zbus/src/object_server/mod.rs
```diff
@@ -767,6 +767,13 @@ impl ObjectServer {
             // error, or deliver the message as though it had an arbitrary one of those
             // interfaces.
             .ok_or_else(|| fdo::Error::Failed("Missing interface".into()))?;
+        // Check that the message has a member before spawning.
+        // Note that an unknown member will still spawn a task. We should instead gather
+        // all the details for the call before spawning.
+        // See also https://github.com/dbus2/zbus/issues/674 for future of Interface.
```
zeenix, line 773: * `FIXME: ` prefix would be nice.
* Feel free to make some changes already here, if needed ([I already made it clear that `Interface` is unstable API](https://github.com/dbus2/zbus/pull/711)). However I don't really think we need to change `Interface` here to be able to check everything upfront. The only/main part we really want to spawn off, is the user-provided handlers.
* This commit is only adding a check using the `hdr` we already have at hand so not sure `💩` is justified. :)

elmarco: 💩 : because we are doing multiple time the same check and the check is very limited: you can still spawn task if you provide a Member, even if it doesn't exist on the interface. Iow, I disagree this check is worthwhile.

zbus/src/object_server/mod.rs
```diff
@@ -766,16 +750,74 @@ impl ObjectServer {
...
+                .ok_or_else(|| fdo::Error::UnknownObject(format!("Unknown object '{path}'")))?;
+
+            node.interface_lock(iface_name.as_ref()).ok_or_else(|| {
+                fdo::Error::UnknownInterface(format!("Unknown interface '{iface_name}'"))
+            })?
+        };
+
+        let with_spawn = {
+            trace!("acquiring read lock on interface `{}`", iface_name);
+            let iface = iface.read().await;
```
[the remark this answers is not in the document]
elmarco: I can move the with_spawn to interface_lock(), that should get rid of the lock.

zbus/src/connection/mod.rs
zeenix: > I am not sure what you want to do.

I don't really see a problem with what you've already in your latest commit. So I'm as confused as you about what we're discussing here anymore. :)

zbus/src/object_server/mod.rs
zeenix: > It would be a major refactoring to move to the proposal I made.

Right, I didn't realize you mean you need **all** of your proposed changes here.

> The gain is great though, because there is no need to lock the interface to get those details: callback, spawn, introspection details etc.

No need to (even read) lock the interface to call method on it? How so? :thinking: Keep in mind that the trait can't be object safe if has associated constants or methods w/o receivers IIRC.

> because we are doing multiple time the same check and the check is very limited

We shouldn't need to check multiple times. Once it's established the the member is present, later code should just assume that it is and `unwrap()`.

> Iow, I disagree this check is worthwhile.

I explained already why I want it but more importantly, the check is already done in the existing code so spawning for messages w/o member present, would be a regression.

zbus/src/object_server/mod.rs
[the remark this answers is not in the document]
zeenix: How? :thinking: You still need to lock the interface's lock to make a call on it.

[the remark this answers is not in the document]
elmarco: To make a call, but not to gather rtti/metadata. that's the main point about #674, so it would be one step in this direction.

zbus/src/connection/mod.rs
elmarco: Can you do it on top? I feel stupid, I don't get what you want.

zbus/src/object_server/mod.rs
elmarco: squashed the 💩

zbus/src/object_server/mod.rs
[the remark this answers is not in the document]
elmarco: We don't need any locking to get the interface static RTTI. See "zb: gather some iface details earlier" patch for the idea. #674 should push this further, and allow to get rid of the read+write locks too.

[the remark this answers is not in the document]
zeenix: I'm not sure what you mean by RTTI and where I find this patch you're talking about but your proposed `Interface` trait in #674 uses associated constants and [hence can't be object-safe](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d838741ba246ea0d03f0031937ce1cf4) and we need object-safety cause we keep various interface impls in the same collection.

zbus/src/object_server/interface.rs
```diff
@@ -134,9 +142,26 @@ pub trait Interface: Any + Send + Sync {
...
+    pub inst: Arc<RwLock<dyn Interface>>,
+    pub spawn_tasks_for_methods: bool,
+}
+
+impl ArcInterface {
+    pub fn new<I>(iface: I) -> Self
+    where
+        I: Interface,
+    {
+        let spawn_tasks_for_methods = iface.spawn_tasks_for_methods();
```
zeenix, line 158: Ah you got the info upfront and cached it, smart!

zeenix:
@elmarco Ok to merge?

zeenix:
Perhaps it would make sense to automatically disable spawning for method calls with `NO_REPLY_EXPECTED` flag set on them?

zbus/src/connection/builder.rs
```diff
@@ -412,13 +415,13 @@ impl<'a> Builder<'a> {
...
-                            .at_ready(path.to_owned(), name.clone(), || iface);
-                    let added = future.await?;
-                    // Duplicates shouldn't happen.
-                    assert!(added);
+                    let added = object_server
+                        .inner()
+                        .at_ready(path.to_owned(), name.clone(), || iface)
+                        .await?;
+                    if !added {
+                        return Err(Error::Unsupported);
```
zeenix, line 422: best add an error variable for this. `Unsupported` seems misplaced since this is a valid failure and not some lack of support of a feature that should be supported.

zeenix (changes requested):
LGTM otherwise.

zbus/src/connection/builder.rs
elmarco: yeah, I wasn't so sure it is not worth to support. But maybe it can be changed anyway another day.

zeenix: We both know that that day will likely never come. :) Let's just add a variant:

```rust
enum Error {
    // ...
}
```

Not the best name or description, I know but I think it's still better than just throwing a generic error.

zeenix: Our error struct is declared `non_exhaustive` so we should make use of that and added variants wherever needed.

zeenix (approved):
Good to go.

zeenix:
So latest change is only the nitpick fix and a rebase on current main? If so, I'll merge.

elmarco:
> So latest change is only the nitpick fix and a rebase on current main? If so, I'll merge.

yes, please merge

PR #683, opened 2024-03-21
⚡️ zb: Pipeline client-side handshake as much as possible

### LINKED ISSUE #493: Pipeline handshakes
The handshake code should send all commands at once, instead of one by one and awaiting a response from the server and hence significantly reducing the connection setup time. This relies on handshakes commands typically (in almost all cases) succeeding.

Given that `cargo bloat` claims handshake method to be the top offender, I'm hoping this will also help with #304.

zeenix:
We now send `DATA`, `NEGOTIATE_UNIX_FD` and `BEGIN` commands at once. This reduces latency and hence improves performance.

It would have been the best to send all the commands at the same time, but seems dbus-daemon kick us out if we send multiple commands after it rejects the auth command. It wants us to negotiate the auth first. We could in theory just reconnect but that would require some architectural changes and rethinking the code.

Moreover cookie auth requires a round-trip to the server to get the cookie context first so we can't fully pipeline that one.

The server side is not possible to pipeline as it the server just responds to commands from the client and it can not assume that clients will always pipeline all commands.

Fixes #493.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -57,9 +57,6 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Build and Test
         run: |
-          mkdir -p /run/user/$UID
-          sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
-          sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
           dbus-run-session --config-file /tmp/dbus-session-abstract.conf -- cargo --locked test --verbose -- basic_connection
           # All features except tokio.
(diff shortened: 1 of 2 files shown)


zeenix:
@kas (I assume you're [this guy](https://gitlab.freedesktop.org/exactly-one-kas)) Hi there, seems the gdbus doesn't really correctly do handshake and pipelining handshake [breaks there](https://github.com/dbus2/zbus/actions/runs/8389693869/job/22976463612?pr=683#step:13:66). While I was more than happy to add the support for that but if supporting the special gdbus session bus means we don't get to do smart things in zbus, I'd rather remove the support (or be OK with it being broken until gdbus is fixed).

zeenix:
Seems I **was** right about GDBus being unable to handle pipelining. It requires BEGIN command to be sent separately after client receive OK on the authentication. Fortunately, the workaround is not too dirty.

PR #697, opened 2024-03-30
Use run() instead of tick()

notgull:
run() is more efficiently and spawns a local task queue for efficiency. As opposed to tick(), which only polls the global queue.

### DIFF
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -527,6 +527,27 @@ impl<'a> Builder<'a> {
 /// tasks until socket reader task kicks in.
 #[cfg(not(feature = "tokio"))]
 fn start_internal_executor(executor: &Executor<'static>, internal_executor: bool) -> Result<()> {
+    use core::{
+        future::Future,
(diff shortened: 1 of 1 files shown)


zbus/src/connection/builder.rs
```diff
@@ -527,16 +527,35 @@ impl<'a> Builder<'a> {
...
         let executor = executor.clone();
         std::thread::Builder::new()
             .name("zbus::Connection executor".into())
             .spawn(move || {
                 crate::utils::block_on(async move {
                     // Run as long as there is a task to run.
-                    while !executor.is_empty() {
-                        executor.tick().await;
-                    }
+                    executor.run(Empty(&executor)).await
```
zeenix, line 558: executor running a dummy future, to run itself seems very weird. I understood from your comment here that you intend to make tick also more efficient, so maybe we just wait for that? It all depends on how much efficiency and under which scenarios would it be sigfican't enough to justify sacrificing code readability and many Future impls.

notgull:
Good point

PR #699, opened 2024-03-30
➖ zm: Remove regex dep

MinusGix:
This removes the unneeded dependency on `regex` in `zbus_macros`. I believe this code should be equivalent to the regular expression.  
- Loops over looking for any `&` as `is_match` just requires it a match anywhere in the string
  - once found, strip off all leading spaces
  - if they're equivalent, then we've found it
  
The only ambiguity I can think of being if the type contained symbols that regex would have considered as part of the name, but I think that'd only happen for tuples and the regex wouldn't match those due to considering it a capture group and not parentheses.

### DIFF
--- a/zbus_macros/src/proxy.rs
+++ b/zbus_macros/src/proxy.rs
@@ -1,7 +1,6 @@
 use crate::utils::{pat_ident, typed_arg, zbus_path, PropertyEmitsChangedSignal};
 use proc_macro2::{Literal, Span, TokenStream};
 use quote::{format_ident, quote, quote_spanned, ToTokens};
-use regex::Regex;
 use syn::{
     fold::Fold, parse_quote, parse_str, spanned::Spanned, AttributeArgs, Error, FnArg, Ident,
     ItemTrait, Path, ReturnType, TraitItemMethod,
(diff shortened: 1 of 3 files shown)


zbus_macros/src/proxy.rs
```diff
@@ -619,12 +618,23 @@ fn gen_proxy_method_call<M: AttrParse + Into<MethodAttrs>>(
         let is_input_type = inputs.iter().any(|arg| {
             // FIXME: We want to only require `Serialize` from input types and `DeserializeOwned`
             // from output types but since we don't have type introspection, we employ this
-            // workaround of regex matching on string reprepresention of the the types to figure out
+            // workaround of matching on string reprepresention of the the types to figure out
```
zeenix, line 621: Might be a good opportunity to also fix the typo here: `reprepresention`. :)

zeenix (changes requested):
Thanks. LGTM otherwise.

zbus_macros/src/proxy.rs
```diff
@@ -619,12 +618,23 @@ fn gen_proxy_method_call<M: AttrParse + Into<MethodAttrs>>(
...
+
+                let mut found = false;
+                for part in pat.split('&').skip(1) {
+                    let part = part.trim_start_matches(' ');
+
+                    if part == param {
+                        found = true;
+                        break;
+                    }
+                }
```
zeenix, line 635: Actually it's strange to split by `&` when it will be the first char. I think this could be simpler. Maybe something like:

```suggestion
                if let Some('&') = pat.chars().next() {
                    let ty_name = pat[1..].trim_start();
                    
                    ty_name == param
                } else {
                    false
                }
```

MinusGix: That's what I initially implemented, but I was mimicking what the code does and I wasn't sure of all the possible uses or how where/when syn inserts random whitespace (since there's only two or so tests for this). I can swap back to that if you want.

zeenix: Given that my suggestion is doing exactly what the regex-based code was doing and is simpler IMO, I'd prefer that, yes.

zeenix (approved):
Nice!

PR #706, opened 2024-04-06
➖ zb: Remove derivative dep 

MaxVerevkin:
Ideally this dependency should be dropped/replaced, because it pulls old syn version, and is very unlikely to be ever updated. It is not as important while `zbus_macros` also depends on syn v1, though.

### DIFF
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -8,7 +8,6 @@ use std::net::TcpStream;
 use std::os::unix::net::UnixStream;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
-    fmt,
     sync::Arc,
 };
 #[cfg(feature = "tokio")]
@@ -58,6 +57,7 @@ enum Target {
 type Interfaces<'a> = HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, ArcInterface>>;
 
 /// A builder for [`zbus::Connection`].
+#[derive(Debug)]
 #[must_use]
 pub struct Builder<'a> {
     target: Option<Target>,
@@ -78,26 +78,6 @@ pub struct Builder<'a> {
 }
 
 assert_impl_all!(Builder<'_>: Send, Sync, Unpin);
-
-impl fmt::Debug for Builder<'_> {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let mut f = f.debug_struct("Builder");
-        f.field("target", &self.target);
-        f.field("max_queued", &self.max_queued);
-        #[cfg(feature = "p2p")]
-        f.field("guid", &self.target);
-        #[cfg(feature = "p2p")]
-        f.field("p2p", &self.p2p);
-        f.field("internal_executor", &self.internal_executor);
-        f.field("names", &self.names);
-        f.field("auth_mechanisms", &self.auth_mechanisms);
-        #[cfg(feature = "bus-impl")]
-        f.field("unique_name", &self.unique_name);
-        f.field("cookie_context", &self.cookie_context);
-        f.field("cookie_id", &self.cookie_id);
-        f.finish_non_exhaustive()
-    }
-}
 
 impl<'a> Builder<'a> {
     /// Create a builder for the session/user message bus connection.
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -30,7 +30,7 @@ use crate::{
     address::{self, Address},
     async_lock::RwLock,
     names::{InterfaceName, WellKnownName},
-    object_server::Interface,
+    object_server::{ArcInterface, Interface},
     Connection, Error, Executor, OwnedGuid, Result,
 };
 
@@ -55,8 +55,7 @@ enum Target {
     Socket(Split<Box<dyn ReadHalf>, Box<dyn WriteHalf>>),
 }
 
-type Interfaces<'a> =
-    HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, Arc<RwLock<dyn Interface>>>>;
+type Interfaces<'a> = HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, ArcInterface>>;
 
 /// A builder for [`zbus::Connection`].
 #[must_use]
@@ -315,7 +314,7 @@ impl<'a> Builder<'a> {
     {
         let path = path.try_into().map_err(Into::into)?;
         let entry = self.interfaces.entry(path).or_default();
-        entry.insert(I::name(), Arc::new(RwLock::new(iface)));
+        entry.insert(I::name(), ArcInterface(Arc::new(RwLock::new(iface))));
 
         Ok(self)
     }
@@ -441,7 +440,7 @@ impl<'a> Builder<'a> {
                     let future =
                         object_server
                             .inner()
-                            .at_ready(path.to_owned(), name.clone(), || iface);
+                            .at_ready(path.to_owned(), name.clone(), || iface.0);
                     let added = future.await?;
                     // Duplicates shouldn't happen.
                     assert!(added);
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -8,6 +8,7 @@ use std::net::TcpStream;
 use std::os::unix::net::UnixStream;
 use std::{
     collections::{HashMap, HashSet, VecDeque},
+    fmt,
     sync::Arc,
 };
 #[cfg(feature = "tokio")]
@@ -58,8 +59,6 @@ type Interfaces<'a> =
     HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, Arc<RwLock<dyn Interface>>>>;
 
 /// A builder for [`zbus::Connection`].
-#[derive(derivative::Derivative)]
-#[derivative(Debug)]
 #[must_use]
 pub struct Builder<'a> {
     target: Option<Target>,
@@ -70,7 +69,6 @@ pub struct Builder<'a> {
     #[cfg(feature = "p2p")]
     p2p: bool,
     internal_executor: bool,
-    #[derivative(Debug = "ignore")]
     interfaces: Interfaces<'a>,
     names: HashSet<WellKnownName<'a>>,
     auth_mechanisms: Option<VecDeque<AuthMechanism>>,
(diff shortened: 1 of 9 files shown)


zeenix:
> Ideally this dependency should be dropped/replaced

Sure, why not do it in this PR? :thinking:

MaxVerevkin:
I can do it in this PR, but I need your input. First of all, drop _or_ replace? If replace, then with _what_?

zeenix:
> First of all, drop _or_ replace? If replace, then with _what_?

Looks like our use case if only for ignoring some fields. Grepping reveals that we've a total of 7 uses for `Derivate` derive and only 3 cases where we make use of the `ignore` attribute. So I think the easiest is to just drop it and manually implement in the few places we need to ignore some fields.

How does that sound?

MaxVerevkin:
Absolutely reasonable.

zbus/src/proxy/mod.rs
```diff
@@ -74,14 +75,22 @@ assert_impl_all!(Proxy<'_>: Send, Sync, Unpin);
 
 /// This is required to avoid having the Drop impl extend the lifetime 'a, which breaks zbus_xmlgen
 /// (and possibly other crates).
-#[derive(derivative::Derivative)]
-#[derivative(Debug)]
 pub(crate) struct ProxyInnerStatic {
-    #[derivative(Debug = "ignore")]
```
zeenix, line 80: This was done by @elmarco in 8a55ca77eb14fb9199f241b04f6bede9cf9fa567 and I see why but I wonder if it would be better to just make `Connection`'s debug repr better instead? :thinking: @elmarco ?

zeenix (changes requested):
Thanks so much. LGTM but maybe we could do a bit better (see comments).

zbus/src/connection/builder.rs
```diff
@@ -80,26 +80,6 @@ pub struct Builder<'a> {
...
-        f.field("names", &self.names);
-        f.field("auth_mechanisms", &self.auth_mechanisms);
-        #[cfg(feature = "bus-impl")]
-        f.field("unique_name", &self.unique_name);
-        f.field("cookie_context", &self.cookie_context);
-        f.field("cookie_id", &self.cookie_id);
-        f.finish_non_exhaustive()
-    }
-}
-
```
zeenix: If you move the `Interface: Debug` commit at the top (first one), you'll not need to remove this code you're adding in a previous commit in the same PR.

zeenix:
just FYI: https://github.com/dbus2/zbus/pull/711

zbus/src/connection/builder.rs
MaxVerevkin: I know, but:
- it is additional work
- the way it is, `Interface: Debug` and `remove custom Debug ...` commits are revertible. If you decide this was a bad idea, you can go back and have the manual impls.

zbus/src/connection/builder.rs
```diff
@@ -82,6 +80,27 @@ pub struct Builder<'a> {
 
 assert_impl_all!(Builder<'_>: Send, Sync, Unpin);
 
+impl fmt::Debug for Builder<'_> {
```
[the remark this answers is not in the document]
MaxVerevkin: Wait, no, this is a very breaking change. No one implements `Interface` manually, but the `Debug` impl is outside of the macros control. I mean, it even breaks the example in the book.

A better option would be to create a wrapper for `Arc<RwLock<dyn Interface>>` with a custom debug impl.

zbus/src/proxy/mod.rs
zeenix: let's assume that with silences, @elmarco signals his agreement. :)

zbus/src/connection/builder.rs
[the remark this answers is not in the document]
zeenix: > No one implements Interface manually, but the Debug impl is outside of the macros control. 

Hmm.. right. I was thinking of `dbus_interface` like `dbus_proxy`. The former doesn't create the type, the latter does.

> A better option would be to create a wrapper for `Arc<RwLock<dyn Interface>>` with a custom debug impl

Sure, sounds good.

zbus/src/connection/builder.rs
zeenix: > * it is additional work

Unfortunately, doing things right, typically is more work.

> * the way it is, `Interface: Debug` and `remove custom Debug ...` commits are revertible. If you decide this was a bad idea, you can go back and have the manual impls.

Ok, now that's a good argument. Let's keep it this way then. :)

zbus/src/proxy/mod.rs
elmarco: I still prefer that Proxy Debug doesn't provide Connection Debug.

zbus/src/connection/builder.rs
MaxVerevkin: > Unfortunately, doing things right, typically is more work.

Not in this case, it seems.

zeenix: > > Unfortunately, doing things right, typically is more work.
> 
> Not in this case, it seems.

"typically" being the keyword here. :) My point was just that "more work" is not a good argument on its own. But you've a better argument so it doesn't matter, here. :)

zbus/src/proxy/mod.rs
zeenix: > I still prefer that Proxy Debug doesn't provide Connection Debug.

Can you elaborate a bit? More specifically, would you still prefer this even if connection debug was super slim? 

Also do you happen to remember, where the debug output was being verbose? If it's the tracing logs, I think we need to tell the intrument macro to skip `self` in places where we aren't already.

zbus/src/object_server/interface.rs
```diff
@@ -130,6 +132,15 @@ pub trait Interface: Any + Send + Sync {
     fn introspect_to_writer(&self, writer: &mut dyn Write, level: usize);
 }
 
+/// A newtype for a reference counted Interface trait-object, with a custom Debug impl.
```
zeenix, line 135: custom -> manual. Otherwise, it gives the impression that we need some special impl.

zeenix (changes requested):
Apart from the nits, it would be great to have slightly more details in the commit logs. 
E.g

```
zb: Introduce internal wrapper type, ArcInterface

This type is a wrapper for `Arc<RwLock<dyn Interface>>`, which adds a `Debug` impl on this type and will allow us to remove 2 custom `Debug` impls for 2 containing types in the following commit.
```

zbus/src/object_server/interface.rs
```diff
@@ -130,6 +132,15 @@ pub trait Interface: Any + Send + Sync {
...
+#[derive(Clone)]
+pub(crate) struct ArcInterface(pub(crate) Arc<RwLock<dyn Interface>>);
+impl fmt::Debug for ArcInterface {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("ArcInterface").finish_non_exhaustive()
```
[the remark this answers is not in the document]
MaxVerevkin: I don't think that faking the struct name in debug impl is a good idea. This is not `Interface` nor `Arc<RwLock<dyn Interface>>`, but an `ArcInterface` with _something_. We can rename `ArcInterface` to something like `InetrfaceObject`, or maybe you have better ideas.

> (especially since it's a private type).

There are a lot of other private types with debug impls that reveal their names. I don't see a problem with that.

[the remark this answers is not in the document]
zeenix: > I don't think that faking the struct name in debug impl is a good idea

It's not faking. The type is just an internal wrapper for `Arc<RwLock<dyn Interface>>` so it's not wrong to show that as the debug impl.

> > (especially since it's a private type).
> 
> There are a lot of other private types with debug impls that reveal their names. I don't see a problem with that.

True that it's not so important that we'll want to add manual `Debug` impl just to hide their names but I'm sure the derive impl, shows the internals as well (so anyone reading the debug repr would get the idea of what the type is about), not just the type name, like you're doing here.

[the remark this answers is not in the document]
zeenix: I won't object to `ArcInterface(Arc<RwLock<dyn Interface>>)` either but if we're manually implementing Debug anyway, might as well hide the name of the internal type that doesn't really matter.

zbus/src/proxy/mod.rs
zeenix: @MaxVerevkin If you agree with me, please make the changes I suggest. Otherwise, let me know.

MaxVerevkin: I don't have an opinion on this one. Can we move it out of scope of this PR?

zeenix: Sure. My main concert is only that there is a lot of LOC in `proxy/mod.rs` already and I don't want to add more to that but perhaps we should split it anyways.

c-git:
Beg your pardon for the noise I just want to put the advisory ID here so for the next person looking they can see it's already been removed here.  Need to now go find out which of my dependencies is bring in this old version and what it would take for them to upgrade.

PS. Thank you very much for this crate and I really enjoyed your interview on rustacean station.

```
warning[unmaintained]: `derivative` is unmaintained; consider using an alternative
   ┌─ /home/one/log-viewer/Cargo.lock:89:1
   │
89 │ derivative 2.2.0 registry+https://github.com/rust-lang/crates.io-index
   │ ---------------------------------------------------------------------- unmaintained advisory detected
   │
   = ID: RUSTSEC-2024-0388
   = Advisory: https://rustsec.org/advisories/RUSTSEC-2024-0388
   = The [`derivative`](https://crates.io/crates/derivative) crate is no longer maintained.
     Consider using any alternative, for instance:
     - [derive_more](https://crates.io/crates/derive_more)
     - [derive-where](https://crates.io/crates/derive-where)
     - [educe](https://crates.io/crates/educe)
   = Announcement: https://github.com/mcarton/rust-derivative/issues/117
   = Solution: No safe upgrade is available!
   = derivative v2.2.0
     └── zbus v3.15.2
         ├── accesskit_unix v0.6.2
         │   └── accesskit_winit v0.16.1
         │       └── egui-winit v0.27.2
         │           └── eframe v0.27.2
         │               └── log_viewer v0.2.0
         ├── atspi-common v0.3.0
         │   ├── atspi v0.19.0
         │   │   └── accesskit_unix v0.6.2 (*)
         │   ├── atspi-connection v0.3.0
         │   │   └── atspi v0.19.0 (*)
         │   └── atspi-proxies v0.3.0
         │       ├── atspi v0.19.0 (*)
         │       └── atspi-connection v0.3.0 (*)
         ├── atspi-connection v0.3.0 (*)
         └── atspi-proxies v0.3.0 (*)
```

PR #712, opened 2024-04-09
♻️ zb: Make tracing an optional dependency

### LINKED ISSUE #331: Make tracing an optional feature
In GitLab by @bilelmoussaoui on Apr 26, 2023, 10:09

It ends up pulling `tracing`/ `tracing-core` / `tracing-attributes` indirectly. As in the libraries I wrote on top of zbus, I explicitly added a `tracing` feature, I would love if that feature would also control zbus's future `tracing` support.

REALERvolker1:
Hey all

I'm here to fix #331. It took me a bit to get to it, I have been pretty busy as of late.

Anyways, I'm pretty sure everything works. I ran the tests on all features, I documented the new macros themselves, changed some of the code structure to better support instrumented futures, and more.

I am using `stable-x86_64-unknown-linux-gnu`, so I don't know 100% if everything works on Windows (I didn't even know Windows had dbus!). I followed the contributor guidelines, so I'm pretty confident with this pr.

I hope my macros make this much easier to maintain in the future.

### DIFF
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["os::unix-apis"]
 readme = "README.md"
 
 [features]
-default = ["async-io"]
+default = ["async-io", "tracing"]
 uuid = ["zvariant/uuid"]
 url = ["zvariant/url"]
 time = ["zvariant/time"]
@@ -37,6 +37,7 @@ async-io = [
 tokio = ["dep:tokio"]
 vsock = ["dep:vsock", "dep:async-io"]
 tokio-vsock = ["dep:tokio-vsock", "tokio"]
+tracing = ["dep:tracing", "tokio?/tracing"]
 
 [dependencies]
 serde = { version = "1.0", features = ["derive"] }
@@ -68,6 +69,11 @@ event-listener = "5.3.0"
 static_assertions = "1.1.0"
 async-trait = "0.1.79"
 async-fs = { version = "2.1.1", optional = true }
+tracing = { version = "0.1.40", optional = true }
+vsock = { version = "0.4.0", optional = true }
+tokio-vsock = { version = "0.4", optional = true }
+xdg-home = "1.1.0"
+
 # FIXME: We should only enable process feature for Mac OS. See comment on async-process below for why we can't.
 tokio = { version = "1", optional = true, features = [
   "rt",
@@ -77,12 +83,8 @@ tokio = { version = "1", optional = true, features = [
   "io-util",
   "process",
   "sync",
-  "tracing",
 ] }
-tracing = "0.1.40"
-vsock = { version = "0.4.0", optional = true }
-tokio-vsock = { version = "0.4", optional = true }
-xdg-home = "1.1.0"
+
 
 [target.'cfg(windows)'.dependencies]
 windows-sys = { version = "0.52", features = [
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["os::unix-apis"]
 readme = "README.md"
 
 [features]
-default = ["async-io", "tracing"]
+default = ["async-io"]
 uuid = ["zvariant/uuid"]
 url = ["zvariant/url"]
 time = ["zvariant/time"]
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -15,7 +15,11 @@ use std::{
     sync::{Arc, OnceLock, Weak},
     task::{Context, Poll},
 };
-use tracing::{debug, info_span, instrument, trace, trace_span, warn, Instrument};
+
+use crate::abstractions::logging::{debug, info, trace, warning};
+#[cfg(feature = "tracing")]
+use tracing::Instrument;
+
 use zbus_names::{BusName, ErrorName, InterfaceName, MemberName, OwnedUniqueName, WellKnownName};
 use zvariant::ObjectPath;
 
@@ -639,49 +643,52 @@ impl Connection {
         let name_lost_fut = if flags.contains(RequestNameFlags::AllowReplacement) {
             let weak_conn = WeakConnection::from(self);
             let well_known_name = well_known_name.to_owned();
-            Some(
-                async move {
-                    loop {
-                        let signal = lost_stream.next().await;
-                        let inner = match weak_conn.upgrade() {
-                            Some(conn) => conn.inner.clone(),
-                            None => break,
-                        };
 
-                        match signal {
-                            Some(signal) => match signal.args() {
-                                Ok(args) if args.name == well_known_name => {
-                                    tracing::info!(
-                                        "Connection `{}` lost name `{}`",
-                                        // SAFETY: This is bus connection so unique name can't be
-                                        // None.
-                                        inner.unique_name.get().unwrap(),
-                                        well_known_name
-                                    );
-                                    inner.registered_names.lock().await.remove(&well_known_name);
+            let name_request_future = async move {
+                loop {
+                    let signal = lost_stream.next().await;
+                    let inner = match weak_conn.upgrade() {
+                        Some(conn) => conn.inner.clone(),
+                        None => break,
+                    };
 
-                                    break;
-                                }
-                                Ok(_) => (),
-                                Err(e) => warn!("Failed to parse `NameLost` signal: {}", e),
-                            },
-                            None => {
-                                trace!("`NameLost` signal stream closed");
-                                // This is a very strange state we end up in. Now the name is
-                                // question remains in the queue
-                                // forever. Maybe we can do better here but I
-                                // think it's a very unlikely scenario anyway.
-                                //
-                                // Can happen if the connection is lost/dropped but then the whole
-                                // `Connection` instance will go away soon anyway and hence this
-                                // strange state along with it.
+                    match signal {
+                        Some(signal) => match signal.args() {
+                            Ok(args) if args.name == well_known_name => {
+                                info!(
+                                    "Connection `{}` lost name `{}`",
+                                    // SAFETY: This is bus connection so unique name can't be
+                                    // None.
+                                    inner.unique_name.get().unwrap(),
+                                    well_known_name
+                                );
+                                inner.registered_names.lock().await.remove(&well_known_name);
+
                                 break;
                             }
+                            Ok(_) => (),
+                            Err(e) => warning!("Failed to parse `NameLost` signal: {}", e),
+                        },
+                        None => {
+                            trace!("`NameLost` signal stream closed");
+                            // This is a very strange state we end up in. Now the name is
+                            // question remains in the queue
+                            // forever. Maybe we can do better here but I
+                            // think it's a very unlikely scenario anyway.
+                            //
+                            // Can happen if the connection is lost/dropped but then the whole
+                            // `Connection` instance will go away soon anyway and hence this
+                            // strange state along with it.
+                            break;
                         }
                     }
                 }
-                .instrument(info_span!("{}", lost_task_name)),
-            )
+            };
+            #[cfg(feature = "tracing")]
+            let name_request_future =
+                name_request_future.instrument(tracing::trace_span!("{}", lost_task_name));
+
+            Some(name_request_future)
         } else {
             None
         };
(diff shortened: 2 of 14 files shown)


zbus/src/connection/socket_reader.rs
```diff
@@ -63,6 +68,9 @@ impl SocketReader {
                         match rule.matches(msg) {
                             Ok(true) => (),
                             Ok(false) => continue,
+                            #[cfg(not(feature = "tracing"))]
+                            Err(_e) => continue,
+                            #[cfg(feature = "tracing")]
```
zeenix, line 73: This turned out to be more intrusive than I was hoping for. Why do we need feature gating in the rest of the code, if you created the macros for both cases (feature enabled and disabled)? :thinking:

zbus/src/connection/socket_reader.rs
```diff
@@ -72,6 +80,7 @@ impl SocketReader {
                     }
                 }
 
+                #[allow(unused_variables)]
```
zeenix, line 83: We shouldn't need these either. The macros should take care of this.

zbus/Cargo.toml
```diff
@@ -77,9 +78,8 @@ tokio = { version = "1", optional = true, features = [
   "io-util",
   "process",
   "sync",
-  "tracing",
 ] }
-tracing = "0.1.40"
+# FIXME: We should only enable process feature for Mac OS. See comment on async-process below for why we can't.
```
zeenix, line 82: This comment is for `tokio` dep so pretty sure this is a mistake.

zeenix (changes requested):
Thanks so much for taking this on and I realize it wasn't an easy task. Overall, the approach looks good to me. Also, thanks for following the commit message rules (somehow everyone misses that at first :laughing:).

Apart from the inline comments, kindly squash the fixups commits into the relevant commits as per the guidelines (see the very last bullet point [here](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests)).

Ideally this should be 3 commits but I'm fine with putting it all in one commit.

zeenix:
> so I don't know 100% if everything works on Windows (I didn't even know Windows had dbus!)

Don't worry about it. The CI will take care of that. :)

REALERvolker1:
Firstly, I am sorry for the number of commits. I usually make numerous small commits on projects that I am working on on my machine, and I simply did not know that the repo style favors larger commits. I usually do multiple, smaller commits because then it is easier to detect when changes were made and by who. I will learn how to use these tools to merge the commits.

Second, the macros that I created for the cases where tracing is not enabled simply expand to `()`, the return type of `println!["{}", var]`. Following the contributor guide, I wanted to remove clippy lints related to dead code, and minimize variable creation when possible. I realize how it bloated the codebase, and I will do my best to minimize the impact of this.

Third, I am sorry about the tokio comment. I have both VSCodium and Neovim set to autoformat, so I am pretty sure that's what happened there. I will move that back where it belongs.

I am still pretty new to contributing to OSS, so thanks for considering my pull request! 🎉

REALERvolker1:
I couldn't figure out how to integrate `#[allow(unused_variables)]` into the logging macros, I got the error `attributes on expressions are experimental`, but besides that, I think I cleaned up pretty well. What do you think?

REALERvolker1:
> Perhaps you can instead tell the compiler to ignore the unused variables instead?
> `let _ = the_variable_to_ignore;`

The problem with this is that clippy does ignore those variables while the tracing feature is disabled, but when it is enabled, those variables are used, which breaks naming convention. Another solution would be to remove the `#[allow(unused_variables)]` markers and introduce clippy warnings into the codebase (which you forbid in CONTRIBUTING.md). The idea I lean towards is breaking variable naming convention with underscore-prefixed names, because that is probably the least burdensome in terms of maintenance.

As for atomic commits, I read the blog post and commit guidelines again, and I think the best course of action would be to manually copy over what I have written into a local copy of the repository. I would be able to succinctly describe what everything does. The problem with this is that we lose the logs of trial and error I dealt with, but I think that it is probably for the best (And in this situation, it seems way easier than dealing with git).

REALERvolker1:
Oh! Another annoyance I had with zbus is that it sends tracing messages at info level. I often use info! for my own purposes, and I think that by default zbus should use trace. (Unless there's an important warning or error).

REALERvolker1:
Alright, I recommitted everything. What do you think?

zeenix:
> > Perhaps you can instead tell the compiler to ignore the unused variables instead?
> > `let _ = the_variable_to_ignore;`
> 
> The problem with this is that clippy does ignore those variables while the tracing feature is disabled, but when it is enabled, those variables are used, which breaks naming convention.

Just to be clear, I didn't propose that you prefix the parameters/names with `_` unconditionally but rather tell the compiler to ignore the variable in the macro for `feature = "tracing"`. Something like we do [here](https://github.com/dbus2/zbus/blob/main/zbus/src/message/builder.rs#L233). 

> As for atomic commits, I read the blog post and commit guidelines again

:+1: 

> I think the best course of action would be to manually copy over what I have written into a local copy of the repository.

Well that's one way but you can also make use of  `rebase`, `reset`, and `commit --amend`. That way you'll also gain some very important git skills, which really should be considered essential but unfortunately are not. :( I don't mind either way, as long as the end result is the same for me as the reviewer. :)

> The problem with this is that we lose the logs of trial and error I dealt with, but I think that it is probably for the best

I know it feels bad that the rest of the world doesn't see your struggles and only the end result, but it's not helpful to have that recorded in the git log. E.g if one of your changes turn out to be problematic and need to be reverted later, with atomic commits, it'll hopefully be just a revert of a specific commit(s).

> Oh! Another annoyance I had with zbus is that it sends tracing messages at info level.

Does it? I only see one use of `tracing::info!` in all the code. In either case, best we deal with this in a separate issue+PR (perhaps after this one is merged).

> Alright, I recommitted everything. What do you think?

Awesome. I see only 4 commits so I'm sure it's definitely much closer to the target already (maybe already there). I'll review the commits soon.

zeenix:
> @REALERvolker1
[Merge branch 'dbus2:main' into main](https://github.com/dbus2/zbus/pull/712/commits/f69084fac1cf2af6b91443547182752a71a684bc)

Please don't merge main but rather rebase on it and force push. Otherwise, we get unrelated changes/commits in the PR.

REALERvolker1:
If the number of botched commits bothers you, don't worry I have 3 different backups locally

zbus/src/connection/mod.rs
```diff
@@ -719,8 +719,8 @@ impl Connection {
                                     // else the name was released in the meantime. :shrug:
                                 }
                                 Ok(_) => (),
-                                Err(e) => {
-                                    warning!("Failed to parse `NameAcquired` signal: {}", e)
+                                Err(_e) => {
+                                    warning!("Failed to parse `NameAcquired` signal: {}", _e)
```
zeenix, line 723: The task is to not have to do this at all. The caller of the macros, should not need to worry about the variables being unused.

REALERvolker1:
Will the compiler optimize away spurious calls to `format_args`? I replaced the `()` in the logging macros with
```rust
{
   format_args!($stuff);
}
```
From what I know, it constructs those at compile-time. I basically don't want to impact runtime performance at all (except removing tracing macros probably makes it faster).

REALERvolker1:
For some reason I only just now realized what you meant by `let _ = ...` 😅

I tried out what you suggested, but the clippy warnings do not go away.
```rust
macro_rules! warning {
    ($first:literal $(, $arg:expr)+$(,)?) => {
        {
            let _ = (&$first $(, &$arg)+);
        }
    };
    ($only:literal) => {
        {
            let _ = &$only;
        }
    };
}
```
If I remove the borrows, the variables are used, but moved into the _ variable, which breaks things.

REALERvolker1:
Also I realized an optimization I can do is to replace the + symbols with * and it works without the $only branch, but I can do that after resolving this issue.

zeenix:
> If I remove the borrows, the variables are used, but moved into the _ variable, which breaks things.

I just checked it locally. The borrow actually works when the variables are passed separately from the literal string. It fails only when the variables are used in the literal string (e.g `{e}`). This will be an issue even if you don't borrow (but you don't get to this warning because there are errors because of the move elsewhere).

Hm.. this is tricky. @bilelmoussaoui since you requested this, do you have some ideas on how to nicely resolve this problem?

zeenix:
One not-so-bad solution that comes to mind is that we require separate arguments to be always passed by reference. IMO the inlined style (`debug!("inlining {e} here")`) should be preferred anyway.

zeenix:
BTW,

```rust
    ($only:literal) => {
        {
            let _ = &$only;
        }
    };
```

You don't need to do anything for the literal case. It can just be `()`.

REALERvolker1:
We could overengineer a solution with zbus_macros

I tried out the format_args call in the playground and with release optimizations, it generates the same asm as if I didn't even have the statement. https://play.rust-lang.org/?version=stable&mode=release&edition=2021&code=fn+main%28%29+%7B%0A++++let+foo+%3D+%22Hello%21%22%3B%0A++++%2F%2F+%7B%0A++++%2F%2F+++++format_args%21%28%22variable+%7Bfoo%7D%22%29%3B%0A++++%2F%2F+%7D%0A%7D

REALERvolker1:
The format_args macro is const in nightly too afaik

zeenix:
> We could overengineer a solution with zbus_macros

This needs to be internal only, sorry. Also it doesn't really fit in zbus_macros.

> I tried out the format_args call in the playground and with release optimizations, it generates the same asm as if I didn't even have the statement. https://play.rust-lang.org/?version=stable&mode=release&edition=2021&code=fn+main%28%29+%7B%0A++++let+foo+%3D+%22Hello%21%22%3B%0A++++%2F%2F+%7B%0A++++%2F%2F+++++format_args%21%28%22variable+%7Bfoo%7D%22%29%3B%0A++++%2F%2F+%7D%0A%7D

Well, yet another reason it's not a good idea. We also can't rely on the fact that it will never gain `#[must_use]` attribute.

zeenix:
> One not-so-bad solution that comes to mind is that we require separate arguments to be always passed by reference. IMO the inlined style (`debug!("inlining {e} here")`) should be preferred anyway.

Just in case it wasn't clear, by this I meant that we can use `let _ = ..` solution but use the args by value. The tracing macros can take references as well.

REALERvolker1:
Hey sorry for the late reply, I was busy. As for variables, I assign everything in tuples into an underscore, and put in `#[allow(unused_variables)]` for good measure.

This does require us to use the args by value, but war is hell

zbus/src/proxy/mod.rs
```diff
@@ -25,6 +24,10 @@ use crate::{
     AsyncDrop, Connection, Error, Executor, MatchRule, MessageStream, OwnedMatchRule, Result, Task,
 };
 
+use crate::abstractions::logging::{debug, trace};
+#[cfg(feature = "tracing")]
+use tracing::Instrument;
```
zeenix, line 29: Best if we just remove the import and use it with full path. Best to minimize the conditional code.

zbus/Cargo.toml
```diff
@@ -77,12 +83,8 @@ tokio = { version = "1", optional = true, features = [
   "io-util",
   "process",
   "sync",
-  "tracing",
 ] }
-tracing = "0.1.40"
-vsock = { version = "0.4.0", optional = true }
-tokio-vsock = { version = "0.4", optional = true }
-xdg-home = "1.1.0"
+
```
zeenix, line 87: let's not include unneeded changes (talking about moving the deps here).

zeenix (changes requested):
Sorry but still some way to go IMO. Apart from inline comments, you'll need to rebase against the latest git main (to resolve the conflicts) and please squash commits and update the commit log accordingly.

Don't give up though! Despite all that, you've made a lot of progress on this and I'm really looking forward to merging this work soon.

zeenix:
@REALERvolker1 Hi, do you think you'll have the time and motivation to finish this?

zeenix:
@bilelmoussaoui Seems like @REALERvolker1 lost his interest. However, if you want to take up rebasing and finishing this PR, I'll be very happy to have this included in 5.0. Otherwise, it will have to wait for the next cycle. I'm closing this PR for now. Please do let me know if you think you'll be working on finishing this or not, so I can remove #331 from 5.0 milestone.

bilelmoussaoui:
no i won't have the time for it

zeenix:
> no i won't have the time for it

Thank you for letting us know.

Issue #231, opened 2021-12-24
Name collisions in generated Proxy code when propery getter/setter exists as a method


zeenix:
In GitLab by @hansihe on Dec 24, 2021, 19:44

There are several different variants of this, all of which I found in interfaces by `NetworkManager`.

---

### Case 1: Method collision with property

See the [`org.freedesktop.NetworkManager`](https://github.com/NetworkManager/NetworkManager/blob/main/introspection/org.freedesktop.NetworkManager.xml) interface for an example.

This interface has both a signal and a method named `State`:

```rust
    /// state method
    fn state(&self) -> zbus::Result<u32>;

    /// State property
    #[dbus_proxy(property)]
    fn state(&self) -> zbus::Result<u32>;
```

This will generate the following error when compiled:

```
error[E0201]: duplicate definitions with name `state`:
  --> service/src/dbus/network_manager/gen/org_freedesktop_NetworkManager.rs:15:1
   |
15 | #[dbus_proxy(interface = "org.freedesktop.NetworkManager")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | previous definition of `state` here
   | duplicate definition
   |
   = note: this error originates in the attribute macro `dbus_proxy` (in Nightly builds, run with -Z macro-backtrace for more info)
```

### Case 2: Method/property/signal collision with [signal]_changed function

See the [`org.freedesktop.NetworkManager`](https://github.com/NetworkManager/NetworkManager/blob/main/introspection/org.freedesktop.NetworkManager.xml) interface for an example.

The code generation will generate a function named `receive_state_changed` for both of these:

```rust
    /// StateChanged signal
    #[dbus_proxy(signal)]
    fn state_changed(&self, state: u32) -> zbus::Result<()>;

    /// State property
    #[dbus_proxy(property)]
    fn state(&self) -> zbus::Result<u32>;
```

This will generate the following error when compiled:

```
error[E0201]: duplicate definitions with name `receive_state_changed`:
  --> service/src/dbus/network_manager/gen/org_freedesktop_NetworkManager.rs:15:1
   |
15 | #[dbus_proxy(interface = "org.freedesktop.NetworkManager")]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   | |
   | previous definition of `receive_state_changed` here
[...]

Issue #169, opened 2021-06-12
xmlgen: property setter can conflict with target interface provided setter method


zeenix:
As an example, the following command:

```
zbus-xmlgen --system org.freedesktop.login1 /org/freedesktop/login1
```

will produce uncompilable code as `set_wall_message` method will be generated for `SetWallMessage` D-Bus method but it'll also generate the same as property setter for the writable `WallMessage` property.

Issue #312, opened 2023-02-01
*Dict macros don't support nested dictionaries 'a{sa{sv}}'


zeenix:
In GitLab by @seiwill on Feb 1, 2023, 17:58

Dictionary of Dictionaries is not supported and returns an `Error::PaddingNot0` when decoded by `Message::body()`.

This code attempts to get the ActiveConnection Settings from NetworkManager. It reads the `a{sa{sv}}` ConnectionSettings which `Option`ally contains other Dict Structs.

```rust
use zbus::{blocking::{Connection, Proxy}, zvariant::{DeserializeDict, Type, OwnedObjectPath}};

#[derive(Debug, Clone, PartialEq, Default, Type, DeserializeDict)]
#[zvariant(signature = "dict")]
pub struct SettingsConnection {
    pub autoconnect: Option<bool>,
    pub id: Option<String>,
    pub interface_name: Option<String>,
    pub read_only: Option<bool>,
    pub timestamp: Option<u64>,
    #[zvariant(rename = "type")]
    pub connection_type: Option<String>,
    pub uuid: Option<String>,
}

#[derive(Debug, Clone, PartialEq, Default, Type, DeserializeDict)]
#[zvariant(signature = "dict")]
pub struct SettingsIp {
    pub method: Option<String>,
    pub addresses: Vec<String>,
}

#[derive(Debug, Clone, PartialEq, Default, Type, DeserializeDict)]
#[zvariant(signature = "a{sa{sv}}")]
pub struct ConnectionSettings {
    pub connection: Option<SettingsConnection>,
    pub ipv4: Option<SettingsIp>,
}

fn main() {
    let connection = Connection::system().unwrap();
    let proxy = Proxy::new(
        &connection,
        "org.freedesktop.NetworkManager",
        "/org/freedesktop/NetworkManager",
        "org.freedesktop.NetworkManager"
    ).unwrap();

    let active_connections: Vec<OwnedObjectPath> = proxy.get_property("ActiveConnections").unwrap();
    for ac in active_connections {
        println!("Looking up {ac:?}");

        let ac_proxy = Proxy::new(
            &connection,
            "org.freedesktop.NetworkManager",
            &ac,
            "org.freedesktop.NetworkManager.Connection.Active"
        ).unwrap();
        let conn_path: OwnedObjectPath = ac_proxy.get_property("Connection").unwrap();
        println!("  Connection Settings {conn_path:?}");

        let settings_proxy = Proxy::new(
            &connection,
[...]

PR #722, opened 2024-04-17
🐛 zx: Prevent collisions between methods, signals and properties

bachp:
Prevent collisons by adding an explicit suffix to properties and signals.

If a collision is detected a `_` is added to the identifier until there is no more collision.

Closes:
- #231
- #169

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -1,5 +1,6 @@
 use snakecase::ascii::to_snakecase;
 use std::{
+    collections::HashMap,
     error::Error,
     fmt::{Display, Formatter, Write},
     process::{Command, Stdio},
@@ -172,11 +173,13 @@ impl<'i> GenTrait<'i> {
         writeln!(w, ")]")?;
         writeln!(w, "trait {name} {{")?;
 
+        let mut identifiers = Identifier::default();
+
         let mut methods = iface.methods().to_vec();
         methods.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for m in &methods {
             let (inputs, output) = inputs_output_from_args(m.args());
-            let name = to_identifier(&to_snakecase(m.name().as_str()));
+            let name = identifiers.make_unique(&to_identifier(&to_snakecase(m.name().as_str())));
             writeln!(w)?;
             writeln!(w, "    /// {} method", m.name())?;
             if pascal_case(&name) != m.name().as_str() {
@@ -190,7 +193,8 @@ impl<'i> GenTrait<'i> {
         signals.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for signal in &signals {
             let args = parse_signal_args(signal.args());
-            let name = to_identifier(&to_snakecase(signal.name().as_str()));
+            let name =
+                identifiers.make_unique(&to_identifier(&to_snakecase(signal.name().as_str())));
             writeln!(w)?;
             writeln!(w, "    /// {} signal", signal.name())?;
             if pascal_case(&name) != signal.name().as_str() {
@@ -214,18 +218,20 @@ impl<'i> GenTrait<'i> {
             writeln!(w)?;
             writeln!(w, "    /// {} property", p.name())?;
             if p.access().read() {
+                let getter_name = identifiers.make_unique(&name);
                 writeln!(w, "{}", fn_attribute)?;
                 let output = to_rust_type(p.ty(), false, false);
                 hide_clippy_type_complexity_lint(w, p.ty().signature())?;
-                writeln!(w, "    fn {name}(&self) -> zbus::Result<{output}>;",)?;
+                writeln!(w, "    fn {getter_name}(&self) -> zbus::Result<{output}>;",)?;
             }
 
             if p.access().write() {
+                let setter_name = identifiers.make_unique(&format!("set_{}", &name));
                 writeln!(w, "{}", fn_attribute)?;
                 let input = to_rust_type(p.ty(), true, true);
                 writeln!(
                     w,
-                    "    fn set_{name}(&self, value: {input}) -> zbus::Result<()>;",
+                    "    fn {setter_name}(&self, value: {input}) -> zbus::Result<()>;",
                 )?;
             }
         }
@@ -434,6 +440,23 @@ fn to_identifier(id: &str) -> String {
         format!("{id}_")
     } else {
         id.replace('-', "_")
+    }
+}
+
+#[derive(Default)]
+struct Identifier {
+    used_identifiers: HashMap<String, usize>,
+}
+
+impl Identifier {
+    fn make_unique(&mut self, identifier: &str) -> String {
+        let use_count = self
+            .used_identifiers
+            .entry(identifier.to_owned())
+            .or_default();
+        let new_identifier = format!("{}{}", &identifier, "_".repeat(*use_count));
+        *use_count += 1;
+        new_identifier
     }
 }
 
--- a/zbus_xmlgen/tests/data/sample_object0.rs
+++ b/zbus_xmlgen/tests/data/sample_object0.rs
@@ -45,6 +45,12 @@ trait SampleInterface0 {
         calypso: &zbus::zvariant::Value<'_>,
     ) -> zbus::Result<()>;
 
+    /// SetWallMessage method
+    fn set_wall_message(&self) -> zbus::Result<()>;
+
+    /// State method
+    fn state(&self) -> zbus::Result<()>;
+
     /// Changed signal
     #[zbus(signal)]
     fn changed(&self, new_value: bool) -> zbus::Result<()>;
@@ -67,6 +73,10 @@ trait SampleInterface0 {
     /// SignalValue signal
     #[zbus(signal)]
     fn signal_value(&self, value: zbus::zvariant::Value<'_>) -> zbus::Result<()>;
+
+    /// State signal
+    #[zbus(signal)]
+    fn state_(&self) -> zbus::Result<()>;
 
     /// Bar property
     #[zbus(property)]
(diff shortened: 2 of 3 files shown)


zbus_xmlgen/src/lib.rs
```diff
@@ -204,7 +204,7 @@ impl<'i> GenTrait<'i> {
         let mut props = iface.properties().to_vec();
         props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for p in props {
-            let name = to_identifier(&to_snakecase(p.name().as_str()));
+            let name = to_identifier(&to_snakecase(p.name().as_str())) + "_property_";
```
zeenix, line 207: while adding `signal` and `property` prefix resolve the issues in question, doing so unconditionally even when there's no conflict, isn't nice. The would be a pretty big behavior change and gives no benefit to cases where there are no conflicts. 

I think I good solution here would only change generation of conflicted parts and leave the rest, as is.

zeenix (changes requested):
Thanks so much for contributing this PR and also to follow our dev practices from the start. However, I think a good solution here would require more work than changing all generated names.

zbus_xmlgen/src/lib.rs
bachp: Thanks for the feedback. I think conditionally adding the suffix in case of conflict would work.

However, I also like the new convention as it gives a clear distinction between the three types.
Would you accept a command line option that always enforces the more verbose convention even when no conflict is detected?

zeenix: > However, I also like the new convention as it gives a clear distinction between the three types.

Sorry, no. The attributes already make it clear, what's a property, a signal and the lack of any attribute means it's a method. There is no need for further clarity.

> Would you accept a command line option that always enforces the more verbose convention even when no conflict is detected?

If you can provide compelling reason for it, we can discuss it. :)

bachp: I update the PR to just add `_` until there is no more collision.

zbus_xmlgen/src/lib.rs
```diff
@@ -190,7 +194,8 @@ impl<'i> GenTrait<'i> {
         signals.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for signal in &signals {
             let args = parse_signal_args(signal.args());
-            let name = to_identifier(&to_snakecase(signal.name().as_str()));
+            let name = collision_hander
+                .avoid_collision(to_identifier(&to_snakecase(signal.name().as_str())));
```
zeenix, line 198: I think the 2 steps should be combined into one. We don't really need a separate type for name collision, just one for creating identifiers.

zeenix (changes requested):
LGTM otherwise.

zbus_xmlgen/src/lib.rs
bachp: First I also tought this, but the `to_identifier` is also used in other places, eg. function parameters, and they don't collide with signals and properties. but

I could move the `to_identifier` function inside the `avoid_collision` but I thought this was it is more explict

zeenix: > First I also tought this

Ah, you're ahead of me. :)

> I could move the `to_identifier` function inside the `avoid_collision` but I thought this was it is more explict

Or you could have the new API use `to_identifier` itself so that caller doesn't need to care. I'd call the type `UniqueNameGenerator` and method `identifier_for_name`, which takes the name as arg, also taking care of `to_snakecase`. You can suggest a better name, just don't like collision in the name (sounds negative :laughing:).

> I was just wondering if there is a case where we also have collisions in arguments?

Doubtful but even if there was, it won't be global (i-e args of different methods can have the same name).

bachp: I tried severl variants of combining operations. In the end I found it's best to keep the operations separate so I can combine these in the different cases:

- Functions and signals i need `to_snakecase`, `to_identifier` and `make_unique`
- For Properties I need  `to_snakecase`, `to_identifier`, and I need `make_unique` as a separate operation depending on read or write case.
- For arguments I only need `to_identifier`, optionally we could apply `make_unique` per function here but I don't see this needed.

So my solution now is to just rename the CollisionHandler to Idenfifier and have a make_unique method.

zeenix: > * Functions and signals i need `to_snakecase`, `to_identifier` and `make_unique`

As I wrote, it would be best to just abstract it all. The caller should just ask for a unique identifier for the name and that's it.

> * For Properties I need  `to_snakecase`, `to_identifier`, and I need `make_unique` as a separate operation depending on read or write case.

From what I can tell from your commit, it's basically doing the same thing. Of course, it would make sense to have a separate API for properties in `UniqueNameGenerator`, e.g `property_getter` and `property_setter` (or it could just be `property` with a boolean arg that tells the direction. Then `identifier_for_name` could just be `method`. 

> * For arguments I only need `to_identifier`, optionally we could apply `make_unique` per function here but I don't see this needed.

:+1:

bachp: I am not sure what you are asking now.

As I see it the current implementation has several functions/methods (`to_snakecase`, `to_identifier` and `make_unique`) that can be combined ot form the final identifier.

The rules how these need to be combined for the different cases are applied locally where the element (`method`, `property`, `signal` and `arguments`) are processed.

The way I understand your proposal is to put the rules how to combine these into separate functions/methods and just call them? While I can do that I'm not convinced it is a good idea as it basically splits the rules into two locations.

zeenix: > The way I understand your proposal is to put the rules how to combine these into separate functions/methods and just call them, e.g. `to_method`,`to_signal`, `to_property` and `to_argument`

At the very least, it should have been clear, which names I proposed. :)

> While I can do that I'm not convinced it is a good idea as it basically splits the rules into two locations.

How so? They'll do pretty much the same thing as you're doing right now, just in one place/type. They will all use `to_snakecase` and `to_identifier`, for example.

zeenix: Let me know, if you'd want me to make the proposed changes in a separate commit on top.

bachp: I did a test with these changes and the NetworkManager dbus xml, and it doesn't solve all the conflicts. I need to dig a bit deeper. I think I will add a test case using the NetworkManager interface description to make the issue more reproducible.

zeenix:
@bachp oh I'm so sorry. I never got a notification that you requested a review after updating the pr. I'll look into this very soon. 

Thanks for your patience. 🙏

bachp:
@zeenix As mentioned in https://github.com/dbus2/zbus/pull/722#discussion_r1581259396 my change didn't completely solve the issue with the networkmanager dbus specification. Unfortunately I was busy with other things.

I want to add a test based on the real world example, not sure when I will get to it tough.

zeenix:
@bachp i hope you've not forgotten about your existing PR. 😉

bachp:
@zeenix No I did not but I ran into other issues while trying to interact with network manager, mainly https://github.com/dbus2/zbus/issues/312 which is more of a blocker for me.I will come back to this once I found a solution for the more blocking issues.

zeenix (changes requested):
GH thinks this is a pending review from me so let me fix that. 😊 Also, it would be great if you could finish this soon. 🙏

bachp:
@zeenix Not sure what is still open and why GH thinks there is a change required. From my point of view this is ready, I also rebased it on main.

It doesn't fix all naming collisions I encountered with NetworkManager, but I can address the remaining issues in a sparate PR.

zeenix:
> @zeenix Not sure what is still open

Oh? I understood from your previous two [comments](https://github.com/dbus2/zbus/pull/722#discussion_r1581259396) here that it's not ready and you've to fix something still.
 
> why GH thinks there is a change required. From my point of view this is ready

Because there is still a conversation that's not marked as "resolved" yet.

> It doesn't fix all naming collisions I encountered with NetworkManager, but I can address the remaining issues in a sparate PR.

For sure. I thought they were a blocker for you and you want to address them all here.

zbus_xmlgen/src/lib.rs
zeenix: @bachp this is the part that is still not resolved. You never answered my last question here. I know it's been many months since so I don't blame you for not remembering but let's try to conclude this only remaining issue,

zeenix (changes requested):
Based on the latest discussion here, I wanted to set the status right.

zbus_xmlgen/src/lib.rs
zeenix: hello??

bachp: @zeenix I'm not sure what the best solution is. If you have a proposal please apply it on top of this branch.

zeenix: > @zeenix I'm not sure what the best solution is. If you have a proposal please apply it on top of this branch.

Perhaps we can find one if you'd follow up with the discussion here? I don't see a problem mentioned here. It's more about semantics afaict and bike-shedding (sorry that's my bad but I care about code quality perhaps a bit too much) about how the functions should work/be named etc.

zeenix: > If you have a proposal

My proposal is above btw.

>  please apply it on top of this branch.

It'd be easier for me to start from scratch and your effort here and my review time will be wasted.

bachp: Ok maybe we need to take a step back as I'm not sure what question you are referring too.

From my point of view this change works as it prevents collisions. It's just that there seems to be additional conflicts with hardcoded name when generating a NetworkManager interface. However I think this can be solved later.

What is missing from your point of view?

zeenix: > Ok maybe we need to take a step back as I'm not sure what question you are referring too.

The "How so" line [in the comment above](https://github.com/dbus2/zbus/pull/722#discussion_r1574542893) where the logical flow just ended.

> What is now still open from your point of view?

The changes I asked for here. If you have already addressed them, please let me know and I'll review again.

---

As for making everything work for NetworkManager, that indeed can wait and was not what this discussion/thread was original about.

zeenix: Also, the branch needs a manual rebase.

zeenix:
@bachp Anyway, I've spent a lot of time already pointed out the obvious (if you'd care to read) again and again and each time you drop the ball. I've better things to do. I'm banning you from this project.

Issue #281, opened 2022-10-12
zvariant::Dict please allow iterating the key/value pairs


zeenix:
In GitLab by @estokes on Oct 12, 2022, 22:46

Hi, at the moment there is no way to just iterate the key/value pairs of a zvariant::Dict. If you want to do it you need to convert it to a hashmap, and to do that you need to read the signature and pick a concrete type that implements Basic, it's actually quite a lot of code to do this, and what I (and probably others) really want is just to iterate the values. I see that it's stored as a Vec of DictEntries, which are just the key and value, so it would be trivial to allow this iteration.

Thanks, zbus is awesome!

adamcavendish:
Hi @zeenix, I would still suggest that we should expose a generic iterator for `zvariant::Dict`. It is because `TryFrom` signature is listed as following:

```rust
impl<'k, 'v, K, V, H> TryFrom<Dict<'k, 'v>> for HashMap<K, V, H>
where
    K: Basic + TryFrom<Value<'k>> + std::hash::Hash + std::cmp::Eq,
    V: TryFrom<Value<'v>>,
    H: BuildHasher + Default,
    K::Error: Into<crate::Error>,
    V::Error: Into<crate::Error>
```

and K requires Hash. For `f64`, it is a valid D-Bus key but it cannot be converted from Dict to HashMap because f64 doesn't implement std::hash::Hash.

Feel free to try:

```rust
let key_signature = zvariant::Signature::from_str_unchecked("d");
let val_signature = zvariant::Signature::from_str_unchecked("s");
let dict = zvariant::Dict::new(key_signature, val_signature);
let hashmap = HashMap::<f64, String>::try_from(v).unwrap();
```

adamcavendish:
For anyone who meets the same issue or possibly locked into a legacy zbus version can use ordered_float package and make a f64 wrapper to workaround this issue:

```rust
/// F64Wrapper is a wrapper for ordered_float::OrderedFloat package.
/// We would need this wrapper is because of a zvariant issue that its Dict doesn't implement iterations.
/// Nor can it be converted into HashMap<f64, BuscliVariant> because f64 doesn't implement Hash trait.
/// Remove F64Wrapper after zvariant::Dict supports iteration.
#[derive(PartialEq, Eq, hash::Hash, PartialOrd, Ord)]
struct F64Wrapper(ordered_float::OrderedFloat<f64>);

impl zvariant::Basic for F64Wrapper {
[...]

PR #734, opened 2024-04-25
🚸 zv: Dict add iterator support

adamcavendish:
This change aims to partially solve #281 to expose an API for `zvariant::Dict` to iterate via a tuple of key and value, aka. (Value, Value) type.

However, because we have migrated to use `BTreeMap<K, V>` for `zvariant::Dict`'s internal representation, we still cannot use `f64` as keys based on the restrictions of `Ord` required by `BTreeMap::<K, V>::insert()` function signature.

```rust
pub fn insert(&mut self, key: K, value: V) -> Option<V>
where
    K: Ord;
```

Since all `Basic` types are allowed in D-Bus specification, we still need to think of a way to allow users to correctly build up a `zvariant::Dict` whose signature is similar to `a{dv}`.

### DIFF
--- a/zvariant/src/dict.rs
+++ b/zvariant/src/dict.rs
@@ -151,12 +151,29 @@ impl<'k, 'v> Dict<'k, 'v> {
         }
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = (&Value<'k>, &Value<'v>)> {
+        self.map.iter()
+    }
+
+    pub fn iter_mut(&mut self) -> impl Iterator<Item = (&Value<'k>, &mut Value<'v>)> {
+        self.map.iter_mut()
+    }
+
     // TODO: Provide more API like https://docs.rs/toml/0.5.5/toml/map/struct.Map.html
 }
 
 impl Display for Dict<'_, '_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         dict_display_fmt(self, f, true)
+    }
+}
+
+impl<'k, 'v> IntoIterator for Dict<'k, 'v> {
+    type Item = (Value<'k>, Value<'v>);
+    type IntoIter = <BTreeMap<Value<'k>, Value<'v>> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.map.into_iter()
     }
 }
 
--- a/zvariant/src/lib.rs
+++ b/zvariant/src/lib.rs
@@ -1156,6 +1156,30 @@ mod tests {
         let map: BTreeMap<i64, String> = dict.try_into().unwrap();
         assert_eq!(map[&1], "123");
         assert_eq!(map[&2], "456");
+        // Use iterator
+        let mut dict = Dict::from(map);
+        let expect = vec![
+            (Value::from(1i64), Value::from("123")),
+            (Value::from(2i64), Value::from("456")),
+        ];
+        let expect_iter = expect.iter().map(|(k, v)| (k, v)).collect::<Vec<_>>();
+        let actual = dict.iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = (&dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = (&mut dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        for (_, v) in dict.iter_mut() {
+            if let Value::Str(vv) = v {
+                *vv = Str::from(vv.to_string() + "-hello");
+            }
+        }
+        let actual = dict.into_iter().collect::<Vec<_>>();
+        let expect = vec![
+            (Value::from(1i64), Value::from("123-hello")),
+            (Value::from(2i64), Value::from("456-hello")),
+        ];
+        assert_eq!(actual, expect);
 
         #[cfg(feature = "gvariant")]
         {


zvariant/src/dict.rs
```diff
@@ -151,6 +154,14 @@ impl<'k, 'v> Dict<'k, 'v> {
         }
     }
 
+    pub fn iter(&self) -> Iter<'_, 'k, 'v> {
+        Iter(self.map.iter())
+    }
+
+    pub fn iter_mut(&mut self) -> IterMut<'_, 'k, 'v> {
```
zeenix, line 161: I think best to return `impl Iterator<Item = .. >` so we don't have to do wrap and implement all the traits.

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,158 @@ mod tests {
         assert_eq!(dict1, dict2);
     }
 
+    fn mock_fd<'a>(value: i32) -> Fd<'a> {
+        use std::os::fd::BorrowedFd;
+        use std::os::fd::RawFd;
+        let fd = unsafe { BorrowedFd::borrow_raw(value as RawFd) };
+        Fd::from(fd)
+    }
```
zeenix, line 1301: You can just use `std::stdin().as_fd().into()` and then we don't need this function or at least it can be even simpler/safer?

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,158 @@ mod tests {
...
+
+    dict_iter_mut_tests_gen! {
+        // basic -> str
+        dict_iter_mut_01, p2s: u8,
+        dict_iter_mut_02, p2s: u16,
+        dict_iter_mut_03, p2s: i16,
+        dict_iter_mut_04, p2s: u32,
+        dict_iter_mut_05, p2s: i32,
+        dict_iter_mut_06, p2s: u64,
+        dict_iter_mut_07, p2s: i64,
```
zeenix, line 1433: I don't think we need tests for each type. Just the case of a key: randomly selected basic type and string, and then viceversa is enough and then we don't need the macros.

zeenix (changes requested):
LGTM otherwise. Thanks so much for contributing this.

zvariant/src/lib.rs
adamcavendish: We cannot use `std::io::stdin().as_fd()` is because `std::io::stdin()` creates a temporary Stdin object (with a once-lock mutex internally) so it will be dropped without assignment.

zvariant/src/lib.rs
adamcavendish: Testing with a good coverage of these makes us possible to find the f64 issue as early as possible which is the reason I added all these test cases for compatibility if we might have future signature changes.
If you would still like to have only 1 for `basic -> string` and 1 for `string -> basic`, I can update the PR.

zvariant/src/lib.rs
zeenix: ah right but you can always [clone it](https://doc.rust-lang.org/std/os/fd/struct.BorrowedFd.html#method.try_clone_to_owned).

zvariant/src/lib.rs
zeenix: > Testing with a good coverage of these makes us possible to find the f64 issue

Sure but that's a one time thing and try it once and then you know (we don't need to test for it all the time). You can even leave the test case for `f64` as commented out with a comment on top.

> If you would still like to have only 1 for `basic -> string` and 1 for `string -> basic`, I can update the PR.

Yes please.

zvariant/src/dict.rs
zeenix: > ```rust
> type IntoIter = impl Iterator<Item = Self::Item>;
> ```

Hmm.. but we do you need this to be generic? Wouldn't this just be `BTreeMap::IntoIterator<Self::Item>`?

zvariant/src/lib.rs
adamcavendish: Oh, sorry I didn't make it clear. The temporary value is the `Stdin` object but not the `BorrowedFd` Object. The Stdin object is not cloneable.

Even if the Stdin object is somehow resolved, I still don't suggest cloning the BorrowedFd object because cloning a BorrowedFd object requires an extra syscall `fcntl(fd, F_DUPFD)` and close the cloned FD on `Drop`, which, in my humble option, is unnessary for testing.

zvariant/src/dict.rs
adamcavendish: Yes, that's doable. I've updated the logics in the second commit.
Note: I'll squash the commits into one after all reviews are finished and turn the PR into ready.

zvariant/src/lib.rs
adamcavendish: I've removed the macro and only use u8 and f64 as `zvariant::Basic` test sample.
The code is available in the 3rd commit (adamcavendish/zbus@c2475a549f1f05d2b54f0de1cde5aa4a9b9d2da9)
Note: I'll squash the commits into one after all reviews are finished and turn the PR into ready.

zvariant/src/lib.rs
zeenix: > Even if the Stdin object is somehow resolved, I still don't suggest cloning the BorrowedFd object

I think it was me who as not clear I think cause I meant that you can create an `OwnedFd` from it (which is why I linked to `try_clone_to_owned`).

> Even if the Stdin object is somehow resolved, I still don't suggest cloning the BorrowedFd object because cloning a BorrowedFd object requires an extra syscall `fcntl(fd, F_DUPFD)` and close the cloned FD on `Drop`, which, in my humble option, is unnessary for testing.

The fact that it's a test, actually means we shouldn't care about the extra syscalls. I'd prefer less/simpler code in over optimizations in this context.

zvariant/src/lib.rs
zeenix: > Note: I'll squash the commits into one after all reviews are finished and turn the PR into ready.

Please always squash the commits and keep them atomic (in terms of  "one logical commit per commit" rather than "each development step in a separate commit). That helps with the review since then I know which commits with what log, I'm approving. It also makes it easy to do per-commit review.

Github provides a facility to compare the diff with the previous version of the commit(s), in case the reviewer is interested in that.

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,354 @@ mod tests {
...
+                    let actual = (&dict).iter().collect::<Vec<_>>();
+                    assert_eq!(actual, expect_iter);
+                    let actual = (&mut dict).iter().collect::<Vec<_>>();
+                    assert_eq!(actual, expect_iter);
+                    let actual = dict.into_iter().collect::<Vec<_>>();
+                    assert_eq!(actual, expect);
+                }
+            )*
+        }
+    }
```
zeenix, line 1328: Now we have a macro that is not used? :thinking:

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,354 @@ mod tests {
...
+        let actual = (&dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = (&mut dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = dict.into_iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect);
+    }
+
+    #[test]
+    fn dict_iter_str_u8() {
```
zeenix, line 1353: * I'd really prefer not to have separate tests for each combo. One test for maps is enough.
* As I previous eluded to, it would actually be great if you can modify the existing test we already have for dict (`dict_value`) and try to add as little as possible.

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,354 @@ mod tests {
...
+        assert_eq!(actual, expect_iter);
+        let actual = (&dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = (&mut dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = dict.into_iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect);
+    }
+
+    // @TODO : Would need to review that `f64: Ord is not satisfied` for BTreeMap.
```
zeenix, line 1374:
```suggestion
    // TODO : Currently not possible to use f64 as `f64: Ord is not satisfied` for BTreeMap.
```

zvariant/src/lib.rs
```diff
@@ -1293,6 +1293,354 @@ mod tests {
...
+        let actual = (&dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = (&mut dict).iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect_iter);
+        let actual = dict.into_iter().collect::<Vec<_>>();
+        assert_eq!(actual, expect);
+    }
+
+    #[test]
+    fn dict_iter_mut_u8_str() {
```
zeenix, line 1564: we especially don't want separate tests just for `mut`.

zvariant/src/dict.rs
```diff
@@ -151,6 +154,14 @@ impl<'k, 'v> Dict<'k, 'v> {
         }
     }
 
+    pub fn iter(&self) -> impl Iterator<Item = (&'_ Value<'k>, &'_ Value<'v>)> {
```
zeenix, line 157: Are the explicit lifetime elisions (`'_`) really needed?

zvariant/src/dict.rs
```diff
@@ -1,3 +1,6 @@
+use std::collections::btree_map::IntoIter as BTreeMapIntoIter;
+use std::collections::btree_map::Iter as BTreeMapIter;
+use std::collections::btree_map::IterMut as BTreeMapIterMut;
```
zeenix, line 3: * This will fail the fmt job because of `imports_granularity = "Crate"` in our `.rustfmt.toml`. Please run `cargo +nightly fmt` and `cargo clippy` (apart from `cargo test`) before pushing your changes to avoid ping-pongs with the CI. :)

* I don't even see them being used in this file. Did you check if this builds (w/o warnings)? Perhaps I'm being blind? :thinking:

zeenix (changes requested):
Slowly and steadily, we'll get there. Don't give up please! :)

zvariant/src/lib.rs
adamcavendish: `mock_fd()` is now removed since we migrate to use non-macro code in `dict_value()` test case.

zvariant/src/lib.rs
adamcavendish: Removed the redundant macro.

zvariant/src/lib.rs
adamcavendish: Updated.

zvariant/src/lib.rs
adamcavendish: Updated and moved to `dict_value()`.

zvariant/src/lib.rs
adamcavendish: Moved to `dict_value()` test case.

zvariant/src/dict.rs
adamcavendish: Tested locally, not required and removed.

zvariant/src/dict.rs
adamcavendish: Updated according to format and clippy.

zeenix:
@adamcavendish Sorry for the delay on this but I just started a new job so have been a bit busy. Will have a look very soon..

adamcavendish:
> @adamcavendish Sorry for the delay on this but I just started a new job so have been a bit busy. Will have a look very soon..

Don't worry. I'll keep an eye on it. The current failures in the CI seems not related to this PR.

zeenix:
> The current failures in the CI seems not related to this PR.

Yeah, they were caused by the latest rust release catching some previously uncaught warnings. :shrug: It's been fixed now so you can rebase and CI should be good.

adamcavendish:
Rebased according to suggestions on the CI failures.

zvariant/src/dict.rs
```diff
@@ -160,6 +168,33 @@ impl Display for Dict<'_, '_> {
...
+
+impl<'k, 'v> IntoIterator for Dict<'k, 'v> {
+    type Item = (Value<'k>, Value<'v>);
+    type IntoIter = <BTreeMap<Value<'k>, Value<'v>> as IntoIterator>::IntoIter;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.map.into_iter()
+    }
+}
+
```
zeenix, line 197: I'm not sure we really need these. I doubt there will be enough users of this to justify this code, just to help avoid `.iter()` call. Besides, it a bit strange to implement `IntoIterator` if we don't implement `Iterator`. I don't think any type does that (at least in std).

zeenix (changes requested):
LGTM otherwise. Thanks so much for you patience with me, specially about the nitpicks and my sense of perfection. :) In this specific case, I just want to keep the LOC as small as possible, since I will be maintaining this code for years once it's merged and released.

zvariant/src/dict.rs
adamcavendish: I suggest that we should keep these because it improves the completeness of `Iterator` related code for `zvariant::Dict` type.

To clarify the statement of
> I doubt there will be enough users of this to justify this code, just to help avoid `.iter()` call.
It might seem so for all reference related types which in current case are `&Dict` and `&mut Dict`. However, `IntoIterator` for `Dict` is important which consumes the variable and allows the user to avoid calling `.clone()` for each value especially when the `zvariant::Value` is an `OwnedValue`.

To further clarify the completeness of `IntoIterator` trait implementation for reference types and value types, I would suggest the sample implementation for `HashMap` in standard library.
https://github.com/rust-lang/rust/blob/1.78.0/library/std/src/collections/hash/map.rs#L2175

zvariant/src/lib.rs
```diff
@@ -1277,6 +1278,327 @@ mod tests {
             decoded.unwrap_err(),
             Error::Message("unknown field `user`, expected `process_id` or `group_id`".to_string())
         );
+
+        // Iterator (u8 -> str / str -> u8)
```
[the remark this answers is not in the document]
adamcavendish: Before my modifications today, please allow me double confirm that you would prefer simpler cases than coverage, right? If so, I'll update the PR to least amount of unit test code to cover only primitive cases, i.e. to reuse your `i32 -> String` and add a minimum case there.

zvariant/src/dict.rs
zeenix: > However, `IntoIterator` for `Dict` is important which consumes the variable and allows the user to avoid calling `.clone()` for each value especially when the `zvariant::Value` is an `OwnedValue`.

Sure, that's a good point. Let's keep the last `IntoIter` impl then and remove the first 2 (the reference ones).

adamcavendish: Removed the first 2.

zvariant/src/lib.rs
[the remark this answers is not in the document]
adamcavendish: Sure. Minimized the test cases to only have `i64 -> &str` version.

zeenix:
@adamcavendish Thanks again for your contribution and patience. I think it's now good for merging. :+1:

PR #758, opened 2024-05-01
:memo: zv: document serde_repr usage for enums

### LINKED ISSUE #686: zvariant: Better document serde_repr need with `repr`
Minimal repro: https://github.com/ChocolateLoverRaj/zbus/blob/297e92097f818a56580a1ab1f48f2548a65b0cf8/zbus/examples/return_enum.rs

What happens:
If I return a string or other "standard" type, the function works as expected. If I use an enum and I call
```
busctl --user call org.zbus.ReturnEnum /org/zbus/ReturnEnum org.zbus.ReturnEnum GetSomething 
```
I get the error
```
Call failed: Remote peer disconnected
```
It seems like the interface just stops because if I call it again I get the error
```
Call failed: The name is not activatable
```

jonathanhood:
Previously it was unclear that the `serde_repr` crate was required for unit enums to be serialized with the proper width as provided by a given `#[repr(_)]` annotation. Just adding a quick shout out to hopefully avoid confusion in the future.

Fixes #686

### DIFF
--- a/zvariant/README.md
+++ b/zvariant/README.md
@@ -100,8 +100,11 @@ let encoded = to_bytes(ctxt, &e).unwrap();
 let decoded: Enum = encoded.deserialize().unwrap().0;
 assert_eq!(decoded, e);
 
-#[derive(Deserialize, Serialize, Type, PartialEq, Debug)]
-// W/o `repr` spec, `u32` is assumed.
+// Enum encoding can be adjusted by using the `serde_repr` crate
+// and by annotating the representation of the enum with `repr`.
+use serde_repr::{Serialize_repr, Deserialize_repr};
+
+#[derive(Deserialize_repr, Serialize_repr, Type, PartialEq, Debug)]
 #[repr(u8)]
 enum UnitEnum {
     Variant1,


zvariant/README.md
```diff
@@ -100,8 +100,11 @@ let encoded = to_bytes(ctxt, &e).unwrap();
 let decoded: Enum = encoded.deserialize().unwrap().0;
 assert_eq!(decoded, e);
 
-#[derive(Deserialize, Serialize, Type, PartialEq, Debug)]
-// W/o `repr` spec, `u32` is assumed.
+// Enum encoding can be adjusted by using the serde_repr crate
```
zeenix, line 103:
```suggestion
// Enum encoding can be adjusted by using the `serde_repr` crate
```

zvariant/README.md
```diff
@@ -113,6 +116,7 @@ assert_eq!(UnitEnum::signature(), "y");
 let encoded = to_bytes(ctxt, &UnitEnum::Variant2).unwrap();
 let e: UnitEnum = encoded.deserialize().unwrap().0;
 assert_eq!(e, UnitEnum::Variant2);
+assert_eq!(encoded.len(), 1);
```
zeenix, line 119: Don't need this in example.

zeenix (changes requested):
Thanks, LGTM apart from 2 nits.

zeenix (approved):
LGTM. I'll approve the CI run after I've fixed the windows CI (#759).

zeenix:
done. You'll need to rebase first though.

Issue #743, opened 2024-04-30
Struct fields with variant type signature serialized as concrete types


jonathanhood:
When serializing struct members the zvariant serializer doesn't properly check if a member should be serialized as a variant with a type signature or as just the basic type.

## Reproduction

```rust
use serde::{ Deserialize, Serialize};
use zvariant::{serialized::Context, Type};

#[derive(Serialize, Deserialize, Type, Debug)]
#[zvariant(signature = "(vv)")]
struct VariantMembers {
    a: i32,
    b: i32
}

#[derive(Serialize, Deserialize, Type, Debug)]
#[zvariant(signature = "(ii)")]
struct ConcreteMembers {
    a: i32,
    b: i32
}

fn main() {
    let input = VariantMembers { a: 1, b: 2 };

    let context = Context::new_dbus(zvariant::Endian::Little, 0);
    let serialized = zvariant::to_bytes(context, &input).unwrap();
    let (deserialized, _): (ConcreteMembers, usize) = serialized.deserialize().unwrap();

    println!("Deserialized: {:?}", deserialized);
}
```

## Expected Outcome

The program should fail (at the `deserialize().unwrap()` call) since a struct of type `(vv)` cannot be decoded as a struct of type `(ii)`.

## Actual Behavior

The struct with type signature `(vv)` is actually encoded as a struct with type `(ii)`. Since this is the case - the program does not fail and it outputs:

```
Deserialized: ConcreteMembers { a: 1, b: 2 }
```

## Proposed Resolution

Add a peek at the type signature when serde calls serialization methods for primitive types and:
- If the type signature doesn't match (e.g. encoding a string `s` when the type signature `i`) then fail in some well defined / debuggable way.
- If the type signature is for for a variant `v` type - encode accordingly.
- If the type signature is for a concrete type that matches the call - encode encodring.

jonathanhood:
While working on this issue I stumbled onto #749. These two issues aren't directly related - but _in code_ they end up interacting for various reasons. I'll solve that issue first and then return to this one.

zeenix:
In general, the custom signature is only for special cases where `Type` can't be used. It's a bit like `unsafe` keyword and you're on your own if you use it. 

However, the dict type is an exception. It would be great to support that without our special derive macros indeed. So if this can be supported as a side-effect of (or rather a step towards) #746, then let's do it.

jonathanhood:
[...]

Issue #772, opened 2024-05-06
zvariant: Refactor deserializers to use boxed variant struct fields


jonathanhood:
I’m seeing an issue with how `zvariant::Value` is handled during deserialization. A `Value` is essentially the rust representation of a DBus variant. This is a value which has a type signature on the wire and some following data that is encoded per the type signature preceding. This requires some extra smarts in the serializer and deserializer to handle properly.

During serialization this is handled by boxing the values in a structure with fields `zvariant::Value::Signature` and `zvariant::Value::Value`. When encoding to DBus format these field names serve as hints to the serializer about what’s going on (please encode a variant’s signature and value). When encoding in other formats (e.g. JSON) this allows the variant information to be retained by simply boxing its two components together in a way that is straightforward.

During deserialization a similar approach is not applied. Calling `Value::deserialize` instead calls `deserialize_any` which tries to its best to map the type in the type signature to some call on the deserializer. This allows any concrete value to be turned in a `Value` whether or not is has a wire encoded type signature. `deserialize_any` _does_ make an attempt to handle real variant types correctly by calling `deserialize_seq` and taking the first element as the signature and the second element as the value - but it gets confused rather easily - especially in the case where the contained variant type _is an array_. It also doesn’t handle the boxing given by the serializer at all. 

There _is_ a sidecar deserialize called `DeserializeValue` which tries to deal with that boxing, but it is specifically geared towards situations where the concrete type of the variant is known/expected by the client code. As a result, it’s not a replacement for proper handling of variant types during deserialization.

By changing the deserialization of `Value` to adhere to the boxing we can use the same hinting strategy that the serializer uses for the deserializer to simplify that logic. We’ll also be able to handle encodings to non-DBus formats in more transparent way via the boxed representation. I think that’s a better overall approach - and the same approach be used to better handle things like `ObjectPath` as well.

So, I’d like to propose that the `Value` deserializer changes to use and expect this boxing strategy. I’d also propose that it should _only_ handle deserialization for true variant types and not call `deserialize_any` to try and map concrete types. This, unfortunately, is a breaking behavior change but is “more correct” and should lead to simpler code for handling these use cases. I suspect that’s a large enough change that it would be considered API breaking. If necessary, some kind of compatibility could be be built where non-variant encodings are handled but describing the behavior might be hard and would ideally the compatibility code would be pulled in a future major rev for simplicity.

Looking for feedback before proceeding down this path. Thanks for reading all this and considering!

[...]

PR #776, opened 2024-05-07
✨ zv: support manually specified variant sigs

jonathanhood:
This change supports directly serializing concretely values into a variant type signature by manually specifying the type signature in some way (either with the `#[zvariant(signature = "<sig>")]` macro or by calling `serialize_with_signature` directly). It also provides similar behavior in reverse - supporting deserialization of a variant into a concretely typed value whenever the signature has been manually set in similar ways.

Fixes
- #743 
- #515

Obsoletes
- #520

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2506,7 +2506,7 @@ checksum = "09041cd90cf85f7f8b2df60c646f853b7f535ce68f85244eb6731cf89fa498ec"
 
 [[package]]
 name = "zbus"
-version = "4.2.0"
+version = "4.2.1"
 dependencies = [
  "async-broadcast",
  "async-executor",
@@ -2552,7 +2552,7 @@ dependencies = [
 
 [[package]]
 name = "zbus_macros"
-version = "4.2.0"
+version = "4.2.1"
 dependencies = [
  "async-io 2.3.2",
  "futures-util",
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "zbus"
-version = "4.2.0"
+version = "4.2.1"
 authors = ["Zeeshan Ali Khan <zeeshanak@gnome.org>"]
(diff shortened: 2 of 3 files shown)


jonathanhood:
Looking into the CI issue now.

zeenix:
Cool. Thanks for doing this. However, I'd prefer this PR adds support for **all** types, not just u8.

zeenix:
> way (either with the `#[zvariant(signature = "")]`

Why empty? It should be `v`, no?

jonathanhood:
> Cool. Thanks for doing this. However, I'd prefer this PR adds support for **all** types, not just u8.

Raised this way to make sure I'm not off on approach before submitting something larger. Happy to just wait for the larger one to merge anything.

zeenix:
> Raised this way to make sure I'm not off on approach before submitting something larger. Happy to just wait for the larger one to merge anything.

Ah ok. Just re-use this PR. Doesn't need to be another PR at all.

jonathanhood:
I've pushed support for all basic (non-string and non-container) types. Strings will come next and then containers. Based on my experimentation things will get harder as this progresses and the PR will grow.

jonathanhood:
I've pushed handling for string-like types. The raw bytes on the wire look good, but there are test failures though in other areas indicated I've messed up something along the way - I'll get to those soon.

jonathanhood:
Alright, I'm now butting up against the issue documented in #772 in various ways. That work will have to get pulled in, but I'll put focus on avoiding any compatibility breakages.

jonathanhood:
I made a big oof locally and force pushed. I've got my work back, luckily, by looking back in my reflog. I'll raise another PR soon once I've made more progress on this.

zeenix:
> I made a big oof locally and force pushed. I've got my work back, luckily, by looking back in my reflog. I'll raise another PR soon once I've made more progress on this.

That's ok but you didn't need to close the PR. You can always force push to fix.

PR #779, opened 2024-05-08
🚑️ zb,zm: Don't invalidate all properties on change

### LINKED ISSUE #765: zbus 4.x.x: getting a property from a GJS DBus proxy client in gnome-shell returns `null`
[system76-power](https://github.com/pop-os/system76-power) provides a drop-in replacement for `net.hadess.PowerProfiles` for integration with gnome-shell 46. Strangely, when gnome-shell gets the `ActiveProfile` property [here](https://github.com/GNOME/gnome-shell/blob/main/js/ui/status/powerProfiles.js#L107), the value that it receives is `null`. Despite that, gnome-shell is able to correctly set the `ActiveProfile` property. When I add logs to the `net.hadess.PowerProfiles` daemon implementation, it does show that it's returning this value as requested by gnome-shell. And if I add logs directly to gnome-shell, it seems that `ActiveProfile` returns `null` everywhere except immediately after setting to `ActiveProfile`, where it will revert back to `null` when the scope that set the property ends.

Yet if I run `busctl --system get-property net.hadess.PowerProfiles /net/hadess/PowerProfiles net.hadess.PowerProfiles ActiveProfile`, it correctly returns the value. The power settings page in GNOME Settings also correctly gets the value. So there may be some quirk in the GJS DBus proxy client implementation.

Can test with system76-power using zbus 4.1.2 from the `zbus-4.1.2` branch [here](https://github.com/pop-os/system76-power/tree/zbus-4.1.2). With zbus 3.15.2, there is no issue in getting a property from gnome-shell's gjs dbus proxy clients.

zeenix:
This fixes a regression in c1aa5c72cee6ba78f5b803f44418c3fc1826200b, where we started to invalidate properties in every `PropertyChanged` signal. This is wrong anyway but it also meant zbus interfaces breaking against gio.

Fixes #765.

### DIFF
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -830,10 +830,7 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     let changed = props_changed.next().await.unwrap();
     let expected_property_key = "EmitsChangedDefault";
     let args = changed.args()?;
-    assert_eq!(
-        args.invalidated_properties(),
-        &vec![expected_property_key.to_string()]
-    );
+    assert!(args.invalidated_properties().is_empty());
 
     let changed_prop_key = *args.changed_properties().keys().next().unwrap();
     assert_eq!(changed_prop_key, expected_property_key);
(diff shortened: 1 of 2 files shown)


craigcabrey:
thank you!

PR #780, opened 2024-05-08
💩 zb: Hack to make connection handshake work inside Flatpak

zeenix:
`xdg-dbus-proxy` [can't handle pipelining](https://github.com/flatpak/xdg-dbus-proxy/issues/21), hence we need to handle `NEGOTIATE_UNIX_FD` command's response before sending out `BEGIN` command and `Hello` method call message.
    
We should remote this as soon as flatpak is fixed and fix is available in major distros.

### DIFF
--- a/zbus/src/connection/handshake/client.rs
+++ b/zbus/src/connection/handshake/client.rs
@@ -1,6 +1,6 @@
 use async_trait::async_trait;
 use std::collections::VecDeque;
-use tracing::{debug, instrument, trace};
+use tracing::{debug, instrument, trace, warn};
 
 use sha1::{Digest, Sha1};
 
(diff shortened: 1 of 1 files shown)


zbus/src/connection/handshake/client.rs
```diff
@@ -179,7 +179,22 @@ impl Client {
 
         let can_pass_fd = self.common.socket_mut().read_mut().can_pass_unix_fd();
         if can_pass_fd {
-            commands.push(Command::NegotiateUnixFD);
+            // xdg-dbus-proxy can't handle pipelining, hence this special handling.
+            // FIXME: Remote this as soon as flatpak is fixed and fix is available in major distros.
```
sdroege, line 183:
```suggestion
            // FIXME: Remove this as soon as flatpak is fixed and fix is available in major distros.
```

maybe also link to the xdg-dbus-proxy issue here for future reference

matthiasclasen:
> We should remote this as soon as flatpak is fixed and fix is available in major distros.

If you consider the microoptimization of pipelining auth to be essential, your first step should be to get the dbus spec clarified to make it clear that it needs to be supported.

zeenix:
> If you consider the microoptimization of pipelining auth to be essential, your first step should be to get the dbus spec clarified to make it clear that it needs to be supported.

Well, the spec maintainer is well-aware of this ambiguity and I'm not the first one to implement this.

PR #782, opened 2024-05-09
✅ zv: add roundtrip integration tests

jonathanhood:
This change adds some integration tests for roundtrips through serde to ensure that, when used via the public API, items that have been through this process are transparent (in the logical equality sense) when compared before and after.

This also finds and fixes a papercut bug with the gvariant implementation of `deserialize_unit_struct`. This could have caused strange issues when decoding these kinds of structures within some nested container.

These test largely cover some behavior that I've been breaking off and on while digging through this code, so I figured I'd go ahead and raise a PR for them.

### DIFF
--- a/zvariant/src/gvariant/de.rs
+++ b/zvariant/src/gvariant/de.rs
@@ -274,7 +274,7 @@ impl<'de, 'd, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> de::Deseriali
     where
         V: Visitor<'de>,
     {
-        visitor.visit_unit()
+        self.deserialize_unit(visitor)
     }
 
     fn deserialize_newtype_struct<V>(self, _name: &'static str, visitor: V) -> Result<V::Value>
--- a/zvariant/tests/bytes_serde_roundtrip_test.rs
+++ b/zvariant/tests/bytes_serde_roundtrip_test.rs
@@ -1,0 +1,329 @@
+use endi::Endian;
+use serde::{Deserialize, Serialize};
+use zvariant::{
+    serialized::{Context, Data},
+    Type,
+};
+
+fn generate_contexts() -> Vec<Context> {
+    vec![
+        Context::new_dbus(Endian::Little, 0),
+        Context::new_dbus(Endian::Big, 0),
+        #[cfg(feature = "gvariant")]
+        Context::new_gvariant(Endian::Little, 0),
+        #[cfg(feature = "gvariant")]
+        Context::new_gvariant(Endian::Big, 0),
+    ]
+}
+
+#[test]
+fn serde_i8() {
+    for context in generate_contexts() {
+        let value: i8 = 42;
+        let serialized: Data<'_, '_> = zvariant::to_bytes(context, &value).unwrap();
+        let (deserialized, decoded): (i8, usize) = serialized.deserialize().unwrap();
+        assert_eq!(value, deserialized);
+        assert_eq!(decoded, serialized.len());
+    }
+}
+
+#[test]
+fn serde_i16() {
+    for context in generate_contexts() {
+        let value: i16 = 42;
+        let serialized: Data<'_, '_> = zvariant::to_bytes(context, &value).unwrap();
+        let (deserialized, decoded): (i16, usize) = serialized.deserialize().unwrap();
+        assert_eq!(value, deserialized);
+        assert_eq!(decoded, serialized.len());
+    }
+}
+
+#[test]
+fn serde_i32() {
+    for context in generate_contexts() {
+        let value: i32 = 42;
+        let serialized: Data<'_, '_> = zvariant::to_bytes(context, &value).unwrap();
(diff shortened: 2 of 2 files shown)


zvariant/src/gvariant/de.rs
```diff
@@ -274,7 +274,7 @@ impl<'de, 'd, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> de::Deseriali
     where
         V: Visitor<'de>,
     {
-        visitor.visit_unit()
+        self.deserialize_unit(visitor)
```
zeenix, line 277: Separate logical change most definitely since it's nothing to do with the tests and commit title doesn't cover this change. Please split into its own commit with it's own title and details.

zvariant/tests/bytes_serde_roundtrip_test.rs
```diff
@@ -0,0 +1,329 @@
+use endi::Endian;
+use serde::{Deserialize, Serialize};
+use zvariant::{
+    serialized::{Context, Data},
+    Type,
+};
+
+fn generate_contexts() -> Vec<Context> {
```
zeenix, line 8: Instead of adding a lot of new code in a separate file, can you please modify and update the big bunch of tests we already have in `lib.rs`, keeping additions to minimum? Feel free to move those tests under `tests/tests.rs` (or even more fine-grained split if you like) first, but you don't have to.

zeenix (changes requested):
Thanks for these but some changes needed.

zvariant/tests/bytes_serde_roundtrip_test.rs
jonathanhood: I'm unsure if the tests in `lib.rs` are integration tests or unit tests. It seems like maybe they are some of both based on my most recent look?

My reasoning with this change: The purpose of putting things in `tests` is to make them integration tests which - crucially - means they can only use the publicly exposed API of the crate. IMO it's good to have _both_ low level unit tests within the crate (ideally right with the modules they are testing) _and_ have a set of seperate integration tests of this nature which expose and cover functionality in the same way that a client program might.

If we were to split up the tests in `lib.rs` into smaller chunks / move them around / etc - IMO they would need to be split up such that things that are real low-level unit tests go with the module they are testing and things that are really integration tests would go under `tests`.

I'd like to add more integration level tests for this stuff in the future. So, my argument here isn't as much about this _specific_ change request but trying to understand patterns so I can adapt to them.

zeenix: > I'm unsure if the tests in `lib.rs` are integration tests or unit tests. It seems like maybe they are some of both based on my most recent look?

I typically don't care about that distinction. IMO the important thing is what we're testing and a lot of the existing tests are doing exactly what you're doing here: serializing and the deserializing.

zeenix: > If we were to split up the tests in `lib.rs` into smaller chunks / move them around / etc - IMO they would need to be split up such that things that are real low-level unit tests go with the module they are testing and things that are really integration tests would go under `tests`.

I'm fine with either. I also don't have a strong preference for splitting. Up to you. :)

jonathanhood: I will look into either cleaning up what I need to to get these tests merged, or splitting them up, or adding my tests to the existing pile. Not sure which path I want to take yet.

Since the discussion is started: there is an additional concern I have here. I believe the tests in `lib.rs` need to be refactored to make them a bit more straightforward / readable / etc for folks coming into the codebase and making changes.

When something goes wrong in those tests it can be pretty hard to identify what _exactly_ is being asserted and what the intention of the test is. This is largely because many of these tests would more ideally be many seperate tests each asserting an individually important - but separately addressable - behavior. By the same coin it an also make it hard to identify where _exactly_ is the best place to add tests for behavior to them.

The easiest clue that this is happening is to just look at the length of some of the tests and how deeply nested they are with logic and different assertions. It also becomes apparent when trying to change this code - it's just pretty hard to identify what exactly a test is trying to do without reaching for the debugger.

This isn't a criticism really. This happens to every well-tested project I've worked on at some point. It's a good problem to have 🤣 - just expressing another motivation for (me/you/someone?) to break things up and clean this up some.

I'll file an issue to track this.

jonathanhood:
Closing this for now. It's not a critical change and It'll be a while before I can get to these test refactors.

zeenix:
> Closing this for now. It's not a critical change and It'll be a while before I can get to these test refactors.

Wait, what about the fix?

zeenix:
> > Closing this for now. It's not a critical change and It'll be a while before I can get to these test refactors.
> 
> Wait, what about the fix?

Also "test refactor" was not required for this PR at all. That was an option if you wanted to get that done already. You could just:

1. add the failing test case to `lib.rs` in the appropriate place.
2. add the fix.

You already have the 2. even.

PR #787, opened 2024-05-10
🔨 add githook scripts 

### LINKED ISSUE #442: Cosmetic: factor out the suggested git hooks into actual scripts
`CONTRIBUTING.md` currently suggests a `pre-commit` and a `pre-push` hook that is effectively a shell script (and that script appears to be identical in busd, gimoji and probably others).

It'd be great to have those as actual scripts provided by the repo so they can be updated as-needed without the user's git hooks getting stale when a command changes. This way a user needs to only symlink to the hook file or invoke the script from their custom hook (if any). So basically what I'm suggesting is that `CONTRIBUTING.md` suggests something like:
```
$ ln -s $PWD/utils/pre-commit.hook .git/hooks/pre-commit
$ ln -s $PWD/utils/pre-push.hook .git/hooks/pre-push
```
And the few users that want to have a more custom hook setup can copy those commands out of the scripts.

Or possibly even as pre-commit hooks (those may even exist already, didn't check yet).

jonathanhood:
This change adds githook scripts to the .githooks folder so they can be more trivially enabled by developers working on this repo by using the command `git --local core.hooksPath .githooks/`. They do the same basic function as the scripts documented in `Contributing.md` but with a little extra flair. They also add a step to build (with --all-features) before pushing.

Additionally, the original scripts were removed from the contributing guide in favor of an explanation of how to enable the scripts.

Closes #442 

A quick screenshot of what execution looks like when executed on my machine.

Screenshot 2024-05-10 at 1 10 51 PM

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -89,65 +89,12 @@ Please note that there are times when clippy is wrong and you know what you are 
 cases, it's acceptable to tell clippy to
 [ignore the specific error or warning in the code](https://github.com/rust-lang/rust-clippy#allowingdenying-lints).
 
-If you intend to contribute often or think that's very likely, we recommend you setup the following git
-hooks:
+If you intend to contribute often or think that's very likely, we recommend you setup the git hook
+scripts contained within this repository. You can enable them with:
 
-* Pre-commit hook that goes in the `.git/hooks/pre-commit` file:
-
-  ```sh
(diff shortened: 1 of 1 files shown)


zeenix (changes requested):
Nice! I :heart: it in general.

jonathanhood:
Review comments should be addressed. Thanks for taking a look at this!

PR #792, opened 2024-05-14
✨ zv: support (de)serialization of `Value` directly

jonathanhood:
This change is inspired by similiar functionality in serde-json which allows for types to be serialized/deserialized into an AST directly. This is useful for a variety of reasons, and I think would be of benefit to zbus as well.

Luckily zvariant already has a type pretty close to this - the zvariant::Value. This thing can represent any value that dbus can represent on the wire. It can also model any variant received from some other service. By serializing and deserializing directly to this thing - it gives clients stronger capability to deal with variants they receive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1191,6 +1191,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "minimal-lexical"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68354c5c6bd36d73ff3feceb05efa59b6acb7626617f4962be322a825e61f79a"
+
+[[package]]
 name = "miniz_oxide"
 version = "0.7.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1233,6 +1239,16 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "memoffset 0.9.0",
+]
+
+[[package]]
+name = "nom"
+version = "7.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d273983c5a657a70a3e8f2a01329822f3b8c8172b73826411a55751e404a0a4a"
+dependencies = [
+ "memchr",
+ "minimal-lexical",
 ]
 
 [[package]]
@@ -2610,6 +2626,7 @@ dependencies = [
  "endi",
  "enumflags2",
  "glib",
+ "nom",
  "rand",
  "serde",
  "serde_bytes",
--- a/zvariant/Cargo.toml
+++ b/zvariant/Cargo.toml
@@ -34,6 +34,7 @@ time = { version = "0.3.36", features = ["serde"], optional = true }
 chrono = { version = "0.4.38", features = [
     "serde",
 ], default-features = false, optional = true }
+nom = { version = "7.1.3" }
 
 [dev-dependencies]
 serde_json = "1.0.116"
--- a/zvariant/src/array.rs
+++ b/zvariant/src/array.rs
@@ -4,7 +4,10 @@ use serde::{
     ser::{Serialize, SerializeSeq, Serializer},
 };
 use static_assertions::assert_impl_all;
-use std::fmt::{Display, Write};
+use std::{
+    collections::VecDeque,
+    fmt::{Display, Write},
+};
 
 use crate::{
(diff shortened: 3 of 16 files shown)


zeenix:
Thanks for doing this work. As discussed on the Matrix channel, I had (past tense cause I got convinced on some points) some reservations:

1. We already provide transformations to/from `Value` for all common types and provide derive macros for easily implementing these transformations for custom types. Moreover these transformations don't require any signature parsing and hence are a lot more efficient that any implementation that needs signature parsing. So even if we do add `to_value/from_value`, it should be just make use of those transformations and not require signature/`Type`. However, thinking more about this, I realized that signature-based transformations would allow any `Type` (it should actually be `DynamicType` if not already) implementing type to be transformed to/from. This would allow us to simplify a few things and eliminate the need to require `Value` transformation implementations in addition to `Type` for properties.  This is currently a common footgun. Having said that, it's not clear to me looking over the commits if this PR really does indeed implement the feature promised in the title: for example giving `to_bytes_with_signature` an `i32` with signature `v` and it being able to handle it. Is that the case?

2. While you're correct that `zvariant::Value` is a DST just like `serde_json::Value`, in practice there is a fundamental difference that could easily lead to interesting footguns if the user isn't aware of this: The former has a very specific encoding that is different from that of the contained type but the latter does not. I regret the name `Value` (it should really have been `Variant`). At least we [document this difference](https://docs.rs/zvariant/latest/zvariant/enum.Value.html). However, if zvariant was able to treat any value as a variant and viceversa, perhaps this distinction wouldn't be too important.

3. Since we already depend on `serde` (among many other libs), I'd rather want to avoid adding dep on another parser. However, giving it more thought and looking the small size of deps being added, I feel that this is a good price to pay for a big gain. The current one is a bit of a performance bottleneck. However, if we're adding a new signature parser, it should just replace the existing one and be used for ser/de everywhere.

jonathanhood:
Thanks for taking the time to review this!

> Having said that, it's not clear to me looking over the commits if this PR really does indeed implement the feature promised in the title: for example giving to_bytes_with_signature an i32 with signature v and it being able to handle it. Is that the case?

It can in some cases, but that is not the core feature of this work nor what was promised in the title. The purpose of this PR is to allow serialization and deserialization to `Value` directly. It does not reliably magic a concrete rust type into a dbus variant type signature, but it is definitely part of "the plan" to do this work.

> However, if we're adding a new signature parser, it should just replace the existing one and be used for ser/de everywhere.

I think that's a fine goal, but I also think the change can and should be incremental.

zeenix:
> It can in some cases, but that is not the core feature of this work nor what was promised in the title. The purpose of this PR is to allow serialization and deserialization to `Value` directly. It does not reliably magic a concrete rust type into a dbus variant type signature, but it is definitely part of "the plan" to do this work.

In that case, I find the titles a bit misleading. I'm sure that's not intentional but that's how they sounded like and got me too excited. :) Anyway, that's very easy to fix so kindly update the PR title and commit messages so it's clearer.

While I've nothing against incremental development (in fact I prefer it), I do also prefer each PR to be self-contained and be able to justify itself on its own.

I guess the selling point of this PR on its own could have been that now you can transform any type to/from Value that implements 'Type' (basically eliminating the need for further [trait constraints on properties](https://dbus2.github.io/zbus/client.html#trait-bounds-for-property-values) and using `Type` everywhere for all values). However, this will only work for `Into<Value>` since we can have a generic `impl<T: DynamicType> TryFrom<T> for Value` but not `impl<T: DynamicType> TryFrom<Value> for T` and hence not super useful either as it only covers half side of the story and using generic tranformations for one direction and not the other, would not make for a great API.

> > However, if we're adding a new signature parser, it should just replace the existing one and be used for ser/de everywhere.
> 
> I think that's a fine goal, but I also think the change can and should be incremental.

Same here about PR needing to be self-contained. I don't want 2 different parsers doing the same exact thing in the main branch. So please either replace the existing parsers in this PR completely or don't add the new parsers yet and use the existing one we already have for now. I leave it up to you which path you want to take.

zeenix:
@jonathanhood do you think you'll continue working on this? If not, I understand and still grateful for your contribution but kindly close the PR in that case. 🙏

jonathanhood:
Thanks for your time reviewing! Going to go ahead and close this. I'll keep our fork with these changes open in case they are useful to the next person who looks into these issues.

PR #796, opened 2024-05-17
feat: used new signature parser in dbus serde

jonathanhood:


jonathanhood:
This was raised on accident. Please ignore.

zeenix:
> This was raised on accident. Please ignore.

It's ok to make a mistake but please make sure it doesn't happen again.

PR #798, opened 2024-05-18
⬆️ zm,zv,zu: Update syn to v2.0.64

### LINKED ISSUE #401: Switch to syn2
Currently zbus/zvariant macros still depend on syn 1.0

zecakeh:
Built on top of and replaces #404.
Fixes #401.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,7 +261,7 @@ checksum = "3b43422f69d8ff38f95f1b2bb76517c91589a924d1559a0e935d7c8ce0274c11"
 dependencies = [
  "proc-macro2",
  "quote",
- "syn 2.0.49",
+ "syn 2.0.64",
 ]
 
 [[package]]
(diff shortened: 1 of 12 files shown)


bilelmoussaoui:
By using async-std (the async-attributes crate specifically) ends up pulling syn1... maybe it is worth investigating dropping the usage of async-std as is and use the sub-crates individually

zeenix:
> By using async-std

It's only a dev dep so no biggie.

Sophie already tried to remove this dep and I [agreed with her what needed to be done](https://github.com/dbus2/zbus/pull/531#issuecomment-1872930834) but then she dropped the ball. 🤷‍♂️

zeenix (changes requested):
Nice! Thanks for doing this work. Please also put the description to the commit. Then I can merge.

zecakeh:
Done.

zeenix (approved):

PR #800, opened 2024-05-19
🚑️ zm: Fix default value of `Interface::spawn_task_for_method`

### LINKED ISSUE #799: Regression - interface functions cannot run concurrently
4.2.1 has a bug that causes interface functions be unable to run concurrently. It happens whether the functions are on the same interface or not. It is a regression because it worked fine in 4.1.2. I don't know exactly which version it started on, but I heard on matrix that there was a refactor recently that could have done it.
I put repro instructions in the zbus_regression branch of the repo where I found it:
[https://github.com/dvhar/pikeru/tree/zbus_regression](https://github.com/dvhar/pikeru/tree/zbus_regression)

zeenix:
This fixes a regression introduced by 91abff76a964d9c2ba7907435886ee9810395d02, resulting in tasks not being spawned for methods by default.

Fixes #799.

### DIFF
--- a/zbus/src/lib.rs
+++ b/zbus/src/lib.rs
@@ -199,6 +199,7 @@ mod tests {
 
     use crate::utils::block_on;
     use enumflags2::BitFlags;
+    use event_listener::Event;
     use ntest::timeout;
     use test_log::test;
     use tracing::{debug, instrument, trace};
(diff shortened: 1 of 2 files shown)


elmarco:
oops, looks like I messed this up when I changed no_spawn for spawn =...

zeenix:
> oops, looks like I messed this up when I changed no_spawn for spawn =...

Yeah. I'll add testcase so it doesn't happen in the future.

PR #805, opened 2024-05-20
✨zv: auto resolve variant type signatures to/from concrete types

jonathanhood:


jonathanhood:
Sorry, did it again. Github's default is to raise a PR against master and this behavior keeps getting me.

zeenix:
> Sorry, did it again. Github's default is to raise a PR against master and this behavior keeps getting me.

I do understand but you are ending up unintentionally spamming. Please keep in mind if this happens again in **the near future**, I'll be forced to **temporarily** block you.

PR #815, opened 2024-05-28
🚑️ zb: Server should handle FDs received during handshake

### LINKED ISSUE #813: 4.2 fails with "Unexpected FDs during handshake" on 32 bit linux
Upgrade from 4.0.1 to 4.2.2 seems to make `zbus::ConnectionBuilder::unix_stream(unix_stream).p2p()` fail with `Handshake("Unexpected FDs during handshake")` in my 32 bit CI.

I can add more details and tests if the issue is not obvious.

zeenix:
If the client pipelines the handshake or is super fast to send the first message after the handshake, the server may receive FDs while still reading the handshake commands from the client. This is not possible to happen if the server is a bus impl but can happen in case of a pure p2p server.
    
We fix this by not erroring out on receiving FDs from the socket during the handshake but instead storing them in a list and processing them later when we receive/process the associated message.
    
Unfortunately, this requires an API break but given that there is no way to fix this without an API break and it's extremely unlikely that anyone is using/implementing these brand new low-level trait methods in question, directly, I think it's OK.
    
Fixes #813.


zbus/src/connection/socket/mod.rs
```diff
@@ -179,6 +181,26 @@ pub trait ReadHalf: std::fmt::Debug + Send + Sync + 'static {
...
+            let fields: crate::message::Fields<'_> = encoded_fields.deserialize()?.0;
+            let num_required_fds = match fields.get_field(crate::message::FieldCode::UnixFDs) {
+                Some(Field::UnixFDs(num_fds)) => *num_fds as usize,
+                _ => 0,
+            };
+            let pending = num_required_fds
+                .checked_sub(fds.len())
+                .ok_or_else(|| crate::Error::ExcessData)?;
+            let to_take = std::cmp::min(pending, already_received_fds.len());
+            fds.extend(already_received_fds.drain(..to_take));
```
zeenix, line 201: fds need to be prepended here. Marking this as draft until I sort this out.

Issue #643, opened 2024-02-21
⬆️ Update tokio-vsock to 0.5


zeenix:
`tokio-vsock::{ReadHalf,WriteHalf}` in 0.5 has lifetimes now and provides owned versions of them for folks who want them independent. Unless we support both 0.4 and 0.5 (ughh.), this will be an API break so likely I'll only do this at the next API break.

zeenix:
@elmarco given that I doubt anyone uses vsock support, I've reconsidered and I'm OK with updating tokio-vsock in 4.x. Of course this depends on whether or not you can finish #846 in a timely manner (making #830 obsolete).

plugwash:
I just patched zbus in Debian to build against tokio-vsock 0.5, our patch is at

https://salsa.debian.org/rust-team/debcargo-conf/-/blob/e09e68f2d7dcce1ab70e8d32b3224706cccbc8bb/src/zbus/debian/patches/tokio-vsock-0.5.patch

zeenix:
Thanks. Can you create a PR for this? Theoretically this is a breaking change but i think it's safe to assume it won't actually break anything for anyone and even if it does, they'll probably want to use the latest tokio-vsock themselves.

PR #839, opened 2024-06-12
Some dependencies updates

elmarco:


zeenix:
Thanks but did you see #636 and #643 ? The dependency updates have already been automated so if a dep hasn't been update, there is usually a reason.

elmarco:
No, I didn't see those. And it's hard to see it's an api break.. why?

zeenix:
> No, I didn't see those. And it's hard to see it's an api break.. why?

Because users of old tokio-vsock would not be able to build with newer zbus anymore, if they construct the connection using the socket. This can be avoided using a hack but given that I don't want hacks for vsock.

zeenix:
So can we close this then?

PR #844, opened 2024-06-14
Fix `println` in client.md

Jedsek:


zeenix (changes requested):
Thanks but this is not a mistake. This is a general term.

Jedsek:
@zeenix but... :)
image

Jedsek:
I blame me for not making it clear. :)

zeenix:
> @zeenix but... :)

Ah, there was a discrepancy.  

> I blame me for not making it clear. :)

No worries. I you had followed [our contributing guideline](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md), it would have encouraged you to write more descriptive/details commit log. Although I won't enforce that for such a simple change but it would be good to make it clear. :)

zeenix (approved):

Jedsek:
Thanks.

Issue #830, opened 2024-06-06
RFC: Drop vsock support


zeenix:
We added vsock support back a few years ago in anticipation of it being accepted in the [D-Bus spec and reference implementation](https://gitlab.freedesktop.org/dbus/dbus/-/merge_requests/200). However, this never happened and it's very unlikely to happen as the discussion has died out.

Moreover, I do not know if anyone is actually using it at all. Given that it's a maintenance burden and we've more work than developer/maintenance time available, I'm leaning towards removing this. The fact that we don't have any tests for this, means that we don't even know if this support actually works and when/if it breaks. We also sometimes have [issues with it](https://github.com/dbus2/zbus/pull/809#issuecomment-2129699731).

If you disagree, this is your chance to speak up.

zeenix:
@elmarco ^

elmarco:
I still wish some day a guest bus can be accessed from a host easily, or some guest services be on a VM host bus. Unfortunately, working on VM & desktop is not very popular.. Even ssh over vsock is very recent, thanks to Lennart systemd-ssh-proxy..

zeenix:
> I still wish some day a guest bus can be accessed from a host easily, or some guest services be on a VM host bus.

I wish that too.

> How bad is maintainance of the vsock feature?

Not a lot but given we've no testing for this in the CI, I don't even know if it actually works and wouldn't know if I (or anyone) breaks it.

zeenix:
@elmarco btw you didn't answer to most important question: are you relying on the vsock support of zbus?

elmarco:
> @elmarco btw you didn't answer to most important question: are you relying on the vsock support of zbus?

not atm

elmarco:
But there are some tests for vsock in CI already.

zeenix:
> But there are some tests for vsock in CI already.

Compile-only. It doesn't tell us the impl really works.

zeenix:
Thanks to @elmarco we now have tests for vsock that run in the CI (I hope?). So we can close this.

PR #846, opened 2024-06-17
add vsock connection tests

elmarco:
The test fails with tokio-vsock, which doesn't seem to send or flush the BEGIN command. More investigation needed.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -1728,6 +1728,51 @@ mod p2p_tests {
     ))]
     #[test]
     #[timeout(15000)]
+    fn vsock_connect() {
+        let _ = crate::utils::block_on(test_vsock_connect()).unwrap();
+    }
+
+    #[cfg(any(
+        all(feature = "vsock", not(feature = "tokio")),
+        feature = "tokio-vsock"
+    ))]
+    async fn test_vsock_connect() -> Result<(Connection, Connection)> {
+        let guid = Guid::generate();
+
+        #[cfg(all(feature = "vsock", not(feature = "tokio")))]
+        let listener = vsock::VsockListener::bind_with_cid_port(vsock::VMADDR_CID_LOCAL, u32::MAX)?;
+        #[cfg(feature = "tokio-vsock")]
+        let listener = tokio_vsock::VsockListener::bind(1, u32::MAX)?;
+
+        let addr = listener.local_addr()?;
+        let addr = format!("vsock:cid={},port={},guid={guid}", addr.cid(), addr.port());
+
+        let server = async {
+            #[cfg(all(feature = "vsock", not(feature = "tokio")))]
+            let server = crate::Task::spawn_blocking(move || listener.incoming().next(), "").await;
+            #[cfg(feature = "tokio-vsock")]
+            let server = listener.incoming().next().await;
+            Builder::vsock_stream(server.unwrap()?)
+                .server(guid)?
+                .p2p()
+                .auth_mechanism(AuthMechanism::Anonymous)
+                .build()
+                .await
+        };
+
+        let client = crate::connection::Builder::address(addr.as_str())?
+            .p2p()
+            .build();
+
+        futures_util::try_join!(server, client)
+    }
+
+    #[cfg(any(
+        all(feature = "vsock", not(feature = "tokio")),
+        feature = "tokio-vsock"
+    ))]
+    #[test]
+    #[timeout(15000)]
     fn vsock_p2p() {
         crate::utils::block_on(test_vsock_p2p()).unwrap();
     }
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -1785,6 +1785,7 @@ mod p2p_tests {
             Builder::vsock_stream(client).p2p().build(),
         )
     }
+
     #[cfg(any(unix, not(feature = "tokio")))]
     #[test]
     #[timeout(15000)]
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -1722,14 +1722,12 @@ mod p2p_tests {
         )
     }
 
-    // Compile-test only since we don't have a VM setup to run this with/in.
     #[cfg(any(
         all(feature = "vsock", not(feature = "tokio")),
         feature = "tokio-vsock"
     ))]
     #[test]
     #[timeout(15000)]
-    #[ignore]
     fn vsock_p2p() {
         crate::utils::block_on(test_vsock_p2p()).unwrap();
     }
@@ -1749,7 +1747,8 @@ mod p2p_tests {
     async fn vsock_p2p_pipe() -> Result<(Connection, Connection)> {
         let guid = Guid::generate();
 
-        let listener = vsock::VsockListener::bind_with_cid_port(vsock::VMADDR_CID_ANY, 42).unwrap();
+        let listener =
+            vsock::VsockListener::bind_with_cid_port(vsock::VMADDR_CID_LOCAL, u32::MAX).unwrap();
         let addr = listener.local_addr().unwrap();
         let client = vsock::VsockStream::connect(&addr).unwrap();
         let server = listener.incoming().next().unwrap().unwrap();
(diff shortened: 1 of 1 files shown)


zeenix:
We already have an automatically created pr for nix update and cargo.lock isn't really something that affects users. Please don't touch it directly. Bots take care of it.

elmarco:
@zeenix  any idea why tokio-vsock doesn't receive BEGIN.. or how to debug that.. I am bit stuck.

zbus/src/connection/mod.rs
```diff
@@ -1534,6 +1534,81 @@ mod p2p_tests {
...
+        all(feature = "vsock", not(feature = "tokio")),
+        feature = "tokio-vsock"
+    ))]
+    #[test]
+    #[timeout(15000)]
+    fn vsock_connect() {
+        crate::utils::block_on(test_vsock_connect()).unwrap();
+    }
+
+    #[cfg(all(feature = "vsock", not(feature = "tokio")))]
```
zeenix, line 1547: most of the code for both `vsock` and `vsock-tokio` seems identical. Best to have one function.

zbus/src/connection/mod.rs
```diff
@@ -1534,6 +1534,81 @@ mod p2p_tests {
...
+    #[cfg(feature = "tokio-vsock")]
+    async fn test_vsock_connect() -> Result<()> {
+        let guid = Guid::generate();
+
+        let listener = tokio_vsock::VsockListener::bind(1, u32::MAX).unwrap();
+        let addr = listener.local_addr().unwrap();
+        let addr = format!("vsock:cid={},port={},guid={guid}", addr.cid(), addr.port());
+
+        let server = std::thread::spawn(move || {
+            crate::utils::block_on(async {
```
zeenix, line 1589: why? spawn a task. might be the problem we're seeing (would be strange though). See [here](https://github.com/dbus2/zbus/blob/d87d4fc1dfa5a20be91bc7befe39aa73606a7db4/zbus/src/lib.rs#L1100).

Although I'd [join the futures](https://github.com/dbus2/zbus/blob/99e48fdd9d975e30331803f8bca10577d1024b93/zbus/src/connection/mod.rs#L1500).

zeenix:
> @zeenix any idea why tokio-vsock doesn't receive BEGIN.. or how to debug that.. I am bit stuck.

Seems to be only an issue for `tokio-vsock` (`vsock` works). I left a comment for improvement and that might help but otherwise, I'm stumped too. Maybe a bug in `tokio-vsock`?

zeenix:
@elmarco do you see this PR going forward?

zbus/src/connection/mod.rs
elmarco: yes it works without extra thread, I'll update the MR.

zbus/Cargo.toml
```diff
@@ -133,6 +133,7 @@ tracing-subscriber = { version = "0.3.18", features = [
   "ansi",
 ], default-features = false }
 tempfile = "3.10.1"
+smol = "2.0"
```
zeenix, line 136: What's this for?

zeenix: We already have runtime abstractions that are already in use by tests. Spawning is part of them.

elmarco: But you need the executor, which is part of Connection.

zeenix: So? What's the issue?

zeenix: You can always add API to the abstraction. In any case, we definitely don't need another runtime, when we've one guaranteed to be available.

elmarco: There is no connection before spawning the task. smol is only added as dev-dependency. I welcome alternative solutions.

zeenix: > There is no connection before spawning the task

As I wrote:

> You can always add API to the abstraction. 

I just checked. There is `Executor::new` that's currently `pub(crate)`. Just make it `pub` and `#[doc(hidden)]`.

zeenix: > smol is only added as dev-dependency.

I don't mind adding the dev-dep but only if there is a justification to add it. There isn't any AFAICT.

elmarco: There is something inelegant about using block_on and a custom executor to run some standard async code test.

zeenix: > a custom executor

? It's just a think wrapper.

zbus/src/connection/mod.rs
```diff
@@ -1559,14 +1559,79 @@ mod p2p_tests {
...
+                    let _conn = crate::connection::Builder::address(addr.as_str())
+                        .unwrap()
+                        .p2p()
+                        .build()
+                        .await
+                        .unwrap();
+
+                    let _ = server.await;
+                })
+                .await;
```
zeenix, line 1625: This could be even simpler and more consistent with other tests, if you'd make use of `futures::*join`, see [here](https://github.com/dbus2/zbus/blob/main/zbus/src/connection/mod.rs#L1693).

zbus/src/connection/mod.rs
```diff
@@ -1620,6 +1689,7 @@ mod p2p_tests {
             Builder::vsock_stream(client).p2p().build(),
         )
     }
+
```
zeenix, line 1692: irrelevant change.

zbus/src/connection/mod.rs
```diff
@@ -1559,14 +1559,79 @@ mod p2p_tests {
...
+                    );
+                    #[cfg(feature = "tokio-vsock")]
+                    let server = executor.spawn(
+                        async {
+                            let server = listener.incoming().next().await.unwrap().unwrap();
+                            Builder::vsock_stream(server)
+                                .server(guid)
+                                .unwrap()
+                                .p2p()
+                                .auth_mechanism(AuthMechanism::Anonymous)
```
zeenix, line 1608: vsock can only do anonymous? :thinking:

zeenix (changes requested):
LGTM otherwise.

zbus/src/connection/mod.rs
elmarco: it could do cookie, I suppose with shared directory. external is not really possible, SO_PEERCRED is still lacking

zeenix: hmm.. ok.

zeenix:
@elmarco if you can please complete this, I can close #830.

elmarco:
looking at it

zbus/src/connection/mod.rs
elmarco: done

zbus/Cargo.toml
```diff
@@ -140,6 +140,7 @@ tracing-subscriber = { version = "0.3.18", features = [
 ], default-features = false }
 tempfile = "3.10.1"
 criterion = "0.5.1"
+blocking = "1.6.1"
```
zeenix, line 143: Is this new? I don't recall see thing change before. 

> vsock crate is not async, but the following added p2p test needs to have
both client&server side async connect to each other.

We've abstractions for both running async in sync (block_on) and sync in async (spawn_blocking). Please make use of those instead. The spawn_blocking actually uses `blocking` when `async-io` (default) is enabled. `spawn_blocking` is currently `pub(crate)` though but feel free to change that to `pub`.

zeenix (changes requested):

zbus/Cargo.toml
elmarco: done

zbus/src/connection/mod.rs
```diff
@@ -1722,6 +1722,52 @@ mod p2p_tests {
...
     }
 
+    #[cfg(any(
+        all(feature = "vsock", not(feature = "tokio")),
+        feature = "tokio-vsock"
+    ))]
+    #[test]
+    #[timeout(15000)]
+    fn vsock_connect() {
+        crate::utils::block_on(async {
```
zeenix, line 1732: Last nitpick, I promise. :) If you look at other tests, I've been splitting the async part into a separate function and then block_on on the whole function call. That avoid unnecessary indentation level, that always makes the code a bit harder to read.

zeenix (changes requested):
One last thing and we're good to go. :+1:

zeenix (approved):

PR #847, opened 2024-06-17
⬆️ : Bump nix to v0.29

elmarco:
supersedes  #809


zeenix:
Thanks. Keep in mind that #809 mainly fails because vsock still grabs old nix so users of vsock will end up with both version but since there aren't any users yet that we know of, it's not an issue in practice. :)

zeenix (approved):

PR #859, opened 2024-06-22
💩 zb: Workaround for xdg-dbus-proxy's monotonic serial requirement

zeenix:
While [the underlying issue in xdg-dbus-proxy][1] has already been fixed, it will take some time before the fix is released and is widely available. So let's add a workaround for now that makes message creation and sending an atomic operation, using an async semaphore if flatpak portal is detected.

[1]: https://github.com/flatpak/xdg-dbus-proxy/issues/46


zeenix:
Converting to a draft to avoid accidental merge before it can be tested with different flatpak apps.

haecker-felix:
Can confirm that this PR fixes the the "Invalid client serial" issue for Key Rack / oo7 :+1:

PR #867, opened 2024-06-26
🐛 zbus: Fix deadlock during name registration

dvrabel-ntnx:
When registering a name, zbus will deadlock if it receives many messages before the RequestName reply. This occurs because:

1. The RequestName() call means the dbus server can send messages for that name _before_ it sends the RequestName() reply.

2. While name registration is in progress the registered names lock is held.

3. These messages are queued on a sender channel.

4. The queued messages are not dispatched because the object server task blocks waiting for the registered names lock (see 2 above).

5. This sender channel can become full, blocking the receiver task when it attempts to enqueue on a full channel.

6. If the receiver task blocks it does not receive the RequestName() reply and name registration.

7. Since name registration never completes the registered names lock is never released, none of the pending messages are dispatched, the sender remains full and the receiver never receives any more messages.

This deadlock is most likely to occur if the dbus service is a dbus-activated systemd service. The dbus server will hold calls while the service is starting, releasing them when RequestName() is called but before the RequestName() reply is sent.

Avoid this deadlock by not attempting to take the registered names lock in the object server task. The check that the destination is a registered name is unnecessary as the dbus server will not send messages for unregistered names.

This allows the object server to dispatch messages while name registration is in progress so the sender does not become full and thus the receiver does not block.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -974,20 +974,25 @@ impl Connection {
                     }) {
                         if let Some(conn) = weak_conn.upgrade() {
                             let hdr = msg.header();
-                            match hdr.destination() {
-                                // Unique name is already checked by the match rule.
-                                Some(BusName::Unique(_)) | None => (),
-                                Some(BusName::WellKnown(dest)) => {
-                                    let names = conn.inner.registered_names.lock().await;
-                                    // destination doesn't matter if no name has been registered
-                                    // (probably means name it's registered through external means).
-                                    if !names.is_empty() && !names.contains_key(dest) {
-                                        trace!(
-                                            "Got a method call for a different destination: {}",
-                                            dest
-                                        );
+                            // If we're connected to a bus, skip the destination check as the
+                            // server will only send us method calls destined to us.
+                            if !conn.is_bus() {
+                                match hdr.destination() {
+                                    // Unique name is already checked by the match rule.
+                                    Some(BusName::Unique(_)) | None => (),
+                                    Some(BusName::WellKnown(dest)) => {
+                                        let names = conn.inner.registered_names.lock().await;
+                                        // destination doesn't matter if no name has been registered
+                                        // (probably means the name is registered through external
+                                        // means).
+                                        if !names.is_empty() && !names.contains_key(dest) {
+                                            trace!(
+                                                "Got a method call for a different destination: {}",
+                                                dest
+                                            );
 
-                                        continue;
+                                            continue;
+                                        }
                                     }
                                 }
                             }


zbus/src/connection/mod.rs
```diff
@@ -974,23 +974,6 @@ impl Connection {
                     }) {
                         if let Some(conn) = weak_conn.upgrade() {
                             let hdr = msg.header();
-                            match hdr.destination() {
-                                // Unique name is already checked by the match rule.
-                                Some(BusName::Unique(_)) | None => (),
-                                Some(BusName::WellKnown(dest)) => {
-                                    let names = conn.inner.registered_names.lock().await;
```
zeenix, line 981: 1. I'm not sure I follow. This lock is being help for a very short time in this block only. It's dropped almost immediately. So how does it create a deadlock? :thinking: 
2. We've the p2p case and bus implementation (busd) cases as well and that's what this logic was added for. The names are registered internally in those cases and there is external broker involved to help us.

zeenix: Never mind about 1.  Should have read your description better before replying here. :)

As for 2. That's still a concern:

> Avoid this deadlock by not attempting to take the registered names lock in the object server task. The check that the destination is a registered name is unnecessary as the dbus server will not send messages for unregistered names.

This isn't true, as I explained above. In fact, I deliberately added this check for busd.

Gotta figure out another solution. Btw kudos for adding a very detailed description on the PR description and commit message. In the future, I'd also request you add emoji prefix to the commit as per [our contribution guidelines](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-bug-reports-and-feature-requests) (feel free to skip it if you don't think you'll be contributing in the future again).

dvrabel-ntnx: 1. The lock is being held by request_name_with_flags() while waiting for the RequestName() reply. The object server is waiting for this lock, trying to process messages before the RequestName() reply is received.

2. You're expecting messages to be received for an unregistered name as you explicitly want to discard them? This isn't some "just in case" defensive check?

zeenix: > 2\. You're expecting messages to be received for an unregistered name as you explicitly want to discard them? 

I need ObjectServer to ignore the message in this case because otherwise, it can respond to message that are destined for another connection (typically this would mean the ObjectServer sending back an error because it will not find the destination interface+method). The main task of a bus implementation is to forward messages from one connection to another and that's why this is needed.

> This isn't some "just in case" defensive check?

Unfortunately, not.

zeenix: I moved to a different architecture in busd and currently this is probably **not** required by busd at least. However, I have been thinking of reverting back to the old architecture and that would mean we need this again.

We do have a `bus-impl` feature (and busd enables it), so what we can do is to condition this on `#cfg[feature = "bus-impl"]` and perhaps also if it's a p2p server?

dvrabel-ntnx: How about dropping the registered_name lock while waiting for the reply? Something like (the untested/incomplete).

In the object server when looking for the registered name it could discard the message if it's pending and we're not a simple service?

`diff --git a/zbus/src/connection/mod.rs b/zbus/src/connection/mod.rs
index 9a94d7fd..fa29fc96 100644
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -5,7 +5,7 @@ use event_listener::{Event, EventListener};
 use ordered_stream::{OrderedFuture, OrderedStream, PollResult};
 use static_assertions::assert_impl_all;
 use std::{
-    collections::HashMap,
+    collections::{HashMap, hash_map::Entry},
     io::{self, ErrorKind},
     num::NonZeroU32,
     ops::Deref,
@@ -601,18 +601,23 @@ impl Connection {
         let well_known_name = well_known_name.try_into().map_err(Into::into)?;
         // We keep the lock until the end of this function so that the (possibly) spawned task
         // doesn't end up accessing the name entry before it's inserted.
-        let mut names = self.inner.registered_names.lock().await;
+        {
+            let mut names = self.inner.registered_names.lock().await;
 
-        match names.get(&well_known_name) {
-            Some(NameStatus::Owner(_)) => return Ok(RequestNameReply::AlreadyOwner),
-            Some(NameStatus::Queued(_)) => return Ok(RequestNameReply::InQueue),
-            None => (),
-        }
+            match names.get(&well_known_name) {
+                Some(NameStatus::Pending) => return Err(Error::NameTaken),
+                Some(NameStatus::Owner(_)) => return Ok(RequestNameReply::AlreadyOwner),
+                Some(NameStatus::Queued(_)) => return Ok(RequestNameReply::InQueue),
+                None => (),
+            }
 
-        if !self.is_bus() {
-            names.insert(well_known_name.to_owned(), NameStatus::Owner(None));
+            if !self.is_bus() {
+                names.insert(well_known_name.to_owned(), NameStatus::Owner(None));
 
-            return Ok(RequestNameReply::PrimaryOwner);
+                return Ok(RequestNameReply::PrimaryOwner);
+            }
+
+            names.insert(well_known_name.to_owned(), NameStatus::Pending);
         }
 
         let dbus_proxy = fdo::DBusProxy::builder(self)
@@ -674,6 +679,7 @@ impl Connection {
         } else {
             None
         };
+        let mut names = self.inner.registered_names.lock().await;
         let status = match reply {
             RequestNameReply::InQueue => {
                 let weak_conn = WeakConnection::from(self);
@@ -727,9 +733,10 @@ impl Connection {
             RequestNameReply::Exists => return Err(Error::NameTaken),
         };
 
-        names.insert(well_known_name.to_owned(), status);
-
-        Ok(reply)
+        match names.entry(well_known_name.to_owned()).and_modify(|s| *s = status) {
+            Entry::Vacant(_) => Err(Error::Failure("cancelled".to_string())),
+            Entry::Occupied(_) => Ok(reply),
+        }
     }
 
     /// Deregister a previously registered well-known name for this service on the bus.
@@ -1261,6 +1268,8 @@ impl From<&Connection> for WeakConnection {
 
 #[derive(Debug)]
 enum NameStatus {
+    // Waiting for a reply.
+    Pending,
     // The task waits for name lost signal if owner allows replacement.
     Owner(#[allow(unused)] Option<Task<()>>),
     // The task waits for name acquisition signal.
`

zeenix: > How about dropping the registered_name lock while waiting for the reply? Something like (the untested/incomplete).

Looking through the code, that seems good to me if it works. Thanks.

dvrabel-ntnx: I've taken a more detailed look at how p2p connections are used in busd and there's seems to be no real difference between these connections and regular connections to a bus wrt name registration and method call reception/dispatch.

Thus I think the checks for the destination is _always_ redundant.

busd as an completely separate name registry for peers registering names with it -- this is not shared with the registered_name hashmap in zbus's Connection.

dvrabel-ntnx: The current implementation of the check is also buggy (or at least has inconsistent or counter-intuitive behaviour) because it:

1. Accepts any method if there are no registered name.
2. Accepts methods if the name registration is only queued and the service is not yet the owner.

These could possibly be fixed but I also think that is impossible to make a destination check that is correct. Consider the following sequences (service on the LHS, dbus server on the RHS).

```
<-- Call(1) -- [reject]
-- RequestName -->
<-- Call(2) -- [accept]
<-- RequestName reply (Owner) --
<-- Call(3) -- [accept]
```
and
```
-- RequestName -->
<-- Call(4) -- [reject]
<-- RequestName reply (AlreadyOwned) --
```
It is impossible to correctly accept Call(2) and correctly reject Call(4) because the necessary information (the RequestName reply) has not yet been seen.

zeenix: > I've taken a more detailed look at how p2p connections are used in busd and there's seems to be no real difference between these connections and regular connections to a bus wrt name registration and method call reception/dispatch.

I already explained that part above:

> I moved to a different architecture in busd and currently this is probably not required by busd at least. However, I have been thinking of reverting back to the old architecture and that would mean we need this again.

> busd has a completely separate name registry for peers registering names with it -- this is not shared with the registered_name hashmap in zbus's Connection.

You're looking at a completely wrong place and in the wrong revision. The relevant part is bus registering its own name on its connection. Since you really want to know and won't take my word for it: busd used to add the freedesktop standard interface directly on each p2p connection so any method calls from the other side will be answered by the object server on these connections, regardless of whether they were destined for the bus or another peer. Obviously this meant method calls to other peers won't go through but just get rejected.

If you are still curious, see the code before any self-dial connection was introduced.

> The current implementation of the check is also buggy (or at least has inconsistent or counter-intuitive behaviour) because it:
> 
> 1. Accepts any method if there are no registered name.
> 2. Accepts methods if the name registration is only queued and the service is not yet the owner.

I don't really think that's going to cause any problems in practice, especially the first one since we're just not doing the check if we don't have a name registered. But as I wrote before, we could make the check specific to p2p/bus-impl case (will need to document it though).

dvrabel-ntnx: Apologies, I misinterpreted one of your earlier comments as saying the changes were on busd main, but not yet released.

I've changed the PR to skip the destination checking if `conn.is_bus()`.

zeenix: > > I've changed the PR to skip the destination checking if `conn.is_bus()`.
> 
> Cool but that is not correct. `is_bus` is a client-side check if the client is connected to a bus on the other end.

Oh nm. I had the logic reversed in my head. :) Basically, you're only enabling the destination check if the conn is a p2p conn, right? Then that's correct AFAICT. :+1:

zbus/src/connection/mod.rs
```diff
@@ -974,20 +974,24 @@ impl Connection {
...
-                                    // (probably means name it's registered through external means).
-                                    if !names.is_empty() && !names.contains_key(dest) {
-                                        trace!(
-                                            "Got a method call for a different destination: {}",
-                                            dest
-                                        );
-
-                                        continue;
+                            // If we're connected to a bus, skip the destination check as the
+                            // server will only send method calls for us.
```
zeenix, line 978: Just a nitpick on the language:

```suggestion
                            // server will only send us method calls that are destined to us.
```

zbus/src/connection/mod.rs
dvrabel-ntnx: The code is:
```
if !conn.is_bus() {
    match hdr.destination() {
        // ...
}
```
i.e., If the peer is a service (not a bus), then check the destination. Which I think is what you want?

zeenix (changes requested):
LGTM otherwise.

zbus/src/connection/mod.rs
zeenix: > i.e., If the peer is a service (not a bus), then check the destination. Which I think is what you want?

Yeah, sorry. Please see my previous comment.

zeenix (approved):
Nice!

PR #899, opened 2024-07-11
✨ zv: support kebab-case in rename_all

bachp:
Also add some tests for case conversion
to ensure renamings are consistent

### DIFF
--- a/zvariant_utils/src/case.rs
+++ b/zvariant_utils/src/case.rs
@@ -1,6 +1,6 @@
 //! Contains utilities used to convert strings between different cases.
 
-/// Convert to pascal or camel case, assuming snake case.
+/// Convert to pascal or camel case, assuming snake or kebab case.
 ///
 /// If `s` is already in pascal or camel case, should yield the same result.
 pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
@@ -8,7 +8,7 @@ pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
     let mut capitalize = is_pascal_case;
     let mut first = true;
     for ch in s.chars() {
-        if ch == '_' {
+        if ch == '_' || ch == '-' {
             capitalize = true;
         } else if capitalize {
             result.push(ch.to_ascii_uppercase());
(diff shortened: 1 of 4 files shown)


zvariant_utils/src/case.rs
```diff
@@ -26,16 +26,60 @@ pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
...
         snake.push(ch.to_ascii_lowercase());
     }
     snake
 }
+
+/// Convert to kebab case.
+///
+/// If `s` is already in kebab case, should yield the same result.
+pub fn kebab_case(s: &str) -> String {
+    snake_case(s).replace('_', "-")
```
zeenix, line 47: This commit ends up parsing and modifying the string 3 times. I'd rather not do that, even if it means a bit of duplication. This is macro code so binary size isn't a concern at least.

zeenix (changes requested):
LGTM otherwise. Thanks!

zvariant_utils/src/case.rs
bachp: My assumtion was that this code it is not executed on the hot path. So I didn't pay attention to makeing it fast but rather to make it simple.

Your comment that this is macro code seems to confirm this.

What is your concern with modifying the string multiple times?

zeenix: > My assumtion was that this code it is not executed on the hot path. So I didn't pay attention to makeing it fast but rather to make it simple.

Yeah, it wouldn't be an issue if there was some big benefit from it but I don't see any. I don't think it's simpler with the current approach. It's not more complicated/less-readable with single pass.

zvariant_utils/src/case.rs
```diff
@@ -26,16 +26,60 @@ pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
     result
 }
 
-/// Convert to snake case, assuming pascal case.
```
zeenix, line 29: You're passing `true` to `pascal_or_camel_case`, which tells it that passed string is pascal case, so isn't pascal case still assumed?

zvariant_utils/src/case.rs
```diff
@@ -8,7 +8,7 @@ pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
     let mut capitalize = is_pascal_case;
     let mut first = true;
     for ch in s.chars() {
-        if ch == '_' {
+        if ch == '_' || ch == '-' {
```
zeenix, line 11: This makes the function, also handle `kebab-case` so docs should be updated.

zvariant_utils/src/case.rs
zeenix: Oh, I got confused because of multiple case changes. Still, this function now assume pascal or came case now, doesn't it?

zvariant_utils/src/case.rs
```diff
@@ -26,16 +26,60 @@ pub fn pascal_or_camel_case(s: &str, is_pascal_case: bool) -> String {
...
 }
 
-/// Convert to snake case, assuming pascal case.
+/// Convert to snake case.
 ///
 /// If `s` is already in snake case, should yield the same result.
 pub fn snake_case(s: &str) -> String {
     let mut snake = String::new();
-    for ch in s.chars() {
+    for ch in pascal_or_camel_case(s, true).chars() {
```
zeenix, line 34: I'm not sure I'm able to follow the logic here. `pascal_or_camel_case` takes snake_case string so why would we call it on the input to this function, that's supposed to create snake case?

I think not doing multiple conversions will help with readability here.

zvariant_utils/src/case.rs
bachp: No not really anymore, it actually just converts the input to PascalCase  using `pascal_or_camel_case(s, true)`. So it doesn't matter what the input is anymore.

zvariant_utils/src/case.rs
bachp: Calling `pascal_or_camel_case` just ensures that it processes PascalCase instead of assuming Pascal case, this is also why I removed the comment.

zvariant_utils/src/case.rs
bachp: You are right I will update this.

zvariant_utils/src/case.rs
zeenix: So I can pass it a string in any format and it can handle it? `pascal_or_camel_case` itself can't handle all cases so I'm even more confused now. :thinking:

zvariant_utils/src/case.rs
zeenix: Thanks for explaining! Yeah, now I think I get it. :laughing: However, I really think it would be easier to follow for a dumb person like myself to follow the code if each conversion did just one conversion directly.

bachp: I re worked the functions and made it work similar to the `pascal_or_camal_case`. And don't have them depend on each other.

zvariant_utils/src/case.rs
bachp: @zeenix The function is now more explicit. Let me know if this better fits.

zvariant_utils/src/case.rs
bachp: I update the comments. But basicaly yes all the functions can handle all forms of valid casing. See the test cases.

zeenix (approved):
Cool, thanks! It's so much easier to follow now. :) Although, I think we should change the API to only expose `snake_case` and `kebab_case` methods in the public API that are thin wrappers around `snake_or_kebab_case`, which isn't then exposed anymore. Same thing for `pascal_or_camel_case`. However, this can be done in an separate PR and not necessarily by you. :)

Also for future reference, I'd have split this commit into 3: the new zvariant_utils API, use of it in zbus_macros and finally the new tests.

PR #914, opened 2024-07-25
🐛 zm: do not invert spawn property value

elmarco:


zeenix:
LGTM

zeenix (approved):

zbus/src/lib.rs
```diff
@@ -913,7 +913,7 @@ mod tests {
             "org.zbus.Issue260",
         )
         .await?
-        .call("Whatever", &())
+        .call::<_, _, ()>("Whatever", &())
```
zeenix, line 916: It's best to keep the `clippy::` prefix in the subject or mentioned in the description of the commit so the context is clear. I generally tend to add the clipping warning/error message to the description even so it's clear from the commit itself, what the change is all about.

zeenix: hmm.. actually I can't find this lint [in clippy](https://rust-lang.github.io/rust-clippy/master/index.html#/). It's not clippy? This would be why it's good to add details to the commit.

zbus_macros/tests/tests.rs
```diff
@@ -109,6 +109,7 @@ fn test_proxy() {
 
 #[test]
 fn test_derive_error() {
+    #[allow(unused)]
```
zeenix, line 112: this is a compile-only test so best to [add `ignore` annotation on it](https://doc.rust-lang.org/book/ch11-02-running-tests.html#ignoring-some-tests-unless-specifically-requested).

zvariant/src/array.rs
```diff
@@ -67,7 +67,7 @@ impl<'a> Array<'a> {
     /// Get the value at the given index.
     pub fn get<V>(&'a self, idx: usize) -> Result<Option<V>>
     where
-        V: ?Sized + TryFrom<&'a Value<'a>>,
+        V: TryFrom<&'a Value<'a>>,
```
zeenix, line 70: > warning: `?Sized` bound is ignored because of a `Sized` requirement

Where is the `Sized` requirement coming from? Please copy&past the full warning/error so it's clear.

Protip: best to also specify the syntax as `rust` even if the error/warning isn't exactly all Rust. The syntax highlighting still makes the message much nicer/easier to read.

Cargo.toml
```diff
@@ -10,3 +10,7 @@ members = [
     "zbus_xmlgen",
 ]
 resolver = "2"
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "warn", check-cfg = ['cfg(tokio_unstable)'] }
```
zeenix, line 15: would be nice to have more detail in the commit log, especially what `check-cfg` is and why it needs to say `tokio_unstable`.

zbus/src/lib.rs
elmarco: ok, added compiler warning

zbus_macros/tests/tests.rs
elmarco: ok, added

zvariant/src/array.rs
elmarco: `TryFrom` has the bound `Sized`, update commit comment with clippy warning

Cargo.toml
elmarco: ok

zbus/src/connection/mod.rs
```diff
@@ -838,7 +838,7 @@ impl Connection {
     /// ```
     /// # // Disable on windows because somehow it triggers a stack overflow there:
     /// # // https://gitlab.freedesktop.org/zeenix/zbus/-/jobs/34023494
-    /// # #[cfg(not(target_os = "unix"))]
+    /// # #[cfg(not(target_os = "windows"))]
```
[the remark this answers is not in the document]
elmarco: ok, let's drop the condition

zbus/src/lib.rs
zeenix: Cool but in the future, please mark it as the verbatim output, just like you did for the clippy warning in c1789b6cbdbef1fc59d750e1ee262055c97b81ca.

PR #921, opened 2024-07-29
✨ zb: Add Connection::graceful_shutdown

### LINKED ISSUE #309: Support for a shutdown signal

cjriches:
Fixes #309.

I ended up naming it `Connection::graceful_shutdown` since `shutdown` felt too ambiguous with `close` to me. Happy to change if desired.

The included test is a little more tokio-integrated than most in the project, but since tokio was already a dev-dependency I'm hoping this is fine. I'm not aware of a different/better way to assert that an async function *doesn't* return.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -75,6 +75,17 @@ pub(crate) struct ConnectionInner {
 
     object_server: OnceLock<blocking::ObjectServer>,
     object_server_dispatch_task: OnceLock<Task<()>>,
+
+    drop_event: Event,
+}
+
+impl Drop for ConnectionInner {
+    fn drop(&mut self) {
+        // Notify anyone waiting that the connection is going away. Since we're being dropped, it's
+        // not possible for any new listeners to be created after this notification, so this is
+        // race-free.
+        self.drop_event.notify(usize::MAX);
+    }
 }
 
 type Subscriptions = HashMap<OwnedMatchRule, (u64, InactiveReceiver<Result<Message>>)>;
@@ -110,6 +121,13 @@ pub(crate) type MsgBroadcaster = Broadcaster<Result<Message>>;
 /// respectively.
 ///
 /// For sending messages you can either use [`Connection::send`] method.
+///
+/// To gracefully close a connection while waiting for any outstanding method calls to complete,
+/// use [`Connection::graceful_shutdown`]. To immediately close a connection in a way that will
+/// disrupt any outstanding method calls, use [`Connection::close`]. If you do not need the
+/// shutdown to be immediate and do not care about waiting for outstanding method calls, you can
+/// also simply drop the `Connection` instance, which will act similarly to spawning
+/// `graceful_shutdown` in the background.
 ///
 /// [method calls]: struct.Connection.html#method.call_method
 /// [signals]: struct.Connection.html#method.emit_signal
@@ -1159,6 +1177,7 @@ impl Connection {
                 msg_receiver,
                 method_return_receiver,
                 registered_names: Mutex::new(HashMap::new()),
+                drop_event: Event::new(),
             }),
         };
 
@@ -1215,6 +1234,55 @@ impl Connection {
             .close()
             .await
             .map_err(Into::into)
+    }
+
+    /// Gracefully close the connection, waiting for all other references to be dropped.
+    ///
+    /// This will not disrupt any incoming or outgoing method calls, and will await their
+    /// completion.
+    ///
+    /// # Caveats
+    ///
+    /// * This will not prevent new incoming messages from keeping the connection alive (and
+    ///   indefinitely delaying this method's completion).
+    ///
+    /// * The shutdown will not complete until the underlying connection is fully dropped, so beware
+    ///   of deadlocks if you are holding any other clones of this `Connection`.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use std::error::Error;
+    /// # use zbus::connection::Builder;
+    /// # use zbus::interface;
+    /// #
+    /// # struct MyInterface;
+    /// #
+    /// # #[interface(name = "foo.bar.baz")]
+    /// # impl MyInterface {
+    /// #     async fn do_thing(&self) {}
+    /// # }
+    /// #
+    /// # #[tokio::main]
+    /// # async fn main() -> Result<(), Box<dyn Error>> {
+    /// let conn = Builder::session()?
+    ///     .name("foo.bar.baz")?
+    ///     .serve_at("/foo/bar/baz", MyInterface)?
+    ///     .build()
+    ///     .await?;
+    ///
+    /// # let some_exit_condition = std::future::ready(());
+    /// some_exit_condition.await;
+    ///
+    /// conn.graceful_shutdown().await;
+    /// #
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn graceful_shutdown(self) {
+        let listener = self.inner.drop_event.listen();
+        drop(self);
+        listener.await;
     }
 
     pub(crate) fn init_socket_reader(
@@ -1303,6 +1371,7 @@ mod tests {
     use super::*;
     use crate::fdo::DBusProxy;
     use ntest::timeout;
+    use std::{pin::pin, time::Duration};
     use test_log::test;
 
     #[cfg(windows)]
@@ -1369,6 +1438,100 @@ mod tests {
         // Let's still make sure the name is gone.
         let name_has_owner = dbus.name_has_owner(name.try_into().unwrap()).await.unwrap();
         assert!(!name_has_owner);
+    }
+
+    #[tokio::test(start_paused = true)]
+    #[timeout(15000)]
+    async fn test_graceful_shutdown() {
+        // If we have a second reference, it should wait until we drop it.
+        let connection = Connection::session().await.unwrap();
+        let clone = connection.clone();
+        let mut shutdown = pin!(connection.graceful_shutdown());
+        // Due to start_paused above, tokio will auto-advance time once the runtime is idle.
+        // See https://docs.rs/tokio/latest/tokio/time/fn.pause.html.
+        tokio::select! {
+            _ = tokio::time::sleep(Duration::from_secs(u64::MAX)) => {},
+            _ = &mut shutdown => {
+                panic!("Graceful shutdown unexpectedly completed");
+            }
+        }
+
+        drop(clone);
+        shutdown.await;
+
+        // An outstanding method call should also be sufficient to keep the connection alive.
+        struct GracefulInterface {
+            method_called: Event,
+            wait_before_return: Option<EventListener>,
+            announce_done: Event,
+        }
+
+        #[crate::interface(name = "dev.peelz.TestGracefulShutdown")]
+        impl GracefulInterface {
+            async fn do_thing(&mut self) {
+                self.method_called.notify(1);
+                if let Some(listener) = self.wait_before_return.take() {
+                    listener.await;
+                }
+                self.announce_done.notify(1);
+            }
+        }
+
+        let method_called = Event::new();
+        let method_called_listener = method_called.listen();
+
+        let trigger_return = Event::new();
+        let wait_before_return = Some(trigger_return.listen());
+
+        let announce_done = Event::new();
+        let done_listener = announce_done.listen();
+
+        let interface = GracefulInterface {
+            method_called,
+            wait_before_return,
+            announce_done,
+        };
+
+        let name = "dev.peelz.TestGracefulShutdown";
+        let obj = "/dev/peelz/TestGracefulShutdown";
+        let connection = Builder::session()
+            .unwrap()
+            .name(name)
+            .unwrap()
+            .serve_at(obj, interface)
+            .unwrap()
+            .build()
+            .await
+            .unwrap();
+
+        // Call the method from another connection - it won't return until we tell it to.
+        let client_conn = Connection::session().await.unwrap();
+        tokio::spawn(async move {
+            client_conn
+                .call_method(Some(name), obj, Some(name), "DoThing", &())
+                .await
+                .unwrap();
+        });
+
+        // Avoid races - make sure we've actually received the method call before we drop our
+        // Connection handle.
+        method_called_listener.await;
+
+        let mut shutdown = pin!(connection.graceful_shutdown());
+        tokio::select! {
+            _ = tokio::time::sleep(Duration::from_secs(u64::MAX)) => {},
+            _ = &mut shutdown => {
+                // While that method call is outstanding, graceful shutdown should not complete.
+                panic!("Graceful shutdown unexpectedly completed");
+            }
+        }
+
+        // If we let the call complete, then the shutdown should complete eventually.
+        trigger_return.notify(1);
+        shutdown.await;
+
+        // The method call should have been allowed to finish properly.
+        done_listener.await;
     }
 }
 
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -126,6 +126,8 @@ tokio = { version = "1.37.0", features = [
   "io-util",
   "net",
   "sync",
+  "time",
+  "test-util",
 ] }
 tracing-subscriber = { version = "0.3.18", features = [
   "env-filter",
--- a/zbus/src/blocking/connection/mod.rs
+++ b/zbus/src/blocking/connection/mod.rs
@@ -257,6 +257,14 @@ impl Connection {
     pub fn close(self) -> Result<()> {
         block_on(self.inner.close())
     }
+
+    /// Gracefully close the connection, waiting for all other references to be dropped.
+    ///
+    /// Blocking version of [`crate::Connection::graceful_shutdown`]. See docs there for
+    /// more details and caveats.
+    pub fn graceful_shutdown(self) {
+        block_on(self.inner.graceful_shutdown())
+    }
 }
 
 impl From<crate::Connection> for Connection {


zbus/src/connection/mod.rs
```diff
@@ -1370,6 +1402,112 @@ mod tests {
...
+    #[timeout(15000)]
+    async fn test_graceful_shutdown() {
+        // If we have the only reference, it should complete immediately.
+        let connection = Connection::session().await.unwrap();
+        let shutdown = connection.graceful_shutdown();
+        tokio::select! {
+            biased;
+            _ = shutdown => {},
+            _ = std::future::ready(()) => {
+                panic!("Graceful shutdown did not immediately complete");
```
zeenix, line 1416: I wonder if this could be potentially problematic. If we receive any method call (e.g a generic client like d-feet running on the bus and calling introspect on all new peers) in the very short lifetime of this connection. :thinking: I don't think we need to test this even since it's not the problematic part. If people want immediate shutdown, they can always `drop` the connection instead.

zbus/src/connection/mod.rs
```diff
@@ -1370,6 +1402,112 @@ mod tests {
...
+
+        // If we have a second reference, it should wait until we drop it.
+        let connection = Connection::session().await.unwrap();
+        let clone = connection.clone();
+        let mut shutdown = pin!(connection.graceful_shutdown());
+        tokio::select! {
+            _ = tokio::time::sleep(Duration::from_secs(u64::MAX)) => {},
+            _ = &mut shutdown => {
+                panic!("Graceful shutdown unexpectedly completed");
+            }
```
zeenix, line 1428: I'm not following. Only 2 futures and you don't want either of them to complete in our lifetimes? :smile:

zbus/src/connection/mod.rs
```diff
@@ -1370,6 +1402,112 @@ mod tests {
         let name_has_owner = dbus.name_has_owner(name.try_into().unwrap()).await.unwrap();
         assert!(!name_has_owner);
     }
+
+    #[tokio::test(start_paused = true)]
```
zeenix, line 1406: * We can't assume the optional tokio feature/dep. This doesn't need to be tokio-specific anyway. We've `future*`, `event-listener`, `async-broadcast` deps as well as async runtime abstractions in `abstractions` module that provide everything you need. The exception would be the `Barrier` but I think it's not necessarily needed.
* Why start_paused?

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1229,24 @@ impl Connection {
             .map_err(Into::into)
     }
 
+    /// Gracefully close the connection, waiting for all other references to be dropped.
+    ///
+    /// This will not disrupt any incoming or outgoing method calls, and will await their
+    /// completion.
+    ///
+    /// # Caution
```
zeenix, line 1237: nitpick: `Caveats` is the typical heading for such docs. :)

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1229,24 @@ impl Connection {
...
+    ///
+    /// * Calling this will not prevent new incoming calls from arriving and keeping the connection
+    ///   alive, so if this is undesirable you may want to deregister any well-known names first.
```
zeenix, line 1240: hmm.. I think this is a problem that we can solve ourselves in zbus so let's do that as part of this method and document that we do this. The names will be lost anyway once this method is done, so it's very much a desired effect and something caller will want in all cases.

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1229,24 @@ impl Connection {
             .map_err(Into::into)
     }
 
+    /// Gracefully close the connection, waiting for all other references to be dropped.
+    ///
+    /// This will not disrupt any incoming or outgoing method calls, and will await their
+    /// completion.
```
zeenix, line 1235: Great job on documenting this. :+1: 

* It would be good to add a simple sample code too.
* Would be nice to update the `Connection` docs too.

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1229,24 @@ impl Connection {
...
+    ///
+    /// # Caution
+    ///
+    /// * Calling this will not prevent new incoming calls from arriving and keeping the connection
+    ///   alive, so if this is undesirable you may want to deregister any well-known names first.
+    ///
+    /// * This shutdown will not complete until the underlying connection is fully dropped, so
+    ///   beware of deadlocks if you are holding any other clones of this `Connection`.
+    pub async fn graceful_shutdown(self) {
```
zeenix, line 1244: Please don't forget to also the blocking wrapper in `blocking::Connection`.

zeenix (changes requested):
Overall, looks really good (especially given you're new to the code) but some fixes/improvements needed (mostly just nitpicks :laughing:).

zbus/src/connection/mod.rs
zeenix: I have been thinking about this. Some thoughts: 

- I'm not sure this is really an issue worth mentioning even. The user is explicitly asking for shutdown **after** all pending methods are handled and they can wait for that. Except in case of some spamming client it long running method handles, this is unlikely to take long in most cases. 
- Well-known names aren't the only way for clients to make method calls into the service, so deregistering names isn't an effective way to stop further incoming calls.
- if we really want to effectively stop further method calls, we should find a way to cancel the ObjectServer task instead.

My suggestion would be to not bother with this for now. If this actually becomes an issue, we can solve it then. For now, just remove this paragraph.

zeenix:
@cjriches I'm sorry for ghosting you. I had a very intense week at work. I'll take a look soon..

zbus/src/connection/mod.rs
```diff
@@ -111,6 +122,10 @@ pub(crate) type MsgBroadcaster = Broadcaster<Result<Message>>;
 ///
 /// For sending messages you can either use [`Connection::send`] method.
 ///
+/// To gracefully close a connection while waiting for any outstanding method calls to complete,
+/// see [`Connection::graceful_shutdown`]. To immediately close a connection in a way that will
+/// disrupt any outstanding method calls, see [`Connection::close`].
```
zeenix, line 127: * Let's add this sentence: "If you do not need the connection shutdown to be immediate and do not care about outstanding method calls to services, you can simply drop the `Connection` instance.
* nit: "see" -> "use" (in both places).

zbus/src/connection/mod.rs
zeenix: > > We can't assume the optional tokio feature/dep.
> 
> Can't we? It's in dev-dependencies so it's guaranteed to be there for testing regardless of other feature selection.

Let me rephase: we **can** (as in ability) but we shouldn't so that the feature is tested for both async runtime backends.

zbus/src/connection/mod.rs
zeenix: Thanks so much for explaining in detail. I appreciate you trying so hard to be very precise. :+1: I learnt something new today. :) I'm not sure we've a similar API/mechanism available in the smol-rs/async-io world, so this part will have to be tokio-specific but that's no biggie since the main part can be backend-agnostic. However, please add a short comment (perhaps above the `sleep` arm of the `select`). Something like "Tokio will advance the time to ensure this future get resolved: https://docs.rs/tokio/latest/tokio/time/fn.pause.html#auto-advance" should suffice.

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1233,56 @@ impl Connection {
...
+    ///     let conn = Builder::session()?
+    ///         .name("foo.bar.baz")?
+    ///         .serve_at("/foo/bar/baz", MyInterface)?
+    ///         .build()
+    ///         .await?;
+    ///
+    ///     // Wait for exit condition here.
+    ///     // some_exit_condition.await;
+    ///
+    ///     conn.release_name("foo.bar.baz").await?;
```
zeenix, line 1274: Not needed.

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1233,56 @@ impl Connection {
...
+    /// use zbus::interface;
+    ///
+    /// struct MyInterface;
+    ///
+    /// #[interface(name = "foo.bar.baz")]
+    /// impl MyInterface {
+    ///     async fn do_thing(&self) {}
+    /// }
+    ///
+    /// async fn graceful_example() -> Result<(), Box<dyn Error>> {
```
zeenix, line 1264: Great job with adding a simple example here. Some notes still:

* Best to [hide non-relevant code](https://github.com/dbus2/zbus/blob/cbeedc62ec9b27867e5fa8ca839a46f4a91b10ab/zbus/src/connection/mod.rs#L842) since this is strictly about the `shutdown` method. 
* Let's use a `main` method so that the code is actually run as a test.

zeenix (changes requested):
So close! I'm pretty confident, we can merge after your next push. :sunglasses:

zbus/src/connection/mod.rs
cjriches: If I've understood correctly, then the fact that tokio is present in dev-dependencies is independent from the feature selection, i.e. it is possible to run this test both with and without tokio as the internal zbus runtime - if the `tokio` feature is not enabled then zbus will still spawn its separate async-io executor for the Connection.

The only sort-of-requirement is that the tokio time-auto-advance feature is useless but harmless if the tight tokio integration is off, as time will immediately advance due to the tokio runtime being idle as the connection is on another runtime - it doesn't actually prove that shutdown doesn't return. But if it's at worst harmless and at best a useful extra check, it seems worth it to me, as the test suite should be run with at least the full feature set, if not again with more combinations. From a quick glance at the CI, it looks like both with and without tokio are indeed tested.

It seems from the other thread that you're on board with this tokio usage now?

cjriches:
> @cjriches I'm sorry for ghosting you. I had a very intense week at work. I'll take a look soon..

No worries, I was on holiday last week so no time lost. I'll try and get back to this once I've caught up.

zbus/src/connection/mod.rs
cjriches: >> if we really want to effectively stop further method calls, we should find a way to cancel the ObjectServer task instead.
>
> This is an interesting idea. I guess we'd have to introduce a second event in ConnectionInner that signals the object server to stop listening.

Thinking about this more, I don't think such a feature would be useful. Presumably the dbus broker would be free to send us more method calls even if we're not listening to the socket anymore, and those method calls would get dropped ungracefully, at which point the whole thing is a bit pointless.

Are you happy with the rephrasing I did?

zbus/src/connection/mod.rs
cjriches: Done, with a little extra about exactly how dropping the Connection behaves.

zbus/src/connection/mod.rs
cjriches: Removed.

zbus/src/connection/mod.rs
cjriches: I learned something about doc-tests today :)

Done.

zbus/src/connection/mod.rs
zeenix: > * If my understanding is correct, releasing a well-known name is in practice _almost_ a guarantee you're not going to receive any more method calls, since the only other way is if someone knew your arbitrarily-assigned unique name, right?

That's not 100% correct. Firstly, there is nothing stopping clients from caching the unique name of the service and calling their methods directly to the unique name. 

Secondly, we need to consider what will happen with and without this change. Clients will still make their calls (most services, even autostart when new messages come for them: a d-bus feature) regardless of whether or not the name in question is taken currently. Let's look at all scenarios here:

  1. If the service has **not** set itself up for autolaunch:
    1. if the service manages to successfully release its name,  the client will get an error from the bus.
    2. if the service does not manage to release its name before it can exit, the client will get an error from the service instead of the bus.
  2. If the server **has** set itself up for autolaunch:
    1. if the service manages to successfully release its name,  the bus will launch the service in a new process (which may not be desirable).
    2. if the service does not manage to release its name before it can exit, the bus will launch the service here too but in this case, there's only going to be one process of the service. I believe this is what people would expect to happen.

So I really don't think we should be recommending releasing of name before exiting.

zbus/src/connection/mod.rs
zeenix: > If I've understood correctly, then the fact that tokio is present in dev-dependencies is independent from the feature selection, i.e. it is possible to run this test both with and without tokio as the internal zbus runtime - if the `tokio` feature is not enabled then zbus will still spawn its separate async-io executor for the Connection.

:heavy_check_mark: 

> The only sort-of-requirement is that the tokio time-auto-advance feature is useless but harmless if the tight tokio integration is off, as time will immediately advance due to the tokio runtime being idle as the connection is on another runtime - it doesn't actually prove that shutdown doesn't return. But if it's at worst harmless and at best a useful extra check, it seems worth it to me, as the test suite should be run with at least the full feature set, if not again with more combinations. From a quick glance at the CI, it looks like both with and without tokio are indeed tested.

Ah ok, sure thing.
 
> It seems from the other thread that you're on board with this tokio usage now?

:+1:

zbus/src/connection/mod.rs
zeenix: Just to add that typically services don't really care that much about messages not getting handled when they're shutting down, because of how D-Bus works (as I explained a bit above). I understand that in some cases, service authors would like to handle the existing queue of calls (and it doesn't hurt to do so in general) so the main point here is to add a best-effort graceful shutdown.

zbus/src/connection/mod.rs
```diff
@@ -1217,6 +1236,56 @@ impl Connection {
...
+    /// Gracefully close the connection, waiting for all other references to be dropped.
+    ///
+    /// This will not disrupt any incoming or outgoing method calls, and will await their
+    /// completion.
+    ///
+    /// # Caveats
+    ///
+    /// * Calling this will not prevent new incoming messages from keeping the collection alive, so
+    ///   you may want to take precautions like deregistering well-known names to reduce the
+    ///   likelihood of this occurring.
```
zeenix, line 1248: As I [explained more now](https://github.com/dbus2/zbus/pull/921#discussion_r1716900578), I think we should remove the part after `alive` here.

zeenix (changes requested):
Just one thing and then we're good to merge. If you still disagree on that part, I propose you remove the part in question from this PR for now and add it back in a separate PR, where we can discuss it more and come to some conclusion. I don't want it to block the rest of the changes. :)

zbus/src/connection/mod.rs
cjriches: > there is nothing stopping clients from caching the unique name of the service and calling their methods directly to the unique name.

True, although this would be brittle to service restarts resulting in a new unique name, so would any client actually do this?

> Let's look at all scenarios here...

I think there's one case missing from your analysis, which is your case 2.1 but with systemd as a service manager and a service of `Type=dbus` - in this case systemd will consider the service to be "stopping" as soon as the name is released. A subsequent call resulting in autolaunch will queue a systemd start behind the stop, bringing the service back up to handle the call as soon as it exits. This results in the stop-start being entirely transparent to the caller and the service effectively being available uninterrupted, and is precisely the example motivating me.

(I think there _might_ be a tiny race still, where if the method call comes in after the dbus broker thinks the name is gone but before systemd has processed the same fact, then the autolaunch could be absorbed as a no-op since systemd thinks the service is still running, resulting in an eventual timeout. But that window seems vanishingly small).

However, you've convinced me that this is niche enough not to be worth making as a general recommendation; I will remove the wording from the docs.

zbus/src/connection/mod.rs
cjriches: Done

zbus/src/connection/mod.rs
zeenix: > > there is nothing stopping clients from caching the unique name of the service and calling their methods directly to the unique name.
> 
> True, although this would be brittle to service restarts resulting in a new unique name, so would any client actually do this?

I agree that they really should not do this but this is the wild world of D-Bus. I hear of people doing all kinds of strange things. :)

> I think there's one case missing from your analysis, which is your case 2.1 but with systemd as a service manager and a service of `Type=dbus`

Right and I've to admit, I didn't remember how systemd handles this. However, this is not the problematic case so we don't need to worry about it IMO.

> (I think there _might_ be a tiny race still, where if the method call comes in after the dbus broker thinks the name is gone but before systemd has processed the same fact, then the autolaunch could be absorbed as a no-op since systemd thinks the service is still running, resulting in an eventual timeout. But that window seems vanishingly small).

True. While the window is small, we shouldn't rely on it being non-problematic. I think no matter what we and service impls do here, unfortunately shutdown can't be guaranteed to be 100% graceful.



> However, you've convinced me that this is niche enough not to be worth making as a general recommendation; I will remove the wording from the docs.

:+1:

zeenix (approved):
:rocket:

Issue #98, opened 2020-10-21
Implement `Type` for commonly used Rust types from external crates


zeenix:
First candidates would be `url::Url` and `uuid::Uuid`. Some of them are [here](https://github.com/launchbadge/sqlx/blob/master/sqlx-core/src/types/mod.rs).

zeenix:
> First candidates would be `url::Url` and `uuid::Uuid`. Some of them are [here](https://github.com/launchbadge/sqlx/blob/master/sqlx-core/src/types/mod.rs).

I think we've pretty decent coverage by now and the `url::Url` and `uuid::Uuid` are already covered. As for better handling of file paths, #977 is tracking that.

Issue #882, opened 2024-07-01
Make `zvariant::Signature` completely static & turn `Type::signature` to trait const


zeenix:
When I wrote zvariant, I was still very new to Rust and didn't make the best choices in some regard. Recently I looked into the `postcard-rpc` code, and that project also seem to have something very similar to a signature: [a Schema](https://docs.rs/postcard/latest/postcard/experimental/schema/trait.Schema.html). However, unlike our `Signature` type, `Schema` is completely static and allocation free (it's targeted for microcontrollers). So it got me thinking, if our `Signature` was similar, we could not only avoid a lot of allocations, but greatly optimize the ser/de process (according to the last benchmarks I conducted, signature parsing is what costs us the most, especially for array types). So something like this:

```rust
#[derive(Debug, Clone, Copy)]
pub enum Signature {
    // Simple types
    U8,
    Bool,
    I16,
    U16,
    I32,
    U32,
    I64,
    U64,
    F64,
    Str,
    Signature,
    ObjectPath,
    Value,
    #[cfg(unix)]
    Fd,

    // Container types
    Array {
        child: &'static Signature,
    },
    Dict {
        key: &'static Signature,
        value: &'static Signature,
    },
    Structure {
        fields: &'static [Signature],
    },
    #[cfg(feature = "gvariant")]
    Maybe {
        child: &'static Signature,
    },
}

pub trait Type {
    const SIGNATURE: Signature;
}

impl<T: Type> Type for [T] {
    const SIGNATURE: Signature = Signature::Array {
        child: &T::SIGNATURE,
    };
}

impl<T: Type> Type for (T,) {
    const SIGNATURE: Signature = Signature::Structure {
        fields: &[T::SIGNATURE],
    };
}
// TODO: Use a macro for for generating all tuple impls
```

All the above code works. :cherries: on top if we can make this work too:

```rust
impl Signature {
    pub const fn as_str(&self) -> &'static str {
        match self {
            Signature::U8 => "y",
            Signature::Bool => "b",
            Signature::I16 => "n",
            Signature::U16 => "q",
            Signature::I32 => "i",
            Signature::U32 => "u",
            Signature::I64 => "x",
            Signature::U64 => "t",
            Signature::F64 => "d",
            Signature::Str => "s",
            Signature::Signature => "g",
            Signature::ObjectPath => "o",
            Signature::Value => "v",
            #[cfg(unix)]
            Signature::Fd => "h",
            Signature::Array { child } => {
[...]

PR #933, opened 2024-08-09
✨ zv: Add more `Type` impls

bash:
See also #98.

This PR implements `Type` on a bunch of types from the standard library and some chrono types.

I also added a helpful macro `impl_type!` that takes care of implementing `Type` and testing that the type and its repr are actually serialize-compatible.

### DIFF
--- a/zvariant/src/type.rs
+++ b/zvariant/src/type.rs
@@ -77,6 +77,106 @@ pub trait DynamicDeserialize<'de>: DynamicType {
         S::Error: Into<zvariant::Error>;
 }
 
+/// Implements the [`Type`] trait by delegating the signature to a simpler type (usually a tuple).
+/// Tests that ensure that the two types are serialize-compatible are auto-generated.
+///
+/// Example:
+/// ```no_compile
+/// impl_type_with_repr! {
+///    // Duration is serialized as a (u64, u32) pair.
+///    Duration => (u64, u32) {
+///        // The macro auto-generates tests for us,
+///        // so we need to provide a test name.
+///        duration {
+///            // Sample values used to test serialize compatibility.
+///            samples = [Duration::ZERO, Duration::MAX],
+///            // Converts our type into the simpler "repr" type.
+///            repr(d) = (d.as_secs(), d.subsec_nanos()),
+///        }
+///    }
+/// }
+/// ```
+macro_rules! impl_type_with_repr {
+    ($($ty:ident)::+ $(<$typaram:ident $(: $($tbound:ident)::+)?>)? => $repr:ty {
+        $test_mod:ident $(<$($typaram_sample:ident = $typaram_sample_value:ty),*>)? {
+            $(signature = $signature:literal,)?
+            samples = $samples:expr,
+            repr($sample_ident:ident) = $into_repr:expr,
+        }
+    }) => {
+        impl $(<$typaram $(: $($tbound)::+)?>)? crate::Type for $($ty)::+ $(<$typaram>)? {
+            #[inline]
+            fn signature() -> crate::Signature<'static> {
+                <$repr>::signature()
+            }
+        }
+
+        #[cfg(test)]
+        #[allow(unused_imports)]
+        mod $test_mod {
+            use super::*;
+            use crate::{serialized::Context, to_bytes, LE};
+
+            $($(type $typaram_sample = $typaram_sample_value;)*)?
+            type Ty = $($ty)::+$(<$typaram>)?;
+
+            const _: fn() = || {
+                fn assert_impl_all<'de, T: ?Sized + serde::Serialize + serde::Deserialize<'de>>() {}
+                assert_impl_all::<Ty>();
+            };
+
+            #[test]
+            fn repr_can_be_deserialized_from_encoded_type() {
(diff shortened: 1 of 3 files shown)


zvariant/src/type.rs
```diff
@@ -77,6 +77,87 @@ pub trait DynamicDeserialize<'de>: DynamicType {
         S::Error: Into<zvariant::Error>;
 }
 
+macro_rules! impl_type {
```
zeenix, line 90: Nice macro! I especially love how it also handles tests. :+1: Having said that, it would be nice to:

* Add a doc comment explaining the usage. It doesn't have to be detailed.
* rename to be a bit more descriptive. `impl_type_for_repr`? Better names welcome of course. :)

zvariant/src/type.rs
```diff
@@ -433,42 +433,54 @@ macro_rules! map_impl {
 map_impl!(BTreeMap<K: Ord, V>);
 map_impl!(HashMap<K: Eq + Hash, V, H: BuildHasher>);
 
-impl Type for Duration {
-    fn signature() -> Signature<'static> {
-        <(u64, u32)>::signature()
+impl_type! {
+    Duration => (u64, u32) {
+        test_duration {
```
zeenix, line 500: Let's remove `test_` prefix from test mod names please. It's redundant given the context.

zvariant/src/type.rs
```diff
@@ -675,6 +675,16 @@ impl_type! {
     }
 }
 
+#[cfg(feature = "chrono")]
+impl_type! {
+    chrono::Month => &str {
```
zeenix, line 761: I really appreciate you splitting the commits. However, I believe addition of `Type` for chrono types is one logical change, so I'd squash the commits and update the commit log. However, it's fine like this too, if you disagree.

zeenix (changes requested):
LGTM other than some minor comments. The new tests also make at least some of the tests in `lib` mod obsolete. It would be great if you could remove those in an additional commit on top (not required so if don't have the time, just let me know and I'll do that at some point after merging).

Reminder to self: it's also high time, `type` mod is split.

zvariant/src/type.rs
```diff
@@ -577,8 +577,6 @@ impl_type! {
     }
 }
 
-// FIXME: Ignoring the `serde-human-readable` feature of `time` crate in these impls:
-// https://github.com/time-rs/time/blob/f9398b9598757508ca3815694f23203843e0011b/src/serde/mod.rs#L110
```
zeenix, line 445: 🔥 or 💡 is a more appropriate emoji for this btw. :wink:

bash:
> Nice macro! I especially love how it also handles tests. 👍

Thaaanks 😊 It took me a few iterations until I knew exactly *what* I wanted to test and how to write the macro :)

Btw I really like your recommended atomic-commit/rebase workflow (it matches my personal workflow pretty closely).

zvariant/src/type.rs
bash: Squashing them makes sense to me 👍🏼

zvariant/src/type.rs
bash: Agreed, the macro name could be a bit more descriptive. How about `impl_type_with_repr`? I'm by no means happy with this name, but I think it's slightly better than `impl_type_for_repr`?

zeenix: sure.

bash:
> The new tests also make at least some of the tests in lib mod obsolete. It would be great if you could remove those in an additional commit on top.

I hope I didn't miss any :)

zeenix:
@bash wow, thanks for the very quick actions. I'll take a second look but if you addressed everything and CI goes green, we can merge already. :)

BTW, I'm slowly working on moving `Type` towards [a parsed representation of signature](https://github.com/zeenix/zbus/commit/f1ae132ceba33e4f99fce6399aa01d2e79365d42) as part of #882. This part of the code will obviously change a bit as part of that. Just FYI. :)

zvariant/src/type.rs
```diff
@@ -77,6 +87,121 @@ pub trait DynamicDeserialize<'de>: DynamicType {
         S::Error: Into<zvariant::Error>;
 }
 
+/// Implements the [`Type`] trait by delegating the signature
+/// to a simpler type (usually a tuple).
+/// Tests that ensure that the two types are serialize-compatible
+/// are auto-generated.
```
zeenix, line 93: Let's only break lines (including comments) at 100 chars limit.

```suggestion
/// Implements the [`Type`] trait by delegating the signature to a simpler type (usually a tuple).
/// Tests that ensure that the two types are serialize-compatible are auto-generated.
```

zeenix (changes requested):
Awesome! One tiny last thing. :)

bash:
> wow, thanks for the very quick actions

Today was a good day in terms of motivation, so I had to capitalize on that :))

> BTW, I'm slowly working on moving Type towards [a parsed representation of signature](https://github.com/zeenix/zbus/commit/f1ae132ceba33e4f99fce6399aa01d2e79365d42)

Oh, that looks really cool ✨

zeenix (approved):

PR #935, opened 2024-08-09
📝 book: update dead and obsolete fd.o links

drendog:
- Update dead notification spec links
- Update obsolete systemd refs links


zeenix (approved):
Thanks!

PR #947, opened 2024-08-14
🚑️ zb: Gracefully handle serial number wrap

### LINKED ISSUE #946: zbus panics once SERIAL_NUM wraps past u32::MAX
We have a zbus server which handles ~high dbus message rates (~thousands/s) and are seeing serial number generation cause panics once the server survives long enough to when the u32 wraps past `u32::MAX` back to zero, which triggers a panic on conversion to `NonZeroU32`

https://github.com/dbus2/zbus/blob/1cc4117f2c25c5914ea2f7452de8b1bbdea0f6a5/zbus/src/message/header.rs#L143

It's exacerbated by the fact that the reply path generates a new serial number only to overwrite it with the reply serial.
https://github.com/dbus2/zbus/blob/1cc4117f2c25c5914ea2f7452de8b1bbdea0f6a5/zbus/src/message/builder.rs#L78-L79
https://github.com/dbus2/zbus/blob/1cc4117f2c25c5914ea2f7452de8b1bbdea0f6a5/zbus/src/message/builder.rs#L40-L41
https://github.com/dbus2/zbus/blob/1cc4117f2c25c5914ea2f7452de8b1bbdea0f6a5/zbus/src/message/builder.rs#L179-L181

opsound:
When `SERIAL_NUM` wraps past `u32::MAX` back to zero, the conversion to `NonZeroU32` panics.

Fixes #946.

### DIFF
--- a/zbus/src/message/header.rs
+++ b/zbus/src/message/header.rs
@@ -134,13 +134,18 @@ assert_impl_all!(PrimaryHeader: Send, Sync, Unpin);
 impl PrimaryHeader {
     /// Create a new `PrimaryHeader` instance.
     pub fn new(msg_type: Type, body_len: u32) -> Self {
+        let mut serial_num = SERIAL_NUM.fetch_add(1, SeqCst);
+        if serial_num == 0 {
+            serial_num = SERIAL_NUM.fetch_add(1, SeqCst);
+        }
+
         Self {
             endian_sig: NATIVE_ENDIAN_SIG,
             msg_type,
             flags: BitFlags::empty(),
             protocol_version: 1,
             body_len,
-            serial_num: SERIAL_NUM.fetch_add(1, SeqCst).try_into().unwrap(),
+            serial_num: serial_num.try_into().unwrap(),
         }
     }
 
@@ -353,7 +358,7 @@ impl<'m> Header<'m> {
     }
 }
 
-static SERIAL_NUM: AtomicU32 = AtomicU32::new(1);
+static SERIAL_NUM: AtomicU32 = AtomicU32::new(0);
 
 #[cfg(test)]
 mod tests {


zeenix (changes requested):
Thanks so much for the very quick action! LGTM. Two things:

1. I took the liberty of editing the title and description of the PR to be consistent with the project's style. If you could kindly edit your commit locally to take these changes in, that would great. If you're not comfortable doing that, just let me know and I'll do it for you.
2. Did you test that this helps?

opsound:
> I took the liberty of editing the title and description of the PR to be consistent with the project's style. If you could kindly edit your commit locally to take these changes in, that would great. If you're not comfortable doing that, just let me know and I'll do it for you.
 
let me try

> Did you test that this helps?

With a hack change that initializes SERIAL_NUM to `u32::MAX`, yes. For the real application to hit the panic requires many days of runtime.

We could also consider initializing it to zero like so, so that the "skipping zero" logic gets traversed + tested at least once as part of the standard case.
```
static SERIAL_NUM: AtomicU32 = AtomicU32::new(0);
```

zeenix (approved):
Looks good. Once you've verified it helps with your issue, please mark the PR as ready and we can merge. 😊

opsound:
Minor update that someone pointed out in our internal review was to update `if serial_num == 0` to `while serial_num == 0` for academic correctness (if intervening threads cause the serial to lap in between the two fetch_add's). Not a practical concern for `u32`, but there's ~no additional cost in making it robust to this, so I changed it.

I made a test harness to spam dbus messages, and our dbus server has survived past ~14 billion messages while running overnight, so I think this is pretty settled. Leaving some bake time and then will flip the PR state to non-draft.

Thanks for helping to review! I'm not sure whether our usage is disclosable yet.

zeenix:
> Minor update that someone pointed out in our internal review was to update if serial_num == 0 to while serial_num == 0 for academic correctness (if intervening threads cause the serial to lap in between the two fetch_add's). 

I assume by "lap" you mean another thread overflowing the serial counter again when the current thread got 0? Even if you somehow make this extremely unlikely situation happen (i think the bus will kick you out if you're so fast), how does your code safeguards against the serial collision? The two threads don't sync with each other in any way. The loop only exits once each thread gets > 0.

The same goes if you meant that another thread bumps the serial between the two bumps, that's also not something your commit helps with for the same exact reason. It's also not a problem at all, if you end up with a serial of 2 just after getting 0, since you won't be using 0 anyway and it's just like you got 2 in the first place.

Please let's stick to the issue in question in this PR. 😊 Serial number collision is very much unrelated here IMO.

opsound:
Tying it back to serial collision is a good point. If another thread(s) somehow managed to cause a lap in between the two fetch_adds, there’s nothing addressing serial collision, and panic in that scenario is actually preferable to silently having duplicate outstanding serial.

Will revert to the loop-less version.

I’m not concerned about the 0=>2 case, and agree that the observable behavior would be fine for the thread which gets vended 2.

zeenix:
> If another thread(s) somehow managed to cause a lap in between the two fetch_adds, there’s nothing addressing serial collision, and panic in that scenario is actually preferable to silently having duplicate outstanding serial.

I'm not sure I'm following actually. Why would there be a panic? None of the threads will use the 0. There is the only point where serial is incremented (atomically) so if any thread gets a 0, this new code will ensure they, try again and the next time they're guaranteed to get a non-zero. The collision is only possible if other threads somehow (not possible but let's say it is) manage to overflow the integer immediately after another thread already did so.

I think it's important to keep in mind that, the code is **not** just locally bumping the integer after getting a zero but doing it atomically through the shared static variable.

> Will revert to the loop-less version.
> 
> I’m not concerned about the 0=>2 case, and agree that the observable behavior would be fine for the thread which gets vended 2.

:+1:

zbus/src/message/header.rs
```diff
@@ -353,7 +358,7 @@ impl<'m> Header<'m> {
     }
 }
 
-static SERIAL_NUM: AtomicU32 = AtomicU32::new(1);
+static SERIAL_NUM: AtomicU32 = AtomicU32::new(0);
```
zeenix, line 361: Actually we don't want this. Traditionally all serial numbers start from 1 and since there is no reason not to start with a 1, let's continue doing that. I can imagine this also causing people to think that some message was lost when monitoring the bus traffic for troubleshooting.

zeenix (changes requested):
Can go in immediately after you've removed this one change, unless you've good arguments for the change in question. :)

zbus/src/message/header.rs
zeenix: Oh, I misread. I'm sorry. This is actually fixing the issue of not starting with a 1. :laughing:

zeenix (approved):

opsound:
> The collision is only possible if other threads somehow (not possible but let's say it is) manage to overflow the integer immediately after another thread already did so.

Yes, the panic I was referring to was in this very specific theoretical (and unlikely/impossible) scenario.

```rust
        let mut serial_num = SERIAL_NUM.fetch_add(1, SeqCst); // A
        if serial_num == 0 { // B
            serial_num = SERIAL_NUM.fetch_add(1, SeqCst); // C
        } 
```

In between A and B, fetch_add could theoretically be called an arbitrary number of times by other threads.

If A observes `serial_num == 0`, and then `fetch_add` is called exactly `u32::MAX` more times by intervening thread(s) before C runs, then C will also observe `serial_num == 0`, leading to panic on conversion to `NonZeroU32`

Like you mentioned, there are real-world considerations or invariants elsewhere in other code that would prevent this from happening in practice.

opsound:
Thanks for working with me on this!

zeenix:
> If A observes serial_num == 0, and then fetch_add is called exactly u32::MAX more times by intervening thread(s) before C runs, then C will also observe serial_num == 0, leading to panic on conversion to NonZeroU32

Ah right. 

> Like you mentioned, there are real-world considerations or invariants elsewhere in other code that would prevent this from happening in practice.

Yeah. 😊

PR #948, opened 2024-08-14
📝 zb: Tidy documentation across entire crate

cjriches:
I figured I'd fix a couple of things while I was here and perhaps got carried away :)

### DIFF
--- a/zbus/src/address/transport/tcp.rs
+++ b/zbus/src/address/transport/tcp.rs
@@ -55,22 +55,22 @@ impl Tcp {
         self
     }
 
-    /// Returns the `tcp:` address `host` value.
+    /// The `tcp:` address `host` value.
     pub fn host(&self) -> &str {
         &self.host
     }
 
-    /// Returns the `tcp:` address `bind` value.
+    /// The `tcp:` address `bind` value.
     pub fn bind(&self) -> Option<&str> {
         self.bind.as_deref()
     }
 
-    /// Returns the `tcp:` address `port` value.
+    /// The `tcp:` address `port` value.
     pub fn port(&self) -> u16 {
         self.port
     }
 
-    /// Returns the `tcp:` address `family` value.
+    /// The `tcp:` address `family` value.
     pub fn family(&self) -> Option<TcpTransportFamily> {
         self.family
     }
--- a/zbus/src/blocking/proxy/mod.rs
+++ b/zbus/src/blocking/proxy/mod.rs
@@ -246,8 +246,8 @@ impl<'a> Proxy<'a> {
     /// method flags to control the way the method call message is sent and handled.
     ///
     /// Use [`call`] instead if you do not need any special handling via additional flags.
-    /// If the `NoReplyExpected` flag is passed , this will return None immediately
-    /// after sending the message, similar to [`call_noreply`]
+    /// If the `NoReplyExpected` flag is passed, this will return None immediately
+    /// after sending the message, similar to [`call_noreply`].
     ///
     /// [`call`]: struct.Proxy.html#method.call
     /// [`call_noreply`]: struct.Proxy.html#method.call_noreply
@@ -266,7 +266,7 @@ impl<'a> Proxy<'a> {
         block_on(self.inner().call_with_flags(method_name, flags, body))
     }
 
-    /// Call a method without expecting a reply
+    /// Call a method without expecting a reply.
     ///
     /// This sets the `NoReplyExpected` flag on the calling message and does not wait for a reply.
     pub fn call_noreply<'m, M, B>(&self, method_name: M, body: &B) -> Result<()>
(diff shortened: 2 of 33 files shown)


zbus/src/address/transport/tcp.rs
```diff
@@ -55,22 +55,22 @@ impl Tcp {
         self
     }
 
-    /// Returns the `tcp:` address `host` value.
+    /// Return the `tcp:` address `host` value.
```
zeenix, line 58: Actually "Return" is redundant here so you might as well remove that.

zeenix (changes requested):

zbus/src/address/transport/tcp.rs
cjriches: Done

zbus/src/blocking/proxy/mod.rs
```diff
@@ -470,11 +470,11 @@ where
     T: TryFrom<zvariant::OwnedValue>,
     T::Error: Into<crate::Error>,
 {
-    // Get the value of the property that changed.
-    //
-    // If the notification signal contained the new value, it has been cached already and this call
-    // will return that value. Otherwise (i-e invalidated property), a D-Bus call is made to fetch
```
zeenix, line 476: Hmm.. I always wrote it "i-e". My whole life has been a life. :open_mouth:

zbus/src/connection/socket/mod.rs
```diff
@@ -312,14 +312,14 @@ pub trait WriteHalf: std::fmt::Debug + Send + Sync + 'static {
     /// After this call, it is valid for all reading and writing operations to fail.
     async fn close(&mut self) -> io::Result<()>;
 
-    /// Supports passing file descriptors.
+    /// Return whether passing file descriptors is supported.
```
zeenix, line 315: This `Return` is also unneeded and inconsistent with the other getters here.

zbus/src/fdo.rs
```diff
@@ -272,7 +272,7 @@ macro_rules! gen_object_manager_proxy {
 gen_object_manager_proxy!(true, false);
 assert_impl_all!(ObjectManagerProxy<'_>: Send, Sync, Unpin);
 
-/// Service-side [Object Manager][om] interface implementation.
+/// Server-side [Object Manager][om] interface implementation.
```
zeenix, line 275: This was very intentional. The "server" in D-Bus is typically the broker and most services are just clients of this server.

zbus/src/match_rule/mod.rs
```diff
@@ -21,7 +21,7 @@ pub use builder::Builder;
 
 /// A bus match rule for subscribing to specific messages.
 ///
-/// This is mainly used by peer to subscribe to specific signals as by default the bus will not
+/// This is mainly used for p2p to subscribe to specific signals as by default the bus will not
```
zeenix, line 24: This is incorrect. We really mean a peer here. A match-rule is mainly for the bus case.

zeenix (commented):
Sorry but still some way to go. :)

zbus/src/match_rule/mod.rs
cjriches: Right, the wording "by peer" didn't make sense to me so I assumed this must be talking about p2p. What do peers have to do with things if it's between us and the bus?

For now, I've removed the words entirely ("This is mainly used to subscribe"), since this reads best to me. Let me know if you'd prefer something else.

zbus/src/fdo.rs
cjriches: Right, I wasn't sure, but line 357 says "Server-side" and seems to be talking about the same side, so I picked one to be consistent.

Should line 357 also be changed to "Service-side"? Or is it actually talking about building a broker?

zbus/src/connection/socket/mod.rs
cjriches: Done

zbus/src/blocking/proxy/mod.rs
cjriches: I mean I wouldn't go so far as to call it _wrong_, but `i.e.` is certainly more traditional 😅

zbus/src/fdo.rs
zeenix: Right, so that's our fault. Until recently we were using `server` to mean a service and it was confusing as hell. I recently fixed that in most places, but obviously I didn't do a very thorough job. :cry: 

> Should line 357 also be changed to "Service-side"? Or is it actually talking about building a broker?

Yeah, that's a left-over from my changes. It should be `Service-side`. There are certainly spots where we really are talking about the server-side but they're likely guarded by the `bus-impl` feature so it should be easy to tell those ones apart.

zbus/src/match_rule/mod.rs
zeenix: `peer` is a term used in the D-Bus world frequently for the bus clients, perhaps cause it can get really confusing otherwise if services are also clients.

> For now, I've removed the words entirely ("This is mainly used to subscribe"), since this reads best to me. Let me know if you'd prefer something else.

That sounds good to me. Maybe it will be confusing for others too.

zbus/src/fdo.rs
cjriches: Fixed on lines 45, 112, and 357.

zeenix (approved):

cjriches:
Thanks for the review, are there any other steps needed to get this merged?

Issue #325, opened 2023-03-13
Implement unixexec: addresses


zeenix:
In GitLab by @bouk on Mar 13, 2023, 14:53

> This transport forks off a process and connects its standard input and standard output with an anonymous Unix domain socket. This socket is then used for communication by the transport. This transport may be used to use out-of-process forwarder programs as basis for the D-Bus protocol.

This can be used with [systemd-stdio-bridge](https://www.freedesktop.org/software/systemd/man/systemd-stdio-bridge.html) to connect to the systemd bus on another system over SSH.

dustymabe:
It looks like #976 merged. Is that all that's needed for this feature request?

I was trying something local but kept getting:

```
address error: unsupported transport 'unixexec'
```

When I try something like the following (which matches the test from #976 closely):

```
let addr = "unixexec:path=systemd-stdio-bridge";
let conn = Builder::address(addr)?              
.build()                                        
.await?;                                        
```

zeenix:
> It looks like [#976](https://github.com/dbus2/zbus/pull/976) merged. Is that all that's needed for this feature request?

Yeah, I think @vially just forgot to add the `Fixes #325` line to the PR so this issue didn't get resolved.


> When I try something like the following (which matches the test from [#976](https://github.com/dbus2/zbus/pull/976) closely):

That's very strange cause not only did @vially added tests, [they are still there](https://github.com/dbus2/zbus/blob/main/zbus/tests/unixexec.rs) and running as part of our CI. I ran it locally just now and it seems to work just fine:

```
❯ cargo t -p zbus unixexec
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.06s
     Running unittests src/lib.rs (target/debug/deps/zbus-2c41d3ae3e3b6a93)

running 1 test
test address::transport::unixexec::tests::connect ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 17 filtered out; finished in 0.00s

     Running tests/basic.rs (target/debug/deps/basic-ebff7500576e1a0d)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 7 filtered out; finished in 0.00s

     Running tests/e2e.rs (target/debug/deps/e2e-10c73fe86ae319da)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 1 filtered out; finished in 0.00s

[...]

PR #976, opened 2024-09-05
✨ zb: Add support for `unixexec` transport

vially:
Add initial support for the [unixexec](https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec) transport (#325). ~~Currently this is **only implemented when the `tokio` feature is enabled**.~~ Edit: Support for the `async-io` runtime has also been implemented in later revisions.

### DIFF
--- a/zbus/src/abstractions/mod.rs
+++ b/zbus/src/abstractions/mod.rs
@@ -8,6 +8,6 @@ mod async_drop;
 pub(crate) mod async_lock;
 pub use async_drop::*;
 
-// Not macOS-specific itself but only used on macOS.
-#[cfg(target_os = "macos")]
+// Not unix-specific itself but only used on unix.
+#[cfg(target_family = "unix")]
 pub(crate) mod process;
--- a/zbus/src/abstractions/process.rs
+++ b/zbus/src/abstractions/process.rs
@@ -1,6 +1,8 @@
+#[cfg(target_os = "macos")]
 use std::{ffi::OsStr, io::Error, process::Output};
 
 /// An asynchronous wrapper around running and getting command output
+#[cfg(target_os = "macos")]
 pub async fn run<I, S>(program: S, args: I) -> Result<Output, Error>
 where
     I: IntoIterator<Item = S>,
--- a/zbus/src/abstractions/process.rs
+++ b/zbus/src/abstractions/process.rs
@@ -1,5 +1,82 @@
+#[cfg(not(feature = "tokio"))]
+use async_process::{unix::CommandExt, Child};
 #[cfg(target_os = "macos")]
-use std::{ffi::OsStr, io::Error, process::Output};
+use std::process::Output;
+use std::{ffi::OsStr, io::Error, process::Stdio};
+#[cfg(feature = "tokio")]
+use tokio::process::Child;
+
+/// A wrapper around the command API of the underlying async runtime.
+pub struct Command(
+    #[cfg(not(feature = "tokio"))] async_process::Command,
+    #[cfg(feature = "tokio")] tokio::process::Command,
+);
+
+impl Command {
+    /// Constructs a new `Command` for launching the program at path `program`.
+    pub fn new<S>(program: S) -> Self
+    where
+        S: AsRef<OsStr>,
+    {
+        #[cfg(not(feature = "tokio"))]
+        return Self(async_process::Command::new(program));
+
+        #[cfg(feature = "tokio")]
+        return Self(tokio::process::Command::new(program));
+    }
+
+    /// Sets executable argument.
+    ///
+    /// Set the first process argument, `argv[0]`, to something other than the
+    /// default executable path.
+    pub fn arg0<S>(&mut self, arg: S) -> &mut Self
+    where
+        S: AsRef<OsStr>,
+    {
+        self.0.arg0(arg);
+        self
+    }
+
+    /// Adds multiple arguments to pass to the program.
+    pub fn args<I, S>(&mut self, args: I) -> &mut Self
+    where
+        I: IntoIterator<Item = S>,
+        S: AsRef<OsStr>,
+    {
+        self.0.args(args);
+        self
+    }
+
+    /// Executes the command as a child process, waiting for it to finish and
+    /// collecting all of its output.
+    #[cfg(target_os = "macos")]
+    pub async fn output(&mut self) -> Result<Output, Error> {
+        self.0.output().await
+    }
+
+    /// Sets configuration for the child process's standard input (stdin) handle.
+    pub fn stdin<T: Into<Stdio>>(&mut self, cfg: T) -> &mut Self {
+        self.0.stdin(cfg);
+        self
+    }
+
+    /// Sets configuration for the child process's standard output (stdout) handle.
+    pub fn stdout<T: Into<Stdio>>(&mut self, cfg: T) -> &mut Self {
+        self.0.stdout(cfg);
+        self
+    }
+
+    /// Sets configuration for the child process's standard error (stderr) handle.
+    pub fn stderr<T: Into<Stdio>>(&mut self, cfg: T) -> &mut Self {
+        self.0.stderr(cfg);
+        self
+    }
+
+    /// Executes the command as a child process, returning a handle to it.
+    pub fn spawn(&mut self) -> Result<Child, Error> {
+        self.0.spawn()
+    }
+}
 
 /// An asynchronous wrapper around running and getting command output
 #[cfg(target_os = "macos")]
@@ -8,15 +85,5 @@ where
     I: IntoIterator<Item = S>,
     S: AsRef<OsStr>,
 {
-    #[cfg(not(feature = "tokio"))]
-    return async_process::Command::new(program)
-        .args(args)
-        .output()
-        .await;
-
-    #[cfg(feature = "tokio")]
-    return tokio::process::Command::new(program)
-        .args(args)
-        .output()
-        .await;
+    Command::new(program).args(args).output().await
 }
--- a/zbus/src/abstractions/process.rs
+++ b/zbus/src/abstractions/process.rs
@@ -6,6 +6,8 @@ use std::{ffi::OsStr, io::Error, process::Stdio};
 #[cfg(feature = "tokio")]
 use tokio::process::Child;
 
+use crate::address::transport::Unixexec;
+
 /// A wrapper around the command API of the underlying async runtime.
 pub struct Command(
     #[cfg(not(feature = "tokio"))] async_process::Command,
@@ -23,6 +25,18 @@ impl Command {
 
         #[cfg(feature = "tokio")]
         return Self(tokio::process::Command::new(program));
+    }
+
+    /// Constructs a new `Command` from a `unixexec` address.
+    pub fn for_unixexec(unixexec: &Unixexec) -> Self {
+        let mut command = Self::new(unixexec.path());
+        command.args(unixexec.args());
+
+        if let Some(arg0) = unixexec.arg0() {
+            command.arg0(arg0);
+        }
+
+        command
     }
 
     /// Sets executable argument.
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -1,6 +1,6 @@
 //! D-Bus transport Information module.
 //!
-//! This module provides the trasport information for D-Bus addresses.
+//! This module provides the transport information for D-Bus addresses.
 
 #[cfg(windows)]
 use crate::win32::autolaunch_bus_address;
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -2,6 +2,8 @@
 //!
 //! This module provides the transport information for D-Bus addresses.
 
+#[cfg(unix)]
+use crate::connection::socket::Command;
 #[cfg(windows)]
 use crate::win32::autolaunch_bus_address;
 use crate::{Error, Result};
@@ -20,6 +22,10 @@ use tokio_vsock::VsockStream;
 use uds_windows::UnixStream;
 #[cfg(all(feature = "vsock", not(feature = "tokio")))]
 use vsock::VsockStream;
+#[cfg(unix)]
+mod unixexec;
+#[cfg(unix)]
+pub use unixexec::Unixexec;
 
 use std::{
     fmt::{Display, Formatter},
@@ -77,6 +83,9 @@ pub enum Transport {
     /// The type of `stream` is `vsock::VsockStream` with the `vsock` feature and
     /// `tokio_vsock::VsockStream` with the `tokio-vsock` feature.
     Vsock(Vsock),
+    /// A `unixexec` address.
+    #[cfg(unix)]
+    Unixexec(Unixexec),
 }
 
 impl Transport {
@@ -136,6 +145,8 @@ impl Transport {
                     }
                 }
             }
+            #[cfg(unix)]
+            Transport::Unixexec(unixexec) => unixexec.connect().await.map(Stream::Unixexec),
             #[cfg(all(feature = "vsock", not(feature = "tokio")))]
             Transport::Vsock(addr) => {
                 let stream = VsockStream::connect_with_cid_port(addr.cid(), addr.port())?;
@@ -213,6 +224,8 @@ impl Transport {
     pub(super) fn from_options(transport: &str, options: HashMap<&str, &str>) -> Result<Self> {
         match transport {
             "unix" => Unix::from_options(options).map(Self::Unix),
+            #[cfg(unix)]
+            "unixexec" => Unixexec::from_options(options).map(Self::Unixexec),
             "tcp" => Tcp::from_options(options, false).map(Self::Tcp),
             "nonce-tcp" => Tcp::from_options(options, true).map(Self::Tcp),
             #[cfg(any(
@@ -236,6 +249,8 @@ impl Transport {
 #[derive(Debug)]
 pub(crate) enum Stream {
     Unix(Async<UnixStream>),
+    #[cfg(unix)]
+    Unixexec(Command),
     Tcp(Async<TcpStream>),
     #[cfg(feature = "vsock")]
     Vsock(Async<VsockStream>),
@@ -246,6 +261,8 @@ pub(crate) enum Stream {
 pub(crate) enum Stream {
     #[cfg(unix)]
     Unix(tokio::net::UnixStream),
+    #[cfg(unix)]
+    Unixexec(Command),
     Tcp(TcpStream),
     #[cfg(feature = "tokio-vsock")]
     Vsock(VsockStream),
@@ -336,6 +353,8 @@ impl Display for Transport {
         match self {
             Self::Tcp(tcp) => write!(f, "{}", tcp)?,
             Self::Unix(unix) => write!(f, "{}", unix)?,
+            #[cfg(unix)]
+            Self::Unixexec(unixexec) => write!(f, "{}", unixexec)?,
             #[cfg(any(
                 all(feature = "vsock", not(feature = "tokio")),
                 feature = "tokio-vsock"
--- a/zbus/src/address/transport/unixexec.rs
+++ b/zbus/src/address/transport/unixexec.rs
@@ -1,0 +1,115 @@
+use std::{
+    borrow::BorrowMut, ffi::OsString, fmt::Display, os::unix::ffi::OsStrExt, path::PathBuf,
+    process::Stdio,
+};
+
+use crate::process::Command;
+
+use super::encode_percents;
+
+/// `unixexec:` D-Bus transport.
+///
+/// <https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec>
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Unixexec {
+    path: PathBuf,
+    arg0: Option<OsString>,
+    args: Vec<OsString>,
+}
+
+impl Unixexec {
+    /// Create a new unixexec transport with the given path and arguments.
+    pub fn new(path: PathBuf, arg0: Option<OsString>, args: Vec<OsString>) -> Self {
+        Self { path, arg0, args }
+    }
+
+    pub(super) fn from_options(opts: std::collections::HashMap<&str, &str>) -> crate::Result<Self> {
+        let Some(path) = opts.get("path") else {
+            return Err(crate::Error::Address(
+                "unixexec address is missing `path`".to_owned(),
+            ));
+        };
+
+        let arg0 = opts.get("argv0").map(OsString::from);
+
+        let mut args: Vec<OsString> = Vec::new();
+        let mut arg_index = 1;
+        while let Some(arg) = opts.get(format!("argv{arg_index}").as_str()) {
+            args.push(OsString::from(arg));
+            arg_index += 1;
+        }
+
+        Ok(Self::new(PathBuf::from(path), arg0, args))
+    }
+
+    /// Binary to execute.
+    ///
+    /// Path of the binary to execute, either an absolute path or a binary name that is searched for
+    /// in the default search path of the OS. This corresponds to the first argument of execlp().
+    /// This key is mandatory.
+    pub fn path(&self) -> &PathBuf {
+        &self.path
+    }
+
+    /// The executable argument.
+    ///
+    /// The program name to use when executing the binary. If omitted the same
+    /// value as specified for path will be used. This corresponds to the
+    /// second argument of execlp().
+    pub fn arg0(&self) -> Option<&OsString> {
+        self.arg0.as_ref()
+    }
+
+    /// Arguments.
+    ///
+    /// Arguments to pass to the binary.
+    pub fn args(&self) -> &[OsString] {
+        self.args.as_ref()
+    }
+
+    pub(super) async fn connect(&self) -> crate::Result<crate::connection::socket::Command> {
+        Command::for_unixexec(self)
+            .stdin(Stdio::piped())
+            .stdout(Stdio::piped())
+            .stderr(Stdio::inherit())
+            .spawn()?
+            .borrow_mut()
+            .try_into()
+    }
+}
+
+impl Display for Unixexec {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("unixexec:path=")?;
+        encode_percents(f, self.path.as_os_str().as_bytes())?;
+
+        if let Some(arg0) = self.arg0.as_ref() {
+            f.write_str(",argv0=")?;
+            encode_percents(f, arg0.as_bytes())?;
+        }
+
+        for (index, arg) in self.args.iter().enumerate() {
+            write!(f, ",argv{}=", index + 1)?;
+            encode_percents(f, arg.as_bytes())?;
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::address::{transport::Transport, Address};
+
+    #[test]
+    fn connect() {
+        let addr: Address = "unixexec:path=echo,argv1=hello,argv2=world"
+            .try_into()
+            .unwrap();
+        let unixexec = match addr.transport() {
+            Transport::Unixexec(unixexec) => unixexec,
+            _ => unreachable!(),
+        };
+        crate::utils::block_on(unixexec.connect()).unwrap();
+    }
+}
--- a/zbus/src/connection/socket/command.rs
+++ b/zbus/src/connection/socket/command.rs
@@ -1,0 +1,137 @@
+use std::{io, os::fd::BorrowedFd};
+
+#[cfg(not(feature = "tokio"))]
+use async_process::{Child, ChildStdin, ChildStdout};
+
+#[cfg(feature = "tokio")]
+use tokio::{
+    io::{AsyncReadExt, ReadBuf},
+    process::{Child, ChildStdin, ChildStdout},
+};
+
+use super::{ReadHalf, RecvmsgResult, Socket, Split, WriteHalf};
+
+/// A Command stream socket.
+///
+/// This socket communicates with a spawned child process via its standard input
+/// and output streams.
+#[derive(Debug)]
+pub(crate) struct Command {
+    stdin: ChildStdin,
+    stdout: ChildStdout,
+}
+
+impl Command {
+    fn into_split(self) -> (ChildStdout, ChildStdin) {
+        (self.stdout, self.stdin)
+    }
+}
+
+impl Socket for Command {
+    type ReadHalf = ChildStdout;
+    type WriteHalf = ChildStdin;
+
+    fn split(self) -> Split<Self::ReadHalf, Self::WriteHalf> {
+        let (read, write) = self.into_split();
+
+        Split { read, write }
+    }
+}
+
+impl TryFrom<&mut Child> for Command {
+    type Error = crate::Error;
+
+    fn try_from(child: &mut Child) -> Result<Self, Self::Error> {
+        let stdin = child
+            .stdin
+            .take()
+            .ok_or(crate::Error::Failure("child stdin not found".into()))?;
+
+        let stdout = child
+            .stdout
+            .take()
+            .ok_or(crate::Error::Failure("child stdout not found".into()))?;
+
+        Ok(Command { stdin, stdout })
+    }
+}
+
+#[cfg(not(feature = "tokio"))]
+#[async_trait::async_trait]
+impl ReadHalf for ChildStdout {
+    async fn recvmsg(&mut self, buf: &mut [u8]) -> RecvmsgResult {
+        match futures_util::AsyncReadExt::read(&mut self, buf).await {
+            Err(e) => Err(e),
+            Ok(len) => {
+                #[cfg(unix)]
+                let ret = (len, vec![]);
+                #[cfg(not(unix))]
+                let ret = len;
+                Ok(ret)
+            }
+        }
+    }
+}
+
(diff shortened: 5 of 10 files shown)


zbus/src/address/transport/mod.rs
```diff
@@ -1,6 +1,6 @@
 //! D-Bus transport Information module.
 //!
-//! This module provides the trasport information for D-Bus addresses.
+//! This module provides the transport information for D-Bus addresses.
```
zeenix, line 3: thanks for your attention to detail but kindly split separate logical changes into separate commits (not PRs).

zbus/src/address/transport/mod.rs
```diff
@@ -18,6 +18,8 @@ use tokio::net::TcpStream;
 use tokio_vsock::VsockStream;
 #[cfg(windows)]
 use uds_windows::UnixStream;
+#[cfg(all(unix, feature = "tokio"))]
+use unixexec::UnixExec;
```
zeenix, line 22: Please move this below, after declaration of the module (like other mods here).

zbus/src/address/transport/mod.rs
```diff
@@ -52,6 +54,8 @@ use std::os::linux::net::SocketAddrExt;
     feature = "tokio-vsock"
 ))]
 pub use vsock_transport::Vsock;
+#[cfg(all(unix, feature = "tokio"))]
```
zeenix, line 57: * I don't see any reason for this to be specific to tokio feature. We even have a runtime abstraction for launching processes (it's currently only enabled for macos but it should be easy to change to for all unix systems). Talking of which, this is very similar to `Launchd` address. Please have a look into that for inspiration.
* Unlike launchd address, this address type is not connectable but only listenable. So it should only be enabled when `server` feature is enabled.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,107 @@
...
+}
+
+impl Display for UnixExec {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("unixexec:")?;
+        encode_percents(f, self.path.as_os_str().as_bytes())
+    }
+}
+
+impl From<UnixExec> for std::process::Command {
```
zeenix, line 90: All this can be removed. Please use the API we already have.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,107 @@
...
+        let mut arg_index = 1;
+        while let Some(arg) = opts.get(format!("argv{arg_index}").as_str()) {
+            args.push(arg.to_string());
+            arg_index += 1;
+        }
+
+        Ok(Self::new(PathBuf::from(path), arg0, args))
+    }
+
+    pub(super) async fn connect(self) -> crate::Result<UnixStream> {
```
zeenix, line 48: We can support connection? According to the specs, this is not a connectable address type. :thinking:

zeenix (changes requested):
Thanks so much for this contribution. :+1:  Please see inline comments. :)

zbus/src/address/transport/mod.rs
zeenix: Lol, you already did this. :facepalm: Sorry for the noise.

zbus/src/address/transport/unixexec.rs
vially: If I'm reading the spec right, it's the other way around (connectable, but not listenable):
> unixexec addresses are connectable, but are not listenable. ([source](https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec))

In any case, the current implementation does seem to work as expected and can be tested with minimal changes (https://github.com/vially/zbus/commit/9f8037cde9401b3ed3e67017d18d6fad88c0c9ab) to the `watch-systemd-jobs` example. It even seems to work when connecting to a *remote* D-Bus session over SSH.

zeenix: > If I'm reading the spec right, it's the other way around (connectable, but not listenable):

Oh, you're absolute right. Sorry, I was very tired when I read it. :)

> > unixexec addresses are connectable, but are not listenable. ([source](https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec))
> 
> In any case, the current implementation does seem to work as expected and can be tested with minimal changes ([vially@9f8037c](https://github.com/vially/zbus/commit/9f8037cde9401b3ed3e67017d18d6fad88c0c9ab)) to the `watch-systemd-jobs` example. It even seems to work when connecting to a _remote_ D-Bus session over SSH.

Awesome! Thanks so much again. Once you've addressed other comments, we can quickly merge it hopefully.

zbus/src/address/transport/unixexec.rs
vially: The existing `process::run` function would not be flexible enough for the `unixexec` transport, but in order to move more into _that direction_ I have created a thin wrapper over the async `Command` types that _could_ be used for this purpose.

That being said, I'm not sure if that's what you meant by using the existing APIs. Maybe there are some other existing APIs that could be used here that I overlooked?

zeenix: > That being said, I'm not sure if that's what you meant by using the existing APIs. Maybe there are some other existing APIs that could be used here that I overlooked?

No, I meant the process abstraction exactly. Thanks for extending the abstraction. :+1:

zeenix:
I took the liberty of declaring the PR "ready for review". I hope you don't mind.

zbus/src/address/transport/mod.rs
zeenix: Now the only real thing remaining is making the code make use of async abstractions to make it work on all runtimes.

vially: I would love to make this work for both async runtimes and in fact my first attempt was to implement this for the `async-io` runtime.

However, I wasn't able to find an *easy* way to spawn an async task from the `UnixExec::connect` method _without_ `tokio`.

I did find the [`Executor::spawn`](https://github.com/dbus2/zbus/blob/97813b0b3ec6ad6b774b1ee927c926fdd4aef821/zbus/src/abstractions/executor.rs#L38) method which does this, but that requires access to an `executor` reference which is not available in `Transport::connect` *in the current implementation*. That doesn't mean the code couldn't be refactored to pass an executor, it's just that it wasn't as straightforward as just using `tokio::io::spawn`.

And that was the main reason why this transport is only implemented for the `tokio` feature initially.

zeenix: > . That doesn't mean the code couldn't be refactored to pass an executor, it's just that it wasn't as straightforward as just using `tokio::io::spawn`.

Right and we should. We wouldn't become tokio-specific just for this feature, that's for sure. :) 

Thinking about this, I think actually it would make sense for the actual transport to be a [`Socket`](https://docs.rs/zbus/latest/zbus/connection/socket/trait.Socket.html) implementation. Then instead of having to spawn a task to pipe the data, you'll direct read and write from the stdio of the command.

zeenix: oh and btw, `main` branch just entered API break phase. If you need some API to change, this would be the perfect time to ask for that. Although if it's not a trivial change, I'm afraid I won't have the time to implement it myself. :)

vially: The latest revision supports both async runtimes. The new implementation required a) passing the `executor` reference as an argument to a few functions and b) adding `async-net` as a dependency.

> Thinking about this, I think actually it would make sense for the actual transport to be a [Socket](https://docs.rs/zbus/latest/zbus/connection/socket/trait.Socket.html) implementation. Then instead of having to spawn a task to pipe the data, you'll directly read and write from the stdio of the command.

:+1: This is _exactly_ the kind of feedback I was looking for when I decided to open this pull-request early, so thanks for that.

You're absolutely right, I think `Socket` might be the better abstraction for this. That being said, I get the feeling that the kind of logic that will be required for implementing this piping *properly* is fraught with peril and I don't fully trust myself to get it right.

I might give it a shot though, just to see what it would look like, but can't promise that something will come out of it.

zeenix: > The latest revision supports both async runtimes.

Awesome!

> The new implementation required a) passing the `executor` reference as an argument to a few functions

This is one of the reasons I suggested a `Socket` impl. If you don't need to spawn a task, you don't need any executor then.

> and b) adding `async-net` as a dependency.

Given that other transport types (who also need to connect) don't need `async-net`, I'm pretty sure we don't need it here either. You need to wrap the `std::os::unix::net::UnixStream` in `async_io::Async` like we do in all other places (unix streams are the most important transport for us even).

Also, you don't need any stream if you go with the `Socket` approach.

> You're absolutely right, I think `Socket` might be the better abstraction for this. That being said, I get the feeling that the kind of logic that will be required for implementing this piping _properly_ is fraught with peril and I don't fully trust myself to get it right.

That's the thing, you won't need a pipe at all AFAICT. IMO it would actually be simpler than what you're doing currently. If you look at the existing impls of `Socket`, you'll find that they're not super involved.

zeenix: > I might give it a shot though, just to see what it would look like, but can't promise that something will come out of it.

Looking at your work so far, I'm 100% confident you can do it easily. However, you're new to the project so I can understand if the low-level `Socket` intimidates you. If you need any help in implementing `Socket` (or anything), please feel free to reach out and ask (we have a matrix channel, which would be best for any questions you might have).

Knowing that `Socket` is most likely the best way forward and that going towards it would make at least 2 changes in this PR, redundant, I'm afraid I'll have to insist on going (or at least trying to) that route in this PR.

vially: Makes sense. Thanks for the feedback and I fully agree that the `Socket` implementation seems the better approach. I'll give that a try and ask for help if I get stuck.

vially: A new revision has been pushed that is using a `Socket` implementation which no longer requires passing an `executor` along various functions. I think it's looking much better now, thanks again for the feedback and guiding the solution in the right direction.

PS: The `async-net` dependency has been dropped too.

zeenix: > A new revision has been pushed that is using a `Socket` implementation which no longer requires passing an `executor` along various functions.

Nice!

> I think it's looking much better now, thanks again for the feedback and guiding the solution in the right direction.

You're most welcome. Thank you also for going the extra mile to get things done in the best way.

> PS: The `async-net` dependency has been dropped too.

cool. :)

zbus/src/abstractions/mod.rs
```diff
@@ -9,6 +9,6 @@ pub(crate) mod async_lock;
 pub use async_drop::*;
 pub(crate) mod file;
 
-// Not macOS-specific itself but only used on macOS.
-#[cfg(target_os = "macos")]
+// Not unix-specific itself but only used on unix.
+#[cfg(target_family = "unix")]
```
zeenix, line 13: Could we please split this in a separate commit that comes before use of `process` on unix target?

zbus/src/abstractions/process.rs
```diff
@@ -19,6 +23,18 @@ impl Command {
         return Self(tokio::process::Command::new(program));
     }
 
+    /// Sets executable argument.
+    ///
+    /// Set the first process argument, `argv[0]`, to something other than the
+    /// default executable path.
+    pub fn arg0<S>(&mut self, arg: S) -> &mut Command
```
zeenix, line 30: This and other changes to `Command` (except for `spawn`) are fixup to the commit adding `Command` type. Please move them to that commit.

zbus/src/abstractions/process.rs
```diff
@@ -31,12 +47,117 @@ impl Command {
...
+    }
+
+    /// Executes the command as a child process, returning a handle to it.
+    pub fn spawn(&mut self) -> Result<Child, Error> {
+        Ok(Child::from(self.0.spawn()?))
+    }
+}
+
+/// A wrapper around the child API of the underlying async runtime.
+pub struct Child(
```
zeenix, line 80: * Addition of this type should either be done as a separate commit or as part of `Command` type addition (with the commit message adjusted appropriately).
* Please make sure only parts that are needed by tests, are declared `pub`. Others should be `pub(crate)`. If we add a `pub` API, please mark it as `#[doc(hidden)]`.

zbus/src/connection/socket/unixexec.rs
```diff
@@ -0,0 +1,118 @@
...
+
+#[cfg(feature = "tokio")]
+use tokio::io::{AsyncReadExt, ReadBuf};
+
+use crate::process::{Child, ChildStdin, ChildStdout};
+
+use super::{ReadHalf, RecvmsgResult, Socket, Split, WriteHalf};
+
+#[derive(Debug)]
+pub struct UnixExecStream {
```
zeenix, line 12: * I'm wondering if this is really unix-specific? AFAICT, it's not but only currently used for a unix-specific address type. If so, I'd give it a more generic name but for now only enable the mod/use for unix target. E.g `CommandStream`?
* Would be good to split addition of this socket type in a separate commit (especially if we make it more generic) that comes before its use in `connection::Builder`.
* I don't think this should be public API. Users just need a `Connection` given a unixexec address, and we can handle all that internally.
* Some docs would be good in any case. :)

zbus/src/connection/socket/unixexec.rs
```diff
@@ -0,0 +1,118 @@
...
+    async fn sendmsg(
+        &mut self,
+        buf: &[u8],
+        #[cfg(unix)] fds: &[BorrowedFd<'_>],
+    ) -> io::Result<usize> {
+        #[cfg(unix)]
+        if !fds.is_empty() {
+            return Err(io::Error::new(
+                io::ErrorKind::InvalidInput,
+                "fds cannot be sent with a unixexec stream",
```
zeenix, line 107: these errors will need to be updated if we renamed this type. Perhaps the struct keeps a `name: &'static str` name` field?

zbus/src/connection/socket/unixexec.rs
```diff
@@ -0,0 +1,118 @@
...
+            let ret = (ret, vec![]);
+
+            ret
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl WriteHalf for ChildStdin {
+    #[cfg(not(feature = "tokio"))]
```
zeenix, line 75: If each method will have a different impl for different backend, it'd be best to feature-gate the impl rather. Please see other socket impls for an example (e.g `tcp` mod).

zeenix (changes requested):
Starting to really get in a great shape. Mostly some nitpicks about git hygiene and naming etc. :)

zbus/src/connection/socket/command.rs
```diff
@@ -0,0 +1,126 @@
...
+#[cfg(feature = "tokio")]
+use tokio::io::{AsyncReadExt, ReadBuf};
+
+use crate::process::{Child, ChildStdin, ChildStdout};
+
+use super::{ReadHalf, RecvmsgResult, Socket, Split, WriteHalf};
+
+/// A Command stream socket.
+///
+/// This socket may be used for communicating with a spawned child process via
```
zeenix, line 13: Let's not give an illusion of choice here: `may be used for communicating` -> `communicates`.  :)

zbus/src/connection/socket/command.rs
```diff
@@ -0,0 +1,126 @@
...
+/// its standard input and output streams.
+#[derive(Debug)]
+pub(crate) struct CommandStream {
```
zeenix, line 16: `Stream` is quite a bit overloaded term in async Rust world and certainly also in zbus code. Let's just call it `Command` (the `socket` mod will ensure clarity, as long as it's not re-exported outside of it and it shouldn't IMO).

zbus/src/abstractions/process.rs
```diff
@@ -1,5 +1,82 @@
...
+/// A wrapper around the command API of the underlying async runtime.
+#[allow(unused)]
+pub struct Command(
+    #[cfg(not(feature = "tokio"))] async_process::Command,
+    #[cfg(feature = "tokio")] tokio::process::Command,
+);
+
+impl Command {
+    /// Constructs a new `Command` for launching the program at path `program`.
+    #[allow(unused)]
```
zeenix: It's fine to have unused warnings in the intermediate commits, so please remove all `allow(unused)`.

zbus/src/abstractions/process.rs
```diff
@@ -78,6 +78,101 @@ impl Command {
...
 }
 
+/// A wrapper around the child API of the underlying async runtime.
+pub(crate) struct Child(
+    #[allow(unused)]
+    #[cfg(not(feature = "tokio"))]
+    async_process::Child,
+    #[allow(unused)]
+    #[cfg(feature = "tokio")]
+    tokio::process::Child,
```
zeenix: Do we really need a new type here? AFAICT the APIs are identical, as evident from the `take_stdio` impl that identical for both backends. Moreover, you still end up with different `ReadHalf` and `WriteHalf` impls for each backend in the end, so we don't gain anything from this abstraction.

zeenix (changes requested):
Other than that, getting very close to me merged. While you rework, if could you please also add a bit of context to each commit message in the details section, that'd be perfect?

zbus/src/connection/socket/mod.rs
```diff
@@ -8,6 +8,8 @@ pub use split::{BoxedSplit, Split};
 
 #[cfg(unix)]
 mod command;
+#[cfg(unix)]
+pub(crate) use command::Command;
```
zeenix, line 12: This ended up in the wrong commit. Please move it to the previous one that adds the type.

zeenix (changes requested):
Ok, looks like just one last thing. :)

zbus/src/connection/socket/mod.rs
vially: Good catch. Should be fixed in latest revision.

zeenix: Thanks.

zeenix:
@vially I'm terribly sorry but I overlooked something important in my previous reviews: tests. Any chance you can add a testcase for `unixexec`? Otherwise, there is a good chance I (or someone else) will break the support. It will be good to see that your PR actually works e2e.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,72 @@
...
+            .spawn()?
+            .borrow_mut()
+            .try_into()
+    }
+}
+
+impl Display for UnixExec {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("unixexec:")?;
+        encode_percents(f, self.path.as_os_str().as_bytes())
```
elmarco, line 57: this is not idempotent. You will need to refactor the handling of arguments: `Vec<String>` is not appropriate.

@zeenix
It probably makes more sense to rebase on top of https://github.com/dbus2/zbus/pull/989 instead.

zeenix: @elmarco sure but could you do the rebase for @vially ?

elmarco: The code in #989 leave the handling of missing argv0..N to the caller. I think this makes more sense. "no further argvY for Y > X are taken into account." doesn't mean they are not parsed.

elmarco: it's not explicit about duplicate keys either afaict

elmarco: > @elmarco sure but could you do the rebase for @vially ?

@vially what do you prefer?

vially: Thanks for the review. I will rebase on top of #989, address the feedback and add some tests then.

vially: @elmarco I have rebased on top of #989, please take a look (especially around the changes to `zbus/src/address/transport/unixexec.rs`, as I'm still in the process of learning Rust and I'm not very familiar with the the `Cow` types).

Still planning to add some tests (once I figure out a good way to test the transport).

elmarco: thanks @vially, the ::address modification and usage look good to me!

vially: @elmarco I've pushed another revision where the `From<&Unixexec>` implementation has been moved to `::process`.

@zeenix Can you give me some pointers on how to write some tests for this? I've looked a bit into the `e2e.rs` tests but I haven't found any *easy* way to test this (without significant refactorings in that file which seemed messy and unnecessary).

zeenix: @vially Yeah, I don't think you should mess around with `e2e`. I think where you've currently put the connection test (right in the module) is fine actually. Alternatively it can also go into a separate new file/mod under `tests` dir (which as of yesterday, contains more than `e2e`). I'd prefer the latter since it's not a unit test.

As for how/what to test, I think it'd be sufficient to establish a connection with the bus, perhaps through `systemd-stdio-bridge`. I'm not sure if `systmd-stdio-bridge` is installed on the CI but if it's not, I think we can install it [like we install dbus](https://github.com/dbus2/zbus/blob/bfeb096f3ade97d48b04597f478e303a917e9e3f/.github/workflows/rust.yml#L76).

zeenix: > @elmarco I've pushed another revision where the `From<&Unixexec>` implementation has been moved to `::process`.

cool. Perhaps the last rebase didn't go well cause I see commits from main in this PR, while they don't show in @elmarco's PR branch. :thinking:

zeenix: > Thanks for the feedback.

Np.

> I have added a basic `unixexec` connection test case using `systemd-stdio-bridge`.

Awesome!

> I don't have an easy way to run the CI pipeline to see if `systemd-stdio-bridge` is installed, but if the pipeline fails I'll get it fixed.

That's no issue. Github will run it after I approve it (approval is only needed for first time contributors). The issue is that currently this branch is in conflict with main so CI won't even attempt to build it.

> I have tried moving the socket test into the `tests` directory too, but it would require making the `socket::command::connect` module/method public. If you think it's worth it I can fix that in the next revision.

Sure, that can wait.

vially: I've just rebased on `main` which fixed the merge conflicts.

zbus/src/abstractions/process.rs
```diff
@@ -94,3 +96,21 @@ where
 {
     Command::new(program).args(args).output().await
 }
+
+impl From<&Unixexec<'_>> for Command {
```
zeenix, line 100: This isn't exactly a conversion. `Unixexec` should provide a method to get the corresponding `Command`.

zeenix (changes requested):
LGTM otherwise. Thanks for your patience here! 

BTW, "🔌" would be a good emoji for the internal socket-type addition commits. :)

zbus/src/abstractions/process.rs
vially: True, although it was [previously argued](https://github.com/dbus2/zbus/pull/976#discussion_r1766701996) that this conversion should not live in `address::transport::unixexec` either (which makes sense too):

> [...] Though ::address is really standalone, so `crate::process::Command` part should be moved elsewhere. in ::process perhaps?

Another option would be to just move the logic into the `connection::socket::command::connect` function (which is the only place where this logic is needed).

In any case, it's fine by me either way so please let me know which one would be preferred:
a) move it into a method on `Unixexec` (with the caveat that it introduces some coupling between the `address` and `process` modules)
b) move into the `connection::socket::command::connect` method

PS: If the `a` method is preferred, any suggestion for the method name? I'm still somewhat new to Rust and I haven't familiarized myself with the naming conventions yet.

zeenix (changes requested):

zbus/src/abstractions/process.rs
```diff
@@ -94,3 +96,21 @@ where
...
+
+        if let Some(arg0) = unixexec.find_arg_by_key(0) {
+            command.arg0(arg0.as_ref());
+        }
+
+        let mut arg_index = 1;
+        while let Some(arg) = unixexec.find_arg_by_key(arg_index) {
+            command.arg(arg.as_ref());
+            arg_index += 1;
+        }
```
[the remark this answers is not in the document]
vially: Agree and I would like to get rid of it if possible. However, I haven't found a more idiomatic way to rewrite it (by iterating over `args`) that:
- gets rid of duplicates
- stops at the first "gap" in the index array

While getting rid of duplicates could be done by converting `args` into a `HashMap` first, I'm not sure what's a good way of detecting the first "gap".

TLDR: I haven't found a cleaner way to rewrite it (that's still spec-compliant) but I'm open to suggestions for improving it.

zbus/src/abstractions/process.rs
vially: Thanks, fixed in latest fixup commit (will do a final rebase once all review feedback is addressed).

zbus/src/abstractions/process.rs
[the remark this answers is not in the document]
zeenix: > However, I haven't found a more idiomatic way to rewrite it (by iterating over `args`) that:
> 
> * gets rid of duplicates

Hmm.. I think that's a more general issue since other address types also use key and value. If we want to solve this, we need a more general solution here. However, as I wrote on #989, we should just ignore duplicates and take the last value.

> * stops at the first "gap" in the index array

Ah right, the spec explicitly says how to handle that. Still, I think the code will be nicer if we do it through iterating over all args:

```rust
        let mut last_index = 0;
        for (i, arg) in unixexec.args() {
            if i != last_index + 1 {
                break;
            }

            command.arg(arg.as_ref());
            last_index = i;
        }
```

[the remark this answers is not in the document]
zeenix: The above code assumes that we'll change arg0 to be treated separately (separate field+getter) as I suggested [here](https://github.com/dbus2/zbus/pull/989#discussion_r1784468781).

zbus/src/abstractions/process.rs
zeenix: > will do a final rebase once all review feedback is addressed

I think I may have misunderstood this. You mean feedback addressed in #989? If so, I don't think you need to wait for that. You can already make the only change we need here as part of this PR: arg0 to be treated differently. We now have #1040 awaiting on this and both PRs now have conflicts with main so I can't run the CI on it to test it.

It's pitty that this PR got so delayed. There was a good reason I originally suggested that @elmarco should base his PR on yours, since this one is much less intrusive and simpler but oh well. :)

zbus/src/abstractions/process.rs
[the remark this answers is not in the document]
vially: > ```rust
> let mut last_index = 0;
> for (i, arg) in unixexec.args() {
>     if i != last_index + 1 {
>         break;
>     }
> 
>     command.arg(arg.as_ref());
>     last_index = i;
> }
> ```

The downsides of this code is that a) it assumes the `args()` are sorted (which is somewhat reasonable) and b) it doesn't handle duplicates. Therefore, I've adjusted it a bit in the latest revision so that it handles duplicates.

But (to me at least) the revised code is less readable while still intrusive to the details of a `Unixexec` and C-like, to the point where I"m not sure if it's actually an improvement over the previous implementation.

This is all a bit subjective and the end result should be the same, so it's fine by me either way. I'm still open to further refactor and improve it though, if you have other suggestions.

zbus/src/abstractions/process.rs
vially: > the only change we need here as part of this PR: arg0 to be treated differently.

I've fixed this in the latest revision. The only point still up for discussion is the implementation of `Command::for_unixexec` which is still a bit ugly, but other than that all remaining feedback has been addressed.

zbus/src/abstractions/process.rs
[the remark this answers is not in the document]
zeenix: > The downsides of this code is that a) it assumes that `args()` are sorted (which is somewhat reasonable)

I think they should be sorted, regardless of this code. There is `Vec::sort_by` to ensure that. The performance penalty will be 0 if the address string had the args sorted already, which they really should be. If people don't put args in order in the address, they really deserve any small performance penalty from the sorting. :)

But this should happen in the construction/parsing of `Unixexec`. I see that you've modified the `Command` abstraction to be more specific to `Unixexec` needs. I don't think that's necessary. Except for `for_unixexec` method, `Command` should provide a generic API: `arg` builder method to push each arg into it and `for_unixexec` should make use of that.

> and b) it doesn't handle duplicates. Therefore, I've adjusted it a bit in the latest revision so that it handles duplicates.

We already talked about duplicates above. As I wrote that needs a general solution as it's not specific to unixexec. However, if you really want to solve it only for unixexec, you can filter the duplicates in `Unixexec` construction (just like sorting). There is also `Vec::dedup_by`.

Thinking about it, you could as well take care of gaps in the same place. Doing all the processing upfront, will also ensure we only do all this once and any further uses of a parsed `Unixexec` can just assume everything is in order.

> But (to me at least) the revised code is less readable while still intrusive to the details of a `Unixexec` and C-like,

I can understand completely that you find it less readable but it's not intrusive since it's not asking for specific indexes but rather just getting a slice of all args and then iterating over them. Also, C doesn't have for loops that iterate over a collection or tuples, so not sure what's C-like about it. :)

You're right that it will be assuming there are no duplicates and args being sorted etc. However, those assumptions could be part of the API-contract of `Unixexec::args` (could even be documented).

> This is all a bit subjective and the end result should be the same, so it's fine by me either way. I'm still open to further refactor and improve it though, if you have other suggestions.

Sure and you can always disagree with me. However, in the future please feel free to be blunt about it. :)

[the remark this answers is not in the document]
vially: > However, in the future please feel free to be blunt about it. :)

:+1: While I may disagree with small bits and pieces, I still appreciate the feedback so please keep it coming :)

Pushed another revision which moves the deduplication and the removal of the gaps during the `Unixexec` construction.

This has the nice side-effect that it simplified the `Command::for_unixexec` implementation which is pretty clean and readable now.

zbus/src/address/transport/unixexec.rs
```diff
@@ -48,7 +59,7 @@ impl<'a> TryFrom<&'a Address<'a>> for Unixexec<'a> {
 
     fn try_from(s: &'a Address<'a>) -> Result<Self> {
         let mut path = None;
-        let mut argv = Vec::new();
+        let mut args = HashMap::new();
```
zeenix, line 62: This brings in extras allocation. Why didn't you make use of the API I recommended that gives you de-duplication/resorting for free in the ideal case, even?

zeenix (changes requested):
LGTM otherwise.

zbus/src/address/transport/unixexec.rs
vially: If you mean using the `Vec::dedup_by` method, that would keep the *first* duplicated value and you mentioned you would like to only keep the *last* one.

zeenix: True. I'm sorry for the miscommunication. My main point was that we just pick one of the values if there are duplicates. The last one makes most sense but it's not necessary.

vially: Makes sense, I just wasn't sure if the "last value" was a hard requirement. I've pushed another revision that chooses the first value instead which got rid of the extra `HashMap` allocation.

zeenix: Nice! again, I'm sorry for not communicating well.

zeenix:
@vially Thank you so much for your extreme patience and hard work on this. However, I've to ask for one last thing: I had to revert the new address API as I believe there was one fundamental flaw in its design that I didn't catch in time (https://github.com/dbus2/zbus/pull/1068). 

I understand if you'd not have the time and energy to rebase your work yet again. If so, please let me know and I'll rebase for you soon.

vially:
Sorry for the late response on this, got a bit caught up with other work. I'm planning to rebase this over the weekend but if you want to merge it sooner feel free to rebase it anytime. It's fine by me either way.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,114 @@
...
+use super::encode_percents;
+
+/// `unixexec:` D-Bus transport.
+///
+/// <https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec>
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Unixexec {
+    path: PathBuf,
+    arg0: Option<OsString>,
+    args: Vec<String>,
```
zeenix, line 17: The both should use the same String type, no?

zeenix (changes requested):
LGTM otherwise. Thanks for your extreme perseverance with this. :pray:

zbus/src/address/transport/unixexec.rs
vially: Good catch! Fixed in latest revision.

zeenix: Thanks but seems you went for the opposite of what I'd have expected. :) Wouldn't `OsString` not be better here? :thinking:

zeenix:
Don't worry about fuzzing failing btw, it's just some issue in the gvariant code.

zbus/src/address/transport/unixexec.rs
vially: No worries at all. Pushed a new revision where all the `Unixexec` args are now `OsString`s.

zeenix (approved):

vially:
:tada: @zeenix Thanks a lot for your support, patience and guidance in getting this merged. It took a bit more work than I originally anticipated but it has come a long way since my initial hacky approach of piping the data through a bunch of extra sockets.

zeenix:
> 🎉 @zeenix Thanks a lot for your support, patience and guidance in getting this merged.

No! Thank **you** for your patience and your will to get the work done with the best quality. :pray:

PR #985, opened 2024-09-12
📝 zb: correct wrong documentation links

fmartinsons:


fmartinsons:
Note that I had also the following warnings during the doc build
```
warning: unresolved link to `crate::connection::Builder::server`
  --> zbus/src/guid.rs:29:41
   |
29 |     /// [`connection::Builder::server`](crate::connection::Builder::server).
   |                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `Builder` has no field or associated item named `server`
   |
   = note: `#[warn(rustdoc::broken_intra_doc_links)]` on by default

warning: unresolved link to `Connection::set_unique_name`
   --> zbus/src/connection/mod.rs:805:11
    |
805 |     /// [`Connection::set_unique_name`].
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^ the struct `Connection` has no field or associated item named `set_unique_name`

warning: `zbus` (lib doc) generated 2 warnings
```

But I didn't know how to fix these (set_unique_name only exists if a feature is set and I don't know what crate::connection::Builder::server should refer too

zeenix:
> Note that I had also the following warnings during the doc build

That's very strange, we even have CI setup to fail on any warnings from rustdoc. :thinking: I'll check later why it isn't doing its job.

> But I didn't know how to fix these (set_unique_name only exists if a feature is set and I don't know what crate::connection::Builder::server should refer too

In general, `git log -S IDENTIFIER` is typically a good way to find the appropriate commit. Since we try to keep commits atomic, there is a good chance there is enough context in the commit itself or the surrounding commits (or the PR the commit was part of).

> set_unique_name only exists if a feature is set

Hmm.. the docs should be building with all features enabled:

```toml
[package.metadata.docs.rs]
all-features = true
targets = ["x86_64-unknown-linux-gnu"]
```

so this should be a warning that you see. :thinking: 

> I don't know what crate::connection::Builder::server should refer too

`crate::connection::Builder::server` exists but is also only available on a feature being enabled (`p2p`). Again, this is also not something you should be seeing with the above configuration applied through `Cargo.toml`. So this is strange. I'll have a look soon to check what's going on.

zbus/src/blocking/proxy/mod.rs
```diff
@@ -159,7 +159,7 @@ impl<'a> Proxy<'a> {
 
     /// Introspect the associated object, and return the XML description.
     ///
-    /// See the [xml](xml/index.html) module for parsing the result.
+    /// See the [xml](../../../zbus_xml/index.html) module for parsing the result.
```
zeenix, line 162: This won't work for docs.rs. We need external link to docs.rs. The existing link is from the time when `zbus_xml` was just a mod of `zbus`.

zbus/src/proxy/mod.rs
```diff
@@ -640,7 +640,7 @@ impl<'a> Proxy<'a> {
 
     /// Introspect the associated object, and return the XML description.
     ///
-    /// See the [xml](xml/index.html) module for parsing the
+    /// See the [xml](../../zbus_xml/index.html) module for parsing the
```
zeenix, line 643: same here.

zeenix (changes requested):
Thanks.

zbus/src/blocking/proxy/mod.rs
fmartinsons: Sorry, I just mimicked what is done elsewhere, for example [here](https://github.com/dbus2/zbus/blob/main/zbus/src/error.rs#L28) for zvariant or zbus_names crates

What is the syntax to make the correct links ?

zeenix: Oh! I didn't know/remember docs.rs can apparently [handle this just fine](https://docs.rs/zbus/latest/zbus/enum.Error.html#variant.Variant). :open_mouth: Never mind then. :)

zeenix:
> > > I don't know what crate::connection::Builder::server should refer too
> > 
> > 
> > `crate::connection::Builder::server` exists but is also only available on a feature being enabled (`p2p`). Again, this is also not something you should be seeing with the above configuration applied through `Cargo.toml`. So this is strange. I'll have a look soon to check what's going on.
> 
> For what it worth, I simply clone the repo and do a "cargo doc --open" on it (rust 1.79.0 129f3b996 2024-06-10))

Ah, I think then I know. The configuration is in the subcrate Cargo.toml, not in the root. In the CI, we explicitly pass `--all-features`. I think most likely the solution is simply to copy&paste the config to the root Cargo.toml and remove the `--all-features` from the CI command. If you could do that as part of this PR, that would be great.

zbus/src/blocking/proxy/mod.rs
zeenix: ah wait, the link to zvariant **is** broken even. So that doesn't actually work.

> What is the syntax to make the correct links ?

Simply link to the https://docs.rs/zbus_xml . We need to do that same for the zvariant link.

fmartinsons: Ah, sorry, I didn't see that zvariant link was also broken. I'll fix these too tomorrow.
Thank you for your quick replies.

fmartinsons:
> 
> Ah, I think then I know. The configuration is in the subcrate Cargo.toml, not in the root. In the CI, we explicitly pass `--all-features`. I think most likely the solution is simply to copy&paste the config to the root Cargo.toml and remove the `--all-features` from the CI command. If you could do that as part of this PR, that would be great.

I tried to simply copy/paste but got error
```
fmartinsons@work-pc:~/rust_projects/zbus$ cargo --locked doc --open
error: missing field `name`
  --> Cargo.toml:18:2
   |
18 | [package.metadata.docs.rs]
   |  ^^^^^^^
   |
```

I tried to use `[workspace.metadata.docs.rs]` and simply `[metadata.docs.rs]` but it had no effects (all features was not enabled) so I let things as they are and simply build with the same command as the CI

zeenix:
> I tried to use `[workspace.metadata.docs.rs]` and simply `[metadata.docs.rs]` but it had no effects (all features was not enabled) so I let things as they are and simply build with the same command as the CI

:shrug: I guess that's probably what I tried at some point as well and forgot about it. :)

zbus/src/blocking/proxy/mod.rs
```diff
@@ -159,7 +159,7 @@ impl<'a> Proxy<'a> {
 
     /// Introspect the associated object, and return the XML description.
     ///
-    /// See the [xml](xml/index.html) module for parsing the result.
+    /// See the [xml](https://docs.rs/zbus_xml) module for parsing the result.
```
zeenix, line 162: oh and it's not a module anymore but a separate crate. :) Might as well correct that too. :)

zeenix (changes requested):

zeenix (approved):
:+1:

Issue #807, opened 2024-05-23
zb: Re-export zvariant types & macros directly


zeenix:
I don't see why users have to access (or even know about in most cases) `zvariant` (as in `zbus::zvariant`). We should just directly export the relevant API from `zvariant` under `zbus`. This will hopefully also further discourage people from using `zvariant` directly and sometimes ending up with incompatible combo of `zbus` and `zvariant`.

zeenix:
LOL, "resolve https://github.com/dbus2/zbus/issues/807" in commit message of 842e988067a37a5ae433749d68fa8f55ced3eabe ended up closing this.

Although while implementing this, I realized that exposing all the API directly under zbus really makes the crate root very croweded, with most types there foreign to zbus. So I'm no longer sure, this is a great idea. :thinking:

zeenix:
🤖 ChatGPT on behalf of zeenix: Addressed by the 6.0 crate merge: [the zbus root](https://github.com/z-galaxy/zbus/blob/33dc2a77d4088232430269a645e1830405638b1b/zbus/src/lib.rs#L84-L107) directly re-exports the wire types and derives (including Value, Type, OwnedValue and signature!), and the old `zbus::zvariant` path is a deprecated compatibility module. Closing.

Issue #820, opened 2024-06-02
Convert `zvariant_utils` macros to proc macros


zeenix:
At least `def_attrs`, which is quite complex for `macro_rules`. I was trying to modify it to allow multiple lists but it turned out to be a whole day of cleaning up the macro and still can't figure it out. :angry: 

@turbocool3r Any chance you'll be able and willing to help with this?

turbocool3r:
Sure, would you prefer if I add support for multiple lists or convert it to a proc macro?

zeenix:
> Sure

Awesome! :+1:

> would you prefer if I add support for multiple lists or convert it to a proc macro?

We need both actually but please prioritize conversion to proc macro first please and then it'd be easier for me (or anyone else) to add multiple lists support, in case you run out of time.

The use case for multiple lists is #807, where we would want zvariant_derive macros to also using `zbus` as the attribute name, in addition to `zvariant`.

BTW, don't forget to bump the major version of `zvariant_utils` and bump the required version in using crates. :)

zeenix:
@turbocool3r Hey, any luck?

turbocool3r:
@zeenix hey, didn't have a lot of free time lately. Hopefully will get to it this weekend

zeenix:
The thing is that we're now preparing for 5.0 and it would be a perfect time for #807 but that's a bit difficult with these macros not supporting a list of attributes. I did spend a weekend trying to add this but I found the macro code to be very difficult. Since you're the author of these macros, I was hoping it would be easi**er** task for you to handle this. As I wrote before, conversion to proc macros can wait.

However, if you don't see yourself having any time for this any time soon, that's perfectly understandable (it's your time after all) but I'd appreciate a more realistic answer so I can plan things accordingly. :pray:

turbocool3r:
Hey, sorry for the long delay, totally forgot about this. Sure, I'll look into it this evening. 👍

turbocool3r:
So I took a look and it really looks like adding support for lists is much easier than rewriting the `def_attrs` macro as a proc macro. That'll also need a new separate crate for utils derive macro (`zvariant_derive_derive`?).

I've written a patch to add support for multiple lists here https://github.com/turbocool3r/zbus/commit/0ec3e83c6bbbfb63470df9ee27c131626c2dcffc. It's quite small and simple, the macro is simplified even more now. Can you take a look? If you still prefer a proc macro, I'll try to make it.

zeenix:
[...]

PR #986, opened 2024-09-13
zv: Multiple lists support for derive macros.

turbocool3r:
This PR adds support for multiple attribute list identifiers to `def_attrs!` macro from `zvariant_utils` and uses this in `zvariant_derive` to avoid mentioning `zvariant` when using macros out of `zbus` (as described in #820).

### DIFF
--- a/zvariant_derive/src/lib.rs
+++ b/zvariant_derive/src/lib.rs
@@ -159,7 +159,7 @@ mod value;
 /// [`Serialize`]: https://docs.serde.rs/serde/trait.Serialize.html
 /// [`Deserialize`]: https://docs.serde.rs/serde/de/trait.Deserialize.html
 /// [serde_repr]: https://crates.io/crates/serde_repr
-#[proc_macro_derive(Type, attributes(zvariant))]
+#[proc_macro_derive(Type, attributes(zbus, zvariant))]
 pub fn type_macro_derive(input: TokenStream) -> TokenStream {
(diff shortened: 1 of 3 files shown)


turbocool3r:
Looks like it's ready for review

zeenix:
Cool. Still no emojis? :crying_cat_face: Could you please add those, it helps me as a very visual person to follow the history better and especially in writing release notes from the commit log. Emoji-less commits also look out of place in our git history now a days. :)

zeenix:
Oh and `zu` is the prefix I've been using for `zvariant_utils`.

zeenix (commented):
Awesome. Functionality wise, it looks great. Only nits about commit log of last commit. :laughing: 

* Could just be "zd: Allow `zbus` attribute in macros". This will fit in the 50 chars limit and is sufficient.
* It would be good to mention that this will make it possible to resolve #807.

zeenix:
> Looks like it's ready for review

Indeed. Functionality wise, this is ready for merge. :)

turbocool3r:
Just reworded everything, added emojis. Not sure if these are the best choice... I was only familiar with 🍻 gitmoji before 😄

zeenix:
> Just reworded everything, added emojis. Not sure if these are the best choice... 

The're good, thanks! :) 

> I was only familiar with 🍻 gitmoji before 😄

I think the only person to have used that emoji so far, is @elmarco (85eb406f6363eabc8a8d6edca5aafd838675fae9).

Issue #982, opened 2024-09-10
Remove ::address  


elmarco:
zbus itself needs to parse dbus addresses, but 99% of users will simply use the default buses, or connect with a string.

Hence, zbus shouldn't provide` zbus::address`. 

Furthermore, parsing a dbus address is a very different job from handling the connection itself. There is not much to share.

zbus can offload that specialized responsibility to an external crate, improving code maintainability, reusability, and overall focus.

I propose to rebase #562 and cleanup some zbus-related code.

zeenix:
As I repeatedly explained on the Matrix channel, I don't want to talk about another crate here (that will happen later for sure). Please edit the description to **solely** focus on the API changes you want to see and their benefits.

elmarco:
> As I repeatedly explained on the Matrix channel, I don't want to talk about another crate here (that will happen later for sure). Please edit the description to **solely** focus on the API changes you want to see and their benefits.

Drop ::address is quite explicit. zbus doesn't provide any dbus address parsing API. An external crate (really I can't name it?) can provide the differences provided above with ::address API.

zeenix:
I gave you a clear path towards achieving your goals here but you want to focus on zbus using your crate, which I told you is not going to happen in the beginning.

elmarco:
So what is the plan to get there?

zeenix:
> So what is the plan to get there?

Copy-pasting from #983:

Since you wouldn't budge, I had a look into dbus-addr. Looks mostly good so let's use it. However, I'll want:

* co-ownership so I can maintain it when you disappear for months. :)
* move into zbus space (we'll need to coordinate that).
* renamed to zbus-address, so it's clear that it's part of the project and people don't hate us for having yet another dep. :)

Also some comments:

* Can you please make use of `doc-comment` crate to keep the README and crate-level docs in sync, like we do in zbus crates?
* Add an example code in the README, with a few different examples.

zeenix:
@elmarco so you happy with the above plan then?

elmarco:
> @elmarco so you happy with the above plan then?

sounds good, I'll try to work on it by the end of the week. thanks!

zeenix:
Thinking a lot more about this, I realised I want to go in the complete opposite direction: https://github.com/dbus2/zbus/issues/1006 . I know you'll likely not agree but if you did any maintenance work, you'd know the pain and extra work splitting of crates bring.

[...]

Issue #304, opened 2023-01-06
Affect on the binary size


zeenix:
In GitLab by @MaxVerevkin on Jan 6, 2023, 20:47

From one of my projects:

```sh
$ cargo bloat --crates --release
 File  .text     Size Crate
 7.9%  18.8%   1.8MiB std
 6.4%  15.2%   1.4MiB zbus
 4.0%   9.4% 898.1KiB zvariant
 3.3%   7.9% 754.3KiB i3status_rs
 3.0%   7.0% 670.3KiB serde
 2.1%   5.0% 480.7KiB reqwest
 1.8%   4.3% 408.7KiB tokio
```

In this example, `zbus` + `zvariant` are responsible for almost 25% of .text section size.

I thought this issue can be used as a place to collect info and ideas on how to improve this.

MaxVerevkin:
Today I tried to use git version of `zbus` in `i3status-rs` to get an idea of how many breaking changes there are (not a lot :) ) and the binary increased in size by ~2M (both stripped and not).

zeenix:
> the binary increased in size by ~2M (both stripped and not)

I know. :( I am hoping #493 will help here.

zeenix:
So I sat down with @sdroege at the GNOME Rust hackfest yesterday and looked at what could possibly be improved, while looking at the output of `cargo bloat` in [busd](https://github.com/dbus2/busd) (as an example). Here are some thoughts from that:

* `Interface::introspect_to_writer` appears for most interfaces being implemented.
  * The problem is number of `format!` calls and `format` macro generating a lot of code.
      * My first thought was to just convert the macros to create string literals instead but the issue is that we need some parameters that we don't have access to in the proc macro: indentation level and `Type::signature()`. Now that I'm writing this, this perhaps shouldn't be too hard to workaround (famous last words :laughing:).
      * We could look into using `ufmt` crate but it seems to be not well-maintained and hasn't seen a release for 2 years.
  *  `zbus::message::builder::Builder::build`
    * Remove the closure use.
  * `tracing`
    * Make it an optional dep #331 (#712). This will not help most people in the end but at least zbus-based libraries can then also provide this feature and make their binary size look smaller (it's mostly about the show anyways tbh).
  * Some invisible closures that `cargo bloat` talks about in various places. :shrug:

sdroege:
> Remove the closure use.

[...]

PR #989, opened 2024-09-15
Rewrite ::address (import from dbus_addr) 

elmarco:
Based on the discussion from https://github.com/dbus2/zbus/issues/982, import dbus_addr as zbus_address and rebase https://github.com/dbus2/zbus/pull/562 to make use of it.


zeenix:
Thanks! I'll have a look soon. In the meantime, could you please rebase this on top of https://github.com/dbus2/zbus/pull/976? That one is just waiting for a test case to be added.

zbus_address/src/address.rs
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /// Transport connection details
+    pub fn transport(&self) -> Result<Transport<'_>> {
+        self.try_into()
+    }
+
+    fn validate(&self) -> Result<()> {
+        self.transport()?;
+        let mut set = HashSet::new();
```
zeenix, line 35: So we go through the trouble of avoiding allocations in the `Address` (even though the 99.99% cases we can't avoid allocation) only to allocate for validation?

I'm not too sure anymore if lifetimes are really worth the effort here.

zeenix (changes requested):
You'll need to change some fundamental design here to make it acceptable to me, sorry.

zbus/src/error.rs
```diff
@@ -61,6 +62,8 @@ pub enum Error {
     InvalidSerial,
     /// The given interface already exists at the given path.
     InterfaceExists(InterfaceName<'static>, ObjectPath<'static>),
+    /// Invalid D-Bus server address.
+    DBusAddress(AddressError),
```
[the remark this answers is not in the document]
elmarco: We already discussed this. All the std rust naming convention is "Addr": [IpAddr](https://doc.rust-lang.org/std/net/enum.IpAddr.html), [Ipv4Addr](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html), [Ipv6Addr](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html), [SocketAddr](https://doc.rust-lang.org/std/net/enum.SocketAddr.html), [SocketAddrV4](https://doc.rust-lang.org/std/net/struct.SocketAddrV4.html), and [SocketAddrV6](https://doc.rust-lang.org/std/net/struct.SocketAddrV6.html). Please don't force me to do something I disagree with. You are free to rename after the merge. I will never agree with that.

zbus_address/README.md
```diff
@@ -0,0 +1,33 @@
+# zbus_address
+
+[![](https://docs.rs/zbus_address/badge.svg)](https://docs.rs/zbus_address/) [![](https://img.shields.io/crates/v/zbus_address)](https://crates.io/crates/zbus_address)
+
+Handle [D-Bus server addresses][dba] parsing.
+
+This is used by [`zbus`]. Other D-Bus crates are also encouraged to use this API in the spirit of cooperation. :)
```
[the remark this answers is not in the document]
elmarco: It's from zbus_names.

zbus_address/README.md
```diff
@@ -0,0 +1,33 @@
...
+```rust,no_run
+use zbus_address::DBusAddr;
+
+let addr: DBusAddr = "unix:path=/tmp/dbus.sock".try_into().unwrap();
+```
+
+# Miscellaneous and caveats on D-Bus addresses
+
+* Assumes values are UTF-8 encoded: this should be clarified in the spec
+  otherwise, fail to read them or use a lossy representation for display.
```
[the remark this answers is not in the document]
elmarco: The spec doesn't say what is the encoding of a dbus address and key=values. ok

zbus_address/README.md
```diff
@@ -0,0 +1,33 @@
...
+* Assumes that empty `key=val` is accepted, so `transport:,,guid=...` is valid.
+
+* Allows key only, so `transport:foo,bar` is ok.
+
+* Accept unknown keys and transports.
+
+**Status:** Stable.
+
+[`zbus`]: https://crates.io/crates/zbus
+[`zbus_macros`]: https://crates.io/crates/zbus_macros
```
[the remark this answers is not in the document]
elmarco: c&p, I'll remove

zbus_address/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+            return DBusAddrList::try_from("autolaunch:");
+
+            #[cfg(target_os = "macos")]
+            return DBusAddrList::try_from("launchd:env=DBUS_LAUNCHD_SESSION_BUS_SOCKET");
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
elmarco: ok

zbus_address/src/address.rs
```diff
@@ -0,0 +1,233 @@
...
+            write!(f, "{k}=")?;
+            v.encode(f)?;
+            first = false;
+        }
+        Ok(())
+    }
+}
+
+/// A trait for objects which can be converted or resolved to one or more [`DBusAddr`] values.
+pub trait ToDBusAddrs<'a> {
```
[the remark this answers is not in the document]
elmarco: Why? with a trait, we can accept address arguments which are strings, or Vec<addr> etc..    This is also how std socket address are handled: https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html

zbus_address/src/address.rs
elmarco: validation can be made optional. alternatively, `validate()` could be improved to avoid allocation. I can add a fixme?

zbus/src/error.rs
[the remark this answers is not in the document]
zeenix: * I'm talking about the name of the enum variant here.
* In case of std API, we don't have any context. This is zbus, and the context of D-Bus is quite clear. There is no need for a prefix to provide context. If there is no prefix, there's no need to shorten `Address`.

> Please don't force me to do something I disagree with. You are free to rename after the merge. I can't agree with your taste, I follow patterns.

That's now how reviews work. With this statment, you're telling me that you don't intend to help maintain this code much and it will be my responsibility to fix all the bugs and answer all questions that come in (as always).

zbus_address/README.md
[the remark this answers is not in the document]
zeenix: > It's from zbus_names README..

touche. :laughing:

zbus_address/src/address.rs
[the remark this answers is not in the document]
zeenix: > Why? with a trait, we can accept address arguments which are strings, or Vec etc..

You can do the same with `TryFrom` impls. if `Vec` is expected, you do a `TryInto` for `AddressList` instead of `Address`. I think our case is more specific than std's.

> This is also how std socket address are handled: https://doc.rust-lang.org/std/net/trait.ToSocketAddrs.html

That API predates TryFrom/TryInto. There's a lot of API in std that's like that only for historical reasons.

zbus_address/src/address.rs
zeenix: > validation can be made optional. alternatively, `validate()` could be improved to avoid allocation. I can add a fixme?

None of that addresses my concerns. Sure, if validation can be allocation free, it'd make me feel much better about the API designed to allow reference types. However, the main issue I see is that for most uses (system and session bus), we'll be allocating and that beats the whole point of lifetimes in types.

However, maybe there is a use case that's quite common where users would be able to avoid allocation, that I'm missing here?

zbus/src/error.rs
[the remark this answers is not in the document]
elmarco: about naming of enum: yes, I didn't want to repeat the argument later.

I value consistency more than you, it seems. 

Address is still unclear for me, even if it comes from a zbus namespace. 

I intend to maintain code that I write. But when I strongly disagree with you on taste (it's only taste here I think, and bikeshedding), I can't be forced. You are the main maintainer, I can't force you either.

zbus_address/src/address.rs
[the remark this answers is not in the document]
elmarco: But Vec isn't expected. The implementation avoids Vec allocation... Why should we force Vec?

zbus_address/src/address.rs
elmarco: "or most uses (system and session bus), we'll be allocating" why?

zeenix: because `std::env::var` allocates. I guess that is not typically true for system but session bus address is almost always taken from the env.

zbus/src/error.rs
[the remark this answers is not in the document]
zeenix: > I can't be forced. You are the main maintainer, I can't force you either.

Well one of us needs to be forced. :) I won't sign off on something I disagree with and as the maintainer, I always get the final say. If I change the names after the merge and you actually do help with maintainance this time around (I have serious doubts based on last many years of experience but I want to believe you), you'll have to deal with the names I choose anyway. What's to be gained by you not already changing the names before the merge?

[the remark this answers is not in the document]
elmarco: Ok.., at least you can recognize I am not as stubborn as you and I made my point clear: you do not follow rust naming convention, you have your own. I strongly disagree with this, I'll keep repeating why you are wrong, but you have the last word.

https://doc.rust-lang.org/std/index.html?search=addr :

Nothing is named "address" (for the same reason vec is not "vector" for instance, there are a lot of common programming short name, addr is one of them).

And for clarilty reasons, when such a common term as "addr" is used in a common namespace, by default, you shouldn't pollute the namespace and use a more descriptive name. Hence `DBusAddr` is the right Rust naming choice.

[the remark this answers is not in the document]
zeenix: > you do not follow rust naming convention

Huh? I explained my reasons for the naming. This is **not** any established Rust convention. Just because std uses some naming convention, it doesn't necessarily make it a convention. We already use `Address` in zbus and I didn't hear any compelling reason to change it.

Anyway, naming isn't something you're likely to convince me on. That's something you must change for this PR to be accepted. Plus I'm becoming sure as I think and discuss more about it that I don't want a split but a module.

zbus_address/src/address.rs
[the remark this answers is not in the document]
zeenix: My thinking was the same as here but see the new comment in that thread. Mark this also as resolved once you've read it. :)

zeenix:
I mainly see only 2 blockers here now:

* The split. As discussed before, I'm not at all convinced a split is worth it in practical terms. We should instead make use of cargo features to make zbus modular instead.
* The naming.

elmarco:
> I mainly see only 2 blockers here now:
> 
>     * The split. As discussed before, I'm not at all convinced a split is worth it in practical terms. We should instead make use of cargo features to make zbus modular instead.
> 
>     * The naming.

I disagree with you one those two points, but if it is what it takes to make progress...

zeenix:
> I disagree with you on those two points

That's well established. :)

> but if it is what it takes to make progress...

I'm sorry but yeah. You've to keep in mind that I did try my best to explain my reasoning in detail. I regret that I failed to convince you in the end.

zeenix:
@elmarco is this ready for review?

elmarco:
> @elmarco is this ready for review?

yes please

zbus/src/address/address_list.rs
```diff
@@ -0,0 +1,80 @@
+use std::{borrow::Cow, fmt};
+
+use super::{Address, Error, Result, ToAddresses};
+
+/// A bus address list.
```
zeenix, line 5: I'm getting a deja-vu asking this but where does this concept come from and (more importantly) where is it used? I am asking because I am unable to find any reference to any such concept in the specification.

zeenix: In any case, it would be good to document it here with link(s) to appropriate docs.

elmarco: When connecting to a server, multiple server addresses can be separated by a semi-colon. The library will then try to connect to the first address and if that fails, it'll try to connect to the next one specified, and so forth. For example

unix:path=/tmp/dbus-test;unix:path=/tmp/dbus-test2

elmarco: The most notable examples are related to autolaunch, typically an address is given and if it doesn't exist a bus is spawned. Of course there are other usages, for fallback handling, especially in network environment. This is not typical afaict.

zeenix: I understand how it can be used or useful. I'm asking for actual uses of it. Does anyone ever use it? The API seems designed around supporting this so this is an important question IMO.

zeenix: OTOH, most of it the address API is sort of niche anyway as we'll have very few direct users of it. So maybe this is not such an important question.

zbus/src/address/percent.rs
```diff
@@ -0,0 +1,206 @@
+use std::{
+    borrow::Cow,
+    ffi::{OsStr, OsString},
+    fmt,
+};
+
+use super::{Error, Result};
+
+pub(crate) trait Encodable {
```
zeenix, line 9: Could you please add some docs to this type. Just a basic description is sufficient.

zbus/src/address/address_list.rs
elmarco: Configurabiliy of dbus server address has always been kind of niche. 99% of the case, you connect to a hardcoded system path :)

zbus/src/address/address.rs
```diff
@@ -0,0 +1,233 @@
...
+        let addr = addr.into();
+        let addr = Self { addr };
+
+        addr.validate()?;
+        Ok(addr)
+    }
+
+    fn validate(&self) -> Result<()> {
+        self.transport()?;
+        let mut set = HashSet::new();
```
zeenix, line 51: Wait, aren't we always allocating then for validation (which always happens)? 🤔 The address strings are typically not that long and typically (99% at least IMO) only parsed once so if the API is designed to avoid allocations, it should at least reduce allocations compared to the existing API.

This was my thinking when I decided not to cater for avoiding allocations in 4.0 address API in the end: if in the end I still have allocations, what's the point?

elmarco: I wrote this code the way I thought would be more efficient for parsing, but I never actually went the whole route and checked against nostd etc. I think it's in general superior than always using Vec and String all over.  But yes, as I told you earlier, there should be an escape solution for assuming an address is already validated, and/or a way to use no_std. Although having this merged in zbus is now harder to work on and guaratee (same as guarateeing the namespace remain clear of other internal dependencies..).

zbus/src/address/percent.rs
elmarco: // A trait for types that can be percent-encoded and written to a [`fmt::Formatter`].
 ok?

zeenix: yes

zbus/src/address/address.rs
zeenix: > I wrote this code the way I thought would be more efficient for parsing, but I never actually went the whole route and checked against nostd etc.

I think you mean noalloc but that's not what we need. It'll be a pretty big challenge to make zbus nostd and especially noalloc. I'm not expecting that at all from this API. I have a vague plan of adding nostd and noalloc features through the use of `heapless` crate but that would be a pretty big project and not relevant here.

> I think it's in general superior than always using Vec and String all over.

In general, yes and that's my point: is the new API really reducing allocations if it always allocate for validation?

> But yes, as I told you earlier, there should be an escape solution for assuming an address is already validated,

I disagree. An address should always be validated, unless built through an `unsafe` API. This goes in line with the new signature API: `zvariant::Signature` is now always valid. I don't think this is a strong argument.

> and/or a way to use no_std (hashset could be replaced quite trivially by a vec from alloc, or a fixed array)

nostd is not the target here. Reduction/avoidance of allocations is, since that's one of the main arguments for the new API here.

> Although having this merged in zbus is now harder to work on and guaratee (same as guarateeing the namespace remain clear of other internal dependencies..).

Not sure I understand but let's focus here: Does the new API proposed here, reduces allocations and if so, how much and if it's worth having all the lifetimes and other complicated parts etc.

elmarco: > Not sure I understand but let's focus here: Does the new API proposed here, reduces allocations and if so, how much and if it's worth having all the lifetimes and other complicated parts etc.

Yes if you skip validation (my intention earlier was obviously behind unsafe.) but I don't think that's the main argument for using this code anyway, almost nobody will care about that, especially if it can't be used directly without the rest of zbus. parsing and and storing result is done through &str references whenever possible.

 There has been a lot of reasons listed earlier, if you wish me to make the list again, I could Not to mention it was written before zbus ::address rewrite, which is fundamental in its existence, otherwise, I might have just done something else instead...

elmarco: also hashset usage is really extra, there is nothing that really prevents addresses from having duplicated keys.. It just seems wrong, because the behaviour isn't described. So if we drop that check, we have no hashset, if that's what worries you about this code.

elmarco: and the lifetime usage is one of the most trivial Rust could offer, it references the given string lifetime. Can't be much simpler, you can agree.

zeenix: > Yes if you skip validation (my intention earlier was obviously behind unsafe.)

But why would we want to skip validation at all? That makes no sense to me to bring that up as an argument.

> but I don't think that's the main argument for using this code anyway

It has been one of your main arguments, sorry. That and support for address list, which we already established is not something anyone really misses. When I told you that AddressList could simply be a Vec<Address>, your only argument against that was that it will force allocation on people. So forcing tiny amount of allocation on use of an extremely niche API, is a big deal but forcing it on everyone is not?

> almost nobody will care about that

That's exactly my point.

> especially if it can't be used directly without the rest of zbus.

I told you how you can make that happen, without a lot of effort using cargo features. In fact you can make use of the API break right now to make it happen already in 5.0.

> There has been a lot of reasons listed earlier, if you wish me to make the list again

This is the only advantage I see [in your list](https://github.com/dbus2/zbus/issues/982#issuecomment-2340819164) that would require an intrusive re-write of the API that you're proposing. All others could be addressed with the current API.

> also hashset usage is really extra, there is nothing that really prevents addresses from having duplicated keys.. It just seems wrong, because the behaviour isn't described. So if we drop that check, we have no hashset, if that's what worries you about this code.

It doesn't worry me one bit. It just contradicts your main argument here.

---

Anyway, if the spec doesn't dictate this, then we shouldn't do this. In case of duplicate keys, let's just take the last value and document this behaviour?

zeenix: > Anyway, if the spec doesn't dictate this, then we shouldn't do this. In case of duplicate keys, let's just take the last value and document this behaviour?

If you agree to this, I believe we can merge this once you've addressed this.

zbus/src/address/transport/systemd.rs
```diff
@@ -0,0 +1,26 @@
+use std::marker::PhantomData;
+
+use super::{Address, Error, KeyValFmt, KeyValFmtAdd, Result};
+
+/// `systemd:` D-Bus transport.
```
zeenix, line 5: Could you please elaborate just a little what this is? Link to some doc would be nice.

zbus/src/address/transport/unix.rs
```diff
@@ -1,129 +1,112 @@
...
+                    break;
+                }
+                "runtime" => {
+                    let v = v.ok_or_else(|| Error::MissingValue(k.into()))?;
+                    let v = decode_percents_str(v)?;
+                    if v != "yes" {
+                        return Err(Error::InvalidValue(k.into()));
+                    }
+                    kind = Some(UnixAddrKind::Runtime);
+                    break;
```
zeenix, line 86: while you adress other things, can you please also add an empty line before the breaks here (and anywhere else you've `break` and `return` at the end of a block of code)? You can see the existing code style about that in the code you're deleting in this file at line 104 (I'm currently looking at the diff of the whole PR branch).

zbus/src/address/transport/unix.rs
```diff
@@ -1,129 +1,112 @@
...
+        };
+        for (k, _) in iter {
+            match k {
+                "path" | "dir" | "tmpdir" | "abstract" | "runtime" => {
+                    return Err(Error::Other("invalid address, only one of `path` `dir` `tmpdir` `abstract` or `runtime` expected".into()));
+                }
+                _ => (),
             }
         }
+        Ok(Unix { kind })
```
zeenix, line 104: empty line before this as well. Won't point out all of them here. Please try your best.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,74 @@
...
+                _ => continue,
+            }
+        }
+
+        let Some(path) = path else {
+            return Err(Error::MissingKey("path".into()));
+        };
+
+        argv.sort_by_key(|(num, _)| *num);
+        Ok(Self { path, argv })
```
zeenix, line 62:
```suggestion
        argv.sort_by_key(|(num, _)| *num);

        Ok(Self { path, argv })
```

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+impl KeyValFmtAdd for Unixexec<'_> {
+    fn key_val_fmt_add<'a: 'b, 'b>(&'a self, mut kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
+        kv = kv.add("path", Some(EncOsStr(self.path())));
+        for (n, arg) in self.argv() {
+            kv = kv.add(Argv(*n), Some(arg));
+        }
+        kv
```
zeenix, line 72:
```suggestion

        kv
```

zeenix (changes requested):
Found a few more things, mostly minor nits.

zbus/src/address/transport.rs
```diff
@@ -0,0 +1,104 @@
...
+
+/// A D-Bus transport.
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Transport<'a> {
+    Unix(unix::Unix<'a>),
+    Launchd(launchd::Launchd<'a>),
+    Systemd(systemd::Systemd<'a>),
+    Tcp(tcp::Tcp<'a>),
+    NonceTcp(nonce_tcp::NonceTcp<'a>),
```
zeenix, line 44: In 4.0 I merged this with Tcp and I think that's the right approach.

This reminds me, have you made an effort to ensure all the changes from 3.x to 4.x (that you didn't object to), have been incorporated?

zbus/src/address/transport.rs
```diff
@@ -0,0 +1,104 @@
...
+mod vsock;
+#[cfg(feature = "vsock_address")]
+pub use vsock::Vsock;
+
+/// A D-Bus transport.
+#[derive(Debug, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Transport<'a> {
+    Unix(unix::Unix<'a>),
+    Launchd(launchd::Launchd<'a>),
```
zeenix, line 41: I understand why you originally decided to make all variants available on all platforms/targets, I can't help but wonder that on a practical level, this will only make the compiler do more work (and also add to that binary size, which we're already infamous for #304) while not giving any real benefit, given that use of address API without the actual transport, is at moment completely hypothetical. Moreover, you've not feature-gated the non-parsing code so it's not even possible to only use the address parsing API after this PR.

Thinking about this now (together with my other comment here), I remember that I had very good reasons for most of my decisions for the address API in 4.0. They were mostly driven by practical needs rather than theoretical ones.

zeenix (changes requested):

zbus/src/address/transport.rs
elmarco: They are treated as different transport in the spec. For consistency, it's best tofollow the spec, otherwise it's surprising and people may wonder if nonce-tcp is actually supported. 

Which changes are you talking about? Yes, I looked at your implementation and we have the same tests.. (except duplicate keys you ask me to drop)

zbus/src/address/transport.rs
elmarco: We are repeating ourself, I thought you already agreed (like many other subjects). The address parsing job is not specific to a platform. We could add more features similar to "vsock_address" if you believe it will help with resulting binary size.

zbus/src/address/transport/systemd.rs
elmarco: added link to spec

zbus/src/address/transport/mod.rs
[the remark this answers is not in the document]
elmarco: ok

zbus/src/address/address.rs
elmarco: drop the duplicate key check. This is a change from 4.0 parsing behaviour.

zbus/src/address/transport.rs
zeenix: > We are repeating ourself,

What's new? :)

> I thought you already agreed (like many other subjects). The address parsing job is not specific to a platform.

As I wrote, it sounds good and correct in theory and that's why I agreed. I'm allowed to have more thoughts that can result in a different conclusion after agreeing. This is a pretty intrusive change with a lot of aspects to it and I've been failing to find one big chunk of time to properly analyze all aspects of it in one go.

I really shouldn't need to explain myself for changing my mind and having doubts about things I agreed with before. If my conclusions are final, we wouldn't have this PR since I told you multiple times to give up on this before. :)

> the address parsing job is not specific to a platform.

Repeating your assertion doesn't help. Please address my concerns above.

> We could add more features similar to "vsock_address" if you believe it will help with resulting binary size.

I don't think that's a good solution as it pushes the burdon on the user to disable those features, as they'll have to be the default if we want all of them supported on all platforms. This is easily automatable so we should and disregard a completely hypothetical advantage.

zbus/src/address/transport.rs
elmarco: I didn't revert.. I wrote this first, based on 3.x. You changed it after, I didn't follow, as I think it could lead to further issues.

zbus/src/address/transport.rs
elmarco: My goal was to make a standalone fully compliant address parser. Now, slowly making zbus-specific this is becoming a thing that diverges from the original goal. You are free to make it more specific in further changes. It will make it harder to reuse outside of zbus..

zeenix: > My goal was to make a standalone fully compliant address parser.

Yeah, "was" being the keyword. I pointed out above, that this PR doesn't enable this use case so that goal hasn't succeeded yet. Also it's not something I care about, for the reasons I've described very clearly in detail already.

> It will make it harder to reuse outside of zbus

Not sure why you still keep bringing that up as some very compelling use case, while you've constantly failed to point to any actual possible users of this. You did so much work and had such long conversation to satisfy a use case that has a very high chance of never realising. :cry:

> You are free to make it more specific in further changes.

I'm sorry but I'm not going to do the work for you. I'm perfectly happy not merging this PR. You need to either address on my concerns or address them before this PR is merged.

---

Kindly stop resolving conversations unilaterally w/o addressing the concerns.

elmarco: added #[cfg] attributes where possible.

zbus/src/address/transport/nonce_tcp.rs
```diff
@@ -0,0 +1,94 @@
...
+impl<'a> NonceTcp<'a> {
+    /// DNS name or IP address.
+    pub fn host(&self) -> Option<&str> {
+        self.host.as_ref().map(|v| v.as_ref())
+    }
+
+    /// Used in a listenable address to configure the interface on which the server will listen:
+    /// either the IP address of one of the local machine's interfaces (most commonly `127.0.0.1`),
+    /// or a DNS name that resolves to one of those IP addresses, or `*` to listen on all interfaces
+    /// simultaneously.
```
zeenix, line 30: Thanks for adding detailed docs. :+1: Not necessarily a blocker but if you could change the docs to follow the convention of:

```
A brief description.

Details in a new paragraph.
```

That would be great.

zbus/src/address/transport/mod.rs
```diff
@@ -1,350 +1,118 @@
...
+    Launchd(launchd::Launchd<'a>),
+    #[cfg(target_os = "linux")]
+    Systemd(systemd::Systemd<'a>),
+    Tcp(tcp::Tcp<'a>),
+    NonceTcp(nonce_tcp::NonceTcp<'a>),
+    Unixexec(unixexec::Unixexec<'a>),
+    Autolaunch(autolaunch::Autolaunch<'a>),
+    #[cfg(feature = "vsock_address")]
+    Vsock(vsock::Vsock<'a>),
+    Other(&'a str),
```
zeenix, line 57: We lost the docs for variants.

zeenix (approved):
Let's get it in but please fix the remaining minor things in a follow-up commit. :pray:

zeenix:
Seems it needs another rebase, so you might as well fix the minor issues while you rebase.

zbus/Cargo.toml
```diff
@@ -35,8 +35,9 @@ async-io = [
   "futures-util/io",
 ]
 tokio = ["dep:tokio"]
-vsock = ["dep:vsock", "dep:async-io"]
-tokio-vsock = ["dep:tokio-vsock", "tokio"]
+vsock = ["dep:vsock", "dep:async-io", "vsock_address"]
+tokio-vsock = ["dep:tokio-vsock", "tokio", "vsock_address"]
+vsock_address = []
```
zeenix, line 40: btw, the chances of anyone **directly** enabling this feature w/o the `vsock` or `tokio-vsock`  features are extremely slim IMO. Given we've plenty of features already, I'd strongly suggest we drop this.

zeenix (changes requested):
Sorry, found one more thing but let's merge after that..

zbus/src/connection/connect.rs
```diff
@@ -0,0 +1,62 @@
+use std::{future::Future, pin::Pin};
+use tracing::debug;
+
```
zeenix, line 3: Oh and you forgot to convert this mod to the `connect/mod.rs`. Can you also check that you didn't forget other modules?

zeenix (changes requested):
sorry. :)

zbus/src/connection/connect.rs
elmarco: ok

zeenix (approved):

zbus/src/address/mod.rs
```diff
@@ -3,475 +3,123 @@
...
-                }
+pub mod transport;
 
-                #[cfg(target_os = "macos")]
-                return Self::from_str("launchd:env=DBUS_LAUNCHD_SESSION_BUS_SOCKET");
-            }
-        }
-    }
+#[allow(clippy::module_inception)]
+mod address;
```
zeenix, line 33: I was about to merge the PR and saw this. This file isn't large enough to justify making an exception here. The `Address` type doesn't need to move at all from here. It also makes the overall diff harder to read.

zbus_address/src/address.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub fn guid(&self) -> Option<Cow<'_, str>> {
+        self.get_string("guid").and_then(|res| res.ok())
+    }
+
+    /// Transport connection details
+    pub fn transport(&self) -> Result<Transport<'_>> {
+        self.try_into()
+    }
+
+    fn validate(&self) -> Result<()> {
```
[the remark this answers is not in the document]
zeenix: Thanks to GH, I found out that this was never resolved. :(

.githooks/pre-commit
```diff
@@ -6,3 +6,10 @@ ensure_rustup_installed
 ensure_rustfmt_installed
 
 check_formatting
+
+cd $(git rev-parse --show-toplevel)
+for i in `git ls-tree -d -r --name-only HEAD`; do
+	[ -e $i.rs ] && hook_failure "please rename $i.rs to $i/mod.rs." && exit 1
+done
+
+exit 0
```
zeenix, line 15: There is a clippy warning for this. I do not know why it didn't trigger in your case but recently I tried to create a `foo/foo.rs` at a work project and it complained.

In any case, this is something rust tooling should handle. Please remove this.

zeenix: > I do not know why it didn't trigger

Oh right, you had added an exception in the code. :facepalm:

elmarco: I don't follow "you had added an exception". no my clippy doesn't warn, but we can drop the last commit

zeenix: > I don't follow "you had added an exception".

The `#[allow(clippy::module_inception)]` that I pointed out [not too long ago](https://github.com/dbus2/zbus/pull/989#pullrequestreview-2338445323). :)

elmarco: it's not the same check, this one is for what you asked me last wrt foo/mod.rs.

zbus/src/address/transport/unixexec.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+/// `unixexec:` D-Bus transport.
+///
+/// <https://dbus.freedesktop.org/doc/dbus-specification.html#transports-exec>
+#[derive(Debug, PartialEq, Eq)]
+pub struct Unixexec<'a> {
+    path: Cow<'a, OsStr>,
+    argv: Vec<(usize, Cow<'a, str>)>,
```
zeenix, line 25: why keep the index in a Vec? Vec and slices are indexed already.

zbus/src/connection/connect/win32.rs
```diff
@@ -0,0 +1,41 @@
...
+
+use super::BoxedSplit;
+use crate::{
+    address::{transport::Transport, Address},
+    win32::autolaunch_bus_address,
+    Error, OwnedGuid, Result,
+};
+
+pub(crate) async fn connect(
+    l: &crate::address::transport::Autolaunch<'_>,
```
zeenix, line 11: I'm sure you can come up with a better name than `l`.

zbus/src/address/transport/vsock.rs
```diff
@@ -1,49 +1,67 @@
...
+    pub fn cid(&self) -> Option<u32> {
         self.cid
     }
+}
 
-    /// The port.
-    pub fn port(&self) -> u32 {
-        self.port
-    }
+impl<'a> TryFrom<&'a Address<'a>> for Vsock<'a> {
```
zeenix, line 28: These aren't really conversions. Typically when you implement a conversion from a reference, it's a hint that it's likely not really a conversion. I suggest adding a `Transport` trait in `address::transport` and putting a `from_address` constructor there, as well as moving `key_val_fmt_add` method there (and remove the `KeyValFmtAdd` trait). Also a better name for `key_val_fmt_add` would be nice (or add some docs on it if you can't come up with something better).

zbus/src/address/transport/unixexec.rs
elmarco: First, because you may or may not have argv0... 

You may also have multiple argvN (since we don't enforce this anymore)

And also the spec says " If a specific argvX is not specified no further argvY for Y > X are taken into account." which may be interepreted as the address may be accepted with argvX missing but further argvY.. in which case you should ignore them for exec, but you may still want to retain them?..

zbus/src/address/transport/vsock.rs
elmarco: ok

zbus/src/connection/connect/win32.rs
elmarco: autol ?

zeenix: `autolaunch` isn't super long. :laughing:

zbus/src/error.rs
```diff
@@ -199,6 +199,12 @@ impl From<nix::Error> for Error {
     }
 }
 
+impl From<crate::address::Error> for Error {
+    fn from(val: crate::address::Error) -> Self {
+        Error::Address(val.to_string())
```
zeenix, line 204: What's the point of having a separate Error with variants if we're going to convert it into a string and shove it in a generic error? As you recall, we moved from separate module errors to one error type per crate. Let's merge `address::Error` into `Error` please.

zbus/src/connection/connect/mod.rs
```diff
@@ -0,0 +1,62 @@
...
+    address::{transport::Transport, Address},
+    Error, Guid, OwnedGuid, Result,
+};
+
+use super::socket::{self, BoxedSplit};
+
+mod macos;
+mod win32;
+
+type ConnectResult = Result<(BoxedSplit, Option<OwnedGuid>)>;
```
zeenix, line 14: Could you please go through other code added here and change it to follow the top-down layout (something I've been trying my best to do for a few years now). I handled this file already here: https://github.com/dbus2/zbus/pull/1036/commits/cd799e4790dc987a1912c93a3b99275a67b045f6

zbus/src/error.rs
elmarco: It was to make ::address standalone, but does it make sense anymore, maybe not.. I don't know how we could trim zbus to only expose ::address if desired.

zeenix: > It was to make ::address standalone

I realize that.

> but does it make sense anymore

I don't think so, unless you find a way for address to be used standalone but for me the more important question would always be that if we'd have any users of this API and if so, how many. Given that `dbus-addr` crate didn't find any users, I'm very doubtful. Maybe we'll find 1 or 2 user but do we make efforts for some extremely niche use cases? 

I suggested to make use of cargo features. Even though that would involve annoying the hell out of tokio users, who disable the default features. Typical users having to enable the main functionality of the crate, just for niche use case, isn't the best idea. Still, if you think it's worth it, I won't stop you from it.

If you don't decide to work on it soon, then we should merge the error types.

zbus/src/address/transport/unixexec.rs
zeenix: Unresolving this thread for this part:

> > First, because you may or may not have argv0...
> 
> That just means that argv0 needs special treatment (sperate field/getter).

This is also closer to the spec since arg0 is separate key.

zbus/src/error.rs
elmarco: also desirable is to compile-time check addresses..

how would you like to merge errors? drop all the variants of address::Error and replace them with the generic Error::Address ?

zeenix: > how would you like to merge errors? drop all the variants of address::Error and replace them with the generic Error::Address ?

I think so, yes. I seriously doubt user can programatically do anything different based on the exact variant and recover for it using the same address. Also, this is what is happening currently after this PR anyway when user uses `connection::Builder::address`.

elmarco: While it may not be handled programmatically, a precise error can allow for better error feedback returned to the user. It's internal anyway, is it really worth to change now ? 

I think it's a shame to conflate the various errors that can be exposed to a flat String. Having a single Address error, imho we lose some robustness. We trade a well set of meaningful errors for a wildcard..

Why don't you do this ? It makes me sad to have to disagree with your opinions...

zbus/src/connection/connect/mod.rs
elmarco: I can't really think of a better order, please do it as it suits you.

zbus/src/address/transport/vsock.rs
elmarco: in #1050

zbus/src/error.rs
zeenix: > While it may not be handled programmatically, a precise error can allow for better error feedback returned to the user.

That's what the description string is for. If they print/debug the error, they'll see the string that explains the exact problem.

> I think it's a shame to conflate the various errors that can be exposed to a flat String. Having a single Address error, imho we lose some robustness. We trade a well set of meaningful errors for a wildcard..

Let's be more specific. Can you give an example of a real-world situation where different variants would help the application programmer? I have to remind you again that most users will not even use this API directly, let alone parse the variants.

BTW, you missed a very important point:

>  Also, this is what is happening currently after this PR anyway when user uses connection::Builder::address.

So if you really want a separate error, at least you should then go the whole mile and use this type in the `zbus::Error::Address`. I don't think we should though for reasons mentioned in this comment and before.

> Why don't you do this ? It makes me sad to have to disagree with your opinions...

Same here but I don't see any connection. In the end it's the same result and you promised to maintain this code and this is very much part of the maintenance IMO. I still think that these changes unnecessarily complicated the API with no **real** benefits but I accepted it because you wouldn't let it go and since it will not affect most users in any way.

zbus/src/connection/connect/mod.rs
zeenix: > I can't really think of a better order,

Not sure what's there to think of. It's like this:

* pub > pub(crate) > private.
* Use before declaration/definition.

Think in terms of what you most likely want to see first in a file when you open it? Most likely it's the public API or pub(crate) API. Also, you want to see the high-level code (i-e use) before implementation details.

>  please do it as it suits you.

I'm sorry but you've to follow the conventions of the project. Besides, I did some of the work for you and gave you an example to follow.

zeenix: Having said all that, this is not super important/urgent. But please keep it in mind in the future. I wish I could say the same about the `address::Error` discussion above but currently it's neither my way nor your way.

zbus/src/error.rs
elmarco: We should keep address::Error separate, because it carries semantics that can be used by API users. They can provide relevant information and feedback when parsing an address. And for testing alone, it's superior to have an enum to match against, rather than a string.

zeenix: I know the theoretical benefits. I was asking for more practical ones. I already explained in detail above why I think we should merge it. If you want to argue, please address all my specific points above. Otherwise, it doesn't take the discussion anywhere.

elmarco: "They can provide relevant information and feedback when parsing an address" think a tool that validates the dbus address, it could be a management tool or for example a middlelayer like libvirt which manipulates and takes dbus configuration from users. You can provide different feedback based on the different errors, provide link to reference documentation, examples, suggestions to fix etc... & I don't need to detail the testing scenario.

This is not something you can reliably do if all you have is a single Error::Address(String) variant.

address::Error has to be public already, because we use this type for returning errors for various conversion traits which are publics. Potentially, users may have to deal with it if they do such conversions themself.

I'll modify the code to expose the address::Error to zbus::Error.

zeenix: > "They can provide relevant information and feedback when parsing an address" think a tool that validates the dbus address, it could be a low-level tool, or a management tool or for example a middlelayer like libvirt which manipulates and takes dbus configuration from users. You can provide different feedback based on the different errors, provide link to reference documentation, examples, suggestions to fix etc... & I don't need to detail the testing scenario.

Now we're talking. Actual practical examples. :+1:

> I'll modify the code to expose the address::Error to zbus::Error

Yes please.

elmarco: done in #1050

zbus/src/connection/connect/mod.rs
elmarco: done in #1050, if you want further code ordering, please be specific, or I invite you to do so. thanks

PR #1010, opened 2024-09-20
🏷️ zv: Add FilePath type

### LINKED ISSUE #977: Can't send a non UTF-8 characters over the bus
I was trying to send a path (PathBuf) through a (z/d)bus proxy to (z/d)bus server, but I got this error by trying to send a path with a non UTF-8 character
  ```
  called `Result::unwrap()` on an `Err` value: Variant(Message("path contains invalid UTF-8 characters"))
  ```
  So my question is: 
  Is it possible to send this exact path with no modification over it's encoding, or encoding is forced? and if so, is there a workaround?
  # Note:
  - Although I can have a work around by converting it to a loosy string, but I can't seem to be able to convert it back to the original encoded format, which what I need to access the file.
  - For reference, this is the non UTF-8 part of the path "\xC9vora", and this the loosly encoded one "�vora".

yassinebenarbia:
Add a new `FilePath` type to serve as a thin abstraction that handles (de)serialization of a file path as a byte array, since en/decoding of strings is limited to only UTF-8 characters in both D-Bus format and serde API.

Fixes #977.

### DIFF
--- a/zvariant/src/file_path.rs
+++ b/zvariant/src/file_path.rs
@@ -1,0 +1,254 @@
+use serde::{Deserialize, Serialize};
+use std::{
+    borrow::Cow,
+    ffi::{CStr, CString, OsStr, OsString},
+    path::{Path, PathBuf},
+};
+
+use crate::Type;
+
+/// File name represented as a nul-terminated byte array.
+///
+/// While `zvariant::Type` and `serde::{Serialize, Deserialize}`, are implemented for [`Path`] and
+/// [`PathBuf`], unfortunately `serde` serializes them as UTF-8 strings and that limits the number
+/// of possible characters to use on a file path. This is not the desired behavior since file paths
+/// are not guaranteed to contain only UTF-8 characters.
+///
+/// To solve this problem, this type is provided which encodes the underlying file path as a
+/// null-terminated byte array.
+///
+/// # Examples:
+///
+/// ```
+/// use zvariant::FilePath;
+/// use std::path::{Path, PathBuf};
+///
+/// let path = Path::new("/hello/world\0");
+/// let path_buf = PathBuf::from(path);
+///
+/// let p1 = FilePath::from(path);
+/// let p2 = FilePath::from(path_buf);
+/// let p3 = FilePath::from("/hello/world");
+///
+/// assert_eq!(p1, p2);
+/// assert_eq!(p2, p3);
+/// ```
+#[derive(Type, Debug, Default, PartialEq, Eq, Serialize, Deserialize, Clone, Ord, PartialOrd)]
+#[zvariant(signature = "ay")]
+pub struct FilePath<'f>(Cow<'f, CStr>);
+
+impl<'f> FilePath<'f> {
+    pub fn new(cow: Cow<'f, CStr>) -> Self {
+        Self(cow)
+    }
+}
+
+impl From<CString> for FilePath<'_> {
+    fn from(value: CString) -> Self {
+        FilePath(Cow::Owned(value))
+    }
+}
+
+impl<'f> From<&'f CString> for FilePath<'f> {
+    fn from(value: &'f CString) -> Self {
+        FilePath(Cow::Borrowed(value.as_c_str()))
+    }
+}
+
+impl<'f> From<&'f OsStr> for FilePath<'f> {
+    fn from(value: &'f OsStr) -> FilePath<'f> {
+        FilePath(bytes_with_null(value.as_encoded_bytes()))
+    }
+}
+
+impl<'f> From<&'f OsString> for FilePath<'f> {
+    fn from(value: &'f OsString) -> FilePath<'f> {
+        FilePath(bytes_with_null(value.as_encoded_bytes()))
+    }
+}
+
+impl From<OsString> for FilePath<'_> {
+    fn from(value: OsString) -> Self {
+        FilePath(vec_to_cstr(value.as_encoded_bytes().to_vec()))
+    }
+}
+
+impl<'f> From<&'f PathBuf> for FilePath<'f> {
+    fn from(value: &'f PathBuf) -> FilePath<'f> {
+        FilePath::from(value.as_os_str())
+    }
+}
+
+impl From<PathBuf> for FilePath<'_> {
+    fn from(value: PathBuf) -> FilePath<'static> {
+        FilePath::from(OsString::from(value))
+    }
+}
+
+impl<'f> From<&'f Path> for FilePath<'f> {
+    fn from(value: &'f Path) -> Self {
+        Self::from(value.as_os_str())
+    }
+}
+
+impl<'f> From<&'f CStr> for FilePath<'f> {
+    fn from(value: &'f CStr) -> Self {
+        Self(Cow::Borrowed(value))
+    }
+}
+
+impl<'f> From<&'f str> for FilePath<'f> {
+    fn from(value: &'f str) -> Self {
+        Self::from(OsStr::new(value))
+    }
+}
+
+impl<'f> AsRef<FilePath<'f>> for FilePath<'f> {
+    fn as_ref(&self) -> &FilePath<'f> {
+        self
+    }
+}
+
+impl From<FilePath<'_>> for OsString {
+    fn from(value: FilePath<'_>) -> Self {
+        // SAFETY: user is responsible of handling conversion from [FilePath] to [OsString]
+        // since FilePath is a set of null terminated bytes and it's interpretations mainly
+        // depends on the underlying platform.
+        // see [std::ffi::os_str::OsString::from_encoded_bytes_unchecked]
+        unsafe { OsString::from_encoded_bytes_unchecked(value.0.to_bytes().to_vec()) }
+    }
+}
+
+impl<'f> From<&'f FilePath<'f>> for &'f Path {
+    fn from(value: &'f FilePath<'f>) -> Self {
+        // This method should fail if FilePath does not represent UTF-8 valid chars
+        // since [Path] is akin to [str], hence the unwrap.
+        Path::new(value.0.as_ref().to_str().unwrap())
+    }
+}
+
+impl<'f> From<FilePath<'f>> for PathBuf {
+    fn from(value: FilePath<'f>) -> Self {
+        PathBuf::from(value.0.to_string_lossy().to_string())
+    }
+}
+
+/// Converts a `Vec<u8>` into a null-terminated `CStr`.
+///
+/// Truncates the vector at the first null byte, if present. If no null byte exists, appends one to
+/// ensure proper termination.
+///
+/// # Returns
+///
+/// A `Cow<'_, CStr>` containing a *guaranteed* null-terminated string.
+pub fn vec_to_cstr(mut bytes: Vec<u8>) -> Cow<'static, CStr> {
+    if let Some(pos) = bytes.iter().position(|&b| b == 0) {
+        bytes.truncate(pos + 1);
+    } else {
+        bytes.push(0);
+    }
+    // unwrap is fine here since we append the null byte.
+    Cow::Owned(CString::from_vec_with_nul(bytes).unwrap())
+}
+
+/// Converts a byte slice into a null-terminated [CStr].
+///
+/// Returns a borrowed [CStr] if the slice already contains a null byte; otherwise, returns an
+/// owned [CStr] with a null byte appended.
+///
+/// # Returns
+///
+/// A [Cow<'_, CStr>] containing a *guaranteed* null-terminated string.
+fn bytes_with_null(bytes: &[u8]) -> Cow<'_, CStr> {
+    if let Ok(cstr) = CStr::from_bytes_until_nul(bytes) {
+        return Cow::Borrowed(cstr);
+    }
+    // unwrap is fine, as we handled the null termination case above.
+    Cow::Owned(CString::new(bytes).unwrap())
+}
+
+#[cfg(test)]
+mod file_path_test {
+    use super::*;
+    use crate::zvariant::Signature;
+    use std::path::{Path, PathBuf};
+
+    #[test]
+    fn from_test() {
+        let path = Path::new("/hello/world");
+        let path_buf = PathBuf::from(path);
+        let osstr = OsStr::new("/hello/world");
+        let os_string = OsString::from("/hello/world");
+        let cstr = CStr::from_bytes_until_nul("/hello/world\0".as_bytes()).unwrap_or_default();
+        let cstring = CString::new("/hello/world").unwrap_or_default();
+
+        let p1 = FilePath::from(path);
+        let p2 = FilePath::from(path_buf);
+        let p3 = FilePath::from(osstr);
+        let p4 = FilePath::from(os_string);
+        let p5 = FilePath::from(cstr);
+        let p6 = FilePath::from(cstring);
+        let p7 = FilePath::from("/hello/world");
+
+        assert_eq!(p1, p2);
+        assert_eq!(p2, p3);
+        assert_eq!(p3, p4);
+        assert_eq!(p4, p5);
+        assert_eq!(p5, p6);
+        assert_eq!(p5, p7);
+    }
+
+    #[test]
+    fn filepath_signature() {
+        assert_eq!(
+            &Signature::static_array(&Signature::U8),
+            FilePath::SIGNATURE
+        );
+    }
+
+    #[test]
+    fn into_test() {
+        let first = PathBuf::from("/hello/world");
+        let third = OsString::from("/hello/world");
+        let fifth = Path::new("/hello/world");
+        let p = FilePath::from(first.clone());
+        let p2 = FilePath::from(third.clone());
+        let p3 = FilePath::from(fifth);
+        let second: PathBuf = p.into();
+        let forth: OsString = p2.into();
+        let sixth: &Path = (&p3).into();
+        assert_eq!(first, second);
+        assert_eq!(third, forth);
+        assert_eq!(fifth, sixth);
+    }
+
+    #[test]
+    fn vec_nul_termination() {
+        let v1 = vec![];
+        let v2 = vec![0x0];
+        let v3 = vec![0x1, 0x2, 0x0];
+        let v4 = vec![0x0, 0x0];
+        let v5 = vec![0x1, 0x0, 0x2, 0x0];
+
+        assert_eq!(
+            Cow::Borrowed(CStr::from_bytes_with_nul(&[0x0]).unwrap()),
+            vec_to_cstr(v1)
+        );
+        assert_eq!(
+            Cow::Borrowed(CStr::from_bytes_with_nul(&[0x0]).unwrap()),
+            vec_to_cstr(v2)
+        );
+        assert_eq!(
+            Cow::Borrowed(CStr::from_bytes_with_nul(&[0x1, 0x2, 0x0]).unwrap()),
+            vec_to_cstr(v3)
+        );
+        assert_eq!(
+            Cow::Borrowed(CStr::from_bytes_with_nul(&[0x0]).unwrap()),
+            vec_to_cstr(v4)
+        );
+        assert_eq!(
+            Cow::Borrowed(CStr::from_bytes_with_nul(&[0x1, 0x0]).unwrap()),
+            vec_to_cstr(v5)
+        );
+    }
+}
--- a/zvariant/src/lib.rs
+++ b/zvariant/src/lib.rs
@@ -36,6 +36,9 @@ pub use fd::*;
 mod object_path;
 pub use crate::object_path::*;
 
+mod file_path;
+pub use crate::file_path::*;
+
 mod ser;
 pub use ser::*;
 


zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
...
+        let path = Path::new("/hello/world");
+        let _ = FilePath::from(path);
+        let path_buf = PathBuf::from("/hello/world");
+        let _ = FilePath::from(path_buf);
+    }
+
+    #[test]
+    fn filepath_signature() {
+        assert_eq!(
+            &Signature::Array(zvariant_utils::signature::Child::Static {
```
zeenix, line 92: you can make use of `Signature::static_array` here.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
+use std::{ffi::{OsStr, OsString}, path::{Path, PathBuf}};
+
+use serde::{de::{self, Visitor}, Deserialize, Deserializer, Serialize, Serializer};
+use crate::de::Deserializer as ZbusDeserializer;
+
+use crate::Type;
+
+/// A file name represented as a nul-terminated byte array.
```
zeenix, line 8: Could you please add a bit more description on second paragraph:

```
While `zvariant::Type` and `serde::{Serialize, Deserialize}`, are implemented for [`Path`] and [`PathBuf`], unfortunately `serde` serializes them as UTF-8 strings. This is not the desired behavior in most cases since file paths are not guaranteed to only contain UTF-8 characters.

To solve this problem, this type is provided which encodes the underlying file path as a null-terminated byte array. Encoding as byte array is also more efficient.
```

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
...
+#[derive(Type, Debug, Default, PartialEq)]
+#[zvariant(signature = "ay")]
+pub struct FilePath(OsString);
```
zeenix, line 11: We don't need to force allocation here. This could be:

```rust
pub struct FilePath<'f>(Cow<'f, Path>);
```

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: Deserializer<'de>
+    {
+        struct FilePathVisitor;
+        impl<'de> Visitor<'de> for FilePathVisitor {
+            type Value = FilePath;
+
+            fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                formatter.write_str("unable to deserialize FilePath")
```
zeenix, line 35: The erroring out itself already conveys the failure to deserialize FilePath. This string is about telling what we were expecting: "a byte array".

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
...
+            }
+
+            fn visit_bytes<E>(self, v: &[u8]) -> std::result::Result<Self::Value, E>
+            where
+                E: de::Error,
+            {
+                unsafe {
+                    Ok(FilePath(
+                            OsStr::from_encoded_bytes_unchecked(v).to_os_string()
+                    ))
```
zeenix, line 45: Hmm.. Are we safe here? The docs say that we're violating the safety rules here:

> As the encoding is unspecified, callers must pass in bytes that originated as a mixture of validated UTF-8 and bytes from [OsStr::as_encoded_bytes](https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.as_encoded_bytes) from within the same Rust version built for the same target platform. For example, reconstructing an OsStr from bytes sent over the network or stored in a file will likely violate these safety rules.

In any case, we'll want a `Safety: ` commend above that proves why this is safe.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,106 @@
+use std::{ffi::{OsStr, OsString}, path::{Path, PathBuf}};
+
+use serde::{de::{self, Visitor}, Deserialize, Deserializer, Serialize, Serializer};
+use crate::de::Deserializer as ZbusDeserializer;
```
zeenix, line 4: We should not need to use a specific deserializer here and I don't see any use of this anyway.

zeenix (changes requested):
Thanks so much for this contribution. Looking good in general, just need a few tweaks, I think. Apart from the inline comments, a few things about the commit and PR title:

* Being a thin wrapper is a detail. The title should be "🏷️ zv: Add `FilePath` type" (I'll take care of the PR title).
* Kindly add details to commit log as well, not just the PR. In this case, they could be identical.
* Add `Fixes #977.` line to the commit & PR description.

zeenix (changes requested):
Other than that, it's getting closer to being ready. Oh and the task was to fix the commit itself, as per the last point in the [contribution guidelines](https://github.com/dbus2/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). If you find this task too much and the docs and tools recommended in the contribution guideline don't help, please let me know and I can help you for this first contribution.

Oh and as the CI would point out, you'd need to fix the formatting but that's easy. :)

zvariant/src/file_path.rs
yassinebenarbia: - It was an assumption rather than an assertion i.e. not **necessarily,** but **safe to assume** as a path can contain other non-UTF8 symbols, so we have no reason to say that it only contains UTF8 symbols, thus, the assumption.
- yes, you were right, the comment has nothing to do with the `unsafe` call, by bad ig ;p
- Also I've moved out from the unsafe call to a [more safe approach](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_vec_with_nul)

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,138 @@
...
+/// let path = Path::new("/hello/world");
+/// let path_buf = PathBuf::from(path);
+///
+/// let p1 = FilePath::from(path);
+/// let p2 = FilePath::from(path_buf);
+/// let p3 = FilePath::from("/hello/world");
+///
+/// assert_eq!(p1, p2);
+/// assert_eq!(p2, p3);
+/// ```
```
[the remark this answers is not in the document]
yassinebenarbia: I added some tests for the (de)serialization process, but didn't change the examples other than adding some `Into`s, since I really think that the way the type is (de)serialized should not matter for users, as they should be able to just wrap their file paths in it and send/receive it.

yassinebenarbia:
For now, I'm thinking of a commit comment+description for my changes, if you can, can you please help me figure that out! [these](https://gist.github.com/yassinebenarbia/2d91b50c6c770d01f7d82df2f1e5cb00) are the changes that I'm planing to made and TBH, I'm really inclined towards making a clean new PR instead of adding to this one :) but this is also the PR that was mentioned in the Issue that I've opened #977.

zvariant/src/file_path.rs
zeenix: > * It was an assumption rather than an assertion i.e. not **necessarily,** but **safe to assume** as a path can contain other non-UTF8 symbols, so we have no reason to say that it only contains UTF8 symbols, thus, the assumption.

Right but assuming UTF-8 where bytes could have non-UTF8 would be problematic. The other way around is not a problem and hence need no documentation and only makes it confusing.

> * yes, you were right, the comment has nothing to do with the `unsafe` call, by bad ig ;p

No worries. That's why we've reviews.

> * Also I've moved out from the unsafe call to a [more safe approach](https://doc.rust-lang.org/std/ffi/struct.CString.html#method.from_vec_with_nul)

Cool but seems all it really gives us is a null-byte check (which we can do safely ourselves too). I also realised that we're forcing copying/allocation where we probably don't need to. I think you should be implementing/using `Visitor::visit_borrowed_bytes` instead combined with https://doc.rust-lang.org/std/ffi/struct.OsStr.html#method.from_encoded_bytes_unchecked .

I'm still a bit unsure about safety of it all here. For example, what happens if the invariant of these methods don't hold. Maybe we can keep this type unix-only, since at least for unix platforms, you can safely deserialize `&OsStr` from bytes: https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStrExt.html

zvariant/src/file_path.rs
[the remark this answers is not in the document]
zeenix: > I really think that the way the type is (de)serialized should not matter for users

As I wrote, that's the main use here. Users will not be using this a generic type for file paths, but OK, no biggie.

zvariant/src/file_path.rs
yassinebenarbia: Hi! I've made some changes that you can find [here](https://gist.github.com/yassinebenarbia/aa0c536eaa364d3985e35edb6153736c), I've tried to minimize copying/allocating memory as much as I can tho, but I'm not quite sure that I didn't miss something, so please let me know if you spotted anything :p, also, if you find those changes satisfying, I'll be adding them to the PR

zeenix: Cool. Can you please just push it to this branch so it's easier to review specific parts of the code?

zeenix:
I'm sorry for the delay but I've been travelling. I'll review now..

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+    ffi::{CStr, OsStr, OsString},
+    path::{Path, PathBuf},
+};
+
+use crate::Type;
+
+/// A file name represented as a nul-terminated byte array.
+///
+/// While `zvariant::Type` and `serde::{Serialize, Deserialize}`, are implemented for [`Path`] and [`PathBuf`], unfortunately `serde` serializes them as UTF-8 strings. This is not the desired behavior in most cases since file paths are not guaranteed to contain only UTF-8 characters.
+/// To solve this problem, this type is provided which encodes the underlying file path as a null-terminated byte array. Encoding as byte array is also more efficient. The Prodigy - Breathe (Brooks Aleksander Remix)
```
zeenix, line 15: * The formatting is wrong (hopefully the CI will catch that). It's best to always run `cargo +nightly fmt` and `cargo clippy` before pushing.
* What's the "The Prodigy - Breathe (Brooks Aleksander Remix)" part about? :thinking:

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+/// # Exmples
```
zeenix, line 17: empty line under headings please.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+///
+///    assert_eq!(p1, p2);
+///    assert_eq!(p2, p3);
+///    assert_eq!(p3, p4);
+///    assert_eq!(p4, p5);
+///    assert_eq!(p5, p6);
+///    assert_eq!(p5, p7);
+/// ```
+/// Also you can (de)serialize the [FilePath] as an array of bytes, consider this for example
+///
```
zeenix, line 47: While I do appreciate you taking the time to write detailed examples, I don't think it's necessary. It's a very simple type and user doesn't need that much code to understand how to use it. Just a very few simple examples of building an instance from random paths and then encoding and decoding them, is more than sufficient.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+///        serde_json::json!(108),
+///        serde_json::json!(108),
+///        serde_json::json!(111),
+///        serde_json::json!(47),
+///        serde_json::json!(119),
+///        serde_json::json!(111),
+///        serde_json::json!(114),
+///        serde_json::json!(108),
+///        serde_json::json!(100),
+///    ]);
```
zeenix, line 72: There is certainly no need to involve complicated stuff, like json etc. :)

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+    }
+    vec_with_nul.push(0x0);
+    return vec_with_nul;
+}
+
+/// Chops a slice of bytes until a nul character,
+/// and if it (the nul) does not exist, it appends it at
+/// the end.
+/// This method won't allocate/copy memory unless the nul
+/// byte does not exist in the `bytes` parameter
```
zeenix, line 125: I don't think we need to do all this for the user. We should just simply check for the null byte and error out if it's not there.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+            return &bytes[0..index];
+        }
+    }
+    // otherwise we have no option but to copy to a new slice
+    // since we borrow bytes immutably
+    // SAFETY: see [`GlobalAlloc::alloc`].
+    let slice = unsafe {
+        // consider switching to https://doc.rust-lang.org/std/alloc/trait.Allocator.html#tymethod.allocate
+        // when it becames a part of the stable release
+        let ptr = std::alloc::alloc(Layout::new::<u8>());
```
zeenix, line 143: Why not just use Box?

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+    fn from(value: CString) -> Self {
+        return FilePath(Cow::Owned(value));
+    }
+}
+
+impl<'f> From<&'f OsStr> for FilePath<'f> {
+    fn from(value: &'f OsStr) -> Self {
+        // SAFETY: this call is safe because we guarentee the nul termination
+        // of the `path_bytes`
+        let bytes_with_nul = chop_bytes_with_nul(value.as_encoded_bytes());
```
zeenix, line 161: * The safety comment is above the safe line.
* We could just use `CStr::from_bytes_with_null` and error out if null is missing. Then we don't need chop_bytes_with_null or unsafe even.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+    }
+}
+
+impl From<OsString> for FilePath<'static> {
+    fn from(value: OsString) -> Self {
+        let path_bytes = chop_vec_with_nul(&value.as_encoded_bytes().to_vec());
+        // SAFETY: this call is safe because we guarentee the nul termination
+        // of the `path_bytes`
+        unsafe {
+            return FilePath(Cow::Owned(CString::from_vec_with_nul_unchecked(path_bytes)));
```
zeenix, line 176: Same comment here, except we'd use `from_vec_with_nul`.

zeenix (changes requested):

zvariant/src/file_path.rs
yassinebenarbia: Oppsees, didn't see that, apparently I pressed an additional `p` after editing, that's embarrassing :x (btw, if you like ambient/electro music, check it out :D)

zvariant/src/file_path.rs
yassinebenarbia: Can't we get away with appending a null byte if missing instead of erroring because of it? although I don't like the idea of "modifying" the user input to begin with, but ig it's an idea that deserves consideration nonetheless, no?

yassinebenarbia: Nvm, just read [this](https://github.com/dbus2/zbus/pull/1010#discussion_r1808064635), although I sill don't like the idea of erroring in a  `from` call, but we can go with that ig :p

yassinebenarbia:
Sorry couldn't find time to apply patches, I'll update/apply patches late this week (hope hope)

zvariant/src/file_path.rs
yassinebenarbia: Although, I kinda agree with you, but this type will be usually used with some others like `Path` and `PathBuf` or even a `&str`  probably in the form of `let path = FilePath::from("path/to/file");` which I would assume that it will be kinda inconvenient to error there because we don't have a null byte at the end don't we? ( I'm not quite sure about erroring in a`from` call, so that will likely end up a `try_from`)

zvariant/src/file_path.rs
yassinebenarbia: used it there since `from_vec_with_nul` checks for null byte at the end, which we already do, but if you don't mind it, we can have it sure :p

zvariant/src/file_path.rs
zeenix: Hmm.. good point.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,348 @@
...
+}
+
+/// Chops a slice of bytes until a nul character,
+/// and if it (the nul) does not exist, it appends it at
+/// the end.
+/// This method won't allocate/copy memory unless the nul
+/// byte does not exist in the `bytes` parameter
+///
+/// # Return:
+/// - _*Guaranteed*_ nul terminated slice of bytes
```
zeenix, line 128: Regarding both chop* functions:

* The name `chop` is confusing. We're not chopping things, just ensuring that there is a null byte at the end.
* Please move them below all the trait impls (I try to keep the code top-down: use before definition & pub before private API).
* Regarding the doc comments (or all comments in this commit actually):
  * The line breaks are at 100 chars (just like the code).
  * Please don't break the line just to start a new sentence, only for new paragraphs. Also `.` at the end of sentences.

yassinebenarbia: For the chop name, it's because if we ever had a string of bytes with two nulls or more in it, it `chop`s the first part and throws the others, but now that I think about it, it should be just `vec_with_nul` and `bytes_with_null`.

yassinebenarbia: I used the `chop` word since if we have a set of bytes with more than one null in it, it just takes the first part and returns it back, and tbh, I agree with you about the `chop` part of the function name being confusing, it should be just `bytes_with_null`

zeenix: > For the chop name, it's because if we ever had a string of bytes with two nulls

Ah ok. I understand your reasoning now but that's not a very likely possiblity IMO. If someone gives us bytes of a null-terminated string, they'll give us exactly that, not more. So I think it'd be best to not name it like that.

zeenix: > bytes_with_null

I think this function should just return a Cow<'_, CStr> directly and should be just called bytes_to_cstr.

Skgland: So, something like this?

```rust
pub fn bytes_to_cstr(data: &[u8]) -> Cow<'_, CStr> {
    CStr::from_bytes_until_nul(data)
        .map(Cow::from)
        .unwrap_or(Cow::from(CString::new(data).expect(
            "from_bytes_until_nul failed so data doesn't contain a nul byte",
        )))
}
```
[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5265cd1a8f0db8247be0d7493b7e6114)

yassinebenarbia: > > bytes_with_null
> 
> I think this function should just return a Cow<'_, CStr> directly and should be just called bytes_to_cstr. (you can elide the lifetime btw).

That's exactly what it became tho :D good catch,  and also got rid of most `'static` or "uselses" lifetimes, at least, that's as far as I can see now.

yassinebenarbia: > So, something like this?
> 
> ```rust
> pub fn bytes_to_cstr(data: &[u8]) -> Cow<'_, CStr> {
>     CStr::from_bytes_until_nul(data)
>         .map(Cow::from)
>         .unwrap_or(Cow::from(CString::new(data).expect(
>             "from_bytes_until_nul failed so data doesn't contain a nul byte",
>         )))
> }
> ```
> 
> [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5265cd1a8f0db8247be0d7493b7e6114)

[This does not work tho, no?](https://play.rust-lang.org?version=stable&mode=debug&edition=2024&gist=5265cd1a8f0db8247be0d7493b7e6114), I intend to make it pass bytes with null and without it just fine, also, trim only bytes until the first encountered null and discard the rest

Skgland: Oops, should have been `unwrap_or_else` not `unwrap_or`

```rust
use std::borrow::Cow;
use std::ffi::{CStr, CString};

pub fn bytes_to_cstr(data: &[u8]) -> Cow<'_, CStr> {
    CStr::from_bytes_until_nul(data)
        .map(Cow::from)
        .unwrap_or_else(|_|Cow::from(CString::new(data).expect(
            "from_bytes_until_nul failed so data doesn't contain a nul byte",
        )))
}

fn main() {
    let v = [0x1, 0x0];
    bytes_to_cstr(&v);
}
```
 [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6f916193f261906f8e3b54d33bd259c8)

zeenix:
@yassinebenarbia please fix the CI breakage first. I'll review after. I really hope we can merge this soon. :) Thanks for your perseverance on this. :muscle:

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+/// # Returns
+///
+/// A [Cow<'_, CStr>] containing a *guaranteed* null-terminated string.
+fn bytes_with_null(bytes: &[u8]) -> Cow<'_, CStr> {
+    // SAFETY: unwrap is fine here since in bot cases we guarante the null termination
+    if let Some(pos) = bytes.iter().position(|&b| b == 0) {
+        return Cow::Borrowed(CStr::from_bytes_with_nul(&bytes[..=pos]).unwrap());
+    }
+    Cow::Owned(CString::new(bytes).unwrap())
+}
```
Skgland, line 166: Using [`from_bytes_until_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul) one can get rid of the first unwrap.


```suggestion
fn bytes_with_null(bytes: &[u8]) -> Cow<'_, CStr> {
    if let Ok(cstr) = CStr::from_bytes_until_nul(&bytes) {
        return Cow::Borrowed(cstr);
    }
    // unwrap is fine, as we handled the null containing case above
    Cow::Owned(CString::new(bytes).unwrap())
}
```

yassinebenarbia: Make sense, I'll apply that and force push.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+impl<'f> From<FilePath<'f>> for PathBuf {
+    fn from(value: FilePath<'f>) -> Self {
+        PathBuf::from(value.0.to_string_lossy().to_string())
+    }
+}
+
+/// Converts a `Vec<u8>` into a null-terminated `CStr`.
+///
+/// Truncates the vector at the first null byte, if present.
+/// If no null byte exists, appends one to ensure proper termination.
```
zeenix, line 137: I really appreciate you adding proper docs for even the internal API. :+1: This is why I hate to nitpick here but please don't break lines for starting a new sentence (only for paragraphs). Break them on word boundary at 100 chars, just like the code. :pray: Same goes for other comments.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+        assert_eq!(first, second);
+        assert_eq!(third, forth);
+        assert_eq!(fifth, sixth);
+    }
+
+    #[test]
+    fn de_serialize() {
+        let file_path = FilePath::from("/hello/world");
+        let path_ser = serde_json::json!(file_path);
+        // value = /hell/world
```
zeenix, line 228: why not make it simpler and so you don't have to have this comment even?

```rust
b"/hello/world".to_vec()
```

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+        let sixth: &Path = (&p3).into();
+        assert_eq!(first, second);
+        assert_eq!(third, forth);
+        assert_eq!(fifth, sixth);
+    }
+
+    #[test]
+    fn de_serialize() {
+        let file_path = FilePath::from("/hello/world");
+        let path_ser = serde_json::json!(file_path);
```
zeenix, line 227: Not sure why we're serializing to JSON here. We should focus on the D-Bus format. 

I think it would be best to move the ser/de tests to `zvariant/src/lib.rs` along with all other ser/de tests for various types and use similar/same patterns you see there.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+        let path_deser: FilePath<'static> = FilePath::deserialize(path_ser).unwrap();
+        assert_eq!(file_path, path_deser);
+
+        let path_buf: PathBuf = file_path.into();
+        let string_lossy = Cow::Borrowed("/hello/world");
+        let path_lossy = path_buf.to_string_lossy();
+        assert_eq!(string_lossy, path_lossy);
+
+        // /hello/world/\\xC9blahblah
+        #[rustfmt::skip]
```
zeenix, line 254: please don't skip fmt unless really necessary.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+///
+/// While `zvariant::Type` and `serde::{Serialize, Deserialize}`, are implemented for [`Path`] and
+/// [`PathBuf`], unfortunately `serde` serializes them as UTF-8 strings and that limits the number
+/// of possible characters to use on a file path. This is not the desired behavior since file paths
+/// are not guaranteed to contain only UTF-8 characters.
+///
+/// To solve this problem, this type is provided which encodes the underlying file path as a
+/// null-terminated byte array.
+///
+/// # Exmples
```
zeenix, line 20: typo: Examples.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,293 @@
...
+///
+/// Consider using `from` to convert/cast [FilePath] to other compatible types, see the example
+/// bellow for reference
```
zeenix, line 23: Missing `:` and an emtpy line (the code below is a separate paragraph).

zeenix (changes requested):
Getting very close. Mostly just minor nitpicks to satisfy now. :) Talking of nitpicks, the commit log is different from the PR title. Let's keep them the same if the PR is a single commit.

zeenix:
> Talking of nitpicks, the commit log is different from the PR title. Let's keep them the same if the PR is a single commit.

I took the liberty of modify the commit description here. Please just copy&paste that into the commit description (breaking lines at 74 chars though).

zvariant/src/file_path.rs
yassinebenarbia: It looks like you've moved the tests away from `zvariant/src/lib.rs`, do you still want them there?

yassinebenarbia: I removed the test (since it introduced a conflict) from `zvariant/src/lib.rs`, if you wish to still have it somewhere else, let me know!

yassinebenarbia:
I added some derive macro `PartialOrd` and `Ord` to make sorting easier.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Converts a `Vec<u8>` into a null-terminated `CStr`.
+///
+/// Truncates the vector at the first null byte, if present. If no null byte exists, appends one to
+/// ensure proper termination.
+///
+/// # Returns
+///
+/// A `Cow<'_, CStr>` containing a *guaranteed* null-terminated string.
+pub fn vec_to_cstr<'a>(mut bytes: Vec<u8>) -> Cow<'a, CStr> {
```
zeenix, line 144: The `'a` is really `'static` here, no? Since this method can't not do `CStr` and only `CString`, I think it should just return a `CString` and be named `vec_to_cstring`.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+
+impl<'f> AsRef<FilePath<'f>> for FilePath<'f> {
+    fn as_ref(&self) -> &FilePath<'f> {
+        self
+    }
+}
+
+impl<'f> From<FilePath<'f>> for OsString {
```
zeenix, line 112: lifetime here can be elided.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,254 @@
...
+    }
+}
+
+impl<'f> From<&'f PathBuf> for FilePath<'f> {
+    fn from(value: &'f PathBuf) -> FilePath<'f> {
+        FilePath::from(value.as_os_str())
+    }
+}
+
+impl<'f> From<PathBuf> for FilePath<'f> {
```
zeenix, line 82: where is the `'f` coming from? :thinking: Isn't it just `'static` here?

zeenix (changes requested):
Sooooooooo close!

zvariant/src/file_path.rs
yassinebenarbia: Tho the idea of that fn is for convenient purposes for the `FilePath` type and I really don't like having to some type gymnastics to go from `Cow<'_, CString>` to `Cow<'_, CStr>` each time I use it, also yeah lifetime there is not necessary, so now changing the return signature to `Cow<'static, CStr>`.

zeenix: > I really don't like having to some type gymnastics to go from `Cow<'_, CString>` to `Cow<'_, CStr>` each time I use it

Sure but I only said this because you only use it once (except in tests) and AFAICT, you just need to append a `.into()` to the call. But no biggie.

zeenix (approved):

zeenix:
> a few things about the commit and PR title:
> 
> * Being a thin wrapper is a detail. The title should be "🏷️ zv: Add `FilePath` type" (I'll take care of the PR title).
> * Kindly add details to commit log as well, not just the PR. In this case, they could be identical.

@yassinebenarbia Seems you never addressed this. :( I'll do this now as part of the rebase.

zvariant/src/file_path.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Converts a `Vec<u8>` into a null-terminated `CStr`.
+///
+/// Truncates the vector at the first null byte, if present. If no null byte exists, appends one to
+/// ensure proper termination.
+///
+/// # Returns
+///
+/// A `Cow<'_, CStr>` containing a *guaranteed* null-terminated string.
+pub fn vec_to_cstr(mut bytes: Vec<u8>) -> Cow<'static, CStr> {
```
zeenix, line 144: Ughh.. I had missed that you made this `pub`. Was this on purpose for some reason?

yassinebenarbia: To tell you the truth, I'm not really sure, afair my first intention was to expose some sort of "helper function(s)" that can be used to deal with the `FilePath` type, but now that I look at it, I think it should be private, or removed and implemented directly here https://github.com/z-galaxy/zbus/blob/36f3fe4736e922280d15d75c84fb4edf780a7293/zvariant/src/file_path.rs#L72 (I realllllllly do apologize for that blunder)

zeenix: No worries. We all make mistakes. :) I think what we could do to ensure that no code that decided to use it already (for some strange reason) breaks:

1. We mark it `#[doc(hidden)]` so it's not in the docs and there is no chance of new code using that.
2. We also mark it as `#[deprecated]` so even if new code uses it, they know it's not to relied on.

Any chance you could provide a PR for that? :pray:

zeenix: > Any chance you could provide a PR for that? 🙏

I guess not. :(

zeenix: Created https://github.com/z-galaxy/zbus/issues/1561

yassinebenarbia: I'm so sorry, but life kinda hit hard and got supper busy these days :D I really apologize.

PR #1014, opened 2024-09-22
➕ zb,zv: Add camino feature to impl `Type` for utf-8 path types

peterdelevoryas:
This adds a crate feature to `zbus` and `variant` called `"camino"`, which adds a derivation of `Type` for `camino::Utf8Path` and `camino::Utf8PathBuf`. It also adds `camino` as a dependency of `zvariant`.


zeenix:
Thanks, looks great!

zeenix (approved):

Issue #727, opened 2024-04-18
RFC: Drop COOKIE auth support


zeenix:
COOKIE auth is not needed at all on Unix platforms, has a complicated implementation (especially the server side) and drags in `sha1` dep (see #530). The only real use case I see for it is Windows, where we currently don't support EXTERNAL auth but once that's fixed (#723), we could think about this.

There is also the use case of TCP but I'm not sure, that's worth keeping either. If UDS is available on all platforms (that we care about), there is no good reason to use TCP. For remote use case, TCP should be highly discouraged since it's unencrypted and COOKIE auth won't work for that anyway.

Together with we can #726, we can really simplify the auth code and pipeline the client-side handshake better (making it go much faster).

zeenix:
Dropping the idea in favor of #731.

zeenix:
I'm rethinking this. @elmarco do you think it'd be a big deal?

PR #1039, opened 2024-10-03
💥 zb: Drop support for `DBUS_COOKIE_SHA1` auth mechanism

### LINKED ISSUE #543: Make SHA1 crate dependency feature gated
Zbus crate depends on a crate "SHA1" which is now flagged as "cryptographically broken" by most Rust Component governance tools in major software companies.

I see that SHA1 crate is used for DBUS_SHA1_COOKIE mechanism. I did some reading on the DBUS_SHA1_COOKIE mechanism and looks like it is only needed on legacy unix and windows systems. So, it should be possible to make the DBUS_SHA1_COOKIE feature as optional feature (gated by a feature flag). This will make zbus crate not being flagged by compliance tools and improve its adoption in production.

Thanks,
Hemendra

zeenix:
* It drags the `sha1` crate as a dependency, which can be [problematic for some users].
* It makes the handshake more complex, not allowing to pipeline all the commands.
* It's not widely used. If `EXTERNAL` is not an option, you might as well just use `ANONYMOUS`.
    
Fixes #543, #727.

[problematic for some users]: https://github.com/dbus2/zbus/issues/543


hkr1990:
@zeenix  When is the next release which has SHA1 dependency removed?

elmarco:
@zeenix this breaks the connection with GIO GDBusConnection / GDBusServer on win32.

This makes connection with auto-spawned servers on windows with shm "DBusDaemonAddressInfo" address, using "nonce-tcp" difficult, as the EXTERNAL mechanism is not implemented there: https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gdbusauthmechanismexternal.c?ref_type=heads#L234

Not being able to connect with GDBus servers is an sufficiently annoying regressions. It breaks win32 support for qemu-display for example.

zeenix:
> Not being able to connect with GDBus servers is an annoying regressions. It breaks win32 support for qemu-display for example.

All these points could have been brought up in the months between #727 (note the "RFC" prefix) and it being actually dropped. :disappointed:

For Windows, we still support TCP and last I checked external works there? If anyone really needs windows support, they should work on making UDS with external auth working there.

Windows support has been one of the hardest. I don't even have a Windows machine and have been relying on the CI to tell me each time I had to made changes in windows-specific parts of the code. So at the point, it'd be a lot more easier to convince me to drop Windows support completely then re-adding complexity to support an old auth mechanism that drags in problematic dependencies (see #543).

elmarco:
It's easy to cross-build zbus with --target x86_64-pc-windows-gnu. There isn't much win32-specific code afaik, so he burden should be low, while the benefits are big for cross-platform projects.

I use free-for-developpers VMs, the download has been officially down for a while, although you can still grab it from https://aka.ms/windev_VM_vmware, then convert it to qemu/libvirt with virt-v2v. (https://developer.microsoft.com/en-us/windows/downloads/virtual-machines/ )

The lack of COOKIE_SHA1 auth means that with zbus 5.0, qemu-display can no longer connect to qemu -display dbus. I will attempt to work around it in future qemu releases to allow ANONYMOUS on win32 with p2p.

If I have time left, I will try to implement EXTERN for win32 glib.

zeenix:
> It's easy to cross-build zbus with --target x86_64-pc-windows-gnu. There isn't much win32-specific code afaik, so he burden should be low, while the benefits are big for cross-platform projects.

As the guy who did the (maintainance) work, I can assure you that is not the case. :) Sure, if you don't touch the code and the windows-specific deps don't change, there is nothing much to do. However, as soon as you try to do some refactor or introduce any new feature that touches that code, you've a problem.

I explained the rationale for this change in the description here and in #727.  I'm afraid the ship is sailed on this one.

Issue #964, opened 2024-08-29
Helper for ibus connection creation


zeenix:
We should provide `connection::builder::ibus` for connecting with ibus. The ibus address needs to be fetched from a command: `ibus address`. This can be handled very similarly to [how we handle the launchd address](https://github.com/dbus2/zbus/blob/main/zbus/src/address/transport/launchd.rs#L26).

Issue #965, opened 2024-08-29
Panic when listening for IBus messages


brandonskiser:
When trying to listen for IBus messages on the `"org.freedesktop.IBus.InputContext"` interface, zbus panics because a message is being passed with a sender that isn't a unique name. A sender with something like `"org.freedesktop.IBus.InputContext"` is being passed instead, which causes a panic.

Here is some code for reference that causes the panic - running this and then focusing out of the terminal causes the panic.
```rust
use tokio::process::Command;
use zbus::export::futures_util::TryStreamExt;
use zbus::fdo::DBusProxy;
use zbus::{ConnectionBuilder, MatchRule, MessageStream};

#[tokio::main]
async fn main() {
    // Taking the ibus address from running `ibus address`
    let ibus_address = {
        let output = Command::new("ibus").arg("address").output().await.unwrap();
        String::from_utf8(output.stdout).unwrap().trim().to_owned()
    };
    println!("IBus address: {ibus_address}");

    let ibus_connection = ConnectionBuilder::address(ibus_address.as_str())
        .unwrap()
        .build()
        .await
        .unwrap();
    // Listening on DBus
    DBusProxy::new(&ibus_connection)
        .await
        .unwrap()
        .add_match_rule(
            MatchRule::builder()
                .interface("org.freedesktop.IBus.InputContext")
                .unwrap()
                .build(),
        )
        .await
        .unwrap();
    let mut stream = MessageStream::from(ibus_connection);
    loop {
        // Panics here when focusing out of a terminal.
        match stream.try_next().await {
            Ok(msg) => {
                println!("Message: {:?}", msg);
            }
            Err(err) => eprintln!("Error: {:?}", err),
        }
    }
}
```

Here is a resulting stack trace
```
thread 'zbus::Connection executor' panicked at /home/ANT.AMAZON.COM/bskiser/.cargo/registry/src/index.crates.io-6f17d22bba15001f/zbus-4.4.0/src/message/fields.rs:137:22:
Invalid field reconstruction: InvalidUniqueName("must start with a `:`")
stack backtrace:
   0:     0x5d28d4f2a985 - std::backtrace_rs::backtrace::libunwind::trace::h58eed11393533053
                               at /rustc/051478957371ee0084a7c0913941d2a8c4757bb9/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
[...]

PR #1040, opened 2024-10-04
✨ zb: Add a function to create a builder for IBus

SummerGram:
- Create a function `ibus` in `conn::builder` which calls `ibus address` and then create the builder

Fixs #964.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -74,6 +74,8 @@ jobs:
           sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
           sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
           sudo apt-get install -y dbus
+          sudo apt-get install -y ibus
+          ibus start &
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
@@ -83,7 +85,7 @@ jobs:
           dbus-run-session --config-file /tmp/dbus-session-abstract.conf -- cargo --locked test --profile "$PROFILE" --verbose -- basic_connection
           # All features except tokio.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
-            cargo --locked test --profile "$PROFILE" --verbose --features uuid,url,time,chrono,option-as-array,vsock,bus-impl \
+            cargo --locked test --profile "$PROFILE" --verbose --features uuid,url,time,chrono,option-as-array,vsock,bus-impl,ibus \
               -- --skip fdpass_systemd
           # Test tokio support.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -26,6 +26,8 @@ camino = ["zvariant/camino"]
 bus-impl = ["p2p"]
 # Enables API that is only needed for peer-to-peer (p2p) connections.
 p2p = ["dep:rand"]
+# Enables API that is only needed for IBus Daemon.
+ibus = []
 async-io = [
   "dep:async-io",
   "async-executor",
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -6,6 +6,8 @@ use static_assertions::assert_impl_all;
 use std::net::TcpStream;
 #[cfg(all(unix, not(feature = "tokio")))]
 use std::os::unix::net::UnixStream;
+#[cfg(feature = "ibus")]
+use std::process::Stdio;
 use std::{
     collections::{HashMap, HashSet},
     vec,
@@ -22,6 +24,9 @@ use uds_windows::UnixStream;
 use vsock::VsockStream;
 
 use zvariant::ObjectPath;
+
+#[cfg(feature = "ibus")]
+use crate::process::Command;
 
 use crate::{
     address::{self, Address},
@@ -83,6 +88,36 @@ impl<'a> Builder<'a> {
     /// Create a builder for the system-wide message bus connection.
     pub fn system() -> Result<Self> {
         Ok(Self::new(Target::Address(Address::system()?)))
+    }
+
+    /// Create a builder for the [IBus] daemon.
+    ///
+    /// [IBus]: https://en.wikipedia.org/wiki/Intelligent_Input_Bus
+    #[cfg(feature = "ibus")]
+    pub async fn ibus() -> Result<Self> {
+        let child_process = Command::new("ibus")
+            .args(["address"])
+            .stdout(Stdio::piped())
+            .spawn()
+            .expect("Fail to call `ibus address`");
+
+        #[cfg(not(feature = "tokio"))]
+        let output = child_process
+            .output()
+            .await
+            .expect("Fail to run `ibus address`");
+
+        #[cfg(feature = "tokio")]
+        let output = child_process
+            .wait_with_output()
+            .await
+            .expect("Fail to run `ibus address`");
+
+        let ibus_address = std::str::from_utf8(&output.stdout)
+            .expect("Invalid utf8 when getting stdout")
+            .trim();
+
+        Builder::address(ibus_address)
     }
 
     /// Create a builder for a connection that will use the given [D-Bus bus address].
--- a/zbus/tests/basic.rs
+++ b/zbus/tests/basic.rs
@@ -117,6 +117,19 @@ async fn fdpass_systemd_async() {
     f.metadata().unwrap();
 }
 
+#[cfg(all(target_os = "linux", feature = "ibus"))]
+#[test]
+fn ibus() {
+    use zbus::blocking;
+
+    let ibus_blocking = blocking::connection::Builder::ibus()
+        .unwrap()
+        .build()
+        .unwrap();
+
+    assert!(!ibus_blocking.server_guid().is_empty());
+}
+
 #[test]
 #[instrument]
 #[timeout(15000)]
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -37,6 +37,14 @@ impl<'a> Builder<'a> {
         crate::connection::Builder::system().map(Self)
     }
 
+    /// Create a builder for the [IBus] daemon.
+    ///
+    /// [IBus]: https://en.wikipedia.org/wiki/Intelligent_Input_Bus
+    #[cfg(feature = "ibus")]
+    pub fn ibus() -> Result<Self> {
+        block_on(crate::connection::Builder::ibus()).map(Self)
+    }
+
     /// Create a builder for a connection that will use the given [D-Bus bus address].
     ///
     /// [D-Bus bus address]: https://dbus.freedesktop.org/doc/dbus-specification.html#addresses


zbus/src/connection/builder.rs
```diff
@@ -88,6 +89,22 @@ impl<'a> Builder<'a> {
         Self::address(&crate::address::system()?)
     }
 
+    /// Create a builder for an IBus connection
+    pub fn ibus() -> Result<Self> {
+        let ibus_address = {
+            let output = Command::new("ibus")
+                .arg("address")
+                .output()
+                .expect("Fail to run `ibus address`");
```
zeenix, line 98: This needs to be async. There is a good reason why I asked you to base this on https://github.com/dbus2/zbus/pull/976 . You at least need the first 2 commits from there. Please rebase this branch on that PR's branch and make use of the `Command` API introduced there.

zbus/src/connection/builder.rs
```diff
@@ -88,6 +89,22 @@ impl<'a> Builder<'a> {
         Self::address(&crate::address::system()?)
     }
 
+    /// Create a builder for an IBus connection
```
zeenix, line 92: It's not for *a* IBus connection but **the** IBus daemon.

```suggestion
    /// Create a builder for the [IBus] daemon.
    ///
    /// [IBus]: https://en.wikipedia.org/wiki/Intelligent_Input_Bus
```

zbus/src/connection/builder.rs
```diff
@@ -88,6 +89,22 @@ impl<'a> Builder<'a> {
...
+    pub fn ibus() -> Result<Self> {
```
zeenix, line 93: Most users will not need this (infact until #965 is resolved, we may not have any users of this) so it needs to be feature-gated. Please add a new `ibus` cargo feature and enable this only when that's enabled.

zbus/src/connection/builder.rs
```diff
@@ -88,6 +89,22 @@ impl<'a> Builder<'a> {
...
+        let ibus_address = {
+            let output = Command::new("ibus")
+                .arg("address")
+                .output()
+                .expect("Fail to run `ibus address`");
+            String::from_utf8(output.stdout)
```
zeenix, line 99: `std::str::from_utf8` is sufficient here and avoids an allocation/copy. You just have to break this statement into two, so you save the output somewhere first and then create a `&str` ref to that.

zbus/src/connection/builder.rs
```diff
@@ -88,6 +89,22 @@ impl<'a> Builder<'a> {
...
+                .expect("Invalid utf8 when getting stdout")
+                .trim()
+                .to_owned()
```
zeenix, line 102: you can remove this as well then and save yet another allocation/copy.

zeenix (changes requested):
Thanks so much. It looks good for a first-time contributions but it will need a bit of work still. Would be great to also add a test case and enable that in the CI (only for the Linux part). You'll probably need to try and install the appropriate ibus package.

zeenix:
Oh and I really like that you added an emoji as per our practices. However a 💥 isn't that best choice here since it's not a breaking change.

zbus/src/connection/builder.rs
SummerGram: I thought it is a connection to the IBus from the program calling the ZBus API so that it can retrieve from or send packet to the IBus

zbus/src/connection/builder.rs
SummerGram: What is the major impact if we add feature for this function? Does it only affect the size of the program?

I tried to solve #965 but not quite success.

SummerGram:
> Oh and I really like that you added an emoji as per our practices. However a 💥 isn't that best choice here since it's not a breaking change.

Haha! I just follow the repo practice. I tried to add long and detailed message in a commit but not work.

zbus/src/connection/builder.rs
SummerGram: I make a change. Please have a look!

I learnt Rust and long time not programming with Rust. Thanks for pointing it out!

zbus/src/connection/builder.rs
SummerGram: I see you didn't merge it to your main yet.

Which branch should I rebase on?

zbus/src/connection/builder.rs
zeenix: yeah that's correct but it's the specific IBus daemon you connect to. Hence use of "the" rather than "a".

zbus/src/connection/builder.rs
zeenix: > I see you didn't merge it to your main yet.

Yeah, there is still some minor stuff needed there but that's only about the later commits, which shouldn't affect your work.

> Which branch should I rebase on?

The [repo+branch of the PR](https://github.com/vially/zbus/tree/unixexec).

zbus/src/connection/builder.rs
zeenix: > What is the major impact if we add feature for this function? Does it only affect the size of the program?

Yeah and compilation speed. The impact is small in this case but given that 99% of users won't be using it, best to not enable it by default.

zbus/src/connection/builder.rs
zeenix: > I make a change. Please have a look!

Will do, thanks.



> I learnt Rust and long time not programming with Rust. Thanks for pointing it out!

No worries.

zeenix: > > I make a change. Please have a look!
> 
> Will do, thanks.

You didn't make the change to `std::str::from_utf8` and given you added another scope, you're forcing yourself to use the allocating/copying `String` API. Also see [my other comment](https://github.com/dbus2/zbus/pull/1040/files#r1788132053).

zeenix:
> > Oh and I really like that you added an emoji as per our practices. However a 💥 isn't that best choice here since it's not a breaking change.
> 
> Haha! I just follow the repo practice. I tried to add long and detailed message in a commit but not work.

Cool. Can you tell me how your'e choosing/selecting the emojis? The `➕` one is for addition of new dependency and you're not adding one. I think you want either ✨ or 🔌. If you're using `gimji`, it should have a description to guide your selection.

zbus/Cargo.toml
```diff
@@ -26,6 +26,8 @@ camino = ["zvariant/camino"]
 bus-impl = ["p2p"]
 # Enables API that is only needed for peer-to-peer (p2p) connections.
 p2p = []
+# Enables API that is only needed for IBus Daemon.
+ibus = []
```
zeenix, line 30: Nice! Please make sure the commit to introduce this feature comes before use of it, or alternatively you can just squash this into that commit. Just make sure to mention in the commit message details that the new function is  only enabled by the new cargo feature, `ibus`.

zbus/src/connection/builder.rs
```diff
@@ -89,7 +89,10 @@ impl<'a> Builder<'a> {
         Self::address(&crate::address::system()?)
     }
 
-    /// Create a builder for an IBus connection
+    /// Create a builder for the [IBus] daemon.
+    ///
+    /// [IBus]: https://en.wikipedia.org/wiki/Intelligent_Input_Bus
```
zeenix: this is a fix to your first commit here and should be part of that, not the one introducing the feature. Of course, if you go for squashing them both, then it'll be fixed by that automatically. :)

zbus/tests/basic.rs
```diff
@@ -112,6 +112,15 @@ fn fdpass_systemd() {
     f.metadata().unwrap();
 }
 
+#[cfg(target_os = "linux")]
+#[cfg(feature = "ibus")]
```
zeenix, line 116: you can combine them using `#[cfg(all(..))]`

zbus/src/connection/builder.rs
SummerGram: Oh, should I fix the first commit as a new one? Introducing the feature be another one commit?

zbus/tests/basic.rs
```diff
@@ -112,6 +112,15 @@ fn fdpass_systemd() {
...
+#[test]
+fn test_ibus() {
+    let ibus = block_on(conn::Builder::ibus().unwrap().build()).unwrap();
+
+    assert!(!ibus.server_guid().as_str().is_empty());
```
zeenix, line 121: if we're going to check the guid, we should do a bit better job than just checking if it's empty. :) You can issue the command here to re-fetch the guid and then compare the guids.

zbus/tests/basic.rs
```diff
@@ -112,6 +112,15 @@ fn fdpass_systemd() {
     f.metadata().unwrap();
 }
 
+#[cfg(target_os = "linux")]
+#[cfg(feature = "ibus")]
+#[test]
+fn test_ibus() {
+    let ibus = block_on(conn::Builder::ibus().unwrap().build()).unwrap();
```
zeenix, line 119: this reminds me, you need to also add the blocking variant of the method in `blocking::connection::Builder`. Then use that here.

SummerGram:
> > > Oh and I really like that you added an emoji as per our practices. However a 💥 isn't that best choice here since it's not a breaking change.
> > 
> > 
> > Haha! I just follow the repo practice. I tried to add long and detailed message in a commit but not work.
> 
> Cool. Can you tell me how your'e choosing/selecting the emojis? The `➕` one is for addition of new dependency and you're not adding one. I think you want either ✨ or 🔌. If you're using `gimji`, it should have a description to guide your selection.

I just go to web and pick emoji. Now I install `gimoji`. Definitely use it!

zbus/src/connection/builder.rs
zeenix: > Oh, should I put the fix of first commit as a new one?

No, the fixes to the (merged) commits should be squashed into the commits they're fixing.

> Introducing the feature be another one commit?

Ideally, yes. This specific change is not the one introducing the cargo feature so this (only talking about these lines here) should be part of the commit that adds the new method. If you keep the introduction of the cargo feature are a separate commit, that commit should come first obviously because the method uses it just before this line here. You'll have to split the commit and then move part of it before.

Alternatively, you can just keep it simple and have both changes in one commit but then please add a bit more description to the commit log and mention that fact that it's optional and not enabled by default.

SummerGram:
I rebase it and stash it.

Please comment it

zeenix:
> I rebase it and stash it.

Oh, I hadn't realized you already resolved issues 5 days ago.

Unfortunately, the main branch diverged in the meantime so you've to rebase again. Sorry.
 
> Please comment it

Will do after you've rebased again.

zeenix:
@SummerGram Oh, nm my previous comment (now deleted). I forgot that you were asked to rebase on #325. Unfortunately, that branch/PR needs rebasing too now. I'll ask the author if they'd still be moving the PR forward..

SummerGram:
I see. I guess that PR will merge to the current main. And what I do is rebase to that new merged main.

Please tell me when it is merged here or in matrix. I do not check that PR often.

zeenix:
> I see. I guess that PR will merge to the current main. And what I do is rebase to that new merged main.

Yes. :)

> Please tell me when it is merged here or in matrix. I do not check that PR often.

Sure thing!

zeenix:
@SummerGram seems you're not around on Matrix so just to let you know #976 has been merged. Please rebase and address the comments so we could get this merged too. :pray:

SummerGram:
> @SummerGram seems you're not around on Matrix so just to let you know #976 has been merged. Please rebase and address the comments so we could get this merged too. 🙏

Mind if I create a new PR?

zeenix:
> > @SummerGram seems you're not around on Matrix so just to let you know #976 has been merged. Please rebase and address the comments so we could get this merged too. 🙏
> 
> Mind if I create a new PR?

Even if you start fresh, you can just force push to the same branch so there's no need. It'll only create noise.

SummerGram:
@zeenix 

Can you please check the base branch correct or not?

zeenix:
> @zeenix
> 
> Can you please check the base branch correct or not?

Given that the PR now only has commits from your iBus work, I think it's a pretty safe bet the base is now the `main` branch. :+1: 

I'll have a look soon but please make sure that commits that are fixups to previous commits (in the same PR/branch), are squashed into those previous commits.

zbus/tests/basic.rs
```diff
@@ -118,13 +118,20 @@ async fn fdpass_systemd_async() {
     f.metadata().unwrap();
 }
 
-#[cfg(target_os = "linux")]
-#[cfg(feature = "ibus")]
+#[cfg(all(target_os = "linux", feature = "ibus"))]
```
zeenix: For example, this is just a fixup of the commit that adds the `test_ibus`. Also, if you move the commit that adds the testcase after the blocking version addition, you can just squashed this whole ("zb: update the test code of IBus") commit into that one.

zbus/tests/basic.rs
```diff
@@ -118,13 +118,20 @@ async fn fdpass_systemd_async() {
...
 #[test]
 fn test_ibus() {
```
zeenix, line 123: nit: please name it `ibus`. `test_` prefix is redundant.

zbus/tests/basic.rs
SummerGram: Ha, I guess it is good to record the code improvements.

zeenix: True but when people look at your commits, they'll only be interested in what changed and why. Also if for any reason (doubtful but possible) we need to revert your changes later, it's easier to do so with atomic (as in, one logical change per commit) commits.

Besides, this goes in

zbus/tests/basic.rs
```diff
@@ -117,6 +117,24 @@ async fn fdpass_systemd_async() {
...
+#[cfg(all(target_os = "linux", feature = "ibus"))]
+#[test]
+fn ibus() {
+    use zbus::{blocking, conn};
+
+    let ibus_blocking = blocking::connection::Builder::ibus()
+        .unwrap()
+        .build()
+        .unwrap();
+    let ibus_non_blocking = block_on(block_on(conn::Builder::ibus()).unwrap().build()).unwrap();
```
zeenix, line 129: the blocking wrapper does exactly this, so I don't think this is needed.

zbus/tests/basic.rs
```diff
@@ -117,6 +117,24 @@ async fn fdpass_systemd_async() {
...
+
+    assert!(!ibus_blocking.server_guid().is_empty());
+    assert_eq!(
+        ibus_blocking.server_guid().is_empty(),
+        ibus_non_blocking.server_guid().is_empty()
+    );
```
zeenix, line 135: comparing booleans is always confusing. :) This statement can be removed as well for the same reason mentioned above.

.github/workflows/rust.yml
```diff
@@ -74,6 +74,7 @@ jobs:
           sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
           sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
           sudo apt-get install -y dbus
+          sudo apt-get install -y ibus
```
zeenix, line 77: Nitpick: The containing commit is not touch `zbus` subcrate. So the prefix should be `CI: `.

zeenix (changes requested):
Getting very close, I think. Apart from these minor things, you need to check the CI. The error is very strange since ibus address on my machine does have a `:` in it, as it should.

zbus/tests/basic.rs
SummerGram: The blocking api will call the non-blocking one. It is redundancy to call it 1 more time.

.github/workflows/rust.yml
SummerGram: maybe need to start the ibus `ibus start &`

zbus/tests/basic.rs
zeenix: I'm not sure I follow. You wrote the same thing I did, in different words? :thinking:

zbus/tests/basic.rs
zeenix: As I wrote before, fixups to existing commits should go into those commits. So after making the changes, you need something like `git commit --fixup HASH_OF_TARGET_COMMIT && git rebase --autosquash -i HASH_OF_TARGET_COMMIT^` (notice the use of `^` to refer to the parent commit).

zeenix: after that `git push --force`.

.github/workflows/rust.yml
zeenix: I was talking about the git commit message, not the change itself. I forgot that one can add comments to a file as whole (rather than to specific changes in the file). Wish there was a way to do that on commit as well.

zbus/src/connection/builder.rs
```diff
@@ -85,6 +90,36 @@ impl<'a> Builder<'a> {
...
+        let output = child_process
+            .output()
+            .await
+            .expect("Fail to run `ibus address`");
+
+        #[cfg(feature = "tokio")]
+        let output = child_process
+            .wait_with_output()
+            .await
+            .expect("Fail to run `ibus address`");
```
zeenix, line 114: Sorry, just noticed this part. Why do they need to be different? `tokio::process::Command` also has `output` method, which should be good enough for us?

Actually, `abstractions::process::Command::output` can do most stuff already for us here? :thinking: You don't even need to bother with piping etc.

zeenix:
> Apart from these minor things, you need to check the CI. The error is very strange since ibus address on my machine does have a `:` in it, as it should.

Did you try to fix this?

PR #1047, opened 2024-10-05
⚡️ zb: Remove unnecessary boxing of a future

zeenix:


zeenix:
hmm.. strange, somehow it worked on my machine but the CI tells me [it doesn't ](https://github.com/dbus2/zbus/actions/runs/11195497749/job/31123230082?pr=1047) (and it explains why it was using a Box).

Issue #1006, opened 2024-09-17
Merge zbus_names into zbus


zeenix:
The zbus_names was created in the hope that it will be shared by other D-Bus crates, and help us cooperate.

However this didn't happen and looks unlikely to happen for a variety of reasons. Instead, it only creates maintenance burden each time changes are needed in this crate. 

I think it's best if this is merged with zbus itself.

elmarco:
I think the idea of having common types in a shared crate is good. However, pulling serde & zvariant goes against the simple goal. At the very least, they should be made optional. But yes, it's a niche use case. 

To help with maintainance, are you using workspace release tools?

zeenix:
> I think the idea of having common types in a shared crate is good

I used to believe so but experience has taught me that unless they're actually shared with other projects, it's not worth the trouble. A good example of smol-rs. Its sub-crates are nicely split but they are actually being used separately by various projects, so the split is well worth it. 

OTOH, there is also the approach tokio takes: everything is one big crate but feature-gate everything so if you want to only use a specific API, you only enable that. 

I'd be happy to go towards the tokio approach.

zeenix:
> To help with maintainance, are you using workspace release tools?

I tried a few times but I found them too complicated and haven't found any tool that handles release notes in the git tag, like we currently do (I like that we have release notes both in the tag and GitHub release).

However, even if releasing was easy, given that zbus_names have no other users other than zbus after 3 years, I don't think this split is worth it. 

We have to also keep in mind that split increases the cold build times and gives users the impression that we've too many deps (something people bring up every now and then).

elmarco:
Given the relation of zbus_names with zv, I think it could make sense to bring it back under zbus. I don't think the point about maintainance holds much for zbus_address though.

Split doesn't necessarily increase build time, sometime it's the other way around. Pros and cons... the subject isn't closed and not worth to debate here without strong official recommandations.

Regarding the number of dependencies, I think experienced developpers are used to have a growing list of deps that is in the order of hundreds. So people use tools to enforce rules regarding license and build checks. It's pretty much the norm.

zeenix:
> Regarding the number of dependencies, I think experienced developpers are used to have a growing list of deps that is in the order of hundreds. So people use tools to enforce rules regarding license and build checks. It's pretty much the norm.

[...]

PR #1050, opened 2024-10-07
Refactoring as requested 

### LINKED ISSUE #1064: Remaining issues with new address API
This is a issue to track the progress of issues and concerns identified with the new `address` API. See https://github.com/dbus2/zbus/pull/1050 and https://github.com/dbus2/zbus/pull/989.

The main concerns right now:

* The address string must only be parsed once and never after an `Address` instance is created. Similarly for other inner types, such as `Transort`.
* The `address` module hierarchy should use `zvariant::Str` everywhere, instead of `Cow`.
* `Address::guid` should return `zbus::Guid` type.

elmarco:
Introduce TransportTrait with from_address() and fmt_key_val(), as requested in #982.

Fixes #1064.

### DIFF
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -58,6 +58,24 @@ pub enum Transport<'a> {
     Vsock(vsock::Vsock<'a>),
 }
 
+impl<'a> Transport<'a> {
+    /// Convert into owned version, with 'static lifetime.
+    pub fn into_owned(self) -> Transport<'static> {
+        match self {
+            Transport::Unix(unix) => Transport::Unix(unix.into_owned()),
+            #[cfg(target_os = "macos")]
+            Transport::Launchd(launchd) => Transport::Launchd(launchd.into_owned()),
+            #[cfg(target_os = "linux")]
+            Transport::Systemd(systemd) => Transport::Systemd(systemd.into_owned()),
+            Transport::Tcp(tcp) => Transport::Tcp(tcp.into_owned()),
+            Transport::NonceTcp(nonce_tcp) => Transport::NonceTcp(nonce_tcp.into_owned()),
+            Transport::Unixexec(unixexec) => Transport::Unixexec(unixexec.into_owned()),
+            Transport::Autolaunch(autolaunch) => Transport::Autolaunch(autolaunch.into_owned()),
+            Transport::Vsock(vsock) => Transport::Vsock(vsock.into_owned()),
+        }
+    }
+}
+
 impl fmt::Display for Transport<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -35,7 +35,7 @@ mod vsock;
 pub use vsock::Vsock;
 
 /// A D-Bus transport.
-#[derive(Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 #[non_exhaustive]
 pub enum Transport<'a> {
     /// Unix Domain Sockets transport.
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -2,7 +2,7 @@
 
 use std::fmt;
 
-use super::{percent, Address, Error, KeyValFmt, KeyValFmtAdd, Result};
+use super::{percent, Address, Error, KeyValFmt, Result};
 
 mod autolaunch;
 pub use autolaunch::Autolaunch;
@@ -79,23 +79,9 @@ pub(crate) trait TransportImpl<'a> {
     fn for_address(s: &'a Address<'a>) -> Result<Self>
     where
         Self: Sized;
-}
 
-impl KeyValFmtAdd for Transport<'_> {
-    fn key_val_fmt_add<'a: 'b, 'b>(&'a self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
-        match self {
-            Self::Unix(t) => t.key_val_fmt_add(kv),
-            #[cfg(target_os = "macos")]
-            Self::Launchd(t) => t.key_val_fmt_add(kv),
-            #[cfg(target_os = "linux")]
-            Self::Systemd(t) => t.key_val_fmt_add(kv),
-            Self::Tcp(t) => t.key_val_fmt_add(kv),
-            Self::NonceTcp(t) => t.key_val_fmt_add(kv),
-            Self::Unixexec(t) => t.key_val_fmt_add(kv),
-            Self::Autolaunch(t) => t.key_val_fmt_add(kv),
-            Self::Vsock(t) => t.key_val_fmt_add(kv),
-        }
-    }
+    // add() the transport (key,val) pairs to KeyValFmt
+    fn fmt_key_val<'s: 'b, 'b>(&'s self, kv: KeyValFmt<'b>) -> KeyValFmt<'b>;
 }
 
 impl<'a> TransportImpl<'a> for Transport<'a> {
@@ -113,6 +99,21 @@ impl<'a> TransportImpl<'a> for Transport<'a> {
             "autolaunch" => Ok(Self::Autolaunch(Autolaunch::for_address(s)?)),
             "vsock" => Ok(Self::Vsock(Vsock::for_address(s)?)),
             _ => Err(Error::UnknownTransport),
+        }
+    }
+
+    fn fmt_key_val<'s: 'b, 'b>(&'s self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
+        match self {
+            Self::Unix(t) => t.fmt_key_val(kv),
+            #[cfg(target_os = "macos")]
+            Self::Launchd(t) => t.fmt_key_val(kv),
+            #[cfg(target_os = "linux")]
+            Self::Systemd(t) => t.fmt_key_val(kv),
+            Self::Tcp(t) => t.fmt_key_val(kv),
+            Self::NonceTcp(t) => t.fmt_key_val(kv),
+            Self::Unixexec(t) => t.fmt_key_val(kv),
+            Self::Autolaunch(t) => t.fmt_key_val(kv),
+            Self::Vsock(t) => t.fmt_key_val(kv),
         }
     }
 }
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -75,6 +75,12 @@ impl fmt::Display for Transport<'_> {
     }
 }
 
+pub(crate) trait TransportImpl<'a> {
+    fn for_address(s: &'a Address<'a>) -> Result<Self>
+    where
+        Self: Sized;
+}
+
 impl KeyValFmtAdd for Transport<'_> {
     fn key_val_fmt_add<'a: 'b, 'b>(&'a self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
         match self {
@@ -92,22 +98,20 @@ impl KeyValFmtAdd for Transport<'_> {
     }
 }
 
-impl<'a> TryFrom<&'a Address<'a>> for Transport<'a> {
-    type Error = Error;
-
-    fn try_from(s: &'a Address<'a>) -> Result<Self> {
+impl<'a> TransportImpl<'a> for Transport<'a> {
+    fn for_address(s: &'a Address<'a>) -> Result<Self> {
         let col = s.addr.find(':').ok_or(Error::MissingTransport)?;
         match &s.addr[..col] {
-            "unix" => Ok(Self::Unix(s.try_into()?)),
+            "unix" => Ok(Self::Unix(Unix::for_address(s)?)),
             #[cfg(target_os = "macos")]
-            "launchd" => Ok(Self::Launchd(s.try_into()?)),
+            "launchd" => Ok(Self::Launchd(Launchd::for_address(s)?)),
             #[cfg(target_os = "linux")]
-            "systemd" => Ok(Self::Systemd(s.try_into()?)),
-            "tcp" => Ok(Self::Tcp(s.try_into()?)),
-            "nonce-tcp" => Ok(Self::NonceTcp(s.try_into()?)),
-            "unixexec" => Ok(Self::Unixexec(s.try_into()?)),
-            "autolaunch" => Ok(Self::Autolaunch(s.try_into()?)),
-            "vsock" => Ok(Self::Vsock(s.try_into()?)),
+            "systemd" => Ok(Self::Systemd(Systemd::for_address(s)?)),
+            "tcp" => Ok(Self::Tcp(Tcp::for_address(s)?)),
+            "nonce-tcp" => Ok(Self::NonceTcp(NonceTcp::for_address(s)?)),
+            "unixexec" => Ok(Self::Unixexec(Unixexec::for_address(s)?)),
+            "autolaunch" => Ok(Self::Autolaunch(Autolaunch::for_address(s)?)),
+            "vsock" => Ok(Self::Vsock(Vsock::for_address(s)?)),
             _ => Err(Error::UnknownTransport),
         }
     }
--- a/zbus/src/address/transport/nonce_tcp.rs
+++ b/zbus/src/address/transport/nonce_tcp.rs
@@ -53,6 +53,25 @@ impl<'a> NonceTcp<'a> {
     pub fn noncefile(&self) -> Option<&OsStr> {
         self.noncefile.as_ref().map(|v| v.as_ref())
     }
+
+    /// Convert into owned version, with 'static lifetime.
+    pub fn into_owned(self) -> NonceTcp<'static> {
+        let Self {
+            host,
+            bind,
+            port,
+            family,
+            noncefile,
+        } = self;
+
+        NonceTcp {
+            host: host.map(|h| h.into_owned().into()),
+            bind: bind.map(|b| b.into_owned().into()),
+            port,
+            family,
+            noncefile: noncefile.map(|n| n.into_owned().into()),
+        }
+    }
 }
 
 impl<'a> TransportImpl<'a> for NonceTcp<'a> {
--- a/zbus/src/address/transport/nonce_tcp.rs
+++ b/zbus/src/address/transport/nonce_tcp.rs
@@ -3,7 +3,7 @@ use std::{borrow::Cow, ffi::OsStr};
 use super::{
     percent::{decode_percents_os_str, decode_percents_str, EncOsStr},
     tcp::TcpFamily,
-    Address, Error, KeyValFmt, KeyValFmtAdd, Result, TransportImpl,
+    Address, Error, KeyValFmt, Result, TransportImpl,
 };
 
 /// `nonce-tcp:` D-Bus transport.
@@ -55,16 +55,6 @@ impl<'a> NonceTcp<'a> {
     }
 }
 
-impl KeyValFmtAdd for NonceTcp<'_> {
-    fn key_val_fmt_add<'a: 'b, 'b>(&'a self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
-        kv.add("host", self.host())
-            .add("bind", self.bind())
-            .add("port", self.port())
-            .add("family", self.family())
-            .add("noncefile", self.noncefile().map(EncOsStr))
-    }
-}
-
 impl<'a> TransportImpl<'a> for NonceTcp<'a> {
     fn for_address(s: &'a Address<'a>) -> Result<Self> {
         let mut res = NonceTcp::default();
@@ -94,5 +84,13 @@ impl<'a> TransportImpl<'a> for NonceTcp<'a> {
         }
 
         Ok(res)
+    }
+
+    fn fmt_key_val<'s: 'b, 'b>(&'s self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
+        kv.add("host", self.host())
+            .add("bind", self.bind())
+            .add("port", self.port())
+            .add("family", self.family())
+            .add("noncefile", self.noncefile().map(EncOsStr))
     }
 }
--- a/zbus/src/address/transport/nonce_tcp.rs
+++ b/zbus/src/address/transport/nonce_tcp.rs
@@ -9,7 +9,7 @@ use super::{
 /// `nonce-tcp:` D-Bus transport.
 ///
 /// <https://dbus.freedesktop.org/doc/dbus-specification.html#transports-nonce-tcp-sockets>
-#[derive(Debug, Default, PartialEq, Eq)]
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
 pub struct NonceTcp<'a> {
     host: Option<Cow<'a, str>>,
     bind: Option<Cow<'a, str>>,
--- a/zbus/src/address/transport/nonce_tcp.rs
+++ b/zbus/src/address/transport/nonce_tcp.rs
@@ -3,7 +3,7 @@ use std::{borrow::Cow, ffi::OsStr};
 use super::{
     percent::{decode_percents_os_str, decode_percents_str, EncOsStr},
     tcp::TcpFamily,
-    Address, Error, KeyValFmt, KeyValFmtAdd, Result,
+    Address, Error, KeyValFmt, KeyValFmtAdd, Result, TransportImpl,
 };
 
 /// `nonce-tcp:` D-Bus transport.
@@ -65,10 +65,8 @@ impl KeyValFmtAdd for NonceTcp<'_> {
     }
 }
 
-impl<'a> TryFrom<&'a Address<'a>> for NonceTcp<'a> {
-    type Error = Error;
-
-    fn try_from(s: &'a Address<'a>) -> Result<Self> {
+impl<'a> TransportImpl<'a> for NonceTcp<'a> {
+    fn for_address(s: &'a Address<'a>) -> Result<Self> {
         let mut res = NonceTcp::default();
         for (k, v) in s.key_val_iter() {
             match (k, v) {
--- a/zbus/src/connection/connect/win32.rs
+++ b/zbus/src/connection/connect/win32.rs
@@ -2,7 +2,7 @@
 
 use super::ConnectResult;
 use crate::{
-    address::{transport::Transport, Address},
+    address::{transport::Transport, OwnedAddress},
     win32::autolaunch_bus_address,
     Error,
 };
@@ -17,9 +17,9 @@ pub(crate) fn connect<'l>(
             return Err(Error::Unsupported);
         }
 
-        let addr: Address<'_> = autolaunch_bus_address()?.try_into()?;
+        let addr: OwnedAddress = autolaunch_bus_address()?.try_into()?;
 
-        if let Transport::Autolaunch(_) = addr.transport()? {
+        if let Transport::Autolaunch(_) = addr.transport() {
             return Err(Error::Failure("Recursive autolaunch: address".into()));
         }
 
--- a/zbus/src/connection/connect/win32.rs
+++ b/zbus/src/connection/connect/win32.rs
@@ -9,10 +9,10 @@ use crate::{
 use std::{future::Future, pin::Pin};
 
 pub(crate) fn connect<'l>(
-    l: &'l crate::address::transport::Autolaunch<'_>,
+    autolaunch: &'l crate::address::transport::Autolaunch<'_>,
 ) -> Pin<Box<dyn Future<Output = ConnectResult> + 'l>> {
     Box::pin(async move {
-        if l.scope().is_some() {
+        if autolaunch.scope().is_some() {
             return Err(Error::Address(
                 "autolaunch with scope isn't supported yet".into(),
             ));
--- a/zbus/src/connection/connect/win32.rs
+++ b/zbus/src/connection/connect/win32.rs
@@ -13,15 +13,14 @@ pub(crate) fn connect<'l>(
 ) -> Pin<Box<dyn Future<Output = ConnectResult> + 'l>> {
     Box::pin(async move {
         if autolaunch.scope().is_some() {
-            return Err(Error::Address(
-                "autolaunch with scope isn't supported yet".into(),
-            ));
+            tracing::debug!("autolaunch with scope isn't supported yet");
+            return Err(Error::Unsupported);
         }
 
         let addr: Address<'_> = autolaunch_bus_address()?.try_into()?;
 
         if let Transport::Autolaunch(_) = addr.transport()? {
-            return Err(Error::Address("Recursive autolaunch: address".into()));
+            return Err(Error::Failure("Recursive autolaunch: address".into()));
         }
 
         super::connect(&addr).await
--- a/zbus/src/address/address_list.rs
+++ b/zbus/src/address/address_list.rs
@@ -1,6 +1,6 @@
 use std::{borrow::Cow, fmt};
 
-use super::{Address, Error, Result, ToAddresses};
+use super::{Address, Error, OwnedAddress, Result, ToAddresses, ToOwnedAddresses};
 
 /// A bus address list.
 ///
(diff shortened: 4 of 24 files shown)


zbus/src/address/transport/mod.rs
```diff
@@ -75,6 +75,12 @@ impl fmt::Display for Transport<'_> {
     }
 }
 
+pub(crate) trait TransportTrait<'a> {
```
zeenix, line 78: Too bad `Tranport` name is already taken. I think the convention is `TransportImpl` though.

zbus/src/address/transport/mod.rs
```diff
@@ -75,6 +75,12 @@ impl fmt::Display for Transport<'_> {
...
+    fn from_address(s: &'a Address<'a>) -> Result<Self>
```
zeenix, line 79: `for_address`. Remember, it's not a conversion.

zbus/src/address/transport/nonce_tcp.rs
```diff
@@ -95,4 +85,12 @@ impl<'a> TransportTrait<'a> for NonceTcp<'a> {
 
         Ok(res)
     }
+
+    fn fmt_key_val<'s: 'b, 'b>(&'s self, kv: KeyValFmt<'b>) -> KeyValFmt<'b> {
```
zeenix, line 89: some docs about what this is all about, would be nice.

zeenix (changes requested):
Thanks. Some minor points but great otherwise. Also some of the feedback hasn't yet been addressed (e.g [this](https://github.com/dbus2/zbus/pull/989#discussion_r1784839367) and [this](https://github.com/dbus2/zbus/pull/989#discussion_r1786134746)) but I'm guessing it'll come in a following PR?

zbus/src/connection/connect/win32.rs
```diff
@@ -8,9 +8,9 @@ use crate::{
 };
 
 pub(crate) async fn connect(
-    l: &crate::address::transport::Autolaunch<'_>,
+    autolaunch: &crate::address::transport::Autolaunch<'_>,
```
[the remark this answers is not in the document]
elmarco: ye

zbus/src/address/transport/nonce_tcp.rs
elmarco: `// add() the transport (key,val) pairs to KeyValFmt` ?

zeenix: works for me, if you also document `KeyValFmt`. :)

zeenix:
While working on the busd port, I realized some other issues with the new address API:

* It uses Cow instead of zvariant::Str, which allows for avoiding allocations when built from static string. There is a very good reason it exists and why we're using it everywhere.
* Address::guid returns a string, instead of our Guid type (which makes use of `Str` type), that's meant for exactly this.

Please address these along with other remaining concerns withing a few days time. I need this all sorted out ASAP so I can finish the release. Otherwise, I'll be forced to revert all your work.

elmarco:
- Could you explain why Str is better than Cow with 'static str ? this is against the my design decision to make address parsing a standalone code.
- Similarly Guid has some zbus-specifics which are not desirable for a generic parsing code. Wrapping the returned value should have no additional cost.

Fwiw, your tone has been concerning. Threatening to disregard the time and effort I have already invested, especially after you accepted it, is not an effective way to manage a project or its team members. This approach needs to change if I am to continue working with you. Thank you.

zeenix:
I did some of the work [here](https://github.com/zeenix/zbus/commits/use-str-in-address/) but I need to focus on a bunch of work work now. We need to replace `Cow` with `Str` in other places too then just `Address` struct itself.

elmarco:
> Because `Cow` clones/allocates on `to_owned` even for `&'static str` since it doesn't differentiates between `&str` and `&'static str`. `Str` does not.

Why does it matter in the context of address parsing?

> > this is against the my design decision to make address parsing a standalone code.
> 
> The code we ended up with in the end was not standalone at all, so those design decisions are very irrelevant now.

It is still trivial to adjust the code to compile standalone and I would rather keep it that way. There is still a more valuable future where it can be used without the rest of zbus.
 
> Again, you need to remove all aspects of this design since that's not what we went for.

Because you changed your mind, but the code does not rely on anything from the rest of zbus so far afaik.

> > Threatening to disregard the time and effort I have already invested
> 
> You know very well that you practically forced me to invest a lot of time reviewing changes too, which took a lot of my time as well. And it seems it continues to drain my limited time.

This is not an excuse.

> > especially after you accepted it
> 
> Because you just won't give up about it, despite the fact that you completely failed to provide a single real world use case. Now I'll have to live with a over-complicated API in my project so the least you can do, is adjust to the project conventions in a timely manner.
> 
> I accepted the PR in the end, mainly a personal favour to you. If it was by someone else, I would have definitely rejected it.

It would be a shame to refuse technically superior code. There are real use cases for parsing a dbus address despite you saying not willing to admit it.

> > is not an effective way to manage a project or its team members.
> 
> I'm sorry to inform you but you've not been a team member for a while now. If you were, you wouldn't be questioning `Str`'s use here, which we use everywhere. I'm giving you a way to get back in by adjusting your new API to better fit with the rest of the project and its conventions.

Thanks, that's a bright way to respect contributors and users.

> > Fwiw, your tone has been concerning.
> > This approach needs to change if I am to continue working with you.
> 
> I need to roll-out 5.0 soon. All public API changes need to happen before that but when I mention issues with your changes, you waste my time arguing about it, instead of trusting my judgements as the maintainer and just addressing the concerns.
> 
> I'd not use this tone, if you stop doing that and address the concerns and follow my lead as the maintainer.

You don't communicate well enough your plans. How are we supposed to know 5.0 is going out soon? There are many things that could still be changed during API break period.

And how are the changes you are asking relevant for 5.0? It's only internal refactoring afaict...

You keep me asking for things against my opinions, when you could just do it.

I am tired of being treated like a robot. It's unfair and unpleasant. I have spent a lot of time on zbus, and I plan to continue doing so, but not like this.

zbus/src/address/transport/mod.rs
elmarco: ok

zbus/src/address/transport/nonce_tcp.rs
elmarco: done

zeenix:
> And how are the changes you are asking relevant for 5.0? It's ~only~ mostly internal refactoring afaict...

Sure not all of them are directly relevant but if I were to decide to (even temporarily) revert the new API, I need to decide that before 5.0. Your cooperation and timely actions, can ensure that I don't seriously consider that option.

> > Because `Cow` clones/allocates on `to_owned` even for `&'static str` since it doesn't differentiates between `&str` and `&'static str`. `Str` does not.
> 
> Why does it matter in the context of address parsing?

Not the parsing itself. You think of this as a purely "parsing" API. To me, it's a specific type representing an address and should be used for that instead of strings. I wanted to create a typed const for default address in busd, as an example and that was the reason I thought of this. I'm sure there can be other uses of this.

It's not different than other string-backed specific types we have in the code base.

> I am tired of being treated like a robot. It's unfair and unpleasant. I have spent a lot of time on zbus, and I plan to continue doing so, but not like this.

If I treated you like a robot, I wouldn't have wasted hours of my time, explaining my concerns and reservations, often repeating again and again, would I? :)

zeenix:
> It is still trivial to adjust the code to compile standalone and I would rather keep it that way. There is still a more valuable future where it can be used without the rest of zbus.

Valuable to who? I've been asking you time and again for **any** actual use case for this and you've failed to point out a single one. We keep going in circles with this purely academic goal of yours. Unless you have some new arguments that could possibly convince me, please stop bringing it up. You have wasted a huge amount of my time on this already. Otherwise, this will remain a non-goal for zbus and must not guide any API design.

elmarco:
> > It is still trivial to adjust the code to compile standalone and I would rather keep it that way. There is still a more valuable future where it can be used without the rest of zbus.
> 
> Valuable to who? 

https://github.com/dbus2/zbus/pull/989#discussion_r1791978395

elmarco:
> Sure not all of them are directly relevant but if I were to decide to (even temporarily) revert the new API, I need to decide that before 5.0. Your cooperation and timely actions, can ensure that I don't seriously consider that option.

Can you be more specific on your timeline. We shouldn't put ourself under that kind of stress.

> > Why does it matter in the context of address parsing?
> 
> Not the parsing itself. You think of this as a purely "parsing" API. To me, it's a specific type representing an address and should be used for that instead of strings. I wanted to create a typed const for default address in busd, as an example and that was the reason I thought of this. I'm sure there can be other uses of this.

The job of parsing should remain outside the rest of zbus usage. Why can't it be wrapped later on? This is already an almost 0-cost abstraction, it shouldn't change much to wrap it.

> It's not different than other string-backed specific types we have in the code base.

But other string types are more directly visible on the bus as message data and must be serialized etc. This is arguably also done at later stage and could be independent from zbus in the first place. This is imho one of the reasons why zbus-names can't be adopted outside zbus easily.

zeenix:
> > Sure not all of them are directly relevant but if I were to decide to (even temporarily) revert the new API, I need to decide that before 5.0. Your cooperation and timely actions, can ensure that I don't seriously consider that option.
> 
> Can you be more specific on your timeline.

Not really. It's been in the making for a bit now and people are awaiting the new optimized zbus.

> We shouldn't put ourself under that kind of stress.

Sure but you went silent for a good few days and given you history of disappearing completely for months and not finishing your PRs for years, I thought you might have disappeared again for another few months when you didn't even respond to simple questions on the matrix channel.

> Why can't it be wrapped later on? This is already an almost 0-cost abstraction, it shouldn't change much to wrap it.

Because changing the return type of a method, is a breaking change.

> > It's not different than other string-backed specific types we have in the code base.
> 
> But other string types are more directly visible on the bus as message data and must be serialized etc.

I don't see that as a relevant fact. I presented a very real use case that I was implementing just yesterday.

> This is arguably also done at later stage and could be independent from zbus in the first place.

The :cow: s are visible in public API and there is no reason, this can't use zvariant::Str already. It also makes it inconsistent with the very clear pattern of preferring Str over Cow in the rest of the code.

> This is imho one of the reasons why zbus-names can't be adopted outside zbus easily.

How so? Sure, you drag zvariant dep but we can always/easily split `Str` into its own crate, if there is sufficient demand. For now it's a niche use case and [people who use zbus_names independently of zbus](https://github.com/dbus2/zbus/issues/1006#issuecomment-2392112000), don't mind this dep so much.

elmarco:
> > Can you be more specific on your timeline.
> 
> Not really. It's been in the making for a bit now and people are awaiting the new optimized zbus.

Without timeline, it's hardly possible to prioritize work.  I won't take that pressure without an agreed timeline, with soft/hard freeze periods etc.

> > We shouldn't put ourself under that kind of stress.
> 
> Sure but you went silent for a good few days and given you history of disappearing completely for months and not finishing your PRs for years, I thought you might have disappeared again for another few months when you didn't even respond to simple questions on the matrix channel.

Incorrect, unfair and unrelated.

> > Why can't it be wrapped later on? This is already an almost 0-cost abstraction, it shouldn't change much to wrap it.
> 
> Because changing the return type of a method, is a breaking change.

I am not talking about changing the return type in a future release. I meant at a higher level, lower level types can be wrapped.

> > But other string types are more directly visible on the bus as message data and must be serialized etc.
> 
> I don't see that as a relevant fact. I presented a very real use case that I was implementing just yesterday.

"I wanted to create a typed const for default address in busd".

I share this goal, I said it very early on. I also said I wish to have compile-time check. It was not a priority though, you didn't show much interest in that, and even went against it since parsing code is not a separate anymore.

Without compile time check, you have to provide unsafe ctor API, which should be avoided as it may create future issues.

It's hardly a priority, "const fn" are slowly getting used. 

Can't we simply have?
```
    pub const fn new_static(addr: &'static str) -> Self {
        Address {
            addr: Cow::Borrowed(addr),
        }
    }
```

> 
> > This is arguably also done at later stage and could be independent from zbus in the first place.
> 
> The 🐮 s are visible in public API and there is no reason, this can't use zvariant::Str already. It also makes it inconsistent with the very clear pattern of preferring Str over Cow in the rest of the code.

In general, Cow is not visible in ::address API, but since decode_percents() returns it, it make sense to expose it.

What does Str brings? My understanding is that it provides the zv traits (unnecessary for parsing), and a peculiar handling of &str which leads a bit astray from the std ToOwned design: it doesn't return an "owned" string like all the ToOwned implementors (https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#implementors). But since it doesn't implement ToOwned, I can accept it, I wish it was better documented to avoid the confusion. 

What is the issue with converting Cow to Str as necessary?

zeenix:
> > > Can you be more specific on your timeline.
> > 
> > 
> > Not really. It's been in the making for a bit now and people are awaiting the new optimized zbus.
> 
> Without timeline, it's hardly possible to prioritize work. I won't take that pressure without an agreed timeline, with soft/hard freeze periods etc.

We don't have regular cycles. It will take as long as it takes. I'm not asking you to ensure by any means to deliver quickly. I'm asking you to try. If I see that you're doing the work, I am willing to delay as long as needed. Also, you must answer questions on the matrix channel (especially if they affect my work/planning) and not ignore them.

> > > We shouldn't put ourself under that kind of stress.
> > 
> > 
> > Sure but you went silent for a good few days and given you history of disappearing completely for months and not finishing your PRs for years, I thought you might have disappeared again for another few months when you didn't even respond to simple questions on the matrix channel.
> 
> Incorrect, unfair and unrelated.

If this was incorrect, why didn't you say so when I asserted this last many times? You know it's very much true and that's why you didn't correct me before.

And no, it's very much relevant. If you disappear for months, I wouldn't want to wait for you. That's very fair.

> > Because changing the return type of a method, is a breaking change.
> 
> I am not talking about changing the return type in a future release. I meant at a higher level, lower level types can be wrapped or converted as necessary.

Then I've no idea what you're talking about. The task about guid is simple: the getter should return Guid type. That's it.

> I shared this goal, early on. I also said I wish to have compile-time check. It was not a priority though, you didn't show much interest in that, and even went against it since parsing code is not a separate anymore.

I don't recall what you're referring to but I was just in general not interested in complicating the address API. I'm still not happy about doing so now but at least we can make it better and more consistent with rest of the project code.

> Without compile time check, you have to provide unsafe ctor API, which should be avoided as it may create future issues.

busd is likely the only one that would need this and an unsafe API is fine for the **only** use it will have. Tests can further ensure it is safe.

> Can't we simply have?
> 
> ```
>     pub unasafe const fn new_static(addr: &'static str) -> Self {
>         Address {
>             addr: Cow::Borrowed(addr),
>         }
>     }
> ```

Because Str gives all the benefits of Cow, plus no allocation/copying on to_owned and it's consistent.

> What does Str brings? My understanding is that it provides the zv traits (unnecessary for parsing)

No. Look, I already explained why we use it everywhere and if for no other reason, we use it here for consistency.

> it doesn't return an "owned" string like all the ToOwned implementors (https://doc.rust-lang.org/std/borrow/trait.ToOwned.html#implementors)

We wanted it to implement `ToOwned` but that doesn't work. Try it and you'll find out. Hence why it has a `to_owned` method instead.

> I wish it was better documented to avoid the confusion.

Sure, I an agree with that. Feel free to fix that.

---

Now I've explained the reasons for why `Str` should be used and I also started the work for you. That's as much time I'm willing to spend on this. Now please, modify the code to use Str.

elmarco:
> Now I've explained the reasons for why `Str` should be used and I also started the work for you. That's as much time I'm willing to spend on this. Now please, modify the code to use Str.

You don't provide evidence why Str is necessary and why it can't be converted from Cow or &str when necessary.

I can see your point about consistency, but the downside are also big. It's an overly complex type when simple common type are good enough. It's against the general design to have parser code independent from the rest of zbus. It's unnecessarily complicated when you can just convert to Str as needed.

elmarco:
> We wanted it to implement `ToOwned` but that doesn't work. Try it and you'll find out. Hence why it has a `to_owned` method instead.

It's not ToOwned, because it doesn't return an owned value, it's still a shared reference...

elmarco:
> And no, it's very much relevant. If you disappear for months, I wouldn't want to wait for you. That's very fair.

Please stop arguing, that won't help. We have job, we have life. Can we simply agree on when we do things so we can prioritize?

elmarco:
> If the underlying type is a Cow, I don't automatically get the advantage of `Str` with the type in quesiton (e.g `Address`). If I do `address.to_owned()` and `address` is created from a static string literal, it will allocate/clone.

That's the purpose of ToOwned. It's not to give you a shared reference. You can simply share the reference to Address. You can trivially convert from &str or Cow to Str afaict. I don't see a compelling reason to expose this zb/zv type directly for Address members. 

> 🤦 It is not just a "parser code". You've repeated this like 100s of times now and I've repeated each time that I don't agree with this goal/ideal.

Because you changed your mind on a previously agreed plan. I didn't change mine.

> > It's not ToOwned, because it doesn't return an owned value, it's still a shared reference...
> 
> String is a shared reference? That's news to me.

What are you talking about... Str::to_owned() will return a shared reference to the underlying &'static str., which is not the design of ToOwned. Maybe it should have been called into_static() instead ? I am not sold on the right way to handle that.

I leave it to you to decide what to do. I fundamentally disagree with your design choices here, I don't see why I should do it.

zeenix:
> 🤦 It is not just a "parser code". You've repeated this like 100s of times now and I've repeated each time that I don't agree with this goal/ideal.

> Because you changed your mind on a previously agreed plan. I didn't change mine.

Not at all. I never agreed that this purely parser-only API. It doesn't have to be either-or. I might have changed my mind on other things but not this.

zeenix:
Not looking for a fight but more a calm-head discussion, hence starting a new thread. :)

As I wrote in the PM earlier, I think I may have overlooked something very important in my review of #982, that would have helped me understand why you kept insisting that `Address` is purely a parser. It looks to me that while `Address` does parse the string it's constructed from, it's still not a parsed representation of an address but rather parses the underlying string each time, it is used. E.g each time I call the `transport` method, the string is parsed.

I think for zbus, this wouldn't really matter much but it would suck for busd and any external users of this API as they'd want the address to be pared only once. This realization also explains why you had to switch to String type for address in https://github.com/dbus2/busd/pull/142

If the intention is so that users are expected to cache the return value of `Address::transport`, then that will also not be 0-cost operation since user will have to `to_own` the transport because you can't do self-referential types (easily) in Rust.

I'm curious to know your thoughts here. I hope you can see my concern here and also that you have some solution that could be acceptable to us both. I would also suggest that if we keep the `Address` type as it is, changing the name to `Parser` so it's clear (similarly for other types I guess).

elmarco:
> I think for zbus, this wouldn't really matter much but it would suck for busd and any external users of this API as they'd need to reuse the same address many times but parsing the string form only once. This realization also explains why you had to switch to String type for address in [dbus2/busd#142](https://github.com/dbus2/busd/pull/142)

As I mentionned, it's temporary, I haven't finished the conversion. But I expect busd to be mostly concerned with the string representation of the address(es) once server is listening anyway, and it shouldn't have to rebuild the string when queried.

> If the intention is so that users are expected to cache the return value of `Address::transport`, then that will also not be 0-cost operation since user will have to `to_own` the transport because you can't do self-referential types (easily) in Rust.

True, but into_owned() should be trivial to add if necessary (I am not convinced it is).

zeenix:
> > I think for zbus, this wouldn't really matter much but it would suck for busd and any external users of this API as they'd need to reuse the same address many times but parsing the string form only once. This realization also explains why you had to switch to String type for address in [dbus2/busd#142](https://github.com/dbus2/busd/pull/142)
> 
> As I mentionned, it's temporary, I haven't finished the conversion. But I expect busd to be mostly concerned with the string representation of the address(es) once server is listening anyway, and it shouldn't have to rebuild the string when queried.

I disagree here. Maybe that works for busd but it's not at all a good design to have a type that parses a string and then on subsequent uses, it parses yet again each time. It should be either completely parsing upfront or always dynamic (the former sounds the best). You can disagree with me about other aspects of `Address` API in 4.x, but you've to admit that it had this part done right: it only parsed onced.

I had an idea: Why don't the types cache the indexes to the parts in the string that they're concerned with? That way, they can be reconstructed with zero cost after the first parsing?

> > If the intention is so that users are expected to cache the return value of `Address::transport`, then that will also not be 0-cost operation since user will have to `to_own` the transport because you can't do self-referential types (easily) in Rust.
> 
> True, but into_owned() should be trivial to add if necessary (I am not convinced it is).

? How does the user keep the transport around then?

elmarco:
> I think for zbus, this wouldn't really matter much but it would suck for busd and any external users of this API as they'd need to reuse the same address many times but parsing the string form only once. This realization also explains why you had to switch to String type for address in [dbus2/busd#142](https://github.com/dbus2/busd/pull/142)

As I mentionned, it's temporary, I haven't finished the conversion. But I expect busd to be mostly concerned with the string representation of the address(es) once server is listening anyway, and it shouldn't have to rebuild the string when queried.

> If the intention is so that users are expected to cache the return value of `Address::transport`, then that will also not be 0-cost operation since user will have to `to_own` the transport because you can't do self-referential types (easily) in Rust.

True, but into_owned() should be trivial to add if necessary (I am not convinced it is).

> > As I mentionned, it's temporary, I haven't finished the conversion. But I expect busd to be mostly concerned with the string representation of the address(es) once server is listening anyway, and it shouldn't have to rebuild the string when queried.
> 
> I disagree here. Maybe that works for busd but it's not at all a good design to have a type that parses a string and then on subsequent uses, it parses yet again each time. It should be either completely parsing upfront or always dynamic (the former sounds the best). You can disagree with me about other aspects of `Address` API in 4.x, but you've to admit that it had this part done right: it only parsed once.

Why would it have to parse several times? If you need to keep around the parsed / break down version (what' s the case?), we could have into_owned(). After all, Cow just allows this choice: either borrowed or owned.

> I had an idea: Why don't the types cache the indexes to the parts in the string that they're concerned with? That way, they can be reconstructed with zero cost after the first parsing?

That's a hack, but sometime you don't have the choice. However, it does not keep a reference to the original string (the whole point), which is awkward and error-prone.

> > True, but into_owned() should be trivial to add if necessary (I am not convinced it is).
> 
> ? How does the user keep the transport around then?

What do you mean? like any owned value.

zeenix:
> > I disagree here. Maybe that works for busd but it's not at all a good design to have a type that parses a string and then on subsequent uses, it parses yet again each time. It should be either completely parsing upfront or always dynamic (the former sounds the best). You can disagree with me about other aspects of `Address` API in 4.x, but you've to admit that it had this part done right: it only parsed once.
> 
> Why would it have to parse several times? 

That's what I'm saying. They shouldn't need to. They parse it once and then keep and used the parsed form from then onwards. I hope you don't mean to say that they should only access address details. That would be a needless restriction on the user.

> If you need to keep around the parsed / break down version (what' s the case?), we could have into_owned(). After all, Cow just allows this choice: either borrowed or owned.

That still doesn't solve the issue of having to parse each time you access `Address` components.

> > I had an idea: Why don't the types cache the indexes to the parts in the string that they're concerned with? That way, they can be reconstructed with zero cost after the first parsing?
> 
> That's a hack,

I don't think so. The Address knows and keeps the string and in the first parsing, it figures out where the transport,guide etc is in the string. This string then remains immutable so any information we keep around about it, is guaranteed to stay the valid.

> but sometime you don't have the choice. 

If you have a better idea, I'm all ears. I don't want to be pushy here but this is a problem we need resolved. Otherwise, it's a regression from 4.x API.

> However, it does not keep a reference to the original string (the whole point), which is awkward and error-prone.

I wasn't suggesting that. It's not even possible without awkward unsafe code and we'll need to get `Pin` involved and that could get complicated. 

> > > True, but into_owned() should be trivial to add if necessary (I am not convinced it is).
> > 
> > 
> > ? How does the user keep the transport around then?
> 
> What do you mean? like any owned value.

I meant that your point about into_owned() do not address the main concern here.

zeenix:
> > That's what I'm saying. They shouldn't need to. They parse it once and then keep and use the parsed form from then onwards. I hope you don't mean to say that they should only access address details once. That would be a needless restriction on the user.
> 
> Yes, it's a well thought restriction.

I don't see any thoughts and seems very unnecessary and arbitrary.

> Can you give me a reason why the broken down version needs to be kept around?

I didn't say it needs to be kept around, necessarily. The suggestion I made as a solution, doesn't even involve that. To explain once again: parsing needs to happen once and then user should have a parsed type that they can keep around and use to access attributes of an address w/o any re-parsing (preferably completely 0-cost).

How we achieve that, is up to us. Unless you've any better solutions, I'd suggest we go for the simple one I presented already, which as I explained is infallible due to the underlying string being immutable.

> For zbus itself, it's clearly unneeded.

Yes, I said that already. It's a concern for the zbus users though who create connection for a given address. busd is just one example that's well-known to us and completely in our control.

> For busd, I have claimed that what you need is the actual listenable address string, not the broken down/parsed address.

Yes but that's not something I agree with. I think we should use strong-typing, wherever possible. From the busd pov, going to using a raw string to represent an Address is a regression.

---

It's clear to me that we don't see eye-to-eye on this either. I won't be able to convince you that this is a valid concern and you won't be able to convince me that it isn't. As I wrote before, I don't want to start another endless fight again. So I'd request you focus on addressing my concern (either by going for my proposed solution or proposing your own).

elmarco:
> > > That's what I'm saying. They shouldn't need to. They parse it once and then keep and use the parsed form from then onwards. I hope you don't mean to say that they should only access address details once. That would be a needless restriction on the user.
> > 
> > 
> > Yes, it's a well thought restriction.
> 
> I don't see any thoughts and seems very unnecessary and arbitrary.

Once again, we go in circles. Believe it or not I thought about the various use cases.

> 
> I didn't say it needs to be kept around, necessarily. The suggestion I made as a solution, doesn't even involve that. To explain once again: parsing needs to happen once and then user should have a parsed type that they can keep around and use to access attributes of an address w/o any re-parsing (preferably completely 0-cost).
> 

And my question remain, why do you need to keep the parsed down version around?

> It's clear to me that we don't see eye-to-eye on this either. I won't be able to convince you that this is a valid concern and you won't be able to convince me that it isn't. As I wrote before, I don't want to start another endless fight again. So I'd request you focus on addressing my concern (either by going for my proposed solution or proposing your own).

If I didn't have thought about the problem already, I would probably do as requested and be done. But I did, and I still claim that there are better ways to solve those issues.

elmarco:
> From the busd pov, going to using a raw string to represent an Address is a regression IMO.

I agree with that, that's why Address can only hold a valid address.

zeenix:
> > > > That's what I'm saying. They shouldn't need to. They parse it once and then keep and use the parsed form from then onwards. I hope you don't mean to say that they should only access address details once. That would be a needless restriction on the user.
> > > 
> > > 
> > > Yes, it's a well thought restriction.
> > 
> > 
> > I don't see any thoughts and seems very unnecessary and arbitrary.
> 
> Once again, we go in circles. Believe it or not I thought about the various use cases.

That's great but we still have with this issue.

> > I didn't say it needs to be kept around, necessarily. The suggestion I made as a solution, doesn't even involve that. To explain once again: parsing needs to happen once and then user should have a parsed type that they can keep around and use to access attributes of an address w/o any re-parsing (preferably completely 0-cost).
> 
> And my question remain, why do you need to keep the parsed down version around?

So they can access address attributes directly or indirectly (e.g when constructing a connection from the address). Some users (think systemd codebase) likes to creates lots of connections.

It's also in general a bad design. The string was already parsed. Why would it need to be parsed over and over again after that? It makes no sense to me. I seriously doubt you can change my mind about this.

I really think I've expressed why this is needed. Can we stop arguing about the need for this and focus on solving the problem?

> If I didn't have thought about the problem already, I would probably do as requested and be done. But I did, and I still claim that there are better ways to solve those issues.

I'm not sure what you mean here. If you're going for my suggestion, please let me know and we can stop arguing. If not, please suggest a better way that would address my concern.

zeenix:
Also, you previously said your API patterns are consistent with [similar API in std](https://doc.rust-lang.org/stable/std/net/trait.ToSocketAddrs.html#) and it was actually one argument that had weight. Now you'll find that std API follows the pattern I'm insisting on here: parse once info a specific type and then access its attributes in a zero-cost way.

E.g `SocketAddrV6`. Not only are its getters zero-cost, they're`const` even.

I'm actually extremely surprised that you don't see a problem here. I was sure this is one issue you'll agree with me on.

zeenix:
Been thinking a lot more about this and realized two more things:

> > I had an idea: Why don't the types cache the indexes to the parts in the string that they're concerned with? That way, they can be reconstructed with zero cost after the first parsing?
> 
> That's a hack

I think you're probably correct here. However, this is because how the new API is designed. It's a workaround for a design flow of the new API.

> For zbus itself, it's clearly unneeded.

Actually, this also affects zbus itself: you parse an Address given a string and then to connect to the address, you fetch the transport and guid, and for that you parse yet again. This is suboptimal and there is no good reason for it to be this way.

zeenix:
Nice. Thanks for the rework and addressing the concerns.

Does it already fix the re-parsing issue yet?

elmarco:
> Actually, this also affects zbus itself: you parse an Address given a string and then to connect to the address, you fetch the transport and guid, and for that you parse yet again. This is suboptimal and there is no good reason for it to be this way.

Correct, because I didn't want to break API at the time of writing this series. If builder were to keep a reference of the original unparsed address, we wouldn't have this problem. The connection itself doesn't need to hold it after that.

See the new patches which now go directly from unparsed addresses to OwnedAddress'es.

zeenix:
> > Actually, this also affects zbus itself: you parse an Address given a string and then to connect to the address, you fetch the transport and guid, and for that you parse yet again. This is suboptimal and there is no good reason for it to be this way.
> 
> Correct, because I didn't want to break API at the time of writing this series. If builder were to keep a reference of the original unparsed address, we wouldn't have this problem. 

How so? It still needs to parse the string first and then have Address reparses when the builder accesses the attributes of the address (each time for each attribute).

> See the new patches which now go directly from unparsed addresses to OwnedAddress'es.

The ownership is not a problem here. You're solving the wrong problem. All types (wether owned or not) must only be parsed to, once.

I'm not sure how I can make it any clearer: The issue is reparsing after the Address has already been parsed from a string. That MUST NOT happen. If you've solved this issue or are planning to, please let me know. Otherwise, you're just wasting more of your and my time.

Without this issue resolved, I will not be going with the new API in 5.0 since I consider this a major flaw in the design of the API. I don't want to waste more time in trying to explain this to you.

elmarco:
> How so? It still needs to parse the string first and then have Address reparses when the builder accesses the attributes of the address (each time for each attribute).

There is no reason to reparse if you correctly handle Address<'a> and Transport<'a> lifetimes.

> > See the new patches which now go directly from unparsed addresses to OwnedAddress'es.
> 
> The ownership is not a problem here. You're solving the wrong problem. All types (wether owned or not) must only be parsed to, once.

That's what OwnedAddress::new() does, it parses and store the result as owned values.

zeenix:
> > How so? It still needs to parse the string first and then have Address reparses when the builder accesses the attributes of the address (each time for each attribute).
> 
> There is no reason to reparse if you correctly handle Address<'a> and Transport<'a> lifetimes.

That makes no sense at all. This has nothing to do with lifetimes and owernership. Also, it has nothing to do with how the user uses it. If I do a `let addr = Address::from_str(addr_str)?;' then `addr.transport()` should be zero-cost.

> > > See the new patches which now go directly from unparsed addresses to OwnedAddress'es.
> > 
> > 
> > The ownership is not a problem here. You're solving the wrong problem. All types (wether owned or not) must only be parsed to, once.
> 
> That's what OwnedAddress::new() does, it parses and store the result as owned values.

So you solved the issue for the new owned variants you just added but it does not solve the issue for the reference types. The owned variants should not have any difference to their reference types other than ownership of the internal state.

zeenix:
> I'm not sure how I can make it any clearer: The issue is reparsing after the Address has already been parsed from a string. That MUST NOT happen. If you've solved this issue or are planning to, please let me know. Otherwise, you're just wasting more of your and my time.
> 
> Without this issue resolved, I will not be going with the new API in 5.0 since I consider this a major flaw in the design of the API. I don't want to waste more time in trying to explain this to you.

Since

* I see the main issue here, as a major flaw in the new API
* I have already spent a considerable amount of my time and energy trying to explain things to you
* you refuse to even see the problem, let alone effectively solve it
* now adding new unnecessary API (addres::Guid) that I didn't agree to

I'm now going ahead with the last resort I mentioned above: reverting the new address API for 5.0. We'll have plenty of time to get the API improved for 6.0.

Moreover, since we've already had a lot of fighting over this, I have to ask you to constraint yourself from commenting further on this topic. If my vision for zbus API is not something you can agree with, feel free to fork the project. That is your right and I will not protest. If you comment further on the topic, I might have to consider taking further drastic measures, such as blocking you (temporarily). I'd really hate to do that so please don't force me to do so.

PR #1061, opened 2024-10-11
🩹 zb: Fix missing cfg(blocking-api)

elmarco:
"cargo build" in zbus/zbus (with default features) fails, there are missing cfg conditions.


zeenix (approved):
Thanks.

PR #1069, opened 2024-10-14
⬆️  all: Bump MSRV

zeenix:
We don't really need to currently but it's a good idea to do this in a major version bump.

### DIFF
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -3,7 +3,7 @@ name = "zbus"
 version = "4.4.0"
 authors = ["Zeeshan Ali Khan <zeeshanak@gnome.org>"]
 edition = "2021"
-rust-version = "1.75"
+rust-version = "1.81"
 
 description = "API for D-Bus communication"
 repository = "https://github.com/dbus2/zbus/"
--- a/zbus_macros/Cargo.toml
+++ b/zbus_macros/Cargo.toml
@@ -7,7 +7,7 @@ authors = [
     "Zeeshan Ali Khan <zeeshanak@gnome.org>",
 ]
 edition = "2021"
-rust-version = "1.75"
+rust-version = "1.81"
 
 description = "proc-macros for zbus"
 repository = "https://github.com/dbus2/zbus/"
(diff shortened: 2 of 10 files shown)


ids1024:
I'm not sure bumping `rust-version` for no specific reason as part of a major version bump is really a best practice. At least not when bumping to a fairly recent version.

Although there doesn't really seem to be any universal agreement, I think the convention for projects that care about Rust MSRV would be to:
- Not require too recent of a Rust version
- Bump the `rust-version` only when needed; and not necessarily as part of a major version, since it's not considered a breaking change (crates that are stable and don't bump the major version can still require a newer Rust MSRV)

I noticed this trying to update to zbus 5.0. In COSMIC, we're currently targeting 1.80. And I think there were some complaints from packagers for other distros when we required that (shortly after 1.80 was released) so we are planning to not increase that for now.

zeenix:
Your problem would be the same even if we actually needed the MSRV bump. 😔

IMO requiring the latest stable Rust in a major release should be acceptable to everyone. It's a major version, after all. It ensures that we're unlikely to require any bumps for a long time. 

I know many crates just bump their MSRV in minor releases (and even patch releases) but they also get yelled at for doing so. Typically there's no real need to bump the MSRV since you don't necessarily have to use the new features, so you can get stuck on old Rust for a very long time.

Having said that, which version will be ok for you? Maybe we can find a compromise.

ids1024:
I don't personally think MRSV bumps should have anything to do with major versions. (I mean, I'd really prefer if everyone could agree to just use the latest Rust version and not care about MSRV. But that's not currently how things are.)

But anyway. For COSMIC we're currently targeting Rust 1.80, so dropping the requirement to that would be solve our immediate issue here. I don't think 1.81 added anything too major that Zbus would likely require.

I don't know if there's a good way to see what version of Rust is packaged across distros, but 1.80 seems to be the newest version packaged by Ubuntu LTS currently, for instance.

zeenix:
> But anyway. For COSMIC we're currently targeting Rust 1.80, so dropping the requirement to that would be solve our immediate issue here. I don't think 1.81 added anything too major that Zbus would likely require.

Sure. I'll try to roll out releases with MSRV downgraded soon.

zeenix:
@ids1024 created #1107 to track this. If you could provide a PR for downgrade, that would be very helpful. Then it'd just be about rolling out releases (although there are a few bugs/regressions I want to fix first).

PR #1100, opened 2024-10-27
zm: Don't generate signals for interface const properties

bilelmoussaoui:
Because they never change.

Currently the following
```rust
pub struct AccessInterface {}

#[zbus::interface(name = "org.freedesktop.impl.portal.Access")]
impl AccessInterface {
    #[zbus(property(emits_changed_signal = "const"), name = "version")]
    fn version(&self) -> u32 {
        1
    }
}
```

Would generate a `version_changed` and a `version_invalidate` even if the property never changes. 

This is current situation https://bilelmoussaoui.github.io/ashpd/ashpd/backend/account/struct.AccountInterface.html and with the changes proposed in this PR, the problem is fixed.

### DIFF
--- a/zbus/tests/issue/issue_310.rs
+++ b/zbus/tests/issue/issue_310.rs
@@ -59,10 +59,6 @@ async fn issue_310() {
             {
                 let iface = iface_ref.get().await;
                 iface
-                    .connected_network_invalidate(iface_ref.signal_emitter())
-                    .await
-                    .unwrap();
-                iface
                     .connected_network_changed(iface_ref.signal_emitter())
                     .await
                     .unwrap();
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -689,40 +689,44 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                         quote!(self.#ident()#method_await)
                     };
 
-                    let prop_changed_method = quote!(
-                        pub async fn #prop_changed_method_name(
-                            &self,
-                            signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
-                        ) -> #zbus::Result<()> {
-                            let mut changed = ::std::collections::HashMap::new();
-                            let value = <#zbus::zvariant::Value as ::std::convert::From<_>>::from(#prop_value_handled);
-                            changed.insert(#member_name, value);
-                            #zbus::fdo::Properties::properties_changed(
-                                signal_emitter,
-                                #zbus::names::InterfaceName::from_static_str_unchecked(#iface_name),
-                                changed,
-                                ::std::borrow::Cow::Borrowed(&[]),
-                            ).await
-                        }
-                    );
+                    if p.emits_changed_signal == PropertyEmitsChangedSignal::True {
+                        let prop_changed_method = quote!(
+                            pub async fn #prop_changed_method_name(
+                                &self,
+                                signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
+                            ) -> #zbus::Result<()> {
+                                let mut changed = ::std::collections::HashMap::new();
+                                let value = <#zbus::zvariant::Value as ::std::convert::From<_>>::from(#prop_value_handled);
+                                changed.insert(#member_name, value);
+                                #zbus::fdo::Properties::properties_changed(
+                                    signal_emitter,
+                                    #zbus::names::InterfaceName::from_static_str_unchecked(#iface_name),
+                                    changed,
+                                    ::std::borrow::Cow::Borrowed(&[]),
+                                ).await
+                            }
+                        );
 
-                    generated_signals.extend(prop_changed_method);
+                        generated_signals.extend(prop_changed_method);
+                    }
 
-                    let prop_invalidate_method = quote!(
-                        pub async fn #prop_invalidate_method_name(
-                            &self,
-                            signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
-                        ) -> #zbus::Result<()> {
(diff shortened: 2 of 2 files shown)


zbus_macros/src/iface.rs
```diff
@@ -689,40 +689,42 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
...
-                            changed.insert(#member_name, value);
-                            #zbus::fdo::Properties::properties_changed(
-                                signal_emitter,
-                                #zbus::names::InterfaceName::from_static_str_unchecked(#iface_name),
-                                changed,
-                                ::std::borrow::Cow::Borrowed(&[]),
-                            ).await
-                        }
-                    );
+                    if p.emits_changed_signal != PropertyEmitsChangedSignal::Const {
```
zeenix, line 692: Also on `PropertyEmitsChangedSignal::False`, no? :thinking:

zbus_macros/src/iface.rs
```diff
@@ -689,40 +689,42 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
...
+                        let prop_changed_method = quote!(
```
zeenix, line 693: I think this should only be generated for `PropertyEmitsChangedSignal::True`.

zbus_macros/src/iface.rs
```diff
@@ -689,40 +689,42 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
...
-                                signal_emitter,
-                                #zbus::names::InterfaceName::from_static_str_unchecked(#iface_name),
-                                ::std::collections::HashMap::new(),
-                                ::std::borrow::Cow::Borrowed(&[#member_name]),
-                            ).await
-                        }
-                    );
+                        generated_signals.extend(prop_changed_method);
+
+                        let prop_invalidate_method = quote!(
```
zeenix, line 712: Probably makes sense to only generate this when `p.emits_changed_signal != PropertyEmitsChangedSignal::Invalidates`.

zeenix (changes requested):
Nice catch! I had assumed we were already doing this for some reason. Some changes needed still. While you're addressing the comments, kindly also put an a emoji on the commit. If you don't want to bother with downloading `gimoji`, you can get it from [here](https://gitmoji.dev/) easily. 🐛 is probably the most appropriate here so you can also just c&p this. :)

zbus_macros/src/iface.rs
bilelmoussaoui: I assumed false, means the signal doesn't gets emitted automatically but you still want the helpers to be generated, no?

zbus_macros/src/iface.rs
bilelmoussaoui: Same question as above

zbus_macros/src/iface.rs
zeenix: The spec says:

> If set to false, the org.freedesktop.DBus.Properties.PropertiesChanged signal, .. is not guaranteed to be emitted if the property changes.

which makes it vague. I can't imagine a use case for someone explicitly setting this attribute to `false` and then wanting to emit signals so I'd just suggest we don't generate when set to `false`.

bilelmoussaoui: > If set to false, the org.freedesktop.DBus.Properties.PropertiesChanged signal, .. is not guaranteed to be emitted if the property changes.

The spec seems to correspond to my understanding that I explained above. 
- `const` => never changes, so no need for any signals.
- `true` => changes -> generate changed signal
- `invalidates` => changes -> generate invalidate signal
- `false` => doesn't change, but the value might be computed at a later time, so generate the signal so the user can emit it still.

But then, I don't really see the difference between `true` & `false`.

zeenix: > false => doesn't change, but the value might be computed at a later time, so generate the signal so the user can emit it still.

I don't agree with this interpretation since IMO the "doesn't change" part contradicts the rest. Also as I wrote above, I don't see the use case.

Having said that, if I don't really have a very strong opinion here. Maybe there is a use case that I don't see. :shrug: So if you prefer, we can skip the generation for both `const` and `invalidates` but not for `false`.

bilelmoussaoui: I personally only care about const here, so I will follow whatever you think is correct.

bilelmoussaoui: The problem i see with the commit I just pushed, is there is no way to have both `invalidate` & `changed` signals generated. But then I might be missing something in regards of the spec.

zeenix: > is there is no way to have both `invalidate` & `changed` signals generated

That's actually on purpose. If you declare `invalidates` for your property, the client will not expect to receive values of the property in the signal, so there is really no need for then doing what you declared you won't be doing. :)

zbus/tests/issue/issue_310.rs
```diff
@@ -19,7 +19,7 @@ async fn issue_310() {
 
     #[zbus::interface(name = "net.connman.iwd.Station")]
     impl Station {
-        #[zbus(property)]
+        #[zbus(property(emits_changed_signal = "true"))]
```
zeenix, line 22: This is the default, so not sure why this is needed.

bilelmoussaoui: Indeed. Removed

zeenix (commented):
The second commit looks good.

zeenix (approved):
Just add ✅ to your first commit and we can call it a day. ;)

bilelmoussaoui:
done

PR #1130, opened 2024-11-07
👷 CONTRIBUTING: Suggest to copy the git hooks

zeenix:
Instead of configuring a non-standard location as the hooks dir. Otherwise, gimoji (and any other git hooks user may want to use) will not work.


zeenix:
cc @jonathanhood.

PR #1148, opened 2024-11-17
🐛 zv: Allow f32::NAN to be decoded too

### LINKED ISSUE #1145: Zvariant asserts on f32::NAN values
Seems that f32::NAN get serialized to dbus as f64::NAN, but when read back again, it passes through an assert! statement that the value is in acceptable range for f32 and thus panics the application.

A suitable fix might be to either return an Result error rather than panic in a function that returns an Result, or simply convert f64::NAN to f32::NAN and hope nobody notices that it's not the same NAN value?

Testcase:

```
use zvariant::{self, serialized::Context, serialized::Data, Type};
use serde::{Deserialize, Serialize};
use std::time::SystemTime;

#[derive(Debug, Serialize, Type, Deserialize)]
struct First {
    time: SystemTime,
    value: f64,
}

impl Default for First {
    fn default() -> Self {
        Self {
            time: SystemTime::now(),
            value: f64::NAN,
        }
    }
}
#[derive(Debug, Serialize, Type, Deserialize)]
struct Second {
    time: SystemTime,
    value: f32,
}
impl Default for Second {
    fn default() -> Self {
        Self {
            time: SystemTime::now(),
            value: f32::NAN,
        }
    }
}


fn encoder<T>(source: &T) -> Vec<u8>
where
    T: zvariant::Type + serde::Serialize,
{
    let ctxt = Context::new_dbus(zvariant::LE, 0);
    let bytes = zvariant::to_bytes(ctxt, &source).expect("Failed to serialize");
    bytes.bytes().into()
}


fn decoder<T>(bytes: &[u8]) -> T
where
    T: zvariant::Type + for<'a> serde::Deserialize<'a>,
{
    let ctxt = Context::new_dbus(zvariant::LE, 0);
    let data = Data::new(bytes, ctxt);
    let decoded: T = data.deserialize().expect("Failed deserialization").0;
    decoded
}

fn first_case() {
    let source = First {
        time: SystemTime::now(),
        value: f64::NAN,
    };
    let bytes = encoder(&source);
    let back: First = decoder(&bytes);
    assert_eq!(source.time, back.time);
    assert!(back.value.is_nan());
}
fn second_case() {
    let source = Second {
        time: SystemTime::now(),
        value: f32::NAN,
    };
    let bytes = encoder(&source);
    let back: Second = decoder(&bytes);
    assert_eq!(source.time, back.time);
    assert!(back.value.is_nan());
}


fn main() {
    println!("F64 values");
    first_case();
    println!("F32 values");
    second_case();
}

```

Spindel:
The range check would fail on NAN, INFINITY and NEG_INFINITY even if they are completely valid things to decode.

This does not promise that an encoded NAN will be bit-wise identical as the decoded NAN, but if signalling NAN is used that precisely the user can have a handful of more interesting problems.

Fixes #1145.

### DIFF
--- a/zvariant/src/dbus/de.rs
+++ b/zvariant/src/dbus/de.rs
@@ -188,7 +188,13 @@ impl<'de, 'd, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> de::Deseriali
             .endian()
             .read_f64(self.0.next_const_size_slice::<f64>()?);
 
-        visitor.visit_f32(f64_to_f32(v))
+        if v.is_finite() && v > (f32::MAX as f64) {
+            return Err(de::Error::invalid_value(
+                de::Unexpected::Float(v),
+                &"Too large for f32",
+            ));
+        }
+        visitor.visit_f32(v as f32)
     }
 
     fn deserialize_str<V>(self, visitor: V) -> Result<V::Value>
--- a/zvariant/src/utils.rs
+++ b/zvariant/src/utils.rs
@@ -66,12 +66,6 @@ pub(crate) fn usize_to_u8(value: usize) -> u8 {
     value as u8
 }
 
-pub(crate) fn f64_to_f32(value: f64) -> f32 {
-    assert!(value <= (f32::MAX as f64), "{} too large for `f32`", value,);
-
-    value as f32
-}
-
 /// Slice the given slice of bytes safely and return an error if the slice is too small.
 pub(crate) fn subslice<I, T>(input: &[T], index: I) -> Result<&I::Output>
 where
--- a/zvariant/src/lib.rs
+++ b/zvariant/src/lib.rs
@@ -2017,6 +2017,28 @@ mod tests {
         let _: ZVStruct<'_> = encoded.deserialize_for_signature(signature).unwrap().0;
     }
 
+    #[test]
+    fn issue_1145() {
+        // Ensure f32::NAN can be encoded and decoded.
+        let ctxt = Context::new_dbus(LE, 0);
+        {
+            let encoded = to_bytes(ctxt, &f32::NAN).unwrap();
+            let result: f32 = encoded.deserialize().unwrap().0;
+            assert!(result.is_nan());
+        }
+        // Ensure f32::INFINITY can be encoded and decoded.
+        {
+            let encoded = to_bytes(ctxt, &f32::INFINITY).unwrap();
+            let result: f32 = encoded.deserialize().unwrap().0;
+            assert!(result.is_infinite());
+        }
+        {
+            let encoded = to_bytes(ctxt, &f32::NEG_INFINITY).unwrap();
+            let result: f32 = encoded.deserialize().unwrap().0;
+            assert!(result.is_infinite());
+        }
+    }
+
     #[cfg(feature = "ostree-tests")]
     #[test]
     fn ostree_de() {


zeenix (commented):
Thanks! Is this WIP? I don't see any fixes in the PR.

Spindel:
> Thanks! Is this WIP? I don't see any fixes in the PR.

Should be a Fixes,  Did I mess up and assign it only as an Issue: ref rather than Fixes ref? Meh. I'll rewrite the commit msg and force push if you don't have any other review comments.

zeenix:
> Should be a Fixes, Did I mess up and assign it only as an Issue: ref rather than Fixes ref? Meh. I'll rewrite the commit msg and force push if you don't have any other review comments.

I'm not referring to the wording. I don't see any code changes accept addition of an assertion and a test case. Please check your changes. Perhaps you forgot to commit something?

zeenix:
> That assert change is really all there is to it, as it treated NAN values as "too big" to transform, causing an panic when de-serializing an NAN value

Why not error out instead? I don't think assert is a solution.

Spindel:
> > That assert change is really all there is to it, as it treated NAN values as "too big" to transform, causing an panic when de-serializing an NAN value
> 
> Why not error out instead? I don't think assert is a solution.

Cause the the code already used assert's and I didn't feel that a drive-by bug fix should change the behaviour that much.

zeenix:
> Cause the the code already used assert's and I didn't feel that a drive-by bug fix should change the behaviour that much.

It's not a behaviour change IMO. asserts are for invariants that must always hold. Clearly, that is not the case here.

zvariant/src/utils.rs
```diff
@@ -66,12 +66,6 @@ pub(crate) fn usize_to_u8(value: usize) -> u8 {
     value as u8
 }
 
-pub(crate) fn f64_to_f32(value: f64) -> f32 {
-    assert!(!value.is_finite() || value <= (f32::MAX as f64), "{} too large for `f32`", value,);
```
zeenix: you're undoing the previous commit. Please rebase to remove it.

zvariant/src/dbus/de.rs
```diff
@@ -188,7 +188,13 @@ impl<'de, 'd, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> de::Deseriali
             .endian()
             .read_f64(self.0.next_const_size_slice::<f64>()?);
 
-        visitor.visit_f32(f64_to_f32(v))
+        if v.is_finite() && v > (f32::MAX as f64) {
```
zeenix, line 191: should this be `!v.is_finit()`? You need to update the test code btw.

zeenix (changes requested):
Now we're talking. :)

zvariant/src/dbus/de.rs
Spindel: I'd really hope not. We want _finite_  (not inf or NAN) to compare against f32::MAX  in order to cause an error. 

If we wanted to be really serious, we'd check small f32 values against precision bits in the f64 number, but that would be slow for very little gain, if any.  

ie. what happens if the f64 float number has digits in the precision range that cannot be encoded as f32 and would cause "bad" rounding. Honestly, not worth making "proper" as the error would be one of those like "and then what do we do with it?" ( It probably means that whomever encoded it may not have used an f32...)

Spindel:
Emojis added, commits squashed.  If there are more changes, you're welcome to do them.

zvariant/src/dbus/de.rs
zeenix: Ah right. I just misread the `&&` as `||` for some reason. Thanks for explaining.

zeenix (approved):

Issue #898, opened 2024-07-11
`interface` should allow adding an unexposed setter


zeenix:
Typically you want properties to be read-only on the bus but settable from the service process itself. When you do set them manually, you need to write a bit of a complicated code:

```rust
struct MyIface(u32);

#[interface(name = "org.myiface.MyIface")]
impl MyIface {
     #[zbus(property)]
     async fn count(&self) -> u32 {
         self.0
     }
}

let iface_ref = connection
    .object_server()
    .interface::<_, MyIface>(path).await?;
let mut iface = iface_ref.get_mut().await;
iface.0 = 42;
iface.count_changed(iface_ref.signal_context()).await?;
```

It would be great if allowed adding a non-exposed setter in interface that sets the value and each time it's called, it handles the signal emission for you. So the property setting code above becomes:

```rust
let iface_ref = connection
    .object_server()
    .interface::<_, MyIface>(path).await?
    .set_count(42)?;
```

gnulux:
Hi , 
Did you have a way to do that? 
I also want to update a property from my service but can update it. 
What is "path" in your code  ?

gnulux:
Hi, did you succed to male this code works. 

my version always return InterfaceNotFound . Any suggestion ?

```
use zbus::{Connection, ObjectServer, InterfaceRef, interface};
use zbus::export::futures_util::StreamExt;
use zbus::zvariant::ObjectPath;
use zbus::names::{BusName, UniqueName};
use zbus::dbus_interface;

#[derive(Debug)]
struct MyIface(u32);

#[interface(name = "org.myiface.MyIface")]
impl MyIface {
    #[zbus(property)]
    async fn count(&self) -> u32 {
        self.0
    }

    async fn set_count(&mut self, value: u32) -> zbus::fdo::Result<()> {
        self.0 = value;
        Ok(())
    }
}

#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error>> {
    // Connect to the session bus
    let connection = Connection::session().await?;
    
    // Define the object path
    let path = ObjectPath::try_from("/org/myiface/MyIface")?;
    
    // Register the interface
    let iface_ref: InterfaceRef<MyIface> = connection
        .object_server()
        .interface(path).await?;
    
    // Set the initial value of the property
    {
        let mut iface = iface_ref.get_mut().await;
        iface.0 = 42;
    }
    
    // Emit a signal that the property has changed
[...]

PR #1176, opened 2024-12-16
✨ zb,zm: Support Option<Header<'_>> parameter in property getters

rzvncj:
This allows for fine-grained connecting-UID checking, in order to be able to control who is able to access property values.

### DIFF
--- a/zbus/src/object_server/interface/mod.rs
+++ b/zbus/src/object_server/interface/mod.rs
@@ -17,8 +17,11 @@ use zbus_names::{InterfaceName, MemberName};
 use zvariant::{OwnedValue, Value};
 
 use crate::{
-    async_lock::RwLock, fdo, message::Message, object_server::SignalEmitter, Connection,
-    ObjectServer,
+    async_lock::RwLock,
+    fdo,
+    message::{self, Header, Message},
+    object_server::SignalEmitter,
+    Connection, ObjectServer,
 };
 
 /// This trait is used to dispatch messages to an interface instance.
@@ -45,10 +48,26 @@ pub trait Interface: Any + Send + Sync {
     }
 
     /// Get a property value. Returns `None` if the property doesn't exist.
-    async fn get(&self, property_name: &str) -> Option<fdo::Result<OwnedValue>>;
+    ///
+    /// Note: The header parameter will be None when the getter is not being called as part
+    /// of D-Bus communication (for example, when it is called as part of initial object setup,
+    /// before it is registered on the bus, or when we manually send out property changed
+    /// notifications).
+    async fn get(
+        &self,
+        property_name: &str,
+        server: &ObjectServer,
+        connection: &Connection,
+        header: &Option<message::Header<'_>>,
+    ) -> Option<fdo::Result<OwnedValue>>;
 
     /// Return all the properties.
-    async fn get_all(&self) -> fdo::Result<HashMap<String, OwnedValue>>;
+    async fn get_all(
+        &self,
+        object_server: &ObjectServer,
+        connection: &Connection,
+        header: &Option<message::Header<'_>>,
+    ) -> fdo::Result<HashMap<String, OwnedValue>>;
 
     /// Set a property value.
     ///
@@ -60,8 +79,9 @@ pub trait Interface: Any + Send + Sync {
         property_name: &'call str,
         value: &'call Value<'_>,
         emitter: &'call SignalEmitter<'_>,
+        header: &'call Option<message::Header<'_>>,
     ) -> DispatchResult<'call> {
-        let _ = (property_name, value, emitter);
+        let _ = (property_name, value, emitter, header);
         DispatchResult::RequiresMut
     }
 
@@ -75,6 +95,7 @@ pub trait Interface: Any + Send + Sync {
         property_name: &str,
         value: &Value<'_>,
         emitter: &SignalEmitter<'_>,
+        header: &Option<Header<'_>>,
     ) -> Option<fdo::Result<()>>;
 
     /// Call a method.
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -264,6 +264,21 @@ impl MyIface {
 
     #[instrument]
     #[zbus(property)]
+    fn test_header_prop(
+        &self,
+        #[zbus(header)] header: Option<Header<'_>>,
+        #[zbus(connection)] connection: &Connection,
+        #[zbus(object_server)] object_server: &ObjectServer,
+    ) -> bool {
+        debug!(
+            "`TestHeaderProp` getter called, header: {:?}, connection: {:?}, object_server: {:?}",
+            header, connection, object_server
+        );
+        header.is_some()
+    }
+
+    #[instrument]
+    #[zbus(property)]
     async fn hash_map(&self) -> HashMap<String, String> {
         debug!("`HashMap` getter called.");
         self.test_hashmap_return().await.unwrap()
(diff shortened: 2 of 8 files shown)


zbus_macros/src/iface.rs
```diff
@@ -103,8 +103,8 @@ enum MethodType {
 
 #[derive(Debug, PartialEq, Copy, Clone)]
 enum PropertyType {
-    Inputs,
-    NoInputs,
+    Setter,
+    Getter,
```
zeenix, line 107: I've been meaning to do this for a while now. Thanks for doing it. :)

zbus_macros/src/iface.rs
```diff
@@ -1089,6 +1115,36 @@ fn introspect_method(name: &str, args: &TokenStream) -> TokenStream {
     )
 }
 
+fn is_special_arg(attrs: &[Attribute]) -> bool {
```
zeenix, line 1118: nit: I've been trying to follow top-down code layout approach, which means use before definition. If you could just move it down after its first use, that would be great.

zbus/tests/e2e.rs
```diff
@@ -257,7 +257,7 @@ impl MyIface {
 
     #[instrument]
     #[zbus(property)]
-    fn count(&self) -> u32 {
+    fn count(&self, #[zbus(header)] hea: Option<Header<'_>>) -> u32 {
```
zeenix, line 260: Kudos for adding a test case.

```suggestion
    fn count(&self, #[zbus(header)] header: Option<Header<'_>>) -> u32 {
```

Also let's do this in another (perhaps new) property and actually use the header. (you don't get a warning from the compiler about unused param? :thinking:)

zbus/src/object_server/interface/mod.rs
```diff
@@ -45,10 +48,19 @@ pub trait Interface: Any + Send + Sync {
     }
 
     /// Get a property value. Returns `None` if the property doesn't exist.
```
zeenix, line 50: Perhaps it'd be good to document here briefly when the `header` is `Some` and when it's `None`.

zbus_macros/src/iface.rs
```diff
@@ -852,6 +865,9 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
             async fn get(
                 &self,
                 property_name: &str,
+                s: &#zbus::ObjectServer,
+                c: &#zbus::Connection,
+                h: &Option<#zbus::message::Header<'_>>,
```
zeenix, line 870: I think it was a bad idea to choose such short non-descriptive names in the beginning but now that we're getting more args, it's even worse now. You're just following the existing code, so not at all your fault but if could you use the opportunity to change the code to use more meaningful names, that would be awesome. Bonus points for first updating the existing use in a separate commit that comes before this one. :wink: 

```suggestion
                object_server: &#zbus::ObjectServer,
                c: &#zbus::Connection,
                h: &Option<#zbus::message::Header<'_>>,
```

zeenix (commented):
Really impressive work. I know it's quite difficult to get into the very complex and strange macro code and most contributors don't touch it.

Apart from the inline comments (mostly minor nitpicks), please add the new parameter to the `interface` docs and also to the book's `service` chapter.

zeenix (changes requested):
oops, used the wrong option. :)

zbus/tests/e2e.rs
rzvncj: No unused parameter warning, no. Maybe Rust is more permissive with `Option`s? Not sure.

rzvncj:
Thanks for the prompt and very helpful review, and the kind words!

I've hopefully addressed your comments, other than updating the book and interface docs, and doing the "prequel" commit renaming the short parameter names - I'll be updating the PR with those changes as soon as possible too.

rzvncj:
All done, I think, except for updating the book, because I'm not sure where this new information fits. Would you like me to update the `MyGreeter` example? Or a specific section? Or both?

zeenix:
> All done, 

:+1: 

> I think, except for updating the book, because I'm not sure where this new information fits. Would you like me to update the `MyGreeter` example? Or a specific section? Or both?

I just had a look a the book again and seems we don't introduce all the special args there so never mind.

zbus_macros/src/iface.rs
```diff
@@ -243,11 +243,11 @@ impl MethodInfo {
             let ret = quote!(r);
 
             quote!(match reply {
-                ::std::result::Result::Ok(r) => c.reply(&hdr, &#ret).await,
-                ::std::result::Result::Err(e) => c.reply_dbus_error(&hdr, e).await,
+                ::std::result::Result::Ok(r) => connection.reply(&hdr, &#ret).await,
+                ::std::result::Result::Err(e) => connection.reply_dbus_error(&hdr, e).await,
```
zeenix, line 260: This commit does not touch `zbus` subcrate so `zb` prefix isn't valid here. :)

zbus/src/object_server/interface/mod.rs
```diff
@@ -60,8 +77,9 @@ pub trait Interface: Any + Send + Sync {
         property_name: &'call str,
         value: &'call Value<'_>,
         emitter: &'call SignalEmitter<'_>,
+        header: &'call message::Header<'_>,
```
zeenix, line 80: While this currently can always get a header passed to it, this will hopefully be changing at some point: https://github.com/dbus2/zbus/issues/898 . So best to make this optional as well. It will also be more consistent that way.

zbus_macros/src/lib.rs
```diff
@@ -267,7 +267,7 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///   important. You can use `out_args` to specify their names.
 ///
 /// * `proxy` - Use this to specify the [`macro@proxy`]-specific method sub-attributes (e.g
-///   `object`). The common sub-attributes (e.g `name`) are automatically forworded to the
+///   `object`). The common sub-attributes (e.g `name`) are automatically forwarded to the
```
zeenix, line 270: I appreciate your fixing the typo here but it's not relevant to this commit so if you could split it into a separate commit, that would be great (`git show --oneline && git reset HEAD^ && git add -p && git commit && git commit -a -C <commit-hash-from-git-show-command` should do it). :)

zbus_macros/src/lib.rs
```diff
@@ -289,6 +289,13 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// using this since it will force all interested peers to fetch the new value and hence result in
 /// excess traffic on the bus.
 ///
+/// The getter property arguments support the following `zbus` attributes:
+///
+/// * `header` - This marks the property argument to receive an optional message header associated
+///   with the D-Bus property get operation being handled. It will be `None` when the getter is
+///   being called as part of the initial object setup (before it gets registered on the bus), or
+///   when we send out property changed notifications.
```
zeenix, line 297: Good job but I think this doesn't need a special section. We should just add to the `header` arg docs below that in case of property methods, the header is optional and when its the case.

zeenix (changes requested):
LGTM otherwise. Getting very close now. :)

rzvncj:
> LGTM otherwise. Getting very close now. :)

Great, thanks! Submitted a new round of patches for review.

zbus_macros/src/lib.rs
```diff
@@ -289,16 +289,19 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
-/// The method arguments support the following `zbus` attributes:
+/// The method and property getter arguments support the following `zbus` attributes:
 ///
 /// * `object_server` - This marks the method argument to receive a reference to the
 ///   [`ObjectServer`] this method was called by.
 /// * `connection` - This marks the method argument to receive a reference to the [`Connection`] on
 ///   which the method call was received.
 /// * `header` - This marks the method argument to receive the message header associated with the
-///   D-Bus method call being handled.
+///   D-Bus method call being handled. For reading properties, this will be an `Option<Header<'_>>`,
```
zeenix, line 299: > For reading properties

"For property getter methods"

zeenix (approved):

PR #1179, opened 2024-12-19
✨ zb,zm: Support special args in interface property setters

rzvncj:
Along with all the other special parameter types that methods currently support (connection, object_server and signal_emitter).

### DIFF
--- a/zbus/src/object_server/interface/mod.rs
+++ b/zbus/src/object_server/interface/mod.rs
@@ -58,7 +58,7 @@ pub trait Interface: Any + Send + Sync {
         property_name: &str,
         server: &ObjectServer,
         connection: &Connection,
-        header: &Option<message::Header<'_>>,
+        header: Option<&message::Header<'_>>,
     ) -> Option<fdo::Result<OwnedValue>>;
 
     /// Return all the properties.
@@ -66,7 +66,7 @@ pub trait Interface: Any + Send + Sync {
         &self,
         object_server: &ObjectServer,
         connection: &Connection,
-        header: &Option<message::Header<'_>>,
+        header: Option<&message::Header<'_>>,
     ) -> fdo::Result<HashMap<String, OwnedValue>>;
 
     /// Set a property value.
@@ -79,7 +79,7 @@ pub trait Interface: Any + Send + Sync {
         property_name: &'call str,
         value: &'call Value<'_>,
         emitter: &'call SignalEmitter<'_>,
-        header: &'call Option<message::Header<'_>>,
+        header: Option<&'call message::Header<'_>>,
     ) -> DispatchResult<'call> {
         let _ = (property_name, value, emitter, header);
         DispatchResult::RequiresMut
@@ -95,7 +95,7 @@ pub trait Interface: Any + Send + Sync {
         property_name: &str,
         value: &Value<'_>,
         emitter: &SignalEmitter<'_>,
-        header: &Option<Header<'_>>,
+        header: Option<&Header<'_>>,
     ) -> Option<fdo::Result<()>>;
 
     /// Call a method.
--- a/zbus/src/object_server/interface/mod.rs
+++ b/zbus/src/object_server/interface/mod.rs
@@ -78,10 +78,19 @@ pub trait Interface: Any + Send + Sync {
         &'call self,
         property_name: &'call str,
         value: &'call Value<'_>,
+        object_server: &'call ObjectServer,
+        connection: &'call Connection,
         emitter: &'call SignalEmitter<'_>,
         header: Option<&'call message::Header<'_>>,
     ) -> DispatchResult<'call> {
-        let _ = (property_name, value, emitter, header);
+        let _ = (
+            property_name,
+            value,
+            object_server,
+            connection,
+            emitter,
+            header,
+        );
         DispatchResult::RequiresMut
     }
 
@@ -94,6 +103,8 @@ pub trait Interface: Any + Send + Sync {
         &mut self,
         property_name: &str,
         value: &Value<'_>,
+        object_server: &ObjectServer,
+        connection: &Connection,
         emitter: &SignalEmitter<'_>,
         header: Option<&Header<'_>>,
     ) -> Option<fdo::Result<()>>;
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -279,6 +279,22 @@ impl MyIface {
 
     #[instrument]
     #[zbus(property)]
+    fn set_test_header_prop(
+        &self,
+        value: bool,
+        #[zbus(header)] header: Option<Header<'_>>,
+        #[zbus(connection)] connection: &Connection,
+        #[zbus(object_server)] object_server: &ObjectServer,
+        #[zbus(signal_emitter)] emitter: SignalEmitter<'_>,
+    ) {
+        debug!("`TestHeaderProp` setter called, value: {}, header: {:?}, connection: {:?}, object_server: {:?}, emitter: {:?}",
+            value, header, connection, object_server, emitter
+        );
+        assert!(header.is_some());
+    }
+
+    #[instrument]
+    #[zbus(property)]
     async fn hash_map(&self) -> HashMap<String, String> {
         debug!("`HashMap` getter called.");
         self.test_hashmap_return().await.unwrap()
@@ -585,6 +601,7 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     drop(props_changed_stream);
 
     proxy.ping().await?;
+    proxy.set_test_header_prop(true).await?;
     assert_eq!(proxy.test_header_prop().await?, true);
     assert_eq!(proxy.count().await?, 1);
     assert_eq!(proxy.cached_count()?, None);
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -719,7 +719,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                                 &self,
                                 signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
                             ) -> #zbus::Result<()> {
-                                let header = None as Option<#zbus::message::Header<'_>>;
+                                let header = ::std::option::Option::None::<&#zbus::message::Header<'_>>;
                                 let connection = signal_emitter.connection();
                                 let object_server = connection.object_server();
                                 #args_from_msg
@@ -864,7 +864,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                 property_name: &str,
                 object_server: &#zbus::ObjectServer,
                 connection: &#zbus::Connection,
-                header: &Option<#zbus::message::Header<'_>>,
+                header: Option<&#zbus::message::Header<'_>>,
             ) -> ::std::option::Option<#zbus::fdo::Result<#zbus::zvariant::OwnedValue>> {
                 match property_name {
                     #get_dispatch
@@ -876,7 +876,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                 &self,
                 object_server: &#zbus::ObjectServer,
                 connection: &#zbus::Connection,
-                header: &Option<#zbus::message::Header<'_>>,
+                header: Option<&#zbus::message::Header<'_>>,
             ) -> #zbus::fdo::Result<::std::collections::HashMap<
                 ::std::string::String,
                 #zbus::zvariant::OwnedValue,
@@ -894,7 +894,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                 property_name: &'call str,
                 value: &'call #zbus::zvariant::Value<'_>,
                 signal_emitter: &'call #zbus::object_server::SignalEmitter<'_>,
-                header: &'call Option<#zbus::message::Header<'_>>,
+                header: Option<&'call #zbus::message::Header<'_>>,
             ) -> #zbus::object_server::DispatchResult<'call> {
                 match property_name {
                     #set_dispatch
@@ -907,7 +907,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                 property_name: &str,
                 value: &#zbus::zvariant::Value<'_>,
                 signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
-                header: &Option<#zbus::message::Header<'_>>,
+                header: Option<&#zbus::message::Header<'_>>,
             ) -> ::std::option::Option<#zbus::fdo::Result<()>> {
                 match property_name {
                     #set_mut_dispatch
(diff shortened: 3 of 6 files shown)


zbus_macros/src/lib.rs
```diff
@@ -289,17 +289,17 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
 ///   [`ObjectServer`] this method was called by.
 /// * `connection` - This marks the method argument to receive a reference to the [`Connection`] on
 ///   which the method call was received.
 /// * `header` - This marks the method argument to receive the message header associated with the
-///   D-Bus method call being handled. For property getter methods, this will be an
-///   `Option<Header<'_>>`, which will be `None` when the function is being called as part of the
-///   initial object setup (before it gets registered on the bus), or when we send out property
-///   changed notifications.
+///   D-Bus method call being handled. For property methods, this will be an `Option<Header<'_>>`,
+///   which will be `None` for getters when the function is being called as part of the initial
```
zeenix, line 300: So currently it is always `Some` for setters? If so, here it's not clear why they're getting an `Option` in that case. I guess we want it to be an `Option` for #898.

I'd suggest removing all the details here and just say that: "which will be set to `None` if the method is called for reasons other than to respond to an external property access".

zbus_macros/src/iface.rs
```diff
@@ -1048,12 +1067,23 @@ fn get_args_from_inputs(
...
+            MethodType::Property(PropertyType::Setter) => quote! {
+                let hdr = header.as_ref().unwrap();
+
                 #server_arg_decl
 
                 #conn_arg_decl
 
                 #header_arg_decl
+
+                #signal_emitter_arg_decl
```
zeenix, line 1086: can we refactor to first generate the common bits here? I think best would be to just set `signal_emitter_arg_decl` to be `quote! {}` and similarly for the header init for the setter.

zeenix (changes requested):
Nice! Minor improvement suggestions only. :)

zbus_macros/src/lib.rs
rzvncj: Yes, it's always `Some` for setters.

TBH, I've done it that way at first because I wasn't sure I wasn't going to hit a case where `None` would be required, and ended up just keeping it for symmetry, so as much as I'd like to take credit for foresight related to #898, I can't. :smile:

But it's good to know it's useful that it stayed and `Option`.

Right, I will update the doc comments. Thanks!

zbus_macros/src/iface.rs
rzvncj: Do you mean we should just have a `MethodType::Property(_)` case (so no longer two separate cases for getter or setters) and just have `signal_emitter_arg_decl` expand to nothing for the getter, and so on?

I will certainly explore that.

rzvncj: > > Do you mean we should just have a `MethodType::Property(_)` case (so no longer two separate cases for getter or setters) and just have `signal_emitter_arg_decl` expand to nothing for the getter, and so on?
> 
> Yes and no. :) I meant that we can avoid some duplication by initializing `signal_emitter_arg_decl` and the header declaration differently:
> 
> ```rust
> let (hdr_init, signal_emitter_arg_decl) = match method_type {
>      ...
> }
> ```
> 
> I actually didn't realize that we're duplicating this in case of non-property as well so this refactor would be even more valuable than I realized. :)

Got it. I hope it's better now. Thanks!

zbus_macros/src/iface.rs
```diff
@@ -1047,35 +1066,47 @@ fn get_args_from_inputs(
...
-            MethodType::Property(_) => quote! {
-                #server_arg_decl
+        let (hdr_init, msg_init, signal_emitter_arg_decl, args_decl) = match method_type {
+            MethodType::Property(PropertyType::Getter) => {
+                (quote! {}, quote! {}, quote! {}, quote! {})
+            }
+            MethodType::Property(PropertyType::Setter) => (
+                quote! { let hdr = header.as_ref().unwrap(); },
+                quote! {},
+                quote! { #signal_emitter_arg_decl },
```
zeenix, line 1076: hmm.. while you're at it, you might as well make signal_emitter also available in the getter? :thinking:

rzvncj: I actually wanted and tried to, but it looks like the emitter is taken from the header, and the header will not always be available for getters. Is that a wrong assumption?

rzvncj: > hmm.. while you're at it, you might as well make signal_emitter also available in the getter? 🤔

Here's a quick try:

```
test iface_and_proxy ... FAILED

failures:

---- iface_and_proxy stdout ----
thread 'zbus::Connection executor' panicked at zbus/tests/e2e.rs:95:1:
called `Option::unwrap()` on a `None` value
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread 'iface_and_proxy' panicked at zbus/tests/e2e.rs:875:1:
timeout: the function call took 15000 ms. Max time 15000 ms


failures:
    iface_and_proxy

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 15.00s

error: test failed, to rerun pass `-p zbus --test e2e`
```

rzvncj: What I mean is that the way the signal emitter i scurrently extracted is (note `match hdr.path()`):
```rust
1042             } else if signal_context || signal_emitter {
1043                 if signal_emitter_arg_decl.is_some() {
1044                     return Err(Error::new_spanned(
1045                         input,
1046                         "There can only be one `signal_emitter` or `signal_context` argument",
1047                     ));
1048                 }
1049
1050                 let signal_context_arg = &input.pat;
1051
1052                 signal_emitter_arg_decl = Some(quote! {
1053                     let #signal_context_arg = match hdr.path() {
1054                         ::std::option::Option::Some(p) => {
1055                             #zbus::object_server::SignalEmitter::new(connection, p).expect("Infallible conversion failed")
1056                         }
1057                         ::std::option::Option::None => {
1058                             let err = #zbus::fdo::Error::UnknownObject("Path Required".into());
1059                             return connection.reply_dbus_error(&hdr, err).await;
1060                         }
1061                     };
1062                 });
```

If there's another way we can extract the emitter there, then we can certainly pass the emitter along to getters as well. Otherwise, a `None` header means no emitter.

zeenix: > I actually wanted and tried to, but it looks like getting the emitter requires a header,

It doesn't need to be. It just needs access to connection and the relevant object path, both of which should be available from contexts where getter is called from.

> What I mean is that the way the signal emitter is currently extracted is (note `match hdr.path()`):

Right, yeah.

---

Looking more into this, I think perhaps your approach here is better. Someone needing signal emissions from a getter are quite small. If anyone comes up with a good use case, we can add it later.

However, we should update the docs to reflect that the signal emitter arg is not available for property getters.

rzvncj: > However, we should update the docs to reflect that the signal emitter arg is not available for property getters.

They already do, [zbus_macros/src/lib.rs](https://github.com/dbus2/zbus/pull/1179/commits/1d80d07a1131c25944be2da1fecfef3a62626e45#diff-5110770af24f0b6088a2af69201805b4e31b1f4061ea76c56acd1a9fffc16cca) says (change made in the previous PR): "This argument is not available for property getters."

zbus/tests/e2e.rs
```diff
@@ -266,7 +266,7 @@ impl MyIface {
     #[zbus(property)]
     fn test_header_prop(
         &self,
-        #[zbus(header)] header: Option<Header<'_>>,
+        #[zbus(header)] header: Option<&Header<'_>>,
```
zeenix, line 269: This is inconsistent now with the `header` arg that normal methods get. I think `Option<Header>` is fine but if we want to change this, we need to also change the normal methods' arg type to `&Header`.

zbus/src/object_server/interface/mod.rs
```diff
@@ -58,15 +58,15 @@ pub trait Interface: Any + Send + Sync {
         property_name: &str,
         server: &ObjectServer,
         connection: &Connection,
-        header: &Option<message::Header<'_>>,
+        header: Option<&message::Header<'_>>,
```
zeenix, line 61: This commit now has 2 seperate logical changes. This one (not covered by the commit log) and the change of `interface` arg. This one is a fix and the other is more an improvement that needs to be applied more consistently. Hence they should be split, assuming we keep the other change.

zeenix (changes requested):
Thanks. Apart from the inline comment, I'd suggest also to:

* Keep the commit log summary line short (50 chars ideally but up to 74 is ok). Make use of the details section, where you can explain the change in detail.
* Please don't sign commits (unless you want to GPG sign them, which is fine). It's really unnecessary IMO and makes one think there are more details in the commit message only to find out that there aren't any. I understand other project require this so you may have this enabled in your git configuration but you can change it just for your local git repo. :pray:

rzvncj:
>     * Please don't sign commits (unless you want to GPG sign them, which is fine). It's really unnecessary IMO and makes one think there are more details in the commit message only to find out that there aren't any. I understand other project require this so you may have this enabled in your git configuration but you can change it just for your local git repo. 🙏

I didn't realize I was. That's indeed a global settings because another project required it.

zeenix:
> I didn't realize I was.

:laughing: 

> That's indeed a global settings because another project required it.

I'm not at all surprised. Unfortunately, many projects do. Oh well, their project, their way I guess. :shrug:

zbus/tests/e2e.rs
rzvncj: Is `Option<&Header>` not what you wanted? That's what I believe you asked for on Matrix.

Screenshot From 2024-12-22 13-58-08

I also did notice that while the other _method_ special args are being passed by reference, `Header` wasn't in a Matrix comment:

Screenshot From 2024-12-22 13-59-44

I'm sorry, but I'm slightly confused. What do you want we do here? Change `Option<&Header>` to `Option<Header>` (in spite of the original request for `Option<&T>`)? Or changing the old parameters to take `&Header` instead of `Header`?

zeenix: As I wrote there, I was talking about the `&Option<Header>` arg in the `Interface` property methods. They should be changed to `Option<&Header>`. I also wrote:

> ah that's only the Interface methods but still we should change it

i-e I didn't suggest we change anything in the `interface` macro args since they're not doing `&Option<..>` but `Option<..>` (i-e there is no reference involved).

rzvncj: > As I wrote there, I was talking about the `&Option<Header>` arg in the `Interface` property methods.

OK, I'll see about re-spinning this then.

zbus_macros/src/iface.rs
```diff
@@ -1017,7 +1017,7 @@ fn get_args_from_inputs(
                 let header_arg = &input.pat;
 
                 header_arg_decl = match method_type {
-                    MethodType::Property(_) => Some(quote! { let #header_arg = header.clone(); }),
+                    MethodType::Property(_) => Some(quote! { let #header_arg = header.cloned(); }),
```
zeenix, line 1039: As discussed in the PM, we want to use the full paths in the generated code: `::std::Option::cloned(header);`

zbus/tests/e2e.rs
```diff
@@ -277,6 +277,21 @@ impl MyIface {
...
+        &self,
+        value: bool,
+        #[zbus(header)] header: Option<Header<'_>>,
+        #[zbus(connection)] connection: &Connection,
+        #[zbus(object_server)] object_server: &ObjectServer,
+        #[zbus(signal_emitter)] emitter: SignalEmitter<'_>,
+    ) {
+        debug!("`TestHeaderProp` setter called, value: {}, header: {:?}, connection: {:?}, object_server: {:?}, emitter: {:?}",
+            value, header, connection, object_server, emitter
+        );
```
zeenix, line 292: best to do a `header.is_some();` here too.

zbus_macros/src/lib.rs
```diff
@@ -289,17 +289,16 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// using this since it will force all interested peers to fetch the new value and hence result in
 /// excess traffic on the bus.
 ///
-/// The method and property getter arguments support the following `zbus` attributes:
+/// The method and property arguments support the following `zbus` attributes:
```
zeenix, line 292: I think we can remove the `and property` now since they all support them now.

zeenix (changes requested):
> .. Support Option<&Header<'_>> parameter in property setters

nit: We're adding support for all attributes so better summary would be something like "Support special args in `interface` property setters" IMO.

---

Pretty sure next iteration can be merged. :)

zbus/tests/e2e.rs
zeenix: it would output on the logging but it wouldn't ensure that it's correctly set to `Some`.

zbus_macros/src/iface.rs
zeenix: AFAICT (unless GH is lying to me), this is not resolved. :thinking:

rzvncj: No, sorry, I marked it resolved before actually pushing. I do that to keep track of my work, but since obviously this is not the workflow you're using I will refrain from doing that in future PRs.

zbus/tests/e2e.rs
zeenix: ah, there was a misunderstanding. Seems, also in case of the property getter test you added in the previous PR. I meant, we should check for it: `assert!(header.is_some());`.

zbus_macros/src/iface.rs
zeenix: >  do that to keep track of my work, but since obviously this is not the workflow you're using I will refrain from doing that in future PRs.

I think it's a good thing you track your progress like this. :+1: I do suggest doing it only after pushing, not only for the benefit of the reviewer but also to ensure that you didn't forget to push. :)

zeenix: Hmm.. it's still not resolved btw but the conversation was resolved? I'm getting very confused now so I'll take a break..

zbus/tests/e2e.rs
rzvncj: Right, that makes sense! Done.

zbus_macros/src/iface.rs
zeenix: GH was showing me old commits for some reason. :thinking:

zeenix: > GH was showing me old commits for some reason. 🤔

No, you ended up putting it in the wrong (second) commit. :) This is a fixup of the first commit.

rzvncj: A break sounds good :slightly_smiling_face:.

Here's the commit: https://github.com/dbus2/zbus/pull/1179/commits/e3e0d1105a12415adf65ef12ac0d685502618dbf#diff-43a2e53b8de7776ea27dd141513cf075f24db06d6612dd519ff04780a301807cR1039

I don't know what GitHub's thinking.

rzvncj: > No, you ended up putting it in the wrong (second) commit. :) This is a fixup of the first commit.

You're right, sorry! Fixed.

zeenix (approved):
LGTM. Thanks.

rzvncj:
> LGTM. Thanks.

Thank you!

PR #1183, opened 2024-12-22
✨ zb,zm: Support SignalEmitter<'_> in property getters

rzvncj:


zbus/src/object_server/node.rs
```diff
@@ -214,6 +215,7 @@ impl Node {
     pub(crate) async fn get_managed_objects(
         &self,
         object_server: &ObjectServer,
+        emitter: &SignalEmitter<'_>,
         connection: &Connection,
```
zeenix, line 219: * Looks like we're not very consistent with the order of the arguments (also before this commit). Could we fix the order in a separate commit first and ensure it's always consistent and then in this commit, we keep the consistency?
* I don't think you need to add the emitter argument here. The node knows it's path and hence can create the emitter from the connection and path. This makes me wonder if this is even needed at all in `Interface` methods. Keeping in mind that `Interface::call*` don't take this argument, I'm pretty sure this is not needed. If so, please also remove it from `Interface::set*` and create the emitter in the generated code.

zeenix (changes requested):
Awesome but I think we can simplify it to not touch `Interface` at all.

zbus/src/object_server/node.rs
rzvncj: > * Looks like we're not very consistent with the order of the arguments (also before this commit). Could we fix the order in a separate commit first and ensure it's always consistent and then in this commit, we keep the consistency?

Is there a particular order you prefer? Which function would be an example?

zeenix: > Is there a particular order you prefer? Which function would be an example?

Let's keep it the same as it was before your first PR. E.g in `Interface::call`, it was: server, connection and message/header.

Looking at it now, didn't we say that it would be more consistent with `Interface` property methods would take a `Option<&Message>` instead of `Option<Header>`? :thinking:

zeenix: Looked at our chat and indeed I pointed that out:

> although now that I took a closer look, taking a Option<&Message> may be more consistent

Given that you always get the header from a message, I really think we should do this for consistency (or change `call*` methods to take a header but I don't see a good reason to do that.

rzvncj: > > Is there a particular order you prefer? Which function would be an example?
> 
> Let's keep it the same as it was before your first PR. E.g in `Interface::call`, it was: server, connection and message/header.
> 
> Looking at it now, didn't we say that it would be more consistent with `Interface` property methods would take a `Option<&Message>` instead of `Option<Header>`? 🤔

We did, but I thought we decided that for the last PR it would do to just continue to use the header and look into that later in order to not make it into a bigger piece of work than it strictly needed to be.

I guess since this the main jist of this PR is not a huge patch I could look into that now too.

zeenix: > I guess since the main jist of this PR is not a huge patch I could look into that now too.

You don't have to. :) This was more a general note so feel free to do it after this PR. We just have to ensure that all the Interface changes happen before the next release. It's declared as unstable API but we should still try our best to not change it too often between releases.

rzvncj: >  I don't think you need to add the emitter argument here. The node knows it's path and hence can create the emitter from the connection and path. This makes me wonder if this is even needed at all in `Interface` methods. Keeping in mind that `Interface::call*` don't take this argument, I'm pretty sure this is not needed. If so, please also remove it from `Interface::set*` and create the emitter in the generated code.

`set*` passes its `signal_emitter` argument on to `property_changed()` functions in the generated code, so it would make sense to keep the emitter argument, I think, since it is not only useful for the special property function arguments, but also required for emitting property changed signals.

rzvncj: >   If so, please also remove it from `Interface::set*` and create the emitter in the generated code.

Other than this suggestion (which I've attempted to explain above does not seem feasible, though it's always possible I'm missing something), I think I've addressed the rest of the suggestions.

I'll look at replacing `Option<&Header>` with `Option<&Message>` once this PR goes in.

I'll leave this conversation unresolved if you disagree with anything, please resolve it if I've addressed everything to your satisfaction.

rzvncj: > > I don't think you need to add the emitter argument here. The node knows it's path and hence can create the emitter from the connection and path. This makes me wonder if this is even needed at all in `Interface` methods. Keeping in mind that `Interface::call*` don't take this argument, I'm pretty sure this is not needed. If so, please also remove it from `Interface::set*` and create the emitter in the generated code.
> 
> `set*` passes its `signal_emitter` argument on to `property_changed()` functions in the generated code, so it would make sense to keep the emitter argument, I think, since it is not only useful for the special property function arguments, but also required for emitting property changed signals.

What I mean to say, in other words, is that for methods the emitter argument is optional, but for property setters it is required because of the additional generated code that sends out property changed signals.

We could extract it in the generated code and remove the emitter parameter to `set*`, but IMHO that's extra computation (so the code becomes less efficient) that always needs to happen (again, unlike methods where it only happens when there's an emitter parameter) with not much real benefit (since the `set*` signatures reflect the importance of the emitter parameter correctly).

zeenix: > `set*` passes its `signal_emitter` argument on to `property_changed()` functions in the generated code, so it would make sense to keep the emitter argument, I think, since it is not only useful for the special property function arguments, but also required for emitting property changed signals.

Yeah, `set*` perhaps does. You're right. I'm sorry, for causing the misunderstanding about that. I got carried away.

However, this specific method, doesn't need to be passed the emitter and can create it itself, given the node knows its own path and has a connection passed to it already.

zbus_macros/src/iface.rs
```diff
@@ -1051,37 +1053,38 @@ fn get_args_from_inputs(
...
-                        }
-                        ::std::option::Option::None => {
-                            let err = #zbus::fdo::Error::UnknownObject("Path Required".into());
-                            return connection.reply_dbus_error(&hdr, err).await;
-                        }
-                    };
-                });
+                signal_emitter_arg_decl = match method_type {
+                    MethodType::Property(PropertyType::Getter) => {
+                        Some(quote! { let #signal_context_arg = signal_emitter.clone(); })
```
zeenix, line 1058: Full path in generated code please: `::std::Clone::clone(signal_emitter)`. Don't feel bad about it though. I make that mistake too. It's easy to forget.

zbus_macros/src/iface.rs
```diff
@@ -1051,37 +1053,38 @@ fn get_args_from_inputs(
...
+                    }
+                    _ => Some(quote! {
+                        let #signal_context_arg = match hdr.path() {
+                            ::std::option::Option::Some(p) => {
+                                #zbus::object_server::SignalEmitter::new(connection, p).expect("Infallible conversion failed")
+                            }
+                            ::std::option::Option::None => {
```
zeenix, line 1065: Hmm.. isn't this path going to be taken for the property setter too? If so, we don't use the `signal_emitter` arg to `Interface::set*`. I think it'd be best to do the same for getter and setter here.

zeenix (changes requested):
LGTM otherwise.

zbus/src/object_server/node.rs
rzvncj: > However, this specific method, doesn't need to be passed the emitter and can create it itself, given the node knows its own path and has a connection passed to it already.

I agree and I've already done that in the latest PR update. Did I miss something?

zeenix: > I agree

:+1: 

> and I've already done that in the latest PR update. Did I miss something?

`Node::get_properties` is still gaining an emitter argument.

zbus/src/object_server/node.rs
```diff
@@ -245,14 +247,15 @@ impl Node {
         &self,
         object_server: &ObjectServer,
         connection: &Connection,
+        emitter: &SignalEmitter<'_>,
```
zeenix, line 250: [Here](https://github.com/dbus2/zbus/pull/1183#discussion_r1896710064).

zbus_macros/src/iface.rs
rzvncj: Huh. I thought that in this particular case, since `signal_emitter` is accessible and usable, that the Rust compiler would be able to figure out correctly which `.clone()` we mean, but I admit that's "based" more on the rules of dependent lookup in C++ that I'm familiar with, than any actual Rust documentation. :slightly_smiling_face: 

I will of course update this. Thanks!

zbus_macros/src/iface.rs
rzvncj: Crap, you're right, sorry and good catch! I'll have a think about how that came about, then I'll update the PR.

rzvncj: Not sure how I ended up doing that, it's definitely not intended. I guess that was just part of a larger refactoring effort that just ended up in a typo there. Should be OK now.

zbus_macros/src/iface.rs
zeenix: > I thought that in this particular case, since `signal_emitter` is accessible and usable, that the Rust compiler would be able to figure out correctly which `.clone()` we mean

You're right. In this specific case, I see very little possibility of it being ambiguous in any context but there is no way to be sure about it. In general it's best to just always be explicit in the generated code.

zeenix (approved):
Nice!

rzvncj:
> Nice!

Thanks for the prompt and helpful reviews! Contributing to zbus has been a very enjoyable experience.

zeenix:
> > Nice!
> 
> Thanks for the prompt and helpful reviews! Contributing to zbus has been a very enjoyable experience.

Haha, you just have a very healthy attitude. :+1:  Many people get annoyed by my perfectionism and leave PRs hanging forever. :sob:

zbus/src/object_server/node.rs
rzvncj: > Looked at our chat and indeed I pointed that out:
> 
> > although now that I took a closer look, taking a Option<&Message> may be more consistent
> 
> Given that you always get the header from a message, I really think we should do this for consistency (or change `call*` methods to take a header but I don't see a good reason to do that.

On second thought, I did quickly try to change to `Option<&Message>`, and that ended up requiring that I add a `Message` parameter to `get_property()` (in proxy/mod.rs), which even assuming would be easy, would change the public API.

IOW, I'm unsure of how to produce a `Message` in all the contexts it is required (i.e. where we are currently able to produce a `Some(&header)`), and this is complicated by the fact that we have a special `#[zbus(header)]` macro, but no `message` equivalent.

Obviously I could once more be missing something - if that's the case, suggestions are, as usual, appreciated. Thanks!

zeenix: > On second thought, I did quickly try to change to `Option<&Message>`, and that ended up requiring that I add a `Message` parameter to `get_property()` (in proxy/mod.rs), which even assuming would be easy, would change the public API.

Huh? This is only the `interface` and service side. How is proxy involved at all? :thinking:

zeenix: AFAICT, you just need to change `Option<&Header>` to `Option<&Message>` in `Interface` methods and then in the macro fetch the header from the message (if `Some`), very similar to how it works for `Interface::call*`, just that it's optional.

rzvncj: > Huh? This is only the `interface` and service side. How is proxy involved at all? 🤔

OK, I've stashed my original change and started over so I can illustrate the steps:

1. In iface.rs, change `set<'call>()` and `set_mut()` to use `msg: Option<&#zbus::message::Message>` instead of the optional `Header<'_>`.

2. `cargo build` says I need to look at zbus/src/object_server/interface/mod.rs:85:17. I change `set()` and `set_mut()` in that file as well.

3. `cargo build` says I need to look at zbus/src/fdo/properties.rs:85:18. There's a `set()` there that takes a `#[zbus(header)] header: Header<'_>,` that I previously made use of. Now, if I remove this `header` parameter and add a `msg: Message` parameter (so I can pass it further in `match iface.instance.read().await.set()`) and change the signature of this `set()` function, then I get:

```
error[E0061]: this method takes 4 arguments but 3 arguments were supplied
   --> zbus/src/proxy/mod.rs:813:14
    |
813 |             .set(self.inner.interface.as_ref(), property_name, value.into())
    |              ^^^------------------------------------------------------------ argument #4 of type `message::Message` is missing
|
```

And this is how we get to the proxy code. Is there something I'm missing that would allow this not to happen?

rzvncj: > AFAICT, you just need to change `Option<&Header>` to `Option<&Message>` in `Interface` methods and then in the macro fetch the header from the message (if `Some`), very similar to how it works for `Interface::call*`, just that it's optional.

Right, but the issue seems to be with the interface part of the `org.freedesktop.DBus.Properties` object, which we use a proxy to in order to get and set properties. By design, there's no accessible full `Message` there that I can think of (but we do have a `Header`, received as a special arg now, like for any other interface).

zeenix: > And this is how we get to the proxy code. 

Ah, it's because that code is using a proxy that is generated from `interface` macro use. You'll need to ensure that the special arg is not getting forwarded to the `proxy` macro use in `interface`, just like the other special args are filtered out.

zeenix: Also it's not actually proxy code that uses `interface` itself but rather it uses the proxy generated from the `interface` (indirectly through `proxy` macro). So the proxy code that the error is pointing to, is not itself the trigger for this. I'll have to dig deeper to figure out what goes wrong, but I'm pretty sure this is something in the macro code that needs updating after this change.

zeenix: The important thing to keep in mind here is that the header is coming from a message, so if you can provide the header, you can also provide the message it comes from.

rzvncj: > > And this is how we get to the proxy code.
> 
> Ah, it's because that code is using a proxy that is generated from `interface` macro use. You'll need to ensure that the special arg is not getting forwarded to the `proxy` macro use in `interface`, just like the other special args are filtered out.

Right, but that would mean that we need to introduce `#[zbus(message)]`. Is that worth doing just so we can pass `Option<&Message>` in a couple of functions, or would it be better to just keep the code as is (albeit with that slight inconsistency)?

zeenix: > Right, but that would mean that we need to introduce `#[zbus(message)]`

Ah, yes true.

> Is that worth doing just so we can pass `Option<&Message>` to a couple of functions, or would it be better to just keep the code as is (albeit with that slight inconsistency)?

I guess not. `Interface` is declared unstable after all so if we really need this, we can do it then. So never mind and thanks for looking into it. :+1:

rzvncj: > > Right, but that would mean that we need to introduce `#[zbus(message)]`
> 
> Ah, yes true.
> 
> > Is that worth doing just so we can pass `Option<&Message>` to a couple of functions, or would it be better to just keep the code as is (albeit with that slight inconsistency)?
> 
> I guess not. `Interface` is declared unstable after all so if we really need this, we can do it then. So never mind and thanks for looking into it. 👍

No problem! Thanks for the prompt replies!

PR #1158, opened 2024-11-25
zb: allow direct access to message properties without allocating

TTWNO:
- Creates a handful of new methods (there could potentially be a few more, too) that allow the user of `zbus` to directly access a reference to header properties without cloning the signature or the `PrimaryHeader`.
- Secondly, it allows you to consume the message in exchange for an owned `zbus::message::Body` type.

This increases performance relative to taking the cloned primary header and signature.
[`atspi`, branch = "deserialize-properly"](https://github.com/odilia-app/atspi/tree/deserialize-properly) can be cloned and `cargo bench` run to confirm these results.

Performance improvement for atspi is in the 30-50% range; I'll attach my benchmarks here for reference.

## `zbus-5.x` branch of atspi

```
     Running benches/event_parsing.rs (target/release/deps/event_parsing-08ab76e3e81624d0)
Gnuplot not found, using plotters backend
Benchmarking 1000 Messages into Events
Benchmarking 1000 Messages into Events: Warming up for 3.0000 s
Benchmarking 1000 Messages into Events: Collecting 100 samples in estimated 9.3177 s (10k iterations)
Benchmarking 1000 Messages into Events: Analyzing
1000 Messages into Events
                        time:   [923.37 µs 925.20 µs 927.12 µs]
                        change: [+109.91% +110.21% +110.52%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 21 outliers among 100 measurements (21.00%)
  2 (2.00%) low severe
  2 (2.00%) low mild
  2 (2.00%) high mild
  15 (15.00%) high severe

     Running benches/event_parsing_100k.rs (target/release/deps/event_parsing_100k-da6ad9ef4c4bfa68)
Gnuplot not found, using plotters backend
Benchmarking 100_000 Messages into Events
Benchmarking 100_000 Messages into Events: Warming up for 3.0000 s

Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 9.3s, or reduce sample count to 50.
Benchmarking 100_000 Messages into Events: Collecting 100 samples in estimated 9.3435 s (100 iterations)
Benchmarking 100_000 Messages into Events: Analyzing
100_000 Messages into Events
                        time:   [93.347 ms 93.455 ms 93.579 ms]
                        change: [+136.69% +137.03% +137.44%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  4 (4.00%) high mild
  2 (2.00%) high severe
```

[...]

PR #1188, opened 2024-12-26
⚡️ zb: Make header cloning cheaper

emilio:
Right now cloning a header involves cloning the `PrimaryHeader` (cheap,
mostly POD), and then cloning the `Fields`, which can be somewhat
complex, as it involves cloning all the strings and the signature, which
if dynamic could be rather costly.

From things coming from a message, strings are always `Borrowed` so the
cloning is cheap, so the only expensive thing is cloning the signature.

Using `Cow` for it looks sensible, it's what `Str` is doing internally
already, effectively.

### DIFF
--- a/zbus/src/message/builder.rs
+++ b/zbus/src/message/builder.rs
@@ -1,4 +1,5 @@
 use std::{
+    borrow::Cow,
     io::{Cursor, Write},
     sync::Arc,
 };
@@ -231,7 +232,7 @@ impl<'a> Builder<'a> {
         let ctxt = dbus_context!(self, 0);
         let mut header = self.header;
 
-        header.fields_mut().signature = signature;
+        header.fields_mut().signature = Cow::Owned(signature);
 
         let body_len_u32 = body_size.size().try_into().map_err(|_| Error::ExcessData)?;
         header.primary_mut().set_body_len(body_len_u32);
@@ -285,7 +286,7 @@ impl<'m> From<Header<'m>> for Builder<'m> {
     fn from(mut header: Header<'m>) -> Self {
         // Signature and Fds are added by body* methods.
         let fields = header.fields_mut();
-        fields.signature = Signature::Unit;
+        fields.signature = Cow::Owned(Signature::Unit);
         fields.unix_fds = None;
 
         Self { header }
--- a/zbus/src/message/fields.rs
+++ b/zbus/src/message/fields.rs
@@ -4,7 +4,7 @@ use serde::{
     Deserialize, Deserializer, Serialize, Serializer,
 };
 use static_assertions::assert_impl_all;
-use std::num::NonZeroU32;
+use std::{borrow::Cow, num::NonZeroU32};
 use zbus_names::{BusName, ErrorName, InterfaceName, MemberName, UniqueName};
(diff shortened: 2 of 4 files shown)


zeenix (changes requested):
Sure, why not, even though we've [plans to make `Header` cloning super cheap (or even free) by use of `Cow`](https://github.com/dbus2/zbus/pull/1158#issuecomment-2514973015). Could you please:

* Update the docs (at least the `interface` macro docs need updating).
* Add an emoji to the commit, as per our contributing guide. I'll update the PR title for you, please c&p that to the commit title.

emilio:
Done, and done, thanks!

zeenix:
> Sure, why not,

Oh wait, I know one reason: this will make it inconsistent with the normal methods, which take header by value and changing them would require an API break.

I'm sorry to have wasted your time by not realising this when you submitted this but could you please instead make the header cloning zero-cost (or at least super cheap so cloning is not a concern)?

zeenix:
> I'm sorry to have wasted your time by not realising this when you submitted this but could you please instead make the header cloning zero-cost (or at least super cheap so cloning is not a concern)?

See the discussion I pointed to in #1158 above. I think the task would be just to finish the work of that PR.

emilio:
So just to recap my understanding:

 * Right now cloning a header involves cloning the `PrimaryHeader` (cheap, mostly POD), and then cloning the `Fields`, which can be somewhat complex: Cloning all the strings (well, sure, bumping the refcounts), and the signature, which if dynamic could be rather costly.

 * I guess from things coming from a message, those are always `Borrowed` so the cloning is cheap. So the only expensive thing is cloning the signature. Using `Cow` for it looks sensible, it's what `Str` is doing internally already effectively.

So that sounds good. Take a look above ^.

One non-trivial difference from the PR you linked is that I changed:

```
if self.signature != Cow::Borrowed(&Signature::Unit) {
```

Which seems wrong, to actually dereference the signature, which preserves behavior

zeenix:
> * Right now cloning a header involves cloning the `PrimaryHeader` (cheap, mostly POD), and then cloning the `Fields`, which can be somewhat complex: Cloning all the strings (well, sure, bumping the refcounts), and the signature, which if dynamic could be rather costly.
>
> * I guess from things coming from a message, those are always `Borrowed` so the cloning is cheap. So the only expensive thing is cloning the signature. Using `Cow` for it looks sensible, it's what `Str` is doing internally already effectively.

Right. Signature cloning is the potentially (depending on the signature) expensive bit so that's the primary target here. However, I think we can make the other (non-copy) fields potentially no-cost (they're currently low-cost) as well by use of `Cow`.

zeenix (changes requested):
Nice! LGTM and thanks for the very quick actions. For my side, all that's left is:

* Please keep the commit title short (ideal: 50 chars limit, acceptable: 75). This could just be "⚡️ zb: Make header cloning cheaper". The use of `Cow` is a detail and should go in the details section of the commit.
* Please update the PR title accordingly. Alternatively you can close this PR and create a new one. Whichever your prefer. :)

emilio:
Ok, updated the commit and PR title / description too :)

zeenix (approved):

Issue #1180, opened 2024-12-19
xmlgen generates code that fails to compile with a dictionary-like property


emilio:
(First of all, thanks for this crate, looks amazing)

I was poking at wpa_supplicant's dbus functionality: https://w1.fi/wpa_supplicant/devel/dbus.html

I generated a rust file like:

```
$ zbus-xmlgen system fi.w1.wpa_supplicant1 /fi/w1/wpa_supplicant1/Interfaces/1
```

Which generated a `p2pdevice.rs` file like this (among others):

<details>

```rust
//! # D-Bus interface proxy for: `fi.w1.wpa_supplicant1.Interface.P2PDevice`
//!
//! This code was generated by `zbus-xmlgen` `5.0.1` from D-Bus introspection data.
//! Source: `Interface '/fi/w1/wpa_supplicant1/Interfaces/1' from service 'fi.w1.wpa_supplicant1' on system bus`.
//!
//! You may prefer to adapt it, instead of using it verbatim.
//!
//! More information can be found in the [Writing a client proxy] section of the zbus
//! documentation.
//!
//! This type implements the [D-Bus standard interfaces], (`org.freedesktop.DBus.*`) for which the
//! following zbus API can be used:
//!
//! * [`zbus::fdo::IntrospectableProxy`]
//! * [`zbus::fdo::PropertiesProxy`]
//!
//! Consequently `zbus-xmlgen` did not generate code for the above interfaces.
//!
//! [Writing a client proxy]: https://dbus2.github.io/zbus/client.html
//! [D-Bus standard interfaces]: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces,
use zbus::proxy;
#[proxy(
    interface = "fi.w1.wpa_supplicant1.Interface.P2PDevice",
    default_service = "fi.w1.wpa_supplicant1",
    default_path = "/fi/w1/wpa_supplicant1/Interfaces/1"
)]
pub trait P2PDevice {
    /// AddPersistentGroup method
    fn add_persistent_group(
        &self,
        args: std::collections::HashMap<&str, &zbus::zvariant::Value<'_>>,
    ) -> zbus::Result<zbus::zvariant::OwnedObjectPath>;

    /// AddService method
    fn add_service(
        &self,
        args: std::collections::HashMap<&str, &zbus::zvariant::Value<'_>>,
    ) -> zbus::Result<()>;

    /// Cancel method
    fn cancel(&self) -> zbus::Result<()>;

[...]

PR #1189, opened 2024-12-26
Don't require an owned value for property setters.

emilio:
I wrote this as I was thinking about #1180, but not sure it's worth the complexity. If you think there are some other things that might use it, please go ahead and merge it, otherwise feel free to close :)

### DIFF
--- a/zbus/src/fdo/properties.rs
+++ b/zbus/src/fdo/properties.rs
@@ -59,7 +59,7 @@ impl Properties {
         &self,
         interface_name: InterfaceName<'_>,
         property_name: &str,
-        value: Value<'_>,
+        value: &Value<'_>,
         #[zbus(object_server)] server: &ObjectServer,
         #[zbus(connection)] connection: &Connection,
         #[zbus(header)] header: Header<'_>,
@@ -76,7 +76,7 @@ impl Properties {
 
         match iface.instance.read().await.set(
             property_name,
-            &value,
+            value,
             server,
             connection,
             Some(&header),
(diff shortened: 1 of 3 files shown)


zbus/src/fdo/properties.rs
```diff
@@ -62,7 +62,7 @@ impl Properties {
         &self,
         interface_name: InterfaceName<'_>,
         property_name: &str,
-        value: Value<'_>,
+        value: &Value<'_>,
```
zeenix, line 65: Unfortunately, this will be a breaking change since this is a public API. :(

emilio: Hmm, ok, lmk if / how you want to deal with it and I can update the patch. Alternatively feel free to just close this (or tag it as deferred or whatever you want :))

emilio:
I addressed the (somewhat pedantic) clippy lint :)

zbus/src/fdo/properties.rs
zeenix: I'll just tag it I guess. I hope you're around to rebase when 6.0 is being released. :)

zeenix: Oh I just realised something: we can add another method here, that does what we need and then have the existing method use that one. 

I'm thinking of adding a more generic method though that interface impl can use directly..

zeenix: > I'm thinking of adding a more generic method though that interface impl can use directly..

Ok, nm this part for now (https://github.com/dbus2/zbus/issues/1361). However, I think a new method will help get around the API break.

zeenix:
@emilio Hi, it's finally the time. We're preping for 6.0. Could you please rebase?

emilio:
@zeenix rebased!

PR #1223, opened 2025-01-23
🐛 zm: Fix generated get_all() function

rzvncj:
Naming the optional header parameter 'header' several times for different property getters collides with the 'header` local argument of the generated get_all() function. Obvious workaround is don't name your special parameter 'header'.

This is easily fixed by simply wrapping each getter code block generated for get_all() in its own scope.


zeenix:
Thanks!

zeenix:
> zm: Fix generated get_all() function

Just for future ref: This is the benefit of the change, not the summary of the change itself. IMO commit summary should have the later as it makes it easy to see in the log what commit did what.

rzvncj:
> Thanks!

Thanks for the lightning-quick review and merge! Also, the commit info suggestion has been noted.

PR #1237, opened 2025-02-02
🐛 zb: Correctly format list of accepted auth mechanisms

### LINKED ISSUE #1216: Bad formatting of accepted mechanisms, regression
This is a regression from commit b792340fc31232e6eced6af99f53bc98e991353a "zb: Use string type for auth mechanisms in Rejected cmd". The mechanisms are not parsed anymore, and they are joined without separator when returning an Handshake error, ex from gio connection:

zbus(Handshake("EXTERNAL rejected by the server. Accepted mechanisms: [EXTERNALDBUS_COOKIE_SHA1]")

zeenix:
This fixes a regression from b792340.

Fixes #1216.


elmarco:
ack

PR #1274, opened 2025-02-27
🐛 xmlgen: Raise XML parsing buffer size

kageurufu:
This fixes parsing large dbus specifications
Specifically,
https://github.com/PackageKit/PackageKit/blob/main/src/org.freedesktop.PackageKit.Transaction.xml currently results in "Error: QuickXml(TooManyEvents(1024))"


zeenix:
Thanks!

PR #1285, opened 2025-03-05
🧑‍💻 all: Use workspace dependencies

fmiguelgarcia:
Move dependencies to workspace. 

It will facilitate a future upgrade of dependency.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -128,17 +128,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "async-fs"
-version = "2.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ebcd09b382f40fcd159c2d695175b2ae620ffa5f3bd6f664131efff4e8b9e04a"
-dependencies = [
- "async-lock",
- "blocking",
- "futures-lite",
-]
-
-[[package]]
 name = "async-io"
 version = "2.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


Cargo.lock
```diff
@@ -127,17 +127,6 @@ dependencies = [
...
-name = "async-fs"
-version = "2.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ebcd09b382f40fcd159c2d695175b2ae620ffa5f3bd6f664131efff4e8b9e04a"
-dependencies = [
- "async-lock",
- "blocking",
- "futures-lite",
-]
-
```
zeenix, line 140: Why are deps being removed from the lock file as a result of this? :thinking:

zeenix (changes requested):
Oh nice! This is indeed going to help a lot. :+1: I really appreciate you following the contribution guide as well.

Apart from the inline comment below, one nitpick about the emoji: please copy&paste the emoji itself (rather than the nickname). Github does the right thing but many git tools (especially the CLI ones) don't. :)

Cargo.lock
fmiguelgarcia: I executed `cargo machete` (I get use to that tbh), and I found some unused dependencies. I ran test just to be sure that everything is working, but let me know if I miss something  :ok_hand:

zeenix: Actually never mind. Just rebase on latest main after [this](https://github.com/dbus2/zbus/pull/1286) is merged. :)

zeenix:
Thanks for updating but you ended up merging main and that unnecessarily complicates history. Please rebase it. 🙏

fmiguelgarcia:
Replaced by https://github.com/dbus2/zbus/pull/1291

Cargo.lock
fmiguelgarcia: Sorry @zeenix . I've open a clean new one, thanks.

zeenix: I never understood why people create new PRs. It's just more noise. Just gotta force push to the same branch on the remote.

PR #1291, opened 2025-03-07
🧑‍💻 all: Use workspace dependencies

fmiguelgarcia:
Move dependencies to workspace.

It will facilitate a future upgrade of dependency.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,3 +24,92 @@ repository = "https://github.com/dbus2/zbus/"
 [profile.bench]
 debug = true
 strip = "none"
+
+[workspace.dependencies]
+# Local dependencies
+zvariant_derive = { path = "./zvariant_derive" }
+zvariant_utils = { path = "./zvariant_utils" }
+zvariant = { path = "./zvariant", default-features = false }
+zbus_names = { path = "./zbus_names" }
+zbus_macros = { path = "./zbus_macros" }
+zbus_xml = { path = "./zbus_xml" }
+zbus = { path = "./zbus" }
+
+# 3rd party
+snakecase = "0.1.0"
+pretty_assertions = "1.4"
+clap = { version = "4.5.4", features = ["derive", "wrap_help"] }
+endi = "1.1.0"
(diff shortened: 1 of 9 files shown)


zeenix (approved):
Thanks. Although I'm a bit unsure if we want to put all dependencies into the workspace (e.g `chrono` is only needed in `zvariant`) but maybe it's more future safe this way. :shrug: Now let's see what the CI says. :)

zeenix:
@fmiguelgarcia This turned out to be a bad idea for local deps:

```
zbus/zvariant_derive ❯ cargo publish
    Updating crates.io index
error: all dependencies must have a version specified when publishing.
dependency `zvariant_utils` does not specify a version
Note: The published dependency will use the version from crates.io,
the `path` specification will be removed from the dependency declaration.
```

And if I specify versions in the workspace:

```
❯ cargo publish --dry-run 
    Updating crates.io index
   Packaging zvariant_derive v5.5.0 (/home/zeenix/checkout/dbus2/zbus/zvariant_derive)
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `zvariant = "^5.5.0"`
  candidate versions found which didn't match: 5.4.0, 5.3.0, 5.2.0, ...
  location searched: crates.io index
  required by package `zvariant_derive v5.5.0 (/home/zeenix/checkout/dbus2/zbus/zvariant_derive)`
```

Specifying versions in the subcrate Cargo.toml breaks too:

```
❯ cargo publish --dry-run 
    Updating crates.io index
   Packaging zvariant_derive v5.5.0 (/home/zeenix/checkout/dbus2/zbus/zvariant_derive)
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `zvariant = "^5.5.0"`
  candidate versions found which didn't match: 5.4.0, 5.3.0, 5.2.0, ...
  location searched: crates.io index
  required by package `zvariant_derive v5.5.0 (/home/zeenix/checkout/dbus2/zbus/zvariant_derive)`
```

So I'm going to drop the changes here for local deps.
```

PR #1292, opened 2025-03-07
⚡️zb: don't clone signature when creating body

TTWNO:
We saw performance uplifts in the range of 10-15% in parsing events just by not cloning, reparsing the signature every time a body was created.

Hopefully it's a small enough change that it's easy to review.

This directly reduces the cost of calling `Body::new(...)` by 90-95%.

### DIFF
--- a/zbus/benches/benchmarks.rs
+++ b/zbus/benches/benchmarks.rs
@@ -33,6 +33,9 @@ fn msg_de(c: &mut Criterion) {
         b.iter(|| {
             let header = msg.header();
             black_box(header);
+            let body = msg.body();
+            let signature = body.signature();
+            black_box(signature);
         })
     });
 
@@ -43,6 +46,8 @@ fn msg_de(c: &mut Criterion) {
     g.bench_function("body", |b| {
         b.iter(|| {
             let body = msg.body();
+            let signature = body.signature();
+            black_box(signature);
             let body: BigBoy<'_> = body.deserialize().unwrap();
             black_box(body);
         })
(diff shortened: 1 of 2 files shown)


TTWNO:
We saw performance uplifts in the range of 10-15% in parsing events just by not cloning, reparsing the signature every time a body was created.

Hopefully it's a small enough change that it's easy to review.

zeenix:
> We saw performance uplifts in the range of 10-15% in parsing events just by not cloning, reparsing the signature every time a body was created.
> 
> Hopefully it's a small enough change that it's easy to review.
> 
> This directly reduces the cost of calling `Body::new(...)` by 90-95%.

Wow, cool. Could you please:

1. put this in the description of the PR and commit? 
2. add a benchmark for this (unless it's covered [by the existing ones](https://github.com/dbus2/zbus/blob/main/zbus/benches/benchmarks.rs)?)?

TTWNO:
Added benchmarks. It's measured in nanoseconds so don't worry too much if there is a bit of variation here and there. But if it jumps into the hundreds of ns... then there might be a problem.

zeenix:
> Added benchmarks.

Awesome! I'll move the commit order so it's possible to see the change. :)

> It's measured in nanoseconds so don't worry too much if there is a bit of variation here and there. But if it jumps into the hundreds of ns... then there might be a problem.

Right, we already have a few benches that are in ns so I'm familiar with the drill. :)

TTWNO:
All good by me. Thanks!

PR #1310, opened 2025-03-20
:sparkles: Adds smartstring features and Type impl

TTWNO:
- Adds new feature `smartstring`
- This feature gives a `Type` impl for `smartstring::SmartString`,
- and imports `smartstring` with the `serde` feature flag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -555,7 +555,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "33d852cb9b869c2a9b3df2f71a3074817f01e1844f839a144f5fcef059a4eb5d"
 dependencies = [
  "libc",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -1414,7 +1414,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.9.2",
- "windows-sys 0.52.0",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


zeenix:
Thanks but I don't want to add (even if optional) dep on a crate that's [most likely unmaintained](https://github.com/bodil/smartstring/issues/52).

Besides, I once did some benchmarking to see if using smartstring (or another crate that offers the same functionality) and found no evidence of it helping in performance. IMO it's better to either do allocation or not, and I already added support for `heapless` for that reason.

TTWNO:
> Thanks but I don't want to add (even if optional) dep on a crate that's https://github.com/bodil/smartstring/issues/52.

Didn't realize the status. That's unfortunate.

> Besides, I once did some benchmarking to see if using smartstring (or another crate that offers the same functionality) and found no evidence of it helping in performance.

We use it in Odilia as a hashable string. Since most (but not all) bus names are short, oue cache daw significant kmprovements.

Maybe there is a better way to write the cache.... and that wouldn't require adding this featyre to zbus.

> IMO it's better to either do allocation or not, and I already added support for heapless for that reason.

I wish DBus had a maximum name size... then I could use heapless.

Anyways, I totally understand.

zeenix:
> > IMO it's better to either do allocation or not, and I already added support for heapless for that reason.
> 
> I wish DBus had a maximum name size...

That wouldn't make sense to have a limit on all strings. You can always specify a limit for strings in your own interfaces.

 then I could use heapless.

I think you can make an educated guess in many cases. The great thing is that deserialization will not panic but rather error out if you go beyond the capacity. If the size limit of the string in question is documented, this will be a bug on the sender side. If not, you can always bump up the size.

Just some random thoughts. YMMV. :)

zeenix:
> > > IMO it's better to either do allocation or not, and I already added support for heapless for that reason.
> > 
> > 
> > I wish DBus had a maximum name size...
> 
> That wouldn't make sense to have a limit on all strings.

That said, D-Bus does impose limits where it makes sense. E.g a signature string can not be more than 255 characters.

PR #1319, opened 2025-03-30
Add security documentation and warnings to system methods

### LINKED ISSUE #1275: Consider warning about trusting Address::system
Good afternoon,

Today `zbus::Address` has a convenient `system()` function on it which attempts to resolve where the socket for system-level services on the current platform/host are. However, this is primarily reliant on reading an environment variable:
https://github.com/dbus2/zbus/blob/0b3b6b190080bcab5be2eae795f9791d73a013bd/zbus/src/address/mod.rs#L90-L91

As you probably know, environment variables can be manipulated in a lot of ways by software running even with same user privileges and attempting to start other apps in weird ways. For example this resulted in [CVE-2012-3524 for `libdbus`](https://nvd.nist.gov/vuln/detail/CVE-2012-3524) back in 2012. This case was primarily about attacking `setuid` programs but the general concern of handling untrusted data in sensitive contexts remains.

I would primarily recommend adding an obvious warning to the documentation of the aforementioned function, but also all of the other public methods that return something from the "system" to note that `zbus` doesn't do any kind of security checks on the returned connection, and that individual users are responsible for validating if it is actually owned by the "system" or not. As a secondary recommendation, you could also detect specifically running in a "privileged" binary at runtime and use a well-known hardcoded path instead.

complexspaces:
This PR closes #1275 by adding basic and sufficient, but attention-grabbing, documentation to all `system*` methods in the crate which either are or utilize the result of `zbus::Address::system` to describe their security properties.

I took the approach to centralize the details/important parts on `zbus::Address::system` and then user interdoc links to redirect all of the other "affected" callsites there to reduce duplication. 

Here are some examples of how `rustdoc` renders it when built locally:
image

image

When viewing the documentation through an editor (in my case VSCode), the warning rendering isn't shown (its a [rustdoc specific markdown extension](https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block) but the `Security` header is nicely large:
image


zeenix (changes requested):
Nice! Could you please just add the emoji prefix to the commit and PR titles, as per the contribution guidelines? To help you avoid having to copy&paste it from gitmoji.dev, I'll do that for the PR title and you can just c&p it from there on the git message.

complexspaces:
I've updated the commit's information. Sorry for missing that the first time around.

zbus/src/address/mod.rs
```diff
@@ -87,6 +87,17 @@ impl Address {
     /// Get the address for the system bus respecting the `DBUS_SYSTEM_BUS_ADDRESS` environment
     /// variable. If we don't recognize the value (or it's not set) we fall back to
     /// `/var/run/dbus/system_bus_socket`.
+    ///
+    /// # Security
+    ///
+    /// <div class="warning">
+    ///
+    /// `zbus` does not perform any security checks or trust validation on the address (path)
+    /// returned by this method.
```
zeenix, line 96: sorry but I only noticed now: This method does not return an address. We need to make it more specific to the problematic bit: 

```suggestion
    /// Note that `zbus` does not (and can not) validate the value of the `DBUS_SYSTEM_BUS_ADDRESS`
    /// environment variable before using it. Users in security-sensitive contexts (ex: authentication)
    /// are responsible for ensuring that the value of this environment variable can be trusted.
```

complexspaces: FWIW I don't agree with the exact wording here. I think its better for users to validate the value the function returns and not its inputs independently.

zeenix: Then we disagree on what's expected of the user. :laughing: Perhaps I'm missing something but:

* how does the user verify the connection is valid? I'd imagine it'd be easier to ensure that the env is correct/clean. I think this is what was suggested in [the summary of the CVE you pointed to](https://nvd.nist.gov/vuln/detail/CVE-2012-3524):
>  libdbus maintainers state that this is a vulnerability in the applications that do not cleanse environment variables
* the damage could already be done by the time user has a connection instance, since the connection creation involves a handshake and "Hello" message being sent out to the peer.

zeenix: @complexspaces are you convinced by my arguments? If not, could you please reply? I understand if you don't want to argue about this anymore but in that case, kindly accept my proposed wording so we can move on here. :pray:

complexspaces: Hey again, sorry, I ended up at a conference that took most of my time and attention.

> how does the user verify the connection is valid? I'd imagine it'd be easier to ensure that the env is correct/clean. I think this is what was suggested in...

My argument here is that I'm not usually a fan of recommending users inspect all of the inputs to a function because its more fragile to drift over time or missing a possible branch. Instead, its easier to create an `Address::system` and analyze that as a "final piece" before creating a connection from it.

> the damage could already be done by the time user has a connection instance, since the connection creation involves a handshake and "Hello" message being sent out to the peer.

Even if you did `Connection::system` instead of `Address::system` I don't follow how a handshake "hello" message would be a security risk when using a memory-safe language like Rust.

zeenix: > Hey again, sorry, I ended up at a conference that took most of my time and attention.

No worries.

> > how does the user verify the connection is valid? I'd imagine it'd be easier to ensure that the env is correct/clean. I think this is what was suggested in...
> 
> My argument here is that I'm not usually a fan of recommending users inspect all of the inputs to a function because its more fragile to drift over time or missing a possible branch.

* Sorry but tbh that sounds a bit vague. How does that all translate to this specific case? :thinking: 
* If it makes you feel better, tehcnically there is no input to the function in question. This is about the environment that the function makes an assumption about.

> Instead, its easier to create an `Address::system` and analyze that as a "final piece" before creating a connection from it.

My main question here is how it is even possible to verify that the connection that you already created is pointing to what you intended it to point to? Perhaps it is very much possible and it's very easy (like you wrote here) but then you need to educate me a bit. :)

complexspaces: > Sorry but tbh that sounds a bit vague. How does that all translate to this specific case? 🤔

Essentially I don't think its 100% reliable to just check the environment variable that `zbus` does because either:
- `zbus` _could_ add more logic in the `system` function that needs verified additionally by the end user.
- The default/fallback path should probably be verified as well for completeness.

> My main question here is how it is even possible to verify that the connection that you already created is pointing to what you intended it to point to?

Oh, I see what you are saying now, sorry. AFAIK you _need_ to go `Address` -> `ConnectionBuilder` manually to verify everything right. I could update the documentation to point at that path more strongly.

complexspaces: > However, I'm still wondering if it's even possible to verify the address? If that is easily possible (I understood that's what you claimed in your previous comment?) then we should just do it ourselves, w/o even involving the user/programmer.

I mentioned this in my email previously, but there are some strategies 1P has adopted for path verification but its not anything confident enough in (when trying to analyze the entirety of how Linux filesystem permissions can work) that I would want to put it in `zbus` as a layer of protection for users. We are fine owning the risk as a company but I think its too whack-a-mole for FOSS 😅.

I think if people are doing security sensitive things with `zbus`, then they should be responsible for making this set of risk acceptances and functionality tradeoffs instead of making you responsible for upkeeping complicated security checks forever.

zeenix: > I think if people are doing security sensitive things with `zbus`, then they should be responsible for making this set of risk acceptances and functionality tradeoffs instead of making you responsible for upkeeping complicated security checks forever.

I agree but I feel like this is a lot more inline with my suggestion and the libdbus maintainer's conclusion. Having said that, I'm fine with telling people to verify the address (instead of ensuring their env is clean in general) if we can give them some clue how to achieve that (some link to documentation perhaps?).

complexspaces: Do you have a rough idea what documentation you'd like to see there? I'm not opposed to your idea of doing both here.

zeenix: My idea was simply that we tell people to ensure their environment is clean and why so.

zeenix:
This doesn't seem to be going anywhere anymore.

PR #1320, opened 2025-03-30
Draft: Fix #1317

### LINKED ISSUE #1317: Setting `reply_serial`
Hello! I'm trying to do some low level proxying and I want to rewrite the serial numbers.

I can modify serial in the primary header, but all access to `fields` including `reply_serial` appears to be private.  Could that be opened up?

(Also, semi-related, to generate unique serial numbers I'm doing `Header::new().serial_num()` then discarding the header, which seems a bit like a waste - could a public method for generating serial numbers for low level usage also be exposed? Happy to make a new issue for that)

Edit:

Okay, sorry, I was playing with it more and I think in order for mutating the header to be useful there'd need to be some more infrastructure.  Right now, even if you have a header you can't turn it into a message, so a `Message` constructor that takes a header and a body, or maybe a `Builder` constructor that takes a header, or something like that would be needed.

A number of other libraries have `into_parts()` and `from_parts()` methods for mutation like this (ex: hyper), maybe something could be done in that mold?

Just some more notes if it helps:
- `build_from_raw_body` needs the signature, but the builder already has a signature, and ownership means you have to clone the signature to work around this
- There's no way AFAICT to take ownership of the fds in the original message, and `build_from_raw_body` needs owned fds, not borrowed

andrewbaxter:


zbus/src/message/fields.rs
```diff
@@ -14,7 +14,7 @@ use crate::message::{FieldCode, Header, Message};
 /// [`Field`]: enum.Field.html
 #[derive(Debug, Default, Clone, Type)]
 #[zvariant(signature = "a(yv)")]
-pub(crate) struct Fields<'f> {
+pub struct Fields<'f> {
```
zeenix, line 17: This used to be public but since it's very low-level API, I was glad when I was able to hide it. As I wrote on #1317, I'll need a compelling use case for this.

PR #1321, opened 2025-03-31
Support ProcessFD in fdo::DBus::get_connection_credentials'

swick:
Add support for the ProcessFD key in `org.freedesktop.DBus.GetConnectionCredentials` to zbus (and thus to busd) by adding the socket pidfd to `ConnectionCredentials` in `socket::unix::get_unix_peer_creds_blocking`.

### DIFF
--- a/zbus/src/connection/socket/unix.rs
+++ b/zbus/src/connection/socket/unix.rs
@@ -331,18 +331,41 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
     // TODO: get this BorrowedFd directly from get_unix_peer_creds(), but this requires a
     // 'static lifetime due to the Task.
     let fd = unsafe { BorrowedFd::borrow_raw(fd) };
+    let mut creds = crate::fdo::ConnectionCredentials::default();
 
     #[cfg(any(target_os = "android", target_os = "linux"))]
     {
-        use nix::sys::socket::{getsockopt, sockopt::PeerCredentials};
+        use nix::{
+            sys::socket::{getsockopt, sockopt::PeerCredentials},
+            unistd::{getgrouplist, Gid, Uid, User},
+        };
+        use std::ffi::CString;
 
-        getsockopt(&fd, PeerCredentials)
-            .map(|creds| {
-                crate::fdo::ConnectionCredentials::default()
-                    .set_process_id(creds.pid() as _)
-                    .set_unix_user_id(creds.uid())
-            })
-            .map_err(|e| e.into())
+        let (uid, gid, pid) = {
+            let unix_creds = getsockopt(&fd, PeerCredentials)?;
+            (
+                Uid::from_raw(unix_creds.uid()),
+                Gid::from_raw(unix_creds.uid()),
+                unix_creds.pid() as u32,
+            )
+        };
+        creds = creds.set_unix_user_id(uid.as_raw()).set_process_id(pid);
+
+        // the dbus spec requires groups to be either absent or complete (primary + secondary
+        // groups)
+        let mut groups = User::from_uid(uid)?
+            .map(|user| CString::new(user.name))
+            .transpose()?
+            .map(|user| getgrouplist(&user, gid))
+            .transpose()
+            .ok()
+            .flatten()
+            .unwrap_or(Vec::new());
+        // it also requires the groups to be numerically sorted
+        groups.sort_by_key(|gid| gid.as_raw());
+        for group in groups.iter() {
+            creds = creds.add_unix_group_id(group.as_raw());
+        }
     }
 
     #[cfg(any(
@@ -354,10 +377,13 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
         target_os = "netbsd"
     ))]
     {
-        let uid = nix::unistd::getpeereid(fd).map(|(uid, _)| uid.into())?;
-        // FIXME: Handle pid fetching too.
-        Ok(crate::fdo::ConnectionCredentials::default().set_unix_user_id(uid))
+        let (uid, _gid) = nix::unistd::getpeereid(fd)?;
+        creds = creds.set_unix_user_id(uid.as_raw())
+
+        // FIXME: Handle pid fetching too
     }
+
+    Ok(creds)
 }
 
 // Send 0 byte as a separate SCM_CREDS message.
--- a/zbus/src/connection/socket/unix.rs
+++ b/zbus/src/connection/socket/unix.rs
@@ -336,10 +336,15 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
     #[cfg(any(target_os = "android", target_os = "linux"))]
     {
         use nix::{
-            sys::socket::{getsockopt, sockopt::PeerCredentials},
+            errno::Errno,
+            sys::socket::{
+                getsockopt,
+                sockopt::{PeerCredentials, PeerPidfd},
+            },
             unistd::{getgrouplist, Gid, Uid, User},
         };
         use std::ffi::CString;
+        use zvariant::OwnedFd;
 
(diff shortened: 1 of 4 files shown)


zeenix:
Nice!

> Add support for the ProcessFD key in `org.freedesktop.DBus.GetConnectionCredentials` to busd

You're adding it to zbus (busd is just a beneficiary). ;)

zbus/src/connection/socket/unix.rs
```diff
@@ -102,7 +102,7 @@ impl super::WriteHalf for Arc<Async<UnixStream>> {
     }
 
     async fn peer_credentials(&mut self) -> io::Result<crate::fdo::ConnectionCredentials> {
-        get_unix_peer_creds(self).await
+        super::ReadHalf::peer_credentials(self).await
```
zeenix, line 105: Thanks for splitting this into its own commit. Could you please also add the rationale/short description in the commit for this change?

zbus/src/connection/socket/unix.rs
```diff
@@ -353,10 +360,12 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
         target_os = "netbsd"
     ))]
     {
-        let uid = nix::unistd::getpeereid(fd).map(|(uid, _)| uid.into())?;
-        // FIXME: Handle pid fetching too.
```
zeenix, line 357: did this comment get accidentally deleted? At least it's not being addressed in this commit.

zbus/src/connection/socket/unix.rs
```diff
@@ -348,7 +349,21 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
...
 
+        let mut groups = User::from_uid(uid)?
+            .map(|user| CString::new(user.name))
+            .transpose()?
+            .map(|user| getgrouplist(&user, gid))
+            .transpose()
+            .ok()
+            .flatten()
+            .unwrap_or(Vec::new());
+        groups.sort_by_key(|gid| gid.as_raw());
```
zeenix, line 365: Not sure we want to do any sorting here unless the spec recommends or requires it.

zbus/src/connection/socket/unix.rs
```diff
@@ -348,7 +349,21 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
...
+
         creds = creds.set_unix_user_id(uid.as_raw()).set_process_id(pid);
+
+        for group in groups.iter() {
+            creds = creds.add_unix_group_id(group.as_raw());
+        }
```
zeenix, line 377: probably more efficient to just do this as part of the chain to fetch the groups: `.for_reach(|g| creds.add_unix_group_id(g))` instead of creating a temporary array first. That is, if sorting is not needed.

Cargo.toml
```diff
@@ -97,7 +97,7 @@ static_assertions = "1.1.0"
 async-recursion = "1.1.1"
 winnow = "0.7"
 uds_windows = "1.1.0"
-nix = { version = "0.29", default-features = false, features = [
+nix = { version = "0.29", git = "https://github.com/swick/nix.git", branch = "wip/sopeerpidfd", default-features = false, features = [
```
zeenix, line 100: This will be a version bump after your PR is merged (and hopefully released, although I can be easily bribe..err.. convinced to use git main) so best to split this into a separate commit.

zbus/src/connection/socket/unix.rs
```diff
@@ -359,11 +364,21 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
...
+            Err(e) => return Err(e.into()),
+        };
+
         creds = creds.set_unix_user_id(uid.as_raw()).set_process_id(pid);
 
         for group in groups.iter() {
             creds = creds.add_unix_group_id(group.as_raw());
         }
+
+        if let Some(pidfd) = pidfd {
```
zeenix, line 379: same comment here as for groups: best to just do this directly (in the `match` arm above).

zeenix (changes requested):
LGTM otherwise. Extra points for going the extra mile and adding groups to credentials and the missing nix API. :+1:

zbus/src/connection/socket/unix.rs
swick: The spec requires it. Added a comment.

Cargo.toml
swick: I intended for this to get removed before the PR is merged. Moved to a separate commit anyway.

zbus/src/connection/socket/unix.rs
swick: It did. Added it again.

zbus/tests/basic.rs
```diff
@@ -340,21 +338,16 @@ async fn test_freedesktop_credentials() -> Result<()> {
...
-                .into_iter()
-                .last()
-                .unwrap()
-                .parse()
+            let pidfd = fd.as_raw_fd();
+            let fdinfo = read_to_string(format!("/proc/self/fdinfo/{pidfd}"))
+                .await
                 .unwrap();
+            let pidline = fdinfo.split('\n').find(|s| s.starts_with("Pid:")).unwrap();
+            let pid = pidline.split('\t').next_back().unwrap().parse().unwrap();
```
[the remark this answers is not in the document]
swick: I thought this used std::fs before, but it didn't. Split into two commits, one for the use declarations, one for the chained calls.

zbus/src/connection/socket/unix.rs
swick: Had that before, but then some state was set in there, other was required later in the code, so I split it into fetching the state and applying it and it seemed better to me. I can do it if you feel strongly about it.

zbus/src/connection/socket/unix.rs
zeenix: wasn't resolved.

zbus/src/connection/socket/unix.rs
zeenix: If there is no reason to do it in two phases, I think we shouldn't. The less logic and state, the easier to read. Although I think it should be all done here since pidfd isn't more important than uid IMO.

zeenix (changes requested):
Getting very close to ready (except for the nix part that we'll need to wait on).

zbus/src/connection/socket/unix.rs
swick: Oops. Should be now.

zbus/src/connection/socket/unix.rs
swick: Changed it to mutate creds as early as possible.

swick:
Added one more cleanup commit.

zeenix (approved):
LGTM. Now it's about nix changes.

swick:
The nix PR got merged but it's not in a release yet. How should we proceed?

zeenix:
> The nix PR got merged

Awesome. :)

> but it's not in a release yet. How should we proceed?

I guess either wait for a release or request them to make one? :)

zeenix:
> The nix PR got merged but it's not in a release yet. How should we proceed?

nix 0.30 was released and now updated in both zbus and busd. I believe your path is clear now. :)

swick:
Rebased, depending on upstream nix 0.30 now.

zeenix:
@swick Thanks. Sorry for the delay.

swick:
Thank you!

Biswa96:
This caused build error in Android platform. Please see the full error output in this issue https://github.com/termux/termux-packages/issues/24733 (click on the Show log line).

zeenix:
> This caused build error in Android platform. Please see the full error output in this issue https://github.com/termux/termux-packages/issues/24733 (click on the Show log line).

Sorry about that. I didn't even know people use zbus on Android. 

I think it's likely just a missing cfg.

zeenix:
@Biswa96 #1381

PR #1322, opened 2025-04-01
✨ zb: Expose low-level serial fields in builder

### LINKED ISSUE #1317: Setting `reply_serial`

andrewbaxter:
Allow overriding serial/reply_serial in the message builder. Per the linked issue this is to support more low-level zbus tools, like a bus-to-bus grafter.

Fixes #1317.

### DIFF
--- a/zbus/src/message/builder.rs
+++ b/zbus/src/message/builder.rs
@@ -1,6 +1,7 @@
 use std::{
     borrow::Cow,
     io::{Cursor, Write},
+    num::NonZeroU32,
     sync::Arc,
 };
 #[cfg(unix)]
@@ -118,6 +119,20 @@ impl<'a> Builder<'a> {
     {
         self.header.fields_mut().destination = Some(destination.try_into().map_err(Into::into)?);
         Ok(self)
+    }
+
+    /// Override the generated or inherited serial.  This is a low level modification,
+    /// generally you should not need to use this.
+    pub fn serial(mut self, serial: NonZeroU32) -> Self {
+        self.header.primary_mut().set_serial_num(serial);
+        self
+    }
+
+    /// Override the reply serial. This is a low level modification, generally you should use
+    /// `Message::method_return` instead.
+    pub fn reply_serial(mut self, serial: Option<NonZeroU32>) -> Self {
+        self.header.fields_mut().reply_serial = serial;
+        self
     }
 
     pub(super) fn reply_to(mut self, reply_to: &Header<'_>) -> Result<Self> {


zeenix:
Nice!

> This has two main changes:

Before I do an actual review, if there are two logical changes, then they should be split into two separate commits, as per our contributing guidelines.

andrewbaxter:
I split the commit.

zeenix:
> 👍 I split the commit.

Great. I'll have a look very soon. I've been a bit too busy lately, sorry.

zeenix:
@andrewbaxter one of the clippy warnings here was from the main branch and will be fixed by https://github.com/dbus2/zbus/pull/1324 . Once that's merged, please rebase on top of that and fix the lint issues introduced by this PR. :pray:

andrewbaxter:
Sure thing, just did that now. No rush on the review!

zeenix:
> Sure thing, just did that now. Thanks, and no rush on the review!

Thanks for your understanding. Btw, please use the emoji characters themselves, not their GitHub code so local git tools also show them as emojis. 😊

zbus/src/message/builder.rs
```diff
@@ -120,6 +121,21 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Override the generated or inherited serial.  This is a low level modification,
+    /// generally you should not need to use this. If you do use this, be sure to generate a
+    /// serial with `take_serial()` to avoid conflicts with other messages.
```
zeenix, line 126:
```suggestion
    /// Override the generated or inherited serial.  This is a low level modification.
    /// Generally you should not need to use this.
```

zbus/src/message/header.rs
```diff
@@ -314,6 +309,15 @@ impl<'m> Header<'m> {
 
 static SERIAL_NUM: AtomicU32 = AtomicU32::new(0);
 
+/// Get a new guaranteed-unique serial.
+pub fn take_serial() -> NonZeroU32 {
```
zeenix, line 313: I don't see a reason to expose this. If the user wants control of the serial numbers, then they are very much on their own for it.

zeenix (changes requested):
The rest looks good to me.

zbus/src/message/header.rs
andrewbaxter: In order to guarantee non-collision both zbus and the user need to use serials from the same pool.

zeenix: > In order to guarantee non-collision both zbus and the user need to use serials from the same pool.

So you want to generate your own serials on the same connection while zbus also generates them for you? Why? :thinking:

zeenix: I mean, why not always generate them for all the messages for the connection in question? I'm not sure I want to support this. It will still be on the user to ensure that the serial numbers are unique. I'd rather they either let zbus handle all serial numbers or they do it, not both.

andrewbaxter: Zbus always generates messages... during connection setup, etc.  Is there a reason you don't want to expose this?

zeenix: > Zbus always generates messages... during connection setup, etc.

There is only 1 message during the handshake and you can safely assume that to be `1` (or you can start from say `50` to be sure there is no conflict with that) but even if you had a conflict with the handshake `Hello` message, it's not a problem (or at least it should not be) since that method has already been replied to and the spec doesn't disallow re-use of serial numbers.

> Is there a reason you don't want to expose this?

It's a good practice to keep the API surface small for multiple reasons so new API should ideally be justified with compelling use cases, more so if it's low-level API most people would never need.

andrewbaxter: > for multiple reasons

Which reasons?

andrewbaxter: Done.

zeenix: > Which reasons? 

More maintainance burdon (the public API is then a contract between the maintainer and the user and can't be broken w/o major version bumps), reduced security because of larger attack surface, more API resulting in users having a harder time finding the API they need (i-e worse user experience) etc etc.

> Done

Thank you.

zeenix (approved):
I took the liberty of updating the PR title & description. I hope that's OK. While doing that, I also noticed that the commit is missing the subcrate prefix. That's no biggie but just mentioning it for future contributions.

zbus/src/message/header.rs
zeenix: > but are all implementations fine with it?

The only implementation that had a problem with it, was xdg-dbus-proxy and that was a considered a bug in the proxy and fixed several months ago. Although I'm not sure if the reuse would have been problematic for the buggy versions even, they had problem with serial numbers not being monotonically increasing.

The important bit is that there is nothing in the spec that would even imply that the serial numbers can't be reused. How would it even work? It's a 32-bit number and will have to be wrapped around at some point.

andrewbaxter:
Thanks!

PR #1337, opened 2025-04-22
✨ zd: Make rename & rename_all work on Value/OwnedValue

### LINKED ISSUE #320: Value/OwnedValue macro doesn't support renaming fields
In GitLab by @wynprice999 on Mar 1, 2023, 22:01

It seems like the field names used for deserialization can't be renamed. I would expect using `#[zvariant(rename = "...")]` would rename the field, but it doesn't appear to have an effect:

<details>

<summary>Example expansion</summary>

```rust
#[derive(DeserializeDict, Value, OwnedValue)]
#[zvariant(signature = "dict")]
pub struct TrackMetadata {
    #[zvariant(rename = "otherfield")]
    pub field1: String,

    pub field2: String
}
```

Expands to the following:
```rs
#[zvariant(signature = "dict")]
pub struct TrackMetadata {
    #[zvariant(rename = "otherfield")]
    pub field1: String,
    pub field2: String,
}
...
impl ::std::convert::TryFrom<::zbus::zvariant::OwnedValue> for TrackMetadata {
    type Error = ::zbus::zvariant::Error;
    #[inline]
    fn try_from(
        value: ::zbus::zvariant::OwnedValue,
    ) -> ::zbus::zvariant::Result<Self> {
        let mut fields = <::std::collections::HashMap<
            ::std::string::String,
            ::zbus::zvariant::Value,
        >>::try_from(value)?;
        ::std::result::Result::Ok(Self {
            field1: fields
                .remove("field1")
                .ok_or_else(|| ::zbus::zvariant::Error::IncorrectType)?
                .downcast()
                .ok_or_else(|| ::zbus::zvariant::Error::IncorrectType)?,
            field2: fields
                .remove("field2")
                .ok_or_else(|| ::zbus::zvariant::Error::IncorrectType)?
                .downcast()
                .ok_or_else(|| ::zbus::zvariant::Error::IncorrectType)?,
        })
    }
}
impl From<TrackMetadata> for ::zbus::zvariant::OwnedValue {
    #[inline]
    fn from(s: TrackMetadata) -> Self {
        let mut fields = ::std::collections::HashMap::new();
        fields.insert("field1", ::zbus::zvariant::Value::from(s.field1));
        fields.insert("field2", ::zbus::zvariant::Value::from(s.field2));
        ::zbus::zvariant::Value::from(fields).into()
    }
}
```

</details>

From what I can tell, a solution would involve the following line: https://gitlab.freedesktop.org/dbus/zbus/-/blob/main/zvariant_derive/src/value.rs#L111

Lyndeno:
Previously, setting these values would not apply to implementations of From and TryFrom when converting between a type and Value/OwnedValue. This change makes it so using rename or rename_all will apply to these traits.

Closes #320

### DIFF
--- a/zvariant_derive/src/lib.rs
+++ b/zvariant_derive/src/lib.rs
@@ -424,6 +424,54 @@ pub fn deserialize_dict_macro_derive(input: TokenStream) -> TokenStream {
 /// assert_eq!(e, StrEnum::Variant2);
 /// ```
 ///
+/// # Renaming fields
+///
+/// ## Auto Renaming
+///
+/// The macro supports specifying a Serde-like `#[zvariant(rename_all = "case")]` attribute on
+/// structures. The attribute allows to rename all the fields from snake case to another case
+/// automatically.
+///
+/// Currently the macro supports the following values for `case`:
+///
+/// * `"lowercase"`
+/// * `"UPPERCASE"`
+/// * `"PascalCase"`
+/// * `"camelCase"`
+/// * `"snake_case"`
+/// * `"kebab-case"`
+///
+/// ## Individual Fields
+///
+/// It's still possible to specify custom names for individual fields using the
+/// `#[zvariant(rename = "another-name")]` attribute even when the `rename_all` attribute is
+/// present.
+///
+/// Here is an example using both `rename` and `rename_all`:
+///
+/// ```
+/// # use zvariant::{OwnedValue, Value, Dict};
+/// # use std::collections::HashMap;
+/// #
+/// #[derive(Clone, Value, OwnedValue)]
+/// #[zvariant(signature = "dict", rename_all = "PascalCase")]
+/// struct RenamedStruct {
+///     #[zvariant(rename = "MyValue")]
+///     field1: String,
+///     field2: String,
+/// }
+///
+/// let s = RenamedStruct {
+///     field1: String::from("hello"),
+///     field2: String::from("world")
+/// };
+/// let v = Value::from(s);
+/// let d = Dict::try_from(v).unwrap();
+/// let hm: HashMap<String, String> = HashMap::try_from(d).unwrap();
+/// assert_eq!(hm.get("MyValue").unwrap().as_str(), "hello");
+/// assert_eq!(hm.get("Field2").unwrap().as_str(), "world");
+/// ```
+///
 /// # Dictionary encoding
 ///
 /// For treating your type as a dictionary, you can use the `signature = "dict"` attribute. See
--- a/zvariant_derive/src/value.rs
+++ b/zvariant_derive/src/value.rs
@@ -19,7 +19,11 @@ pub fn expand_derive(ast: DeriveInput, value_type: ValueType) -> Result<TokenStr
     match &ast.data {
         Data::Struct(ds) => match &ds.fields {
             Fields::Named(_) | Fields::Unnamed(_) => {
-                let StructAttributes { signature, .. } = StructAttributes::parse(&ast.attrs)?;
+                let StructAttributes {
+                    signature,
+                    rename_all,
+                    ..
+                } = StructAttributes::parse(&ast.attrs)?;
                 let signature = signature.map(|signature| match signature.as_str() {
                     "dict" => "a{sv}".to_string(),
                     _ => signature,
@@ -32,6 +36,7 @@ pub fn expand_derive(ast: DeriveInput, value_type: ValueType) -> Result<TokenStr
                     &ds.fields,
                     signature,
                     &zv,
+                    rename_all,
                 )
             }
             Fields::Unit => Err(Error::new(ast.span(), "Unit structures not supported")),
@@ -51,6 +56,7 @@ fn impl_struct(
     fields: &Fields,
     signature: Option<String>,
     zv: &TokenStream,
+    rename_all: Option<String>,
 ) -> Result<TokenStream, Error> {
     let statc_lifetime = LifetimeParam::new(Lifetime::new("'static", Span::call_site()));
     let (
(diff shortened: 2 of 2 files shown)


zvariant_derive/src/dict.rs
```diff
@@ -5,7 +5,7 @@ use zvariant_utils::macros;
 
 use crate::utils::*;
 
-fn dict_name_for_field(
+pub(crate) fn dict_name_for_field(
```
zeenix, line 8: It's smart using the existing code we're not using it for a dict, so either this should be moved to utils and renamed accordingly or we shouldn't touch this and use `rename_identifier` instead.

zvariant_derive/src/value.rs
```diff
@@ -130,7 +136,12 @@ fn impl_struct(
                     let (fields_init, entries_init): (TokenStream, TokenStream) = fields
                         .iter()
                         .map(|field| {
+                            let FieldAttributes { rename } =
+                                FieldAttributes::parse(&field.attrs).unwrap_or_default();
                             let field_name = field.ident.to_token_stream();
+                            let dict_name =
```
zeenix, line 142: This needs a better name too. :) shadowing `field_name` would be fine if it's not used again later.

zeenix (changes requested):
Thanks so much for this contribution. :+1: Also, kudos for following our contributors guidelines so well (it's very rare for first time contributors to do that from the beginning :smile:). Some nitpicks but overall looks good. Apart from the line comments:

* Document/list the new attributes in the macros' docs please.
* It would be good to test this. Best to just modify the doc example(s) to achieve that.
* I suggest using `zd` abbreviation for commit prefix but up to you.

zvariant_derive/src/dict.rs
Lyndeno: Switched to rename identifier

zvariant_derive/src/value.rs
Lyndeno: Changed it to key name as it is for reference the keys in the dict.

Lyndeno:
Thank you for your review. I have applied your line suggestions and will get to implementing some tests for this change.

Lyndeno:
I have implemented your suggestions, please let me know what you think.

zvariant_derive/src/lib.rs
```diff
@@ -424,6 +424,48 @@ pub fn deserialize_dict_macro_derive(input: TokenStream) -> TokenStream {
 /// assert_eq!(e, StrEnum::Variant2);
 /// ```
 ///
+///# Auto renaming fields
+///
+/// The macro supports specifying a Serde-like `#[zvariant(rename_all = "case")]` attribute on
+/// structures. The attribute allows to rename all the fields from snake case to another case
```
zeenix, line 430: Really nice job with documentation! So I hate to ask for any more changes but I'm a perfectionist. :see_no_evil: It looks like we don't already document the renaming of fields so probably best to generalize the header and text here to introduce to both kinds of renaming?

zeenix (changes requested):
LGTM otherwise. While you fix that, perhaps 2 other minor changes too:

* this is an enhancement, not a bug fix (so ✨ would be better). 
* this fixes #320, so please add `Fixes #320.` line to the commit message and the PR description. :pray: 

Once these minor things are addressed, I believe we can merge. :+1:

zvariant_derive/src/lib.rs
Lyndeno: I just copied this from the `SerializeDict` docs. I'll definitely refine this to make sure both attributes are well explained.

zvariant_derive/src/lib.rs
```diff
@@ -424,6 +424,54 @@ pub fn deserialize_dict_macro_derive(input: TokenStream) -> TokenStream {
...
+/// * `"snake_case"`
+/// * `"kebab-case"`
+///
+/// ## Individual Fields
+///
+/// It's still possible to specify custom names for individual fields using the
+/// `#[zvariant(rename = "another-name")]` attribute even when the `rename_all` attribute is
+/// present.
+///
+/// Using `rename` and `rename_all`:
```
zeenix, line 450:
```suggestion
/// Here is an example using both `rename` and `rename_all` attributes:
```

zeenix (changes requested):
LGTM otherwise. We're almost there.

zvariant_derive/src/lib.rs
Lyndeno: Done

zeenix (approved):
Thanks!

zeenix:
@Lyndeno btw, does these actually do anything? :thinking: The D-Bus (and GVariant) structs are anonymous so the field names are not reflected in the encoding. In case of *Dict macros, we encode the struct as a map, so the field names are important. However, when encoding the value as a struct of strust inside a variant, I'm not sure it actually does anything meaningful. 

or am I missing something? :thinking: I do apologise for not thinking about this most important aspect of this PR until before hitting the "enable auto-merge" button. :facepalm:

Lyndeno:
I am working on a new [project](https://github.com/Lyndeno/ppd-rs/blob/master/src/main.rs):

In it I have this struct:
```rust
#[derive(SerializeDict, DeserializeDict, Debug, Type, OwnedValue, Value)]
#[zvariant(signature = "dict", rename_all = "PascalCase")]
struct Profile {
    profile: String,
    driver: String,
    platform_driver: Option<String>,
    cpu_driver: String,
}
```

Using this proxy method to get it:
```rust
#[zbus(property)]
fn profiles(&self) -> Result<Vec<Profile>>;
```

The property is defined in this [project](https://gitlab.freedesktop.org/upower/power-profiles-daemon/-/blob/ea0d7504a8b9ec4378cd17db3386efc761195dae/src/org.freedesktop.UPower.PowerProfiles.xml.in#L130):
```xml
<!--
        Profiles:

        An array of key-pair values representing each profile. The key named
        "Driver" (s) identifies the power-profiles-daemon backend code used to
        implement the profile.

        The key named "Profile" (s) will be one of:
        - "power-saver" (battery saving profile)
        - "balanced" (the default  profile)
        - "performance" (a profile that does not care about noise or battery consumption)

        Only one of each type of profile will be listed, with the daemon choosing the
        more appropriate "driver" for each profile type.

        This list is guaranteed to be sorted in the same order that the profiles
        are listed above.
-->
<property name="Profiles" type="aa{sv}" access="read"/>
```

Without this patch, my code does not work unless I make the struct member names PascalCase directly. With this patch, the macro works. So for my use-case, it **does do** something.

> However, when encoding the value as a struct of strust inside a variant, I'm not sure it actually does anything meaningful.

I do not think I know enough about dbus and this project to know what this means.

Lyndeno:
Thank you for your speedy review

Lyndeno:
Do we want to make this work is we use
```rust
#[serde(rename_all = "...")]
```

Right now it only works if we use:
```rust
#[zvariant(rename_all = "...")]
```

PR #1339, opened 2025-04-24
✨ zb,zn,zv: Implement Borrow for Owned* types

refi64:
As-is, using the `Owned*` types as a key is rather limited, because we can't index into e.g. a `HashMap<OwnedInterfaceName>` with a normal `InterfaceName<'_>`. In order to support this, the types need to implement `Borrow`, which is more-or-less designed for this type of scenario anyway. (That does come with some stringent requirements, i.e. traits implemented on the owned type need to have identical behavior to the borrowed type, but all of these types are already rather thin wrappers to begin with.)


refi64:
pushed up a change to make this more generalized (no need to limit the target lifetime to `'static`), and also fix up the commit message because it was a bit wonky before and slightly incorrect (some, but not all, of the types already implement `Borrow<str>` which could be used for key lookups).

zeenix (approved):
Very nice! I love it that this is the second PR from you that I can just approve and merge. Perfection!

PR #1340, opened 2025-04-24
✨ zb: Make OwnerChangedStream ordered

refi64:
NameOwnerChangedStream is *already* ordered, so switching out the .map implementation and implementing OrderedStream is all that's needed.


zeenix (approved):
Nice! Thanks so much. BTW, I noticed that we don't have any test cases for ensuring these streams actually **are** ordered. :( @danieldg perhaps you'd be interested in adding that since you did all the ordering implementation. :wink:

zeenix:
@refi64 Hey, seems this change broke API for people (#1371). Could you please have a look and see if you can provide a fix? :pray:

Issue #73, opened 2020-08-24
zbus_macros: Rename "name" attributes to "rename"


zeenix:
To be consistent with serde (and ourselves), we should replace "name" attributes with "rename". We still support the "name" attribute but deprecate it.

Issue #1344, opened 2025-04-25
zvariant_derive: `#[serde(default)]` for `DeserializeDict`


eaglesemanation:
As far as I understand, the current implementation of proc macro that generates `DeserializeDict` only supports `Option<T>` for handling values that are not necessarily present in `a{sv}`.

I've tried deriving both `Deserialize` and `DeserializeDict` at the same time, but as expected this causes overlapping trait implementation. Would be awesome to be able to just use serde attributes directly, but I don't have enough proc-macro experience to judge if that's actually possible to do. Seems like `serde_repr` somehow does that.

Alternatively, it would be nice to just have `#[zvariant(default)]`, but that would possibly lead down the path of fully reimplementing serde functionality, which doesn't sound like a fun time for foss maintainer.

zeenix:
Yeah, I actually really don't like that we even have these macros. See #1345.

Issue #85, opened 2020-09-12
Supprt skip_serializing_if attribute


zeenix:
In GitLab by @bilelmoussaoui on Sep 12, 2020, 22:40

Currently, having a TypeDict with Vec<T> means that the key associated to the Vec<T> will be serialized with an empty array as a value.

skip_serializing_if attribute would solve such use case as you can have
```rust
#[derive(SerializeDict, DeserializeDict, TypeDict, Debug)]
/// A notification
pub struct Notification {
    /// User-visible string to display as the title.
    pub title: String,
    /// Array of buttons to add to the notification.
    #[zvariant(skip_serializing_if = "Vec::is_empty")]
    pub buttons: Vec<Button>,
}
```

I have looked a bit how to implement this, would that be an interesting feature to have?

zeenix:
Fixed by #1354 .

PR #1345, opened 2025-04-26
🚧 zv: Turn SerializeDict & DeserializeDict into wrappers around serde::{Serialize,Deserialize}

zeenix:
We have discussed this on and off before but I think last time we really looked into it some years back, our ser/de machinery was needlessly over-complicated and therefore very fragile so it wasn't easy to make this work. However, now it seems like it'd be very much possible, especially after the first commit here (e2b2fcd0d2521a7b9a1e5cbf2208fe288948e763).

This will automatically fix various rough edges with our macros, like #1344.

~~Having said that, I'm quite busy these days with other stuff so maybe someone else can help me finish this PR? If not, I'll get to it in few weeks hopefully.~~

### DIFF
--- a/zvariant/src/dbus/ser.rs
+++ b/zvariant/src/dbus/ser.rs
@@ -1,5 +1,5 @@
 use serde::{
-    ser::{self, SerializeSeq, SerializeTuple},
+    ser::{self, SerializeMap, SerializeSeq, SerializeTuple},
     Serialize,
 };
 use std::{
@@ -338,6 +338,7 @@ where
             Signature::Structure(_) => {
                 StructSerializer::structure(self).map(StructSeqSerializer::Struct)
             }
+            Signature::Dict { .. } => self.serialize_map(Some(len)).map(StructSeqSerializer::Map),
             _ => Err(Error::SignatureMismatch(
                 self.0.signature.clone(),
                 "a struct, array or variant".to_string(),
@@ -554,6 +555,7 @@ where
 pub enum StructSeqSerializer<'ser, 'b, W> {
     Struct(StructSerializer<'ser, 'b, W>),
     Seq(SeqSerializer<'ser, 'b, W>),
+    Map(MapSerializer<'ser, 'b, W>),
 }
 
 macro_rules! serialize_struct_anon_fields {
@@ -591,6 +593,7 @@ macro_rules! serialize_struct_anon_fields {
                 match self {
                     StructSeqSerializer::Struct(ser) => ser.$method(value),
                     StructSeqSerializer::Seq(ser) => ser.serialize_element(value),
+                    StructSeqSerializer::Map(_) => unreachable!(),
                 }
             }
 
@@ -598,6 +601,7 @@ macro_rules! serialize_struct_anon_fields {
                 match self {
                     StructSeqSerializer::Struct(ser) => ser.end_struct(),
                     StructSeqSerializer::Seq(ser) => ser.end_seq(),
+                    StructSeqSerializer::Map(_) => unreachable!(),
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


zeenix:
I managed to make the functions in the test code generic, so we could also just:

* Provide these generic ser/de methods in zvariant.
* Deprecated our macros and document (both in the macro docs and the book) how to use them directly with serde macros.

Given how I've very little time these days for active zbus dev, I'm leaning towards this.

@bilelmoussaoui WDYT?

zeenix:
> > Having said that, I'm quite busy these days with other stuff so maybe someone else can help me finish this PR? If not, I'll get to it in few weeks hopefully.
> 
> What else is missing? just what you mentioned in the second comment?

I'm asking if what I described in the second comment, sounds good to you. It'll be a bit more typing for the developer but I think it's worth moving everyone to just using serde directly and it feels more "normal" than having to use some custom zvariant macros:

```rust
        #[derive(Serialize, Deserialize, Type, PartialEq, Debug)]
        #[zvariant(signature = "a{sv}")]
        struct Test {
            #[serde(
                serialize_with = "serialize_optional_value",
                deserialize_with = "deserialize_optional_value",
                skip_serializing_if = "Option::is_none",
                default
            )]
            process_id: Option<u32>,
            #[serde(
                serialize_with = "serialize_optional_value",
                deserialize_with = "deserialize_optional_value",
                skip_serializing_if = "Option::is_none",
                default
            )]
            group_id: Option<u32>,
            #[serde(
                serialize_with = "serialize_value",
                deserialize_with = "deserialize_value"
            )]
            user: String,
        }
```

>  also could this possibly already land in a future 0.x release ?

0.x? We're already at 5.x. :)

bilelmoussaoui:
> I'm asking if what I described in the second comment, sounds good to you. It'll be a bit more typing for the developer but I think it's worth moving everyone to just using serde directly and it feels more "normal" than having to use some custom zvariant macros:

for me that is actually better to be verbose than having a default "unexpected" one.

> 0.x? We're already at 5.x. :)

i meant a minor release.

bilelmoussaoui:
Although, in your example, from where `optional_value` would come from? In any case, looking forward to seeing this land, it would unlock a bunch of simplifications in my crates. 

Maybe you can close https://github.com/dbus2/zbus/issues/73 & https://github.com/dbus2/zbus/issues/85 and possibly others with this change.

zeenix:
> Although, in your example, from where `optional_value` would come from? In any case, looking forward to seeing this land, it would unlock a bunch of simplifications in my crates.

From zvariant (it'll be a new module). See 91a1f9474b55c671e6ca216b59b03f50e11047c4.

> Maybe you can close #73 & #85 and possibly others with this change.

Yup.

---

I bumped into an issue though. :( The following code ends up with stack overflow:

```rust
        #[derive(Serialize, Deserialize, Type, PartialEq, Debug, Default)]
        #[zvariant(signature = "a{sv}")]
        #[serde(default)]
        struct TestSkipUnknown {
            #[serde(with = "optional", skip_serializing_if = "Option::is_none")]
            process_id: Option<u32>,
            #[serde(with = "optional", skip_serializing_if = "Option::is_none")]
            group_id: Option<u32>,
        }
        let _: TestSkipUnknown = encoded.deserialize().unwrap().0;
 ```
 
 If the encoded dict contains unknown keys. However, it works fine if you set `#[serde(deny_unknown_fields)]` attribute on the struct but then of course you get an error. This could be serde bug though. I'll investigate..

zeenix:
Alright, finally done, in #1354. This took many hours because it uncovered a bunch of bugs in our ser/de machinery. I believe now it's all fixed. Closing this in favor of #1354.

zeenix:
> > Maybe you can close #73 & #85 and possibly others with this change.
> 
> Yup.

Actually turns out that #73 is for the zbus macros, not zvariant. So that is not fixed by #1354.

PR #1362, opened 2025-05-05
Warning fixes

elmarco:


zeenix:
Looks great. Thanks for the very quick actions. Did you check why CI isn't failing on these (especially the compiler error)?

.github/workflows/rust.yml
```diff
@@ -119,6 +119,8 @@ jobs:
       RUST_LOG: trace
       PKG_CONFIG: C:\bin\pkg-config.exe
       PKG_CONFIG_PATH: C:\lib\pkgconfig
+      ErrorActionPreference: Stop
```
zeenix, line 122: Interesting, why is this note needed for the Linux job? :thinking:

.github/workflows/rust.yml
```diff
@@ -119,6 +119,8 @@ jobs:
...
+      WarningPreference: Stop
```
zeenix, line 123: Is this really needed since we're enabling `-D warning` above?

zeenix: ah ok

PR #1366, opened 2025-05-10
✨ zm: Copy attributes to `receive_*_changed` and `cached_*` methods

### LINKED ISSUE #1312: Attribute macros (like cfg) should copy to `receive_*_changed` and `cached_*`
When `interface`/`proxy` attribute macro combines with `#[cfg(..)]` macros on the methods, we assume them are work as expected,
```rust
// my-proxy crate in workspace to share the D-Bus API

#[zbus::proxy(..)]
trait MyApi {
    /// Client2, 3, ..., and the server should not see this property!
    /// server will use all-features=true in unit testing
    #[cfg(feature = "client1")]
    #[zbus(property)]
    fn foo(&self) -> zbus::Result<Foo>;

    /// Clients should not see this property!
    #[cfg(feature = "server")]
    #[zbus(property)]
    fn set_foo(&self, foo: Foo) -> zbus::Result<()>;
}
```
But `#[cfg(feature = "client1")]` is not working for `receive_foo_changed` and `cached_foo`, other downstream crates **can see them without enabling "client1" feature.**
Only `set_foo` work as expected.

Another scenario is the `interface` macro plus the `proxy` option. Maybe a `#[zbus(cfg(..))]` attribute should used to limit the client API, but maybe it is another PR.

KmolYuan:
Fix #1312 

In the `proxy` macro, support applying attributes to the related methods of the same property. This ensures that common attributes, such as `#[doc = "..."]` and `#[cfg(...)]`, can be applied to both the getter methods `receive_<property>_changed` and `cached_<property>`.

# Macro expanding preview

```rust
/// Test property for testing.
#[cfg(test)]
#[zbus(property)]
fn property(&self) -> fdo::Result<Vec<String>>;
```
will generates
```diff
 #[doc = " Test property for testing."]
 #[cfg(test)]
 #[allow(clippy::needless_question_mark)]
 pub fn property(&self) -> fdo::Result<Vec<String>> { ... }

+#[doc = " Test property for testing."]
+#[cfg(test)]
 #[doc = "Get the cached value of the `Property` property, or `None` if the property is not cached."]
 pub fn cached_property(
     &self,
 ) -> ::std::result::Result<
     ::std::option::Option<<fdo::Result<Vec<String>> as ::zbus::ResultAdapter>::Ok>,
     <fdo::Result<Vec<String>> as ::zbus::ResultAdapter>::Err,
 > { ... }

+#[doc = " Test property for testing."]
+#[cfg(test)]
 #[doc = "Create a stream for the `Property` property changes. This is a convenient wrapper around [`zbus::blocking::Proxy::receive_property_changed`]."]
 pub fn receive_property_changed(
     &self,
 ) -> ::zbus::blocking::proxy::PropertyIterator<
     'p,
     <fdo::Result<Vec<String>> as ::zbus::ResultAdapter>::Ok,
 > { ... }
```

### DIFF
--- a/zbus_macros/src/proxy.rs
+++ b/zbus_macros/src/proxy.rs
@@ -723,6 +723,7 @@ fn gen_proxy_property(
                 This is a convenient wrapper around [`{proxy_name}::receive_property_changed`]."
                 );
                 quote! {
+                    #(#other_attrs)*
                     #[doc = #gen_doc]
                     pub #usage fn #receive #ty_generics(
                         &self
@@ -747,6 +748,7 @@ fn gen_proxy_property(
                     " Get the cached value of the `{property_name}` property, or `None` if the property is not cached.",
                 );
                 quote! {
+                    #(#other_attrs)*
                     #[doc = #cached_doc]
                     pub fn #cached_getter(&self) -> ::std::result::Result<
                         ::std::option::Option<<#ret_type as #zbus::ResultAdapter>::Ok>,


zbus_macros/src/proxy.rs
```diff
@@ -717,12 +717,13 @@ fn gen_proxy_property(
         let receive_method = match emits_changed_signal {
             PropertyEmitsChangedSignal::True | PropertyEmitsChangedSignal::Invalidates => {
                 let (_, ty_generics, where_clause) = m.sig.generics.split_for_impl();
-                let receive = format_ident!("receive_{}_changed", method_name);
+                let receive = format_ident!("receive_{method_name}_changed");
```
zeenix, line 720: please either remove unrelated changes or split them into their own commits. As our contributing guidelines say, we prefer atomic commits (each logical change goes into its own commit).

zbus_macros/src/proxy.rs
```diff
@@ -717,12 +717,13 @@ fn gen_proxy_property(
...
                 let gen_doc = format!(
                     "Create a stream for the `{property_name}` property changes. \
-                This is a convenient wrapper around [`{proxy_name}::receive_property_changed`]."
+                     This is a convenient wrapper around [`{proxy_name}::receive_property_changed`]."
                 );
                 quote! {
+                    #(#other_attrs)*
```
zeenix, line 726: That easy? :laughing: Would be good to add a test case (doesn't have to be a new separate one but a part of an existing one).

zeenix (changes requested):
Thanks for this.

zbus_macros/src/proxy.rs
KmolYuan: Reverted in f1a09ad7bd6e5465d1e0ffdc9d2926ece7637f3d.

zbus_macros/src/proxy.rs
zeenix: hmm.. good question. Nevermind.

zbus_macros/src/proxy.rs
```diff
@@ -720,9 +720,10 @@ fn gen_proxy_property(
                 let receive = format_ident!("receive_{}_changed", method_name);
                 let gen_doc = format!(
                     "Create a stream for the `{property_name}` property changes. \
-                This is a convenient wrapper around [`{proxy_name}::receive_property_changed`]."
+                     This is a convenient wrapper around [`{proxy_name}::receive_property_changed`]."
```
zeenix, line 723: Is this change really relevant to the commit?

zbus_macros/src/proxy.rs
```diff
@@ -744,9 +745,10 @@ fn gen_proxy_property(
             | PropertyEmitsChangedSignal::Const => {
                 let cached_getter = format_ident!("cached_{}", method_name);
                 let cached_doc = format!(
-                    " Get the cached value of the `{property_name}` property, or `None` if the property is not cached.",
+                    "Get the cached value of the `{property_name}` property, or `None` if the property is not cached.",
```
zeenix, line 748: same here.

zeenix (changes requested):
I think there are still unrelated changes in the commit.

zeenix:
@KmolYuan the lint job failing is not your fault. Just rebase (not merge) on main.

zbus_macros/src/proxy.rs
KmolYuan: Revert in f511f8fd0a50a81b7682ff091ac6323642207fdc.

zbus_macros/src/proxy.rs
KmolYuan: Revert in f511f8fd0a50a81b7682ff091ac6323642207fdc.

PR #1367, opened 2025-05-10
📝 Document generated prop_changed and prop_invalidated methods

duelafn:
Avoids "missing documentation for a method" when -W missing-docs enabled in user code.

The documentation added is similar to the wording from the `#[interface]` macro documentation for the PROP_changed methods.

Thanks for a great crate!

### DIFF
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -725,7 +725,14 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                     };
 
                     if p.emits_changed_signal == PropertyEmitsChangedSignal::True {
+                        let changed_doc = format!(
+                            "Emit the “PropertiesChanged” signal with the new value for the\n\
+                             `{member_name}` property.\n\n\
+                             This method should be called if a property value changes outside\n\
+                             its setter method."
+                        );
                         let prop_changed_method = quote!(
+                            #[doc = #changed_doc]
                             pub async fn #prop_changed_method_name(
                                 &self,
                                 __zbus__signal_emitter: &#zbus::object_server::SignalEmitter<'_>,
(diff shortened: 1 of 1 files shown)


zbus_macros/src/iface.rs
```diff
@@ -726,6 +726,11 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
 
                     if p.emits_changed_signal == PropertyEmitsChangedSignal::True {
                         let prop_changed_method = quote!(
+                            #[doc = "Emit the “PropertiesChanged” signal with the new value for the"]
+                            #[doc = concat!(" `", #member_name, "` property.")]
+                            #[doc = ""]
+                            #[doc = "This method should be called if a property value changes outside its"]
+                            #[doc = " setter method."]
```
zeenix, line 733: Let's not add multiple `#[doc` attributes but rather create the doc string and then just use it here. See how we're doing that for the proxy doc [here](https://github.com/dbus2/zbus/blob/e309b0d11f28247a5c666949b9910ff5e1c90fde/zbus_macros/src/iface.rs#L1611).

zeenix (changes requested):
Thanks so much! Just need a minor adjustment.

zbus_macros/src/iface.rs
duelafn: Ah, yep, that's better - done!

zbus_macros/src/iface.rs
```diff
@@ -750,7 +755,13 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                     }
 
                     if p.emits_changed_signal == PropertyEmitsChangedSignal::Invalidates {
+                        let invalidate_doc = format!(
+                            "Emit the “PropertiesChanged” signal for the `{member_name}` property without including the new value.\n\n\
```
zeenix, line 759: At least some of these lines go beyond 100 chars limit we have. Please split them. Then we can just merge. :+1:

duelafn: Sorry, updated.

zeenix (approved):
:+1:

PR #1381, opened 2025-05-19
🐛 zb: Fix build on Android by limiting PeerPidfd to linux

swick:
For whatever reason the nix::sys::socket::sockopt::PeerPidfd is only available to target_os = "linux", so properly guard the code using it against that.

Fixes: 3602dcbb ("✨ zb: Add ProcessFD/pidfd to peer ConnectionCredentials")

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: ${{ env.MSRV }}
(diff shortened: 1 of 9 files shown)


zeenix:
@swick Thanks for the quick action here. :+1: I'm fixing the CI breakage. In the meantime, could you please add `aarch64-linux-android` to the workflow where we check the build for other platforms? That will also test this fix.

swick:
> could you please add aarch64-linux-android to the workflow where we check the build for other platforms

Added a new commit which I think does what you want it to, but would be good to double check.

Issue #1111, opened 2024-10-30
`PropertyStream` yields initial value at start if service is running, but doesn't yield value when service starts later


ids1024:
As mentioned in https://github.com/dbus2/zbus/issues/1109#issuecomment-2448073824.

If a (non-socket activated) service isn't running when a `PropertyStream` is created, the stream doesn't yield an initial value (as expected), but also doesn't yield a value once a service does start running.

Ideally the stream should also yield any time the name owner is changed.

PR #1389, opened 2025-05-31
🐛 zb: Make `PropertyStream` yield changes if service starts later

ids1024:
This seems like the simplest way to address https://github.com/dbus2/zbus/issues/1111.

Before this, a `PropertyStream` created when a service wasn't running would never yield a value, even if the service started later and send property change signals.

But if a service stopped, and then was restarted, property changes would be received.

It seems `PropertiesCache::init()` was returning an error, from trying to call `GetAll` with a destination that doesn't exist.

Instead, let `init()` complete successfully (without populating the cache), so it returns a `PropertiesChangedStream` which is then monitored for changes.

It may also be good if it yielded values when a service takes ownership of the name, but that's a separate issue since it also applies to service restarts. And an application can handle `OwnerChanged` itself relatively easily.

### DIFF
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -358,6 +358,12 @@ impl PropertiesCache {
                 Some(Either::Left(_update)) => {
                     // discard updates prior to the initial population
                 }
+                Some(Either::Right(Err(zbus::Error::MethodError(name, _, _))))
+                    if name == "org.freedesktop.DBus.Error.ServiceUnknown" =>
+                {
+                    // DBus service doesn't exist. But may appear later on the bus.
+                    break;
+                }
                 Some(Either::Right(populate)) => {
                     populate?.body().deserialize().map(|values| {
                         self.update_cache(&uncached_properties, &values, &[], &interface);


zbus/src/proxy/mod.rs
```diff
@@ -358,6 +358,12 @@ impl PropertiesCache {
                 Some(Either::Left(_update)) => {
                     // discard updates prior to the initial population
                 }
+                Some(Either::Right(Err(zbus::Error::MethodError(name, _, _))))
```
zeenix, line 361: Interesting. So we don't end up with a `zbus::fd::Error::ServiceUnknown`? :cry:

ids1024: I guess that's because this is using `call_method_raw` to call `GetAll`. So it just gets a `zbus::Error` instead of a `zbus::fdo::Error`.

ids1024: We can call `zbus::fdo::Error::from()` here. Though ultimately it needs to convert back to `zbus::Error` somewhere, if it doesn't match `ServiceUnknown`, since `get_property()` returns this error as a `zbus::Error` when `cache.ready()` fails.

zeenix: > We can call `zbus::fdo::Error::from()` here.

That'd be better but I want to solve it more generally so one gets the more specific error and here you can then just match on that. Hence why I laid out the plan I did in the previous comment.

ids1024: I'm not sure exactly what you're suggesting.

If we convert to a `zbus::fdo::Error`, that will consume the `zbus::Error`; so we need to clone or convert back if we want to return the original error. I think we want to just ignore `ServiceUnknown` here, but still return an error for other `zbus::fdo::Error` variants, (both the `zbus::fdo::Error::Zbus` fallthrough case, and the others).

zeenix: > I'm not sure exactly what you're suggesting.

Same here. :sob: I'm not sure I follow. With [my suggestion](https://github.com/dbus2/zbus/pull/1389#discussion_r2118273992), you should be matching against `zbus::Error::Fdo(fdo::Error::ServiceUnknown` here instead of matching the string manually. The conversion will happen [here](https://github.com/dbus2/zbus/blob/3d3f356b06f2f95a7b28a49e590f354e0368742d/zbus/src/error.rs#L248).

zeenix: > so we need to clone or convert back if we want to return the original error.

As I wrote:

> From<Message> impl should convert the Error into an fdo::Error and wrap it in Error::Fdo, unless the conversion ends up with the fall-through fdo::Error::ZBus variant, in which case we discard the fdo::Error and keep the original Error.

This will involve restructuring the Fdo::Error.

ids1024: Oh, I just completely missed that there's already an `FDO` variant of `zbus::Error`.

ids1024: So something like this?

```patch
             match message.body().deserialize_unchecked::<&str>() {
-                Ok(detail) => Error::MethodError(name, Some(String::from(detail)), message),
+                Ok(detail) => {
+                    let err = Error::MethodError(name, Some(String::from(detail)), message);
+                    Error::from(fdo::Error::from(err))
+                }
                 Err(_) => Error::MethodError(name, None, message),
             }
```

(Looks a little funny converting a `zbus::Error` to an `fdo::Error` and back like this, but `From<fdo::Error> for Error` already does a match on whether its an `fdo::Error::Zbus` or not, and otherwise creates a `Error::FDO`)

zeenix: > So something like this?

I think so, yes.

> (Looks a little funny converting a `zbus::Error` to an `fdo::Error` and back like this

True true but maybe we can add some comments and make the types explicit to make it easy to read the code.

> but `From<fdo::Error> for Error` already does a match on whether its an `fdo::Error::Zbus` or not, and otherwise creates a `Error::FDO`)

Ah right, true.

ids1024:
Seems to be working with that updating. (It would be nice if `box` patterns where stable...).

zeenix:
> Seems to be working with that updating.

Nice!

> (It would be nice if `box` patterns where stable...).

Right. Could you please add a FIXEM comment pointing to the relevant issue?

ids1024:
Added a comment. https://github.com/rust-lang/rust/issues/29641 has been open since 2015, but apparently https://github.com/rust-lang/rust/issues/87121 supersedes that, so maybe it will actually be stable this decade.

zeenix:
@ids1024 the CI is failing. I think something goes wrong with the error conversion. Could you please have a look?

ids1024:
With the error conversion commit, the `basic_connection` and `basic_connection_async` tests need to be updated to check for `Error::FDO` instead of `Error::MethodError`. 

But then there's the `error_from_zerror` test. Instead of the expected `fdo::Error::TimedOut("so long")`, it's getting `fdo::Error::ZBus::FDO(zbus::Error::FDO(fdo::Error::TimedOut("so long")))`...

Because we already converted FDO error variants, and the `DBusError` macro used to define the conversion from `zbus::fdo::Error` doesn't handle the `zbus::Error::FDO` case, and just packs that into `fdo::Error::ZBus`.

Presumably changing that would require either a manual implementation of `From<zbus::Error>` that also handles the `FDO` case, or a special case in the macro. Though I don't know if this error conversion is likely to causes issues anywhere in general.

zeenix:
> fdo::Error::ZBus::FDO(zbus::Error::FDO(fdo::Error::TimedOut("so long")))

Hmm.. this shouldn't be happening. This is what I was hoping to avoid with [this part of my suggestion](https://github.com/dbus2/zbus/pull/1389#discussion_r2118273992):

> unless the conversion ends up with the fall-through fdo::Error::ZBus variant, in which case we discard the fdo::Error and keep the original Error.

zeenix:
@ids1024 do you intend to continue this work?

zeenix:
> @ids1024 do you intend to continue this work?

I guess not. :cry:

ids1024:
Well, if changing the definition of `From<zbus::Error> for fdo::Error` to have a branch handling `Error::FDO` is the correct way to do things, that's demonstrated by copying the expanded `DBusError` macro code and modifying the `From` impl.

I guess the only not painfully verbose way to do that would be to add some kind of special case in the `DBusError` macro...

ids1024:
It doesn't seem too hard to add a special case in the macro, using an internal boolean attribute parameter.

zeenix:
@ids1024 Thanks so much for picking this up again! Although I pinged you (was just doing a quick old PR triage), I can't seem to remember exact details here so I'll have to re-read the discussion when I'm not tired (hopefully tomorrow morning). :)

zbus/src/fdo/error.rs
```diff
@@ -2,7 +2,11 @@ use crate::DBusError;
 
 /// Errors from <https://gitlab.freedesktop.org/dbus/dbus/-/blob/master/dbus/dbus-protocol.h>
 #[derive(Clone, Debug, DBusError, PartialEq)]
-#[zbus(prefix = "org.freedesktop.DBus.Error", impl_display = true)]
+#[zbus(
+    prefix = "org.freedesktop.DBus.Error",
+    impl_display = true,
+    zbus_internal_is_fdo_err = true
```
zeenix, line 8: Not ideal but ok. :)

zbus_macros/src/error.rs
```diff
@@ -140,9 +142,19 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
 
     let from_zbus_error_impl = zbus_error_variant
         .map(|ident| {
+            let fdo_conversion = if zbus_internal_is_fdo_err == Some(true) {
```
zeenix, line 147: would be good to add a comment above here, explaining what's going on here.

zbus/src/error.rs
```diff
@@ -271,7 +271,11 @@ impl From<Message> for Error {
         if let Some(name) = header.error_name() {
             let name = name.to_owned().into();
             match message.body().deserialize_unchecked::<&str>() {
-                Ok(detail) => Error::MethodError(name, Some(String::from(detail)), message),
+                Ok(detail) => {
+                    let err = Error::MethodError(name, Some(String::from(detail)), message);
+                    // Convert to `Error::FDO` if method error is an FDO error
```
zeenix: Thanks for adding this comment but I find the wording is a bit confusing. We're converting unconditionally but comment says `if`.

zbus/src/proxy/mod.rs
```diff
@@ -355,6 +355,14 @@ impl PropertiesCache {
                 Some(Either::Left(_update)) => {
                     // discard updates prior to the initial population
                 }
+                // FIXME: Use deref pattern when stable
+                // https://github.com/rust-lang/rust/issues/87121
+                Some(Either::Right(Err(Error::FDO(fdo))))
+                    if matches!(*fdo, fdo::Error::ServiceUnknown(_)) =>
```
zeenix, line 361: Can't this be:

```rust
                Some(Either::Right(Err(Error::FDO(fdo::Error::ServiceUnknown(_))))) =>
```

?

zbus/tests/basic.rs
```diff
@@ -52,7 +52,7 @@ fn basic_connection() {
         "Hello",
         &(),
     ) {
-        Err(zbus::Error::MethodError(_, _, _)) => (),
+        Err(zbus::Error::FDO(_)) => (),
```
zeenix, line 55: Hmm.. so we end up having only `call_method` always returning an `zbus::Error::FDO`? That seems wrong and also while it's not an API break, it's a big behavioural change and could break existing client code out there so we really need to be sure this is what we want. 

I would actually prefer a bit of a redundant cycle in a certain case (properties) over always returning a `FDO` variant. fdo::Error is primarily for errors you get from the bus and ones defined by the spec.

zbus/src/proxy/mod.rs
ids1024: It looks like the separate `if` branch is still needed, since `Error::FDO` contains a `Box`. So avoiding this presumably still needs deref patterns (or `box` patterns, if those somehow happened sooner).

zbus/src/error.rs
ids1024: If will only be an `Error::FDO` if it's an FDO error. If it doesn't match any variant of `fdo::Error`, the conversion will keep it as `Error::MethodError`.

(I guess the comment could mention that part about how it will otherwise be `Error::MethodError`.)

zbus/tests/basic.rs
ids1024: The error here is `zbus::Error::FDO(box zbus::fdo::Error::Failed(_))`, because the error is `org.freedesktop.DBus.Error.Failed`.

If the error string did not match any variant of `zbus::fdo::Error`, the return value would still be `zbus::Error::MethodError(_)`.

zbus/src/proxy/mod.rs
zeenix: Ah right.

zbus_macros/src/error.rs
zeenix: I think we can merge this, once you've addressed this nitpick.

zeenix: Oh and your commits are missing emoji prefixes. Please add them as per [our contribution guidelines](https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). :pray:

ids1024:
Added a comment on the special case in `zbus_macros`, and added a little more detail to the comment in the error conversion code. Presumably 🐛 is appropriate for all three commits, since it seems like a correction to how one might expect things would already work.

zeenix:
Sorry for the delay in reviewing. I've been drowning in work-work lately. 😭

> Added a comment on the special case in zbus_macros, and added a little more detail to the comment in the error conversion code.

Thank you!

> Presumably 🐛 is appropriate for all three commits, since it seems like a correction to how one might expect things would already work.

Thanks for adding that but only the last commit is a bug fix so 🐛 doesn't seem appropriate for the other 2. Better candidates: 🥅 🚸.

Otherwise we're ready to merge.

zeenix:
> Otherwise we're ready to merge.

Spoke too soon. I just realised we're missing a test case here. Would it be possible for you to add one?

zeenix:
@ids1024 Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

Issue #1328, opened 2025-04-08
zbus_xml: Parsing error for large Node XML file


endrift:
Attempting to parse the attached XML file, returned by the Introspect interface, using `Node::from_reader(Cursor::new(xml))` fails with the error "XML error: Deserializer buffers 1024 events, limit exceeded". Using zbus_xml 5.0.2 and quick-xml 0.36.2. Let me know if I need to provide any more information.

[iface.xml.gz](https://github.com/user-attachments/files/19641723/iface.xml.gz)

endrift:
I think you misunderstand. I'm not using this for xmlgen, I'm using it for getting information live from the Introspect interface to be parsed in a running program. The closest other way to do this is by using the ObjectManager interface, but that winds up querying the values of the properties, which is not desired.

Further, I'm not sure that this error is related to security-related limits, though I'm not 100% sure on that.

zeenix:
> I'm using it for getting information live from the Introspect interface to be parsed in a running program. The closest other way to do this is by using the ObjectManager interface, but that winds up querying the values of the properties, which is not desired.

1. I don't see any properties in the [ObjectManager interface](https://docs.rs/zbus/latest/zbus/fdo/struct.ObjectManagerProxy.html).
2. That constraint sounds like something very specific to your use case. Properties are the bread and butter of D-Bus and avoiding to use them isn't something I'd be interested in catering for in our API.

> Further, I'm not sure that this error is related to security-related limits, though I'm not 100% sure on that.

Instead of just rejecting my assertion, perhaps if had dug a bit deeper, you'd have found out that [I'm correct](https://github.com/tafia/quick-xml/blob/a81bac74d6c30b80024a4f9198e501d4aed29d3d/src/de/mod.rs#L2525)? :)

---

Anyway, it seems the [limits are set by us](https://github.com/dbus2/zbus/blob/main/zbus_xml/src/lib.rs#L265) and we're setting them to 4k, which seems pretty small for computers these days. So if you could provide a PR that raises this limit (e.g to 8k or say 64k), I'd accept it and reopen the issue if you provide one.

endrift:
> > I'm using it for getting information live from the Introspect interface to be parsed in a running program. The closest other way to do this is by using the ObjectManager interface, but that winds up querying the values of the properties, which is not desired.
> 
>  1. I don't see any properties in the [ObjectManager interface](https://docs.rs/zbus/latest/zbus/fdo/struct.ObjectManagerProxy.html).

[...]

PR #1395, opened 2025-06-04
🐛 zx: Allow parsing very big introspection xml

JavaLot:
Some applications pass very large lists via d-bus, this fix allows you to successfully complete your tasks by trusting these applications.

### DIFF
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -545,7 +545,7 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     debug!("Created: {:?}", root_introspect_proxy);
 
     let root_xml = root_introspect_proxy.introspect().await?;
-    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes())
+    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes(), 1024)
         .map_err(|e| Error::Failure(e.to_string()))?;
     let mut node = &root_node;
     for name in ["org", "freedesktop", "MyService"] {
@@ -649,8 +649,8 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
 
     let xml = proxy.inner().introspect().await?;
     debug!("Introspection: {}", xml);
-    let node =
-        zbus_xml::Node::from_reader(xml.as_bytes()).map_err(|e| Error::Failure(e.to_string()))?;
+    let node = zbus_xml::Node::from_reader(xml.as_bytes(), 1024)
+        .map_err(|e| Error::Failure(e.to_string()))?;
     let ifaces = node.interfaces();
     let iface = ifaces
         .iter()
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -545,7 +545,9 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     debug!("Created: {:?}", root_introspect_proxy);
 
     let root_xml = root_introspect_proxy.introspect().await?;
-    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes())
+    let limit = zbus_xml::NodeEventLimit::new(4096);
+    let root_node = limit
+        .read(root_xml.as_bytes())
         .map_err(|e| Error::Failure(e.to_string()))?;
     let mut node = &root_node;
     for name in ["org", "freedesktop", "MyService"] {
@@ -649,8 +651,10 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
 
     let xml = proxy.inner().introspect().await?;
     debug!("Introspection: {}", xml);
-    let node =
-        zbus_xml::Node::from_reader(xml.as_bytes()).map_err(|e| Error::Failure(e.to_string()))?;
+    let limit = zbus_xml::NodeEventLimit::new(4096);
+    let node = limit
+        .read(xml.as_bytes())
+        .map_err(|e| Error::Failure(e.to_string()))?;
     let ifaces = node.interfaces();
     let iface = ifaces
         .iter()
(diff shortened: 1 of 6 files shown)


zeenix:
Thanks so much.

> 🐘

I appreciate your humour here but we use a specific set of emojis, each with a established meaning/context. Kindly choose the appropriate one from [here](https://gitmoji.dev/) (or add one using [ `gimoji`](https://github.com/zeenix/gimoji)). If you have suggestions on addition of new emojis, please send a issue/PR with justification to [gitmoji](https://github.com/carloscuesta/gitmoji/pulls).

zbus_xml/src/lib.rs
```diff
@@ -266,6 +266,13 @@ impl<'a> Node<'a> {
         Ok(Node::deserialize(&mut deserializer)?)
     }
 
+    /// Parse the introspection XML document from trusted reader without limiting the number of
+    /// events.
```
zeenix: * As [I explained on the issue](https://github.com/dbus2/zbus/issues/1328#issuecomment-2788895744), the limits are there for a good reason. The fix should be to bump the limit.
* This is not solving the issue but rather adding the workaround in zbus_xml, which zbus-xmlgen doesn't even make use of.

zeenix:
@JavaLot ?

JavaLot:
@zeenix, I will continue tomorrow! :-)

zeenix:
> @zeenix, I will continue tomorrow! :-)

:+1:

zeenix:
@JavaLot if you need help with git, please let me know.

zbus_xml/src/lib.rs
```diff
@@ -260,9 +260,9 @@ pub struct Node<'a> {
 
 impl<'a> Node<'a> {
     /// Parse the introspection XML document from reader.
-    pub fn from_reader<R: Read>(reader: R) -> Result<Node<'a>> {
+    pub fn from_reader<R: Read>(reader: R, limit: usize) -> Result<Node<'a>> {
```
zeenix: This is an API break and IMO not needed. We should just hardcode the limit and if we want to make it configurable, we need to [wait for 6.0](https://github.com/dbus2/zbus/milestone/7) .

zbus/tests/e2e.rs
```diff
@@ -545,7 +545,7 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     debug!("Created: {:?}", root_introspect_proxy);
 
     let root_xml = root_introspect_proxy.introspect().await?;
-    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes())
+    let root_node = zbus_xml::Node::from_reader(root_xml.as_bytes(), 1024)
```
zeenix: I'm confused. This is lower than the previous limit.

zeenix (changes requested):
Cool. We're getting there.

zeenix:
@JavaLot There is no need to revert your commits. When you've a fix, just `git commit --amend` and `git push --force..`.

zeenix:
> @JavaLot
[Merge branch 'dbus2:main' into main](https://github.com/dbus2/zbus/pull/1395/commits/2b3e72a68a30ae48f0d682a94bf47ab14e7712fe)

Kindly not merge the main branch, that just adds redundant commits to the PR. Please rebase instead. :pray:

zeenix:
What you need to do is an interactive rebase on main branch and squashing all commits and then you'll end up with one commit. Then force push.

If you want me to do that for you, please let me know.

zbus_xml/src/lib.rs
JavaLot: API returned back

zbus/tests/e2e.rs
JavaLot: set to 4096

JavaLot:
Need to somehow get it clean
Check if everything is ok

zbus_xml/src/lib.rs
```diff
@@ -260,6 +260,7 @@ pub struct Node<'a> {
 
 impl<'a> Node<'a> {
     /// Parse the introspection XML document from reader.
+    #[deprecated]
```
zeenix, line 263: Why? Please keep in mind that we prefer atomic commits so this should go in a separate commit with a rationale in the details section of the git commit message.

zbus_xml/src/error.rs
```diff
@@ -2,7 +2,7 @@ use quick_xml::de::DeError;
 use std::{convert::Infallible, error, fmt};
 use zvariant::Error as VariantError;
 
-/// The error type for `zbus_names`.
+/// The error type for `zbus_xml`.
```
zeenix, line 5: Oh good catch! It's unrelated to the containing commit through so it also should be split in its own commit. It's just a typo fix though so no need for justification/rationale. :)

zbus/tests/e2e.rs
```diff
@@ -649,8 +651,10 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
 
     let xml = proxy.inner().introspect().await?;
     debug!("Introspection: {}", xml);
-    let node =
-        zbus_xml::Node::from_reader(xml.as_bytes()).map_err(|e| Error::Failure(e.to_string()))?;
+    let limit = zbus_xml::NodeEventLimit::new(4096);
```
zeenix, line 654: let's use a constant for this for if we change it in one place, we change it everywhere.

zbus_xml/src/lib.rs
```diff
@@ -312,6 +313,33 @@ impl<'a> TryFrom<&'a str> for Node<'a> {
     }
 }
 
+/// A parser that can control introspection data.
```
zeenix, line 316: Why are we adding new API? As I wrote, we just need to hardcode the limit and this change should be completely hidden from existing users.

zeenix (changes requested):
I think it needs a lot of work, apart from the cleanup. I can always be wrong about my assumptions of course but if you decide to go a different route, please discuss it first so I know what to expect when looking at the code changes and it also helps you avoid wasting your time.

zeenix:
@JavaLot reminder: this is still waiting on you.

zeenix:
Ok. Thanks for wasting our time on this. 🤦‍♂️

Issue #1125, opened 2024-11-04
New release broke gvariant feature


bilelmoussaoui:
If you build zbus from git using `cargo build --features gvariant` it would fail with the following 
```
user@toolbox ~/P/zbus (main)> cargo build --features gvariant
    Blocking waiting for file lock on build directory
   Compiling zvariant v5.1.0 (/home/user/Projects/zbus/zvariant)
error[E0004]: non-exhaustive patterns: `Format::GVariant` not covered
   --> zvariant/src/serialized/data.rs:234:28
    |
234 |         let mut de = match self.context.format() {
    |                            ^^^^^^^^^^^^^^^^^^^^^ pattern `Format::GVariant` not covered
    |
note: `Format` defined here
   --> /home/user/Projects/zbus/zvariant_utils/src/serialized.rs:5:1
    |
5   | pub enum Format {
    | ^^^^^^^^^^^^^^^
...
12  |     GVariant,
    |     -------- not covered
    = note: the matched value is of type `Format`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
267 ~             .map(Deserializer::DBus)?,
268 ~             Format::GVariant => todo!(),
    |

error[E0004]: non-exhaustive patterns: `Format::GVariant` not covered
   --> zvariant/src/serialized/data.rs:307:28
    |
307 |         let mut de = match self.context.format() {
    |                            ^^^^^^^^^^^^^^^^^^^^^ pattern `Format::GVariant` not covered
    |
note: `Format` defined here
   --> /home/user/Projects/zbus/zvariant_utils/src/serialized.rs:5:1
    |
5   | pub enum Format {
    | ^^^^^^^^^^^^^^^
...
12  |     GVariant,
    |     -------- not covered
    = note: the matched value is of type `Format`
help: ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown
    |
340 ~             .map(Deserializer::DBus)?,
341 ~             Format::GVariant => todo!(),
    |

error[E0004]: non-exhaustive patterns: `Format::GVariant` not covered
  --> zvariant/src/ser.rs:58:21
   |
58 |     let len = match ctxt.format() {
   |                     ^^^^^^^^^^^^^ pattern `Format::GVariant` not covered
   |
note: `Format` defined here
  --> /home/user/Projects/zbus/zvariant_utils/src/serialized.rs:5:1
   |
5  | pub enum Format {
[...]

PR #1398, opened 2025-06-05
✨ zb: Add gvariant feature

A6GibKm:
Allows enabling the gvariant features in the sub-crates without adding them explicitly.

This can be seen working at https://github.com/bilelmoussaoui/oo7/pull/262.

Fixes (for real): https://github.com/dbus2/zbus/issues/1125

### DIFF
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["os::unix-apis"]
 readme = "README.md"
 
 [features]
+gvariant = ["zvariant/gvariant", "zbus_macros/gvariant"]
 default = ["async-io", "blocking-api"]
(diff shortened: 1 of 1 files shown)


zeenix:
As I wrong on Matrix, this will be a last resort and sounds like a workaround. So we should file an issue on Cargo about this and include a FIXME note here pointing to that issue.

zeenix:
There is a possible problem with this, which is why I had removed this feature from zbus. See the log of c49e085275d62f19019b78c7109aea6654e78a26.

zeenix:
Converted to draft to avoid accidental merge. :)

zeenix:
Gvariant support has been deprecated in favour of a separate crate, zgvariant.

PR #1401, opened 2025-06-09
✨ zb: Allow overriding some request name flags

A6GibKm:
By default names are requested with ReplaceExisting and DoNotQueue. We add to builder methods to control whether the ReplaceExisting and AllowReplacement bits are set.

### DIFF
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -1,5 +1,6 @@
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
+use enumflags2::BitFlags;
 use event_listener::Event;
 #[cfg(not(feature = "tokio"))]
 use std::net::TcpStream;
@@ -24,6 +25,7 @@ use zvariant::ObjectPath;
 
 use crate::{
     address::{self, Address},
+    fdo::RequestNameFlags,
     names::{InterfaceName, WellKnownName},
     object_server::{ArcInterface, Interface},
     Connection, Error, Executor, Guid, OwnedGuid, Result,
@@ -54,6 +56,17 @@ enum Target {
 type Interfaces<'a> = HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, ArcInterface>>;
 
 /// A builder for [`zbus::Connection`].
+///
+/// The builder allows setting the flags [`RequestNameFlags::AllowReplacement`] and
+/// [`RequestNameFlags::ReplaceExisting`] when requesting names, but the flag
+/// [`RequestNameFlags::DoNotQueue`] will always be enabled. The reasons are:
+///
+/// 1. There is no indication given to the caller of [`Self::build`] that the name(s) request was
+///    enqueued and that the requested name might not be available right after building.
+///
+/// 2. The name may be acquired in between the time the name is requested and the
+///    [`crate::fdo::NameAcquiredStream`] is constructed. As a result the service can miss the
+///    [`crate::fdo::NameAcquired`] signal.
 #[derive(Debug)]
 #[must_use]
 pub struct Builder<'a> {
@@ -69,6 +82,7 @@ pub struct Builder<'a> {
     auth_mechanism: Option<AuthMechanism>,
     #[cfg(feature = "bus-impl")]
     unique_name: Option<crate::names::UniqueName<'a>>,
+    request_name_flags: BitFlags<RequestNameFlags>,
 }
 
 impl<'a> Builder<'a> {
@@ -288,6 +302,9 @@ impl<'a> Builder<'a> {
     /// of the connection setup ([`Builder::build`]), immediately after interfaces
     /// registered (through [`Builder::serve_at`]) are advertised. Typically this is
     /// exactly what you want.
+    ///
+    /// The methods [`Builder::allow_name_replacements`] and [`Builder::replace_existing_names`]
+    /// allow to set the [`zbus::fdo::RequestNameFlags`] used to request the name.
     pub fn name<W>(mut self, well_known_name: W) -> Result<Self>
     where
         W: TryInto<WellKnownName<'a>>,
@@ -297,6 +314,22 @@ impl<'a> Builder<'a> {
         self.names.insert(well_known_name);
 
         Ok(self)
+    }
+
+    /// Whether the [`zbus::fdo::RequestNameFlags::AllowReplacement`] flag will be set when
+    /// requesting names.
+    pub fn allow_name_replacements(mut self, allow_replacement: bool) -> Self {
+        self.request_name_flags
+            .set(RequestNameFlags::AllowReplacement, allow_replacement);
+        self
+    }
+
+    /// Whether the [`zbus::fdo::RequestNameFlags::ReplaceExisting`] flag will be set when
+    /// requesting names.
+    pub fn replace_existing_names(mut self, replace_existing: bool) -> Self {
+        self.request_name_flags
+            .set(RequestNameFlags::ReplaceExisting, replace_existing);
+        self
     }
 
     /// Set the unique name of the connection.
@@ -388,7 +421,8 @@ impl<'a> Builder<'a> {
         );
 
         for name in self.names {
-            conn.request_name(name).await?;
+            conn.request_name_with_flags(name, self.request_name_flags)
+                .await?;
         }
 
         Ok(conn)
@@ -407,6 +441,7 @@ impl<'a> Builder<'a> {
             auth_mechanism: None,
             #[cfg(feature = "bus-impl")]
             unique_name: None,
+            request_name_flags: BitFlags::default(),
         }
     }
 
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -526,12 +526,9 @@ impl Connection {
         W: TryInto<WellKnownName<'w>>,
         W::Error: Into<Error>,
     {
-        self.request_name_with_flags(
-            well_known_name,
-            RequestNameFlags::ReplaceExisting | RequestNameFlags::DoNotQueue,
-        )
-        .await
-        .map(|_| ())
+        self.request_name_with_flags(well_known_name, BitFlags::default())
+            .await
+            .map(|_| ())
     }
 
     /// Register a well-known name for this connection.
--- a/zbus/src/fdo/dbus.rs
+++ b/zbus/src/fdo/dbus.rs
@@ -20,7 +20,9 @@ use crate::{proxy, OwnedGuid};
 /// The flags used by the bus [`request_name`] method.
 ///
 /// [`request_name`]: struct.DBusProxy.html#method.request_name
-#[bitflags]
+// TODO: Revisit this choice of defaults. See
+// https://github.com/dbus2/zbus/issues/1413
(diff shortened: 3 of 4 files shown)


zbus/src/connection/builder.rs
```diff
@@ -387,8 +406,16 @@ impl<'a> Builder<'a> {
             already_received_fds,
         );
 
+        let mut flags = crate::fdo::RequestNameFlags::DoNotQueue.into();
```
zeenix: would be nice to move the initialization of `RequestNameFlags` in `Connection::request_name` to a `Default` impl and then use it from here and there.

zbus/src/connection/builder.rs
```diff
@@ -387,8 +406,16 @@ impl<'a> Builder<'a> {
...
+        if self.allow_replacement {
+            flags |= crate::fdo::RequestNameFlags::AllowReplacement;
+        }
+        if self.replace_existing {
+            flags |= crate::fdo::RequestNameFlags::ReplaceExisting;
+        }
```
zeenix: This is changing the current behaviour I believe since there is not distinction between user not setting it and disabling the flag explicitly. I think we need these fields to be `Option`-wrapped and then always set/unset the flag when its `Some`.

zbus/src/connection/builder.rs
```diff
@@ -69,6 +69,8 @@ pub struct Builder<'a> {
     auth_mechanism: Option<AuthMechanism>,
     #[cfg(feature = "bus-impl")]
     unique_name: Option<crate::names::UniqueName<'a>>,
+    allow_replacement: bool,
+    replace_existing: bool,
```
zeenix: Would be great to add to the `Builder` docs that why we only allow these params and not `DoNotQueue`.

zbus/src/connection/builder.rs
```diff
@@ -299,6 +304,20 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Whether the [`zbus::fdo::RequestNameFlags::AllowReplacement`] flag will be set when
+    /// requesting names.
+    pub fn allow_replacement(mut self, allow_replacement: bool) -> Self {
```
zeenix: more descriptive name please: allow_name_replacements (notice the plural).

zbus/src/connection/builder.rs
```diff
@@ -299,6 +304,20 @@ impl<'a> Builder<'a> {
...
+        self.allow_replacement = allow_replacement;
+        self
+    }
+
+    /// Whether the [`zbus::fdo::RequestNameFlags::ReplaceExisting`] flag will be set when
+    /// requesting names.
+    pub fn replace_existing(mut self, replace_existing: bool) -> Self {
```
zeenix: same here: replace_existing_names

zbus/src/connection/builder.rs
```diff
@@ -407,6 +434,9 @@ impl<'a> Builder<'a> {
             auth_mechanism: None,
             #[cfg(feature = "bus-impl")]
             unique_name: None,
+            // Keep in sync with Connection::request_name
```
zeenix: Not needed if we go with my `Default` suggestion. :)

zbus/src/connection/builder.rs
A6GibKm: How exactly it changes the behavior? In both cases (user sets the defaults manually vs user does not set any flag) both would trigger the same code path which would essentially match the old one (request_names calls request_name_with_flags).

zbus/src/connection/builder.rs
A6GibKm: I will add a method to generate the "defaults" used by zbus, but imho the only flag that makes sense as a default is no bits set.

zbus/src/connection/builder.rs
zeenix: Right, I didn't notice that you were setting them in the constructor here. In any case, this is a bit disconnected from that. I think it'd be actually best to just keep the flags directly in the Builder struct and then the setters can set/unset the flags accordingly.

zbus/src/connection/builder.rs
zeenix: > I will add a method to generate the defaults,

:+1: 

> but imho the only flag that makes sense as a default is not bits set.

Perhaps but then we need another centralised way to get the default  flags used in `connection`. I'm not sure it's worth it though because I don't really see where would "no flags set" value be useful.

zbus/src/connection/builder.rs
A6GibKm: Please tell me if they are OK.

zbus/src/connection/builder.rs
A6GibKm: We can still implement Defaults if you like, it is even supported as per https://docs.rs/enumflags2/latest/enumflags2/struct.BitFlags.html#customizing-default. But I am not sure it is the best idea for the public API, e.g. why AllowReplacement is not part of the Defaults and ReplaceExisting is?

zeenix: > We can still implement Defaults if you like, it is even supported as per https://docs.rs/enumflags2/latest/enumflags2/struct.BitFlags.html#customizing-default.

Nice.

> But I am not sure it is the best idea for the public API, e.g. why AllowReplacement is not part of the Defaults and ReplaceExisting is?

Then I guess your point is less that there shouldn't be a default value but rather that it doesn't make sense to have one while not the other. You have a point there. I'd refer other impls (e.g gio) and see what default (if any) they choose. If they choose the same, we can just point to the fact that our default is based on that. If not, we still go with the current defaults to not break any existing code relying on these defaults, add an issue about this and a FIXME comment in the code pointing to this issue.

zbus/src/connection/builder.rs
```diff
@@ -54,6 +56,12 @@ enum Target {
 type Interfaces<'a> = HashMap<ObjectPath<'a>, HashMap<InterfaceName<'static>, ArcInterface>>;
 
 /// A builder for [`zbus::Connection`].
+///
+/// The builder allows setting the flags [`RequestNameFlags::AllowReplacement`] and
+/// [`RequestNameFlags::ReplaceExisting`] when requesting names, but
+/// [`RequestNameFlags::DoNotQueue`] will always be enabled. The reason is that otherwise there
+/// might be a window in between creating the connection and getting the requested name in which the
+/// requested name might belong to a different connection.
```
zeenix: Clearly I failed to communicate the problem well. :) Let me try:

If a name is already taken when its requested in `Connection::build`, the bus will put the request in a queue. The service will in this case want to be notified when it acquires the name. This can be accomplished with [fdo::DBusProxy::receive_name_acquired](https://docs.rs/zbus/latest/zbus/fdo/struct.DBusProxy.html#method.receive_name_acquired). However, there are two issues with this:

1. There is currently on indication given to the caller of `Connection::build` that the name(s) request was enqueued so the caller has no idea that the name request was not completed.
2. The name may be acquired between the time the name is requested and the `fdo::NameAcquiredStream` is constructed and as a result the service can miss the `NameAcquired` signal.

The solution to these problems will be complex and will likely require an API break. Given that typically a service will not want to queue for acquiring its name(s), passing of `RequestNameFlags::DoNotQueue` is not currently supported. However, if you have a compelling use case for this, please file a issue, providing detailed rationale and we will try to accommodate your needs.

A6GibKm: Updated the docstring.

zbus/src/connection/builder.rs
A6GibKm: I am not sure if they have any default, they ask for flags explicitly https://docs.gtk.org/gio/func.bus_own_name.html.

zeenix: Right, I guess there are advantages of being a low-level C API. :) I guess then we go for the second option:

> If not, we still go with the current defaults to not break any existing code relying on these defaults, add an issue about this and a FIXME comment in the code pointing to this issue.

A6GibKm: The code is already using our current defaults. Should I also `impl Default` for them?

zeenix: > Should I also `impl Default` for them?

Exactly. Sorry if it wasn't clear.

A6GibKm: Done, thanks!

zeenix: > Done, thanks!

Cool but this part is still missing:

> If they choose the same, we can just point to the fact that our default is based on that. If not, we still go with the current defaults to not break any existing code relying on these defaults, add an issue about this and a FIXME comment in the code pointing to this issue.

A6GibKm: Done at https://github.com/dbus2/zbus/issues/1413. 

One thing I should mention though is that RequestNameFlags did have a default before this change, the following used to compile:

```rust
    use enumflags2::BitFlags;
    let flags: BitFlags<zbus::fdo::RequestNameFlags> = BitFlags::default();
    assert_eq!(flags, BitFlags::empty());
```

so this is effectively an API break for everyone who relied on that default.

zeenix: > Done at #1413.

:pray: 

> One thing I should mention though is that RequestNameFlags did have a default before this change

Oh, I didn't think that it's used through enumflags2::Bitflags so it will get an implicit default. We do assume a default value for it already though and that's not matching the current implicit one and IMO chances of someone depending on this are rather low.

So I'd say let's go with the new explicit Default but #1413  and its commit should hilight this and then it'll be harder for me to forget to mention this in the release notes.

If it turns out anyone was depending on this and we end up breaking things for them, we can always revert the change and go with your earlier impl.

zbus/src/fdo/dbus.rs
```diff
@@ -20,7 +20,9 @@ use crate::{proxy, OwnedGuid};
 /// The flags used by the bus [`request_name`] method.
 ///
 /// [`request_name`]: struct.DBusProxy.html#method.request_name
-#[bitflags]
+// TODO Revisit this choice of defaults. See
```
zeenix: nit:

```suggestion
// TODO: Revisit this choice of defaults. See
```

zbus/src/connection/mod.rs
```diff
@@ -526,12 +526,10 @@ impl Connection {
...
         W::Error: Into<Error>,
     {
-        self.request_name_with_flags(
-            well_known_name,
-            RequestNameFlags::ReplaceExisting | RequestNameFlags::DoNotQueue,
-        )
-        .await
-        .map(|_| ())
+        let flags = BitFlags::default();
+        self.request_name_with_flags(well_known_name, flags)
```
zeenix: future suggestion: best to keep the style of the code intact: there was no need for a new variable.

zeenix (approved):
LGTM

zbus/src/connection/builder.rs
A6GibKm: I updated the commit and issue. As I mentioned in the issue, the fallback should be minimal. Thanks.

zeenix:
The CI is failing because of new clippy lint. Please rebase on main once #1416 is merged.

A6GibKm:
Rebased.

zeenix:
@A6GibKm I had assumed this was merged but it was blocked on a discussion that was not marked as resolved.

PR #1406, opened 2025-06-12
📝 zn: doc typo

duelafn:
Error names have same constraints as *interface* names.

Trivial update resolving tautological statement, now matches dbus spec.


zeenix:
@duelafn Thanks. I took the liberty of splitting the description part of the title into the description section here. I would have updated the commit log too but that'd be too much work for a typo fix. :) Anyway, just to keep in mind for your future contributions. Thanks again!

PR #1408, opened 2025-06-12
✨zm: Support write-only properties

KmolYuan:
Support write-only properties in the `interface` macro.

- When the getter is not explicitly defined, the property will be write-only.
- The `emits_changed_signal` option must still be specified only on the getter method. If no getter is defined, this option will be set to `false`.

p.s.
I've also changed some related code blocks. (wrong indents, wrong semi-colons)
If you think they shouldn't be fixed, I will try to revert them.🩹

---

<details>
<summary>Old contents</summary>

Defined some rules when handling `cfg` attributes.
- `cfg` attributes on properties now determine their accessibility. (read/write/not exist)
- The `emits_changed_signal` option must still be specified only on the getter method. If no getter is defined (i.e., the property is write-only), this option will be set to `false`.
- When the getter is conditionally generated, the `emits_changed_signal` behavior will still respect the write-only fallback.

</details>

### DIFF
--- a/zbus_macros/src/lib.rs
+++ b/zbus_macros/src/lib.rs
@@ -250,7 +250,8 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///     * `"invalidates"` - the change signal is emitted, but the value is not included in the
 ///       signal.
 ///     * `"const"` - the property never changes, thus no signal is ever emitted for it.
-///     * `"false"` - the change signal is not emitted if the property changes.
+///     * `"false"` - the change signal is not emitted if the property changes. If a property is
+///       write-only, the change signal will not be emitted in this interface.
 ///
 /// * `signal` - the method is a "signal". It must be a method declaration (without body). Its code
 ///   block will be expanded to emit the signal from the object path associated with the interface


zeenix:
> I've also changed some related code blocks. (wrong indents, wrong semi-colons)
> If you think they shouldn't be fixed, I will try to revert them.🩹

Such fixes are always welcome but please keep in mind that we follow atomic commits and each commit should only contain one logical change.

zeenix:
Since this is quite an intrusive change, I must first establish a compelling use case. Typically, individual members of an interface to depend on specific config is a sign of a bad API design, since interfaces in D-Bus are very dynamic and theoretically there is no limit on the number of interfaces an object implements so a good design splits the API into separate interfaces so conditional implementation/build can be done on per-interface level.

Anyway, there could be a use case (e.g a well-established interface you have to implement that you have no power over to change) but I'd like to hear the specific use case before considering this change.

KmolYuan:
> Anyway, there could be a use case (e.g a well-established interface you have to implement that you have no power over to change) but I'd like to hear the specific use case before considering this change.

I need to implement a central server for data exchange, where one client is responsible for setting the values of properties, and another client reads them (not in real-time interaction). There are 9 clients connected to this central hub, and some properties are broadcasted and shared among all clients. Some properties are not always peer to peer: Some front-line clients may panic, and we don’t want them to get stuck before a restart. We need a setter from another client to help save the situation.

Without using `cfg` attributes, everything still works functionally. However, as you can imagine, the proxy APIs become quite messy. While the server must implement all properties, it is also used in unit tests to simulate the behavior of other clients.

- If we define separate proxies for each client, it becomes difficult to consistently add or remove properties across both proxy and interface. My teammates and I occasionally forget to remove deprecated parts. Still, this is our current solution.

- If we define separate interfaces for each client and host them together on the server, zbus currently doesn’t provide a good synchronization mechanism across multiple interfaces. Since each interface is a separate struct, I have to manually sync data when one side sets a value. Even though I can use `Arc` and `Mutex`, I still need to figure out how to send the property changed signal.

The benefit of using `cfg` attributes is that we can generate different proxies and their mocked servers for unit testing based on crate features. This approach significantly reduces code size. Of course, I do acknowledge that D-Bus is designed to be more flexible than this.

zeenix:
> * If we define separate interfaces for each client and host them together on the server, zbus currently doesn’t provide a good synchronization mechanism across multiple interfaces.

There are at least 2 options here:

1. shared mutable state (`Arc<Mutex>` you mentioned is a typical way to achieve this)
2. keeping the state in another task/thread and communicate to it via channels.

Moreover this is a very internal detail to your implementation and should not be the deciding factor in the design of your D-Bus API IMO.

> * I still need to figure out how to send the property changed signal.

I can see the problem here if you are exposing the same exact property from multiple interfaces that represent the same underlying state. If so, I think that likely points to an API design problem. You probably want to split your interfaces more/better so while the underlying state is share but only one interface provides the property and you mutate the state through the property. #898 is something you'd want here.

KmolYuan:
> I can see the problem here if you are exposing the same exact property from multiple interfaces that represent the same underlying state. If so, I think that likely points to an API design problem. You probably want to split your interfaces more/better so while the underlying state is share but only one interface provides the property and you mutate the state through the property. https://github.com/dbus2/zbus/issues/898 is something you'd want here.

That would be great to hear that. And #898 provides a clean way to sync across multiple interfaces.

```rust
/* --- server main --- */
let _conn = builder
    .serve_at(PATH, ClientA)? // set foo
    .serve_at(PATH, ClientB)? // get foo
    .build()
    .await?;
std::future::pending::<()>().await;
unreachable!();

/* --- Currently --- */
#[interface(proxy, ..)]
impl ClientA {
    #[zbus(property)]
    pub fn set_foo(
        &self,
        val: String,
        #[zbus(object_server)]
        server: &zbus::ObjectServer,
    ) {
        *self.foo.lock().unwrap() = val;
        // Can be a macro like `sync_across!(server, ClientB, foo_changed)`
        let iface_ref = server.interface::<_, ClientB>(PATH).await.unwrap();
        iface_ref.get().foo_changed(iface_ref.signal_emitter()).await?;
    }
}

/* --- #898 --- */
#[interface(proxy, ..)]
impl ClientA {
    #[zbus(property)]
    pub fn set_foo(
        &mut self,
        val: MyEnumFlag, // Use atomics to use &self
        #[zbus(object_server)]
        server: &zbus::ObjectServer,
    ) {
        self.foo = val;
        (server.interface::<_, ClientB>(PATH).await.unwrap())
          .set_foo(val)?; // should emit foo changed signal of ClientB
    }
}
```

But the write-only properties are not supported yet in the `interface` macro currently, and this PR includes their implementation. Should we add this part to zbus? I can modify this PR to "Support write-only properties."

zeenix:
> That would be great to hear that. And #898 provides a clean way to sync across multiple interfaces.

I'm glad to hear you found a way to make things work w/o this PR.

```
        *self.foo.lock().unwrap() = val;
        // Can be a macro like `sync_across!(server, ClientB, foo_changed)`
        let iface_ref = server.interface::<_, ClientB>(PATH).await.unwrap();
        iface_ref.get().foo_changed(iface_ref.signal_emitter()).await?;
```

With #898, you'll not need to share the state like this anymore and can keep the state in one interface, which exposes that state as a property. Keep in mind also my suggestion that the same exact underlying property shouldn't be exposed by multiple interfaces.

> But the write-only properties are not supported yet in the `interface` macro currently, and this PR includes their implementation.

The reason write-only properties haven't been added all these years is that I've not yet come across any good use case for them. The main reason a property exists as a property is so interested clients can query their values and (in most cases) be notified of changes, which is impossible if the property is write-only.

> Should we add this part to zbus?

That would be great, yes. :) I created that issue myself for my own code so I'm very much for it. :)

> I can modify this PR to "Support write-only properties."

Sure, if you really think you need them.

KmolYuan:
> With https://github.com/dbus2/zbus/issues/898, you'll not need to share the state like this anymore and can keep the state in one interface, which exposes that state as a property. Keep in mind also my suggestion that the same exact underlying property shouldn't be exposed by multiple interfaces.

Oh wow, I forgot this. Yes, we can send the value to the getter's interface directly.

> The reason write-only properties haven't been added all these years is that I've not yet come across any good use case for them. The main reason a property exists as a property is so interested clients can query their values and (in most cases) be notified of changes, which is impossible if the property is write-only.

If the getter is exposed in another interface, it is a slightly more explicit choice than a method call.

> Sure, if you really think you need them.

OK, I'll try to remove the cfg parts of this PR.

zbus_macros/src/iface.rs
```diff
@@ -447,6 +441,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
             args_names,
             reply,
             member_name,
+            cfg_attrs,
```
zeenix, line 437: I appreciate your attention to details a lot here and also how you're keeping the commits atomic. :+1:  I also appreciate that you're adding the emojis but please use the appropriate one for each. E.g this isn't a bug fix but rather an internal refactor so ♻️ would be better.

zbus_macros/src/iface.rs
```diff
@@ -74,11 +74,11 @@ def_attrs! {
 }
 
 #[derive(Debug, Default)]
-struct Property<'a> {
+struct Property {
     read: bool,
     write: bool,
     emits_changed_signal: PropertyEmitsChangedSignal,
-    ty: Option<&'a Type>,
+    ty: Option<Type>,
```
zeenix, line 81: Why is this change needed?

zbus_macros/src/iface.rs
```diff
@@ -716,6 +723,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                              its setter method."
                         );
                         let prop_changed_method = quote!(
+                            #(#cfg_attrs)*
```
zeenix: Some unrelated changes in this commit.

zbus_macros/tests/tests.rs
```diff
@@ -312,6 +312,68 @@ fn test_interface() {
     }
 }
 
+#[test]
+fn test_multi_interface_sync() {
```
zeenix: Kudos for not only adding a unit test but also a more e2e test :+1: but let's instead edit [our zbus e2e test](https://github.com/dbus2/zbus/blob/main/zbus/tests/e2e.rs) to add a write-only property to the interface there and the proxy in there to test it. Reminder to self: that test file really needs to be split.

zbus_macros/src/lib.rs
```diff
@@ -363,6 +364,68 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+/// ```no_run
+/// use zbus_macros::interface;
+/// # const PATH: &str = "/org/freedesktop/zbus_macros/test_iface_sync";
+/// # const WELL_KNOWN_NAME: &str = "org.freedesktop.zbus_macros_iface_sync";
+/// struct ClientA;
+/// struct ClientB {
+///     foo: u32,
+/// }
+///
+/// #[interface(proxy, name = "org.freedesktop.zbus.ClientA")]
```
zeenix: nitpick: interfaces don't represent clients so I think this is not a good name. How about `InterfaceA` instead? Same for `ClientB`.

zbus_macros/src/lib.rs
```diff
@@ -363,6 +364,68 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// # Ok::<_, Box<dyn Error + Send + Sync>>(())
 /// ```
 ///
+/// # Synchronous between multiple interfaces
```
zeenix: Really nice! Although I feel like this would be a good example for #898. IMO we shouldn't advertise write-only properties as an internal synchronization mechanism since the property is still exposed on the bus to everyone.

Moreover, as I wrote, if the underlying property is the same, then it should be exposed by one interface only and clients shouldn't set and get it from different interfaces.

Of course there can be states in different interfaces that can depend on each other and have side-effects but write-only property isn't a good mechanism to handle that. Until #898, user can always manually add a (a normal, not D-Bus) setter method on an interface that other interfaces can call and this setter can emit the changed signal for the property.

zeenix (changes requested):
Really awesome job on this! However, I'm a perfectionist :see_no_evil: so I have a few concerns and wishes.

zbus_macros/src/iface.rs
KmolYuan: The setter's type is inferred from a value from the `method_info` that was dropped in the loop. So I need to copy it rather than reference it.

zbus_macros/src/lib.rs
KmolYuan: OK, I'll change the setter to a method call.

zbus_macros/tests/tests.rs
KmolYuan: Moved to `zbus/tests/e2e.rs`.

zbus_macros/src/iface.rs
KmolYuan: Okay, removing them generates more informative error messages.

zbus_macros/src/lib.rs
KmolYuan: Renamed `ClientA` -> `InterfaceA`, `ClientB` -> `InterfaceB`.

zbus_macros/src/iface.rs
```diff
@@ -393,20 +393,25 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
...
-                return Err(syn::Error::new(
-                    method.span(),
-                    "`emits_changed_signal` cannot be specified on setters",
-                ));
+            } else {
+                property.write = true;
+                if prop_attrs.emits_changed_signal.is_some() {
+                    return Err(Error::new_spanned(
+                        method,
+                        "Should not specify `emits_changed_signal` on setters",
```
zeenix: "Should" implies that it it's still possible. Let's keep the original error.

zbus/tests/e2e.rs
```diff
@@ -1093,3 +1093,72 @@ async fn iface_and_proxy_(#[allow(unused)] p2p: bool) {
     );
     debug!("Bus confirmed that all names were definitely released.");
 }
+
+#[test]
+fn test_multi_interface_sync() {
```
zeenix: Thanks for moving it but I meant that we add to the existing interface here, not a completely new test. Also the `zm` commit prefix need update. :)

zbus_macros/src/lib.rs
zeenix: That's not quite what I was getting at. :) I'd skip this for now and revisit it as part of the PR for #898.

zeenix (approved):
Almost there. Thanks!

zbus_macros/src/lib.rs
```diff
@@ -363,6 +364,67 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+///     .path(PATH)?
+///     .build()
+///     .await?;
+/// proxy.set_foo(42).await?;
+///
+/// let client_b = server
+///     .object_server()
+///     .interface::<_, InterfaceB>(PATH)
+///     .await?;
+/// assert_eq!(client_b.get().await.foo, 42);
```
zeenix:
```suggestion
/// let iface_b = server
///     .object_server()
///     .interface::<_, InterfaceB>(PATH)
///     .await?;
/// assert_eq!(iface_b.get().await.foo, 42);
```

KmolYuan: Fixed it.

zbus/tests/e2e.rs
KmolYuan: I added a write-only setter to that big interface in the `e2e` test.

The `test_multi_interface_sync` is duplicated with the docstring test, so I removed it.

zbus_macros/src/iface.rs
KmolYuan: Reverted.

zbus_macros/src/lib.rs
KmolYuan: I created a private setter for the interface B and called it from the interface A to demonstrate a synchronous mechanism.

zeenix: Thanks but I really don't think we need this here. It's not the only mechanism to synchronise between interfaces anyway and it's not relevant to this PR (anymore). I'd rather instead see a new question in [the FAQ section of our book](https://dbus2.github.io/zbus/faq.html), where we inform people about the different synchronization methods possible but not in this PR and better if it's done after (or as part of) the PR for #898.

zbus/tests/e2e.rs
zeenix: cool but it's not testing the use though. Can you edit the client-side of the test too to set it?

zbus_macros/src/lib.rs
KmolYuan: Sure, I can remove this from this PR.

zbus/tests/e2e.rs
KmolYuan: Added a `set_count` test for them.

PR #1414, opened 2025-06-29
⚡zb,zm: Proxy await properties

### LINKED ISSUE #1412: Proxy API to await specific changes in properties
Reference: [`tokio::sync::watch::Receiver::wait_for`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.wait_for).

In the dynamic version `wait_for` via `Proxy`, we can make a function for waiting for a property "foo" to become the expected value:
```rust
use futures::FutureExt as _;

async fn wait_for<T>(
    proxy: &zbus::proxy::Proxy,
    property_name: &str,
    predicate: impl FnMut(&T) -> bool,
) -> zbus::Result<T>
where
    T: TryFrom<zbus::zvariant::OwnedValue>,
    T::Error: Into<zbus::Error>,
{
    let mut stream = proxy.receive_property_changed::<T>(property_name).await;
    // If the value already meets the requirements, a single property_changed stream is blocked forever.
    // So we need to check it first.
    let mut value = proxy.get_property::<T>(property_name).await?;
    while !predicate(&value) {
        value = stream.next().await.unwrap().get().await?;
    }
    Ok(value)
}

tokio::select! {
    foo = wait_for(proxy, "foo", |foo| *foo != expected_foo).map(|r| r.unwrap()) => {
        /* Cancel the workflow */
        return Err(Error::FooCrash(foo))
    }
    _bar = wait_for(proxy, "bar", |bar| *bar == expected_bar).map(|r| r.unwrap()) => {
        /* Continue the workflow */
    }
}
```
The static version methods are hard-coded in the method name, so we cannot generalize getters and property_changed streams.
I wish to have `proxy.wait_for_foo(predicate)`:
```rust
tokio::select! {
    foo = proxy.wait_for_foo(|foo| *foo != expected_foo).map(|r| r.unwrap()) => {
        /* Cancel the workflow */
        return Err(Error::FooCrash(foo))
    }
    _bar = proxy.wait_for_bar(|bar| *bar == expected_bar).map(|r| r.unwrap()) => {
        /* Continue the workflow */
    }
}
```
when `emits_changed_signal` on the property is `"true"` or `"invalidates"`. Or add an extra option.

KmolYuan:
Fix #1412.

Add `wait_property_for` method for `Proxy`/`blocking::Proxy` and generate `wait_*_for` methods in the `proxy` macro.

### DIFF
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -216,7 +216,29 @@ pub struct PropertyStream<'a, T> {
     name: &'a str,
     proxy: Proxy<'a>,
     changed_listener: EventListener,
+    peek_current: bool,
     phantom: std::marker::PhantomData<T>,
+}
+
+impl<T> PropertyStream<'_, T> {
+    /// Yield the current property value once without waiting for the next update.
+    ///
+    /// This is useful if you want to check the current value of a property without waiting for the
+    /// next update. This option works independently of the [`CacheProperties`] option.
+    ///
+    /// Note that zbus doesn’t queue the updates. If the listener is slower than the receiver, it
+    /// will only receive the last update.
+    pub fn with_current(mut self) -> Self {
+        self.peek_current_once();
+        self
+    }
+
+    /// Inplace version of [`PropertyStream::with_current()`].
+    ///
+    /// This stream will yield the current property value on the next item without waiting changes.
+    pub fn peek_current_once(&mut self) {
+        self.peek_current = true;
+    }
 }
 
 impl<'a, T> stream::Stream for PropertyStream<'a, T>
@@ -232,7 +254,12 @@ where
             // With no cache, we will get no updates; return immediately
             None => return Poll::Ready(None),
         };
-        ready!(Pin::new(&mut m.changed_listener).poll(cx));
+
+        if m.peek_current {
+            m.peek_current = false;
+        } else {
+            ready!(Pin::new(&mut m.changed_listener).poll(cx));
+        }
 
         m.changed_listener = properties
             .values
@@ -984,6 +1011,7 @@ impl<'a> Proxy<'a> {
             name,
             proxy: self.clone(),
             changed_listener,
+            peek_current: false,
             phantom: std::marker::PhantomData,
         }
     }
--- a/zbus/tests/e2e.rs
+++ b/zbus/tests/e2e.rs
@@ -758,6 +758,12 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     );
     my_obj_proxy.ping().await?;
 
+    {
+        // Test peek_current option. If this option is not set, the stream will blocked forever.
+        let mut stream = my_obj_proxy.receive_count_changed().await.with_current();
+        assert_eq!(stream.next().await.unwrap().get().await.unwrap(), 0);
+    }
+
     let mut ifaces_removed_stream = obj_manager_proxy.receive_interfaces_removed().await?;
     debug!("Created: {:?}", ifaces_removed_stream);
     // Must process in parallel, so the stream listener does not block receiving


zbus/src/proxy/mod.rs
```diff
@@ -1018,6 +1018,49 @@ impl<'a> Proxy<'a> {
...
+    ///
+    /// If the property already meets the predicate, this function will return immediately.
+    ///
+    /// If the property is cachable, this function will poll the property stream until the predicate
+    /// is met. If the property is uncacheable, this function will busy await until the predicate is
+    /// met. (not recommended)
+    ///
+    /// Note that zbus doesn't queue the updates. If the listener is slower than the receiver, it
+    /// will only receive the last update.
+    pub async fn wait_property_for<T>(
```
zeenix: I'm now wondering, isn't this achieving the same as you could with [futures API](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.filter) easily? 🤔

KmolYuan: It's a convenience function of
```rust
use futures::{FutureExt as _, StreamExt as _};

let name;
let predicate;
let fut = {
    let event_stream = (proxy.receive_property_changed(name).await)
        .then(|ev| async move { ev.get().await });
    let mut stream = std::pin::pin!(
        futures::stream::once(proxy.get_property::<T>(name).await)
            .chain(event_stream) // impl Stream<Item = zbus::Result<T>>
            .filter(predicate)
    );
    // the future roughly equivalent to `proxy.wait_property_for(name, predicate)`
    stream.next().map(|item| item.unwrap()) // impl Future<Output = zbus::Result<T>>
};
```

zeenix: > It's a convenience function of

Right but do you really need to create a future here? :thinking: 

```rust
use futures::{FutureExt as _, StreamExt as _, TryStreamExt as _};

let name;
let predicate;

let event_stream = (proxy.receive_property_changed(name).await)
    .then(|ev| ev.get())
    .try_filter(predicate);
let next_prop_value = event_stream.try_next().await?;
```

Something like that should work, no? Our binary size is already big (#304) so I'm just wondering if this is a frequently used enough API to justify its existence in zbus, if futures already provide an API to easily do what you need.

KmolYuan: Well, the shortcoming of the property changed stream is that it doesn't return the current value to check if the current value already meets the requirement. Otherwise, the `receive_property_changed` method is enough.

zeenix: > Well, the shortcoming of the property changed stream is that it doesn't return the current value to check if the current value already meets the requirement.

Huh? That's what the `try_filter` in my example do.

zeenix: Just to be clear, I'm not saying that PropertyChangedStream already does the whole thing you need. My point was that combined with futures (which is already our dep) API, what you need can likely be achieved with a few LOC.

KmolYuan: > Just to be clear, I'm not saying that PropertyChangedStream already does the whole thing you need. My point was that combined with futures (which is already our dep) API, what you need can likely be achieved with a few LOC.

Sure, adding a filter or checking values in loops is the solution we are currently using. But based on the reliance on compile-time checking (property names), we made separate methods for different properties due to they are difficult to generalize.

If there are some settings on the `PropertyStream` type, it can also solve the problem.
```rust
let mut stream = proxy.receive_foo_changed().await.with_current();
```

zeenix: > Sure, adding a filter or checking values in loops is the solution we are currently using.

That's not the solution I suggested but that works too. The main point is that as long as it's easy and needs a few LOCs, probably not a good idea to add more API to zbus unless we've reasons to believe that it's going to be useful for a big percentage of our users.

> . But based on the reliance on compile-time checking (property names), we made separate methods for different properties due to they are difficult to generalize.

You can create a generic API that deals with `PropertyStream` so you can use it with the result so any `receive_*_changed` method?

KmolYuan: I made another implementation that is smaller than the old one. Based on the `PropertyStream<'_, T>`.

zeenix: > I made another implementation that is smaller than the old one. Based on the `PropertyStream<'_, T>`.

In my previous comment, I meant how you can easily do this in your code base w/o needing any changes in zbus. Your new changes are completely different from the aim of this PR so should have been a separate PR.

zbus/src/proxy/mod.rs
```diff
@@ -216,9 +216,31 @@ pub struct PropertyStream<'a, T> {
     name: &'a str,
     proxy: Proxy<'a>,
     changed_listener: EventListener,
+    peek_current: bool,
     phantom: std::marker::PhantomData<T>,
 }
 
+impl<T> PropertyStream<'_, T> {
+    /// Yield the current property value once without waiting for the next update.
```
zeenix, line 224: I'm not sure about the usefulness of this API. There is already API provide by the Proxy and the macro-generated proxies to read the current property and get the locally cached value as well (which is immediate). :thinking:

zeenix:
I'm sorry but I'm closing this now. If we agree on the latest commits, they should be split into a separate PR anyway.

zbus/src/proxy/mod.rs
KmolYuan: @zeenix If I remember correctly, if the value is already cached, `PropertyStream`s will wait for the next change instead of retrieving the value immediately. And sometimes we need to re-check the value again.

zeenix: > @zeenix If I remember correctly, if the value is already cached, `PropertyStream`s will wait for the next change instead of retrieving the value immediately.

True. I think we could change that instead of adding new API.

KmolYuan: > True. I think we could change that instead of adding new API.

`tokio` has a [`mark_changed`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.mark_changed)/`mark_unchanged` API for rechecking values from watch channels, which is similar to our circumstance.

zeenix: I don't see how that's the same/similar. That API is for letting you know the value has changed.

KmolYuan: > I don't see how that's the same/similar. That API is for letting you know the value has changed.

Adding an API allows user to control the rechecking behavior. However that maybe unuseful due to streams/iterators are almost wrapped. Without it, yielding the current value on the first element is the simplest way.

zeenix: > Without it, yielding the current value on the first element is the simplest way.

As I wrote, we can add that behaviour to the existing property stream w/o any new API. If the value is cached, it yields that value first, if caching is disabled or the value isn't cached yet, it doesn't do that.

PR #1419, opened 2025-06-30
🚑️ zv: Check signature before serializing struct as a u8

### LINKED ISSUE #1417: org.freedesktop.zbus.Error: invalid length 19, expected >= 20
We are getting message like these when zvariant 5.4 and 5.5 are communicating with each other. I don't know for sure, but I could imagine that the following byte difference is causing the issue.

```
fn main() {
    let ctxt = zvariant::serialized::Context::new_dbus(zvariant::LE, 0);
    let data = zvariant::to_bytes(ctxt, &X::default()).unwrap();

    println!("{:?}", data.bytes());
    // zvariant 5.4.0 output:
    // [0, 0, 0, 0, 0, 0, 0, 0]

    // zvariant 5.5.1, 5.5.2, 5.5.3 output:
    // [0]
}

#[derive(zvariant::DeserializeDict, zvariant::SerializeDict, zvariant::Type, Debug, Default)]
#[zvariant(signature = "dict")]
pub struct X {
    pub scale: Option<u32>,
}
```

zeenix:
This fixes a regression that was exposed by 56cbb1f64.
    
Fixes #1417.

### DIFF
--- a/zvariant/src/dbus/ser.rs
+++ b/zvariant/src/dbus/ser.rs
@@ -326,22 +326,19 @@ where
     }
 
     fn serialize_struct(self, _name: &'static str, len: usize) -> Result<Self::SerializeStruct> {
-        if len == 0 {
-            return StructSerializer::unit(self).map(StructSeqSerializer::Struct);
-        }
-
         self.0
             .add_padding(self.0.signature.alignment(self.0.ctxt.format()))?;
         match &self.0.signature {
             Signature::Variant => StructSerializer::variant(self).map(StructSeqSerializer::Struct),
             Signature::Array(_) => self.serialize_seq(Some(len)).map(StructSeqSerializer::Seq),
+            Signature::U8 => StructSerializer::unit(self).map(StructSeqSerializer::Struct),
             Signature::Structure(_) => {
                 StructSerializer::structure(self).map(StructSeqSerializer::Struct)
             }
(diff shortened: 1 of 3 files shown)


decathorpe:
Do you expect to publish a new release of zvariant with this fix? I'm working on all the Rust dependencies for GNOME 49 and [snapshot pinning to zvariant ~5.4](https://gitlab.gnome.org/GNOME/snapshot/-/merge_requests/382) currently prevents us from building it - and just relaxing the dependency would just expose this bug IIUC?

EDIT: If no immediate release is planned, I can just backport this patch to zvariant 5.5 that we have in Fedora, too.

zeenix:
> EDIT: If no immediate release is planned, I can just backport this patch to zvariant 5.5 that we have in Fedora, too.

Usually I don't have a timeline. I'll release soon now that I know it's blocking stuff..

decathorpe:
Thanks - then there is no urgency, I've included the patch from this PR in the Fedora package for zvariant v5 for now. :)

zeenix:
@decathorpe zvariant 5.6 published to crates.io. I'll prep the release notes and tags soon..

decathorpe:
That works too. Thank you 🙌🏼

zeenix:
> That works too. Thank you 🙌🏼

[All done](https://github.com/dbus2/zbus/releases/tag/zvariant-5.6.0).

PR #1426, opened 2025-07-09
✨ zb: Property stream will yield the current value on the first element

KmolYuan:
Uses a dummy event to make the property stream yield the current value immediately when the cache is available.

### DIFF
--- a/zbus/src/blocking/proxy/mod.rs
+++ b/zbus/src/blocking/proxy/mod.rs
@@ -325,14 +325,7 @@ impl<'a> Proxy<'a> {
             .map(SignalIterator)
     }
 
-    /// Get an iterator to receive owner changed events.
-    ///
-    /// If the proxy destination is a unique name, the stream will be notified of the peer
-    /// disconnection from the bus (with a `None` value).
-    ///
-    /// If the proxy destination is a well-known name, the stream will be notified whenever the name
-    /// owner is changed, either by a new peer being granted ownership (`Some` value) or when the
-    /// name is released (with a `None` value).
+    /// Get an iterator to receive property changed events.
     ///
     /// Note that zbus doesn't queue the updates. If the listener is slower than the receiver, it
     /// will only receive the last update.
@@ -343,7 +336,14 @@ impl<'a> Proxy<'a> {
         PropertyIterator(block_on(self.inner().receive_property_changed(name)))
     }
 
-    /// Get an iterator to receive property changed events.
+    /// Get an iterator to receive owner changed events.
+    ///
+    /// If the proxy destination is a unique name, the stream will be notified of the peer
+    /// disconnection from the bus (with a `None` value).
+    ///
+    /// If the proxy destination is a well-known name, the stream will be notified whenever the name
+    /// owner is changed, either by a new peer being granted ownership (`Some` value) or when the
+    /// name is released (with a `None` value).
     ///
     /// Note that zbus doesn't queue the updates. If the listener is slower than the receiver, it
     /// will only receive the last update.


zeenix:
> Fix #1412

It does not. :) That issue is about awaiting specific changes while this PR is about changing the behaviour of property changed stream a little.

KmolYuan:
> > Fix #1412
> 
> It does not. :) That issue is about awaiting specific changes while this PR is about changing the behaviour of property changed stream a little.

Removed it.

zbus/src/proxy/mod.rs
```diff
@@ -964,21 +964,24 @@ impl<'a> Proxy<'a> {
...
-            Event::new().listen()
-        };
+                .or_default();
+        }
+
+        // Set a dummy event to start the stream.
+        // The stream will yield the current value immediately when the cache is available.
+        let event = Event::new();
+        let changed_listener = event.listen();
+        event.notify(1);
```
zeenix: Maybe I'm just being stupid here but do we really need to change any logic here? Wouldn't it be enough to just add the `event.notify` call here? :thinking:

zbus/tests/e2e.rs
```diff
@@ -758,6 +758,23 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     );
     my_obj_proxy.ping().await?;
 
+    // Test cached property-changed stream
```
zeenix: High marks for taking care to add tests but I think it would be better to just update the existing code here, which already creates a stream for `Count` property change (line 749 just above).

zbus/src/blocking/proxy/mod.rs
```diff
@@ -325,14 +325,7 @@ impl<'a> Proxy<'a> {
...
 
-    /// Get an iterator to receive owner changed events.
-    ///
-    /// If the proxy destination is a unique name, the stream will be notified of the peer
-    /// disconnection from the bus (with a `None` value).
-    ///
-    /// If the proxy destination is a well-known name, the stream will be notified whenever the name
-    /// owner is changed, either by a new peer being granted ownership (`Some` value) or when the
-    /// name is released (with a `None` value).
+    /// Get an iterator to receive property changed events.
```
zeenix, line 328: Nice!

zbus/tests/e2e.rs
KmolYuan: I'm not sure if the new test will break the logic of testing `cache_count`, and the new test is focused on cached properties.

zbus/src/proxy/mod.rs
KmolYuan: > Wouldn't it be enough to just add the event.notify call here? 🤔

`EventListener`'s API is quite difficult to get the source on the above `if let Some(...) = ... {} else {}` to call notify.🥲 (line 972)

Also, the `PropertyStream::poll_next()` resets the `changed_listener` in each yield. (line 237)
And if the event listener's source is the same as the property cache pool and the created stream is not used, I'm afraid of wrongly notifying other streams in different tasks. (or may be no side effects?)

zeenix: I'm not sure if I understand what you're saying but this also breaks the listener for the subsequent updates since you're creating only a transient `Event` and then dropping it and the `changed_listender` is now tied to that so it will never be woken up again after the first event. If you do a more thorough testing, you'll see what I mean.

KmolYuan: OK, updated.

zbus/src/proxy/mod.rs
```diff
@@ -964,7 +964,8 @@ impl<'a> Proxy<'a> {
     /// Note that zbus doesn't queue the updates. If the listener is slower than the receiver, it
     /// will only receive the last update.
     ///
-    /// If caching is not enabled on this proxy, the resulting stream will not return any events.
+    /// The stream will yield the current value on the first element. If caching is not enabled on
```
zeenix: This wording makes it sound like the stream will only return the current value. This needs better docs.

zbus/src/proxy/mod.rs
```diff
@@ -975,7 +976,9 @@ impl<'a> Proxy<'a> {
             let entry = values
                 .entry(name.to_string())
                 .or_insert_with(PropertyValue::default);
-            entry.event.listen()
+            let listener = entry.event.listen();
+            entry.event.notify(1);
```
zeenix: if the value is not yet cached, this will end up notifying unnecessarily, no? :thinking:  Can we check for the value being cached here?

zbus/tests/e2e.rs
zeenix: It doesn't break it if you update the logic appropriately. We also need to keep the code to minimum here (there is a lot of code in this file already and hard to keep track of things).

zbus/src/proxy/mod.rs
KmolYuan: Updated docs. (29e3b045ca466ae7967b0fe7a150299c8c82b76f)

zbus/src/proxy/mod.rs
KmolYuan: Checked `entry.value`. (29e3b045ca466ae7967b0fe7a150299c8c82b76f)

zbus/tests/e2e.rs
KmolYuan: Reuse the stream object on line 749. (3d7377bc3c5860bc86e3cd7b8e437f9446d29466)

zbus/tests/e2e.rs
```diff
@@ -758,6 +758,17 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
 
+    // Test the property-changed stream values
+    assert_eq!(stream.next().await.unwrap().get().await?, 0);
+    for test_case in [55, 66, 77, 88] {
+        let (value, ret) = futures_util::join!(
+            async { stream.next().await.unwrap().get().await.unwrap() },
+            my_obj_proxy.set_count(test_case)
+        );
+        ret.unwrap();
+        assert_eq!(test_case, value);
```
zeenix: Thanks. nitpick: let's not use `unwrap` where we can use `?`.

zbus/tests/e2e.rs
```diff
@@ -758,6 +758,17 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     );
     my_obj_proxy.ping().await?;
 
+    // Test the property-changed stream values
+    assert_eq!(stream.next().await.unwrap().get().await?, 0);
```
zeenix, line 762: Add a comment above so it's clear what's going on:

```rust
// The stream should first yield the current value.
```

zbus/tests/e2e.rs
```diff
@@ -758,6 +758,17 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+    for test_case in [55, 66, 77, 88] {
+        let (value, ret) = futures_util::join!(
+            async { stream.next().await.unwrap().get().await.unwrap() },
+            my_obj_proxy.set_count(test_case)
+        );
+        ret.unwrap();
```
zeenix: We can handle the ret right in the `join` arm and then just do `let (value, _)` on line 764. That's more consistent and easier to read.

zbus/tests/e2e.rs
KmolYuan: Fixed in de1d1d1428d8dde5b05b7622ca1181c533d4935a.

zbus/tests/e2e.rs
KmolYuan: Fixed in de1d1d1428d8dde5b05b7622ca1181c533d4935a.

zbus/tests/e2e.rs
KmolYuan: Changed to `unwrap` in futures. (de1d1d1428d8dde5b05b7622ca1181c533d4935a)

zbus/tests/e2e.rs
```diff
@@ -758,6 +758,16 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
     );
     my_obj_proxy.ping().await?;
 
+    // The stream should first yield the current value.
+    assert_eq!(stream.next().await.unwrap().get().await?, 0);
+    for test_case in [55, 66, 77, 88] {
+        let (value, ()) = futures_util::join!(
```
zeenix: `let (value, _)` is more readable IMO and hence why I suggested that. :)

zbus/tests/e2e.rs
zeenix: That's the opposite of what I asked for. :)

zbus/tests/e2e.rs
KmolYuan: Changed to `let (value, _) = try_join!(..)?;` pattern for this. (2b8079bb89e81a28e2cf50cfb45b602eaa94a85a)

zbus/src/blocking/proxy/mod.rs
zeenix, line 328: For future reference, this isn't a refactor. 📝 would have been a more appropriate emoji prefix for the commit.

PR #1429, opened 2025-07-11
🧑‍💻 zb: add fdo::dbus::StartServiceReply type

valpackett:
In 6.0 this will be the return type of start_service_by_name. For now, just provide a TryFrom<u32>. We want to use it in busd already though.

---

From https://github.com/dbus2/busd/pull/238#discussion_r2201345696

### DIFF
--- a/zbus/src/fdo/dbus.rs
+++ b/zbus/src/fdo/dbus.rs
@@ -105,6 +105,33 @@ pub enum ReleaseNameReply {
     NotOwner = 0x03,
 }
 
+/// The return code of the [`DBusProxy::start_service_by_name`] method.
+///
+/// In zbus 6.0, this will become the return type of `start_service_by_name`.
+/// For now, it's provided separately with a `TryFrom<u32>` implementation
+/// to avoid breaking changes in the API.
+#[repr(u32)]
+#[derive(Deserialize_repr, Serialize_repr, Type, Debug, PartialEq, Eq)]
+pub enum StartServiceReply {
+    /// The service was successfully started.
+    Success = 0x01,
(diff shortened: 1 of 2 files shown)


zbus/src/fdo/dbus.rs
```diff
@@ -105,6 +105,31 @@ pub enum ReleaseNameReply {
     NotOwner = 0x03,
 }
 
+/// The return code of the [`start_service_by_name`] method.
+///
+/// [`start_service_by_name`]: struct.DBusProxy.html#method.start_service_by_name
```
zeenix: * Let's add a note here why this is not yet the return type of `start_service_by_name` and that it will be in 6.0.
* Let's make use of the rustdoc feature to link and use the full path:

```suggestion
/// The return code of the [`DBusProxy::start_service_by_name`] method.
```

zbus/src/fdo/dbus.rs
```diff
@@ -105,6 +105,31 @@ pub enum ReleaseNameReply {
...
+#[repr(u32)]
+#[derive(Deserialize_repr, Serialize_repr, Type, Debug, PartialEq, Eq)]
+pub enum StartServiceReply {
+    /// The service was successfully started.
+    Success = 0x01,
+    /// The service was already running.
+    AlreadyRunning = 0x02,
+}
+
+// The response type is kept as u32 until the next ABI break, so provide this for now.
```
zeenix: nitpick: Add `FIXME: ` prefix. There are highlighters in IDEs for those. :)

zeenix (commented):
LGTM otherwise.

PR #1432, opened 2025-07-14
Add an interface_unchecked function to object_server

sander-machado:
The current interface function requires a read lock to be acquired on the interface before returning the references. It needs this to check if the interface can be safely downcasted. This new function does not contain that check allowing it be used both internally and externally, to acquire a shared reference to the current interface. This is especially useful when annotating the functions with signalreceiver is not possible or desired.

I have some very niche usecases where annotating our function with SignalReceiver is not viable and this function would help a lot for my usecase so I hope you'll consider accepting this PR.

### DIFF
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -272,7 +272,7 @@ impl ObjectServer {
             .instance
             .clone();
 
-        // Ensure what we return can later be dowcasted safely.
+        // Ensure what we return can later be downcasted safely.
         lock.read()
             .await
             .downcast_ref::<I>()
@@ -280,7 +280,94 @@ impl ObjectServer {
 
         let conn = self.connection();
         // SAFETY: We know that there is a valid path on the node as we already converted w/o error.
-        let emitter = SignalEmitter::new(&conn, path).unwrap().into_owned();
+        let emitter = SignalEmitter::new(&conn, path)?.into_owned();
+
+        Ok(InterfaceRef {
+            emitter,
+            lock,
+            phantom: PhantomData,
+        })
+    }
+
+    /// Get the interface at the given path, but without checking if it can later be downcasted
+    /// this can cause the `InterfaceRef` to be invalid if the interface at the path is not of
+    /// the expected type.
+    ///
+    /// # Errors
+    ///
+    /// If the interface is not registered at the given path, an `Error::InterfaceNotFound` error is
+    /// returned.
+    ///
+    /// # Examples
+    ///
+    /// The typical use of this is property changes inside a dispatched handler where you don't
+    /// want to annotate the function with a signal handler:
+    ///
+    /// ```no_run
+    /// # use std::error::Error;
+    /// # use zbus::{Connection, interface};
+    /// # use async_io::block_on;
+    /// #
+    /// struct MyIface {
+    ///     count: u32,
+    ///     connection: Connection,
+    /// }
+    ///
+    /// #[interface(name = "org.myiface.MyIface")]
+    /// impl MyIface {
+    ///      #[zbus(property)]
+    ///      async fn count(&self) -> u32 {
+    ///          self.count
+    ///      }
+    ///
+    ///     async fn update_count(&mut self) -> zbus::Result<()> {
+    ///         // gets a reference to its own interface without having to annotate the function with signal handler
+    ///         // using `interface_unchecked` instead of `interface` allows you to avoid the deadlock
+    ///         let iface_ref = self.connection
+    ///             .object_server()
+    ///             .interface_unchecked::<_, MyIface>("/my/path").await?;
+    ///          self.count = 42;
+    ///         // emits a signal to notify the change
+    ///         self.count_changed(iface_ref.signal_emitter()).await
+    ///     }
+    /// }
+    ///
+    /// # block_on(async {
+    /// # let connection = Connection::session().await?;
+    /// #
+    /// # let path = "/org/zbus/path";
+    /// # connection.object_server().at(path, MyIface {
+    /// #    count: 0,
+    /// #    connection: connection.clone(),
+    /// # }).await?;
+    /// let iface_ref = connection
+    ///     .object_server()
+    ///     .interface::<_, MyIface>(path).await?;
+    /// let mut iface = iface_ref.get_mut().await;
+    /// iface.update_count().await?; // will send a property changed signal as well now
+    /// # Ok::<_, Box<dyn Error + Send + Sync>>(())
+    /// # })?;
+    /// #
+    /// # Ok::<_, Box<dyn Error + Send + Sync>>(())
+    /// ```
+    pub async fn interface_unchecked<'p, P, I>(&self, path: P) -> Result<InterfaceRef<I>>
+    where
+        I: Interface,
+        P: TryInto<ObjectPath<'p>>,
+        P::Error: Into<Error>,
+    {
+        let path = path.try_into().map_err(Into::into)?;
+        let root = self.root().read().await;
+        let node = root.get_child(&path).ok_or(Error::InterfaceNotFound)?;
+
+        let lock = node
+            .interface_lock(I::name())
+            .ok_or(Error::InterfaceNotFound)?
+            .instance
+            .clone();
+
+        let conn = self.connection();
+        let emitter = SignalEmitter::new(&conn, path)?.into_owned();
 
         Ok(InterfaceRef {
             emitter,
--- a/zbus/tests/issue/issue_1430.rs
+++ b/zbus/tests/issue/issue_1430.rs
@@ -1,0 +1,46 @@
+use zbus::{interface, Connection};
+
+struct TestInterface {
+    dbus_connection: Connection,
+    count: u8,
+}
+
+#[interface(name = "org.zbus.MyGreeter1")]
+impl TestInterface {
+    // Can be `async` as well.
+    async fn say_hello(&mut self, name: &str) -> String {
+        self.count += 1;
+        let _interface = self
+            .dbus_connection
+            .object_server()
+            .interface_unchecked::<_, TestInterface>("/test")
+            .await
+            .unwrap();
+        format!("Hello {}! I have been called {} times.", name, self.count)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::issue::issue_1430::TestInterface;
+
+    #[tokio::test]
+    async fn test_deadlock() {
+        let connection = zbus::connection::Connection::session().await.unwrap();
+        let test_interface = TestInterface {
+            count: 0,
+            dbus_connection: connection.clone(),
+        };
+        connection
+            .object_server()
+            .at("/test", test_interface)
+            .await
+            .unwrap();
+        let interface = connection
+            .object_server()
+            .interface::<_, TestInterface>("/test")
+            .await
+            .unwrap();
+        let _ = interface.get_mut().await.say_hello("deadlock").await;
+    }
+}
--- a/zbus/tests/issue/mod.rs
+++ b/zbus/tests/issue/mod.rs
@@ -3,6 +3,7 @@ mod issue_104;
 mod issue_121;
 #[cfg(feature = "blocking-api")]
 mod issue_122;
+mod issue_1430;
 mod issue_173;
 mod issue_260;
 mod issue_466;


zbus/src/object_server/mod.rs
```diff
@@ -272,15 +272,101 @@ impl ObjectServer {
...
+    {
+        let path = path.try_into().map_err(Into::into)?;
+        let root = self.root().read().await;
+        let node = root.get_child(&path).ok_or(Error::InterfaceNotFound)?;
+
+        let lock = node
+            .interface_lock(I::name())
+            .ok_or(Error::InterfaceNotFound)?
+            .instance
+            .clone();
```
zeenix, line 367: I'm puzzled how this can work. The issue is not downcasting but rather multiple parts of the code needing to lock the interface at the same time. When a method is called, it's already locked for you (that's why you can get a `&mut self` receiver) so it shouldn't be possible (at least using safe Rust) to access the interface again. Here you are still locking the interface and returning the lock in the `InterfaceRef` returned. :thinking: I started the CI to see if you tests actually pass somehow. I'd still like an explanation how how it is working.

zeenix (changes requested):
Other than that, please follow our contribution guidelines to the best of you abilities. :pray:

zbus/src/object_server/mod.rs
```diff
@@ -272,15 +272,102 @@ impl ObjectServer {
...
+    ///
+    ///     async fn update_count(&mut self) -> zbus::Result<()> {
+    ///         // gets a reference to its own interface without having to annotate the function with signal handler
+    ///         // using `interface_unchecked` instead of `interface` allows you to avoid the deadlock
+    ///         let iface_ref = self.connection
+    ///             .object_server()
+    ///             .interface_unchecked::<_, MyIface>("/my/path").await?;
+    ///          self.count = 42;
+    ///         // emits a signal to notify the change
+    ///         self.count_changed(iface_ref.signal_emitter()).await
```
zeenix, line 331: Hmm.. even if it works, this is a very strange way of doing it and **a lot** more work than just adding a signal emitter argument. I don't quite understand why you can't add extra arguments to your methods (even if the code is generated) so I'm sorry but I'm not going to be adding it for just one project since the chances of anyone else having the same limitation as you, are extremely low IMO.

zeenix:
Closing based on https://github.com/dbus2/zbus/pull/1432#discussion_r2204336038

zbus/src/object_server/mod.rs
sander-machado: It does look like that here, but I believe this should allow an extension in the future where you can create a set property function that also emits signals from inside and outside the interface without having the pass the signalreceiver as an argument, it's how we use it right now already.

sander-machado: I think it basically just gives you a reference to the signal emitter without needing to pass it as an argument also was quite surprised that it worked. removing the check in the interface function itself also still lets all tests pass

zbus/src/object_server/mod.rs
sander-machado: It's because I only need to acces the signal emitter, maybe if I change it so it doesn't return an InterfaceRef but only the signal emitter that would make more sense?

sander-machado: because in the end all I'm really trying to do is get the signal emitter without having to annotated the function. I want to avoid having the always pass the signal emitter manually when I call the function from an external location.

zbus/src/object_server/mod.rs
zeenix: > but I believe this should allow an extension in the future where you can create a set property function that also emits signals from inside and outside the interface without having the pass the signalreceiver as an argument, it's how we use it right now already.

When you've a solution for that, we can chat. :) I would also love nothing more than removing the need for signal emittor.

> utside the interface without having the pass the signalreceiver as an argument, it's how we use it right now already.

Oh! Are you calling the D-Bus method (methods defined inside the `interface` macro's impl block) from outside? You should not need to do that and should not do. You should instead add a normal method that handle the SignalReceiver creation as a detail (or use #898 when it's implemented).

zeenix: > I think it basically just gives you a reference to the signal emitter without needing to pass it as an argument also was quite surprised that it worked.

I'm too. How does it work? If this can work then we can certainly add better API (or improve the existing one) but we can't just rely on "oh the tests pass so it works" because you never know what you missed in the tests. We need to know how exactly it works.

zbus/src/object_server/mod.rs
zeenix: I know what you end-goal here is but that's not the question. The question is how you're avoiding the deadlock if you're still locking the interface.

zbus/src/object_server/mod.rs
sander-machado: > > I think it basically just gives you a reference to the signal emitter without needing to pass it as an argument also was quite surprised that it worked.
> 
> I'm too. How does it work? If this can work then we can certainly add better API (or improve the existing one) but we can't just rely on "oh the tests pass so it works" because you never know what you missed in the tests. We need to know how exactly it works.

> > but I believe this should allow an extension in the future where you can create a set property function that also emits signals from inside and outside the interface without having the pass the signalreceiver as an argument, it's how we use it right now already.
> 
> When you've a solution for that, we can chat. :) I would also love nothing more than removing the need for signal emittor.
> 
> > utside the interface without having the pass the signalreceiver as an argument, it's how we use it right now already.
> 
> Oh! Are you calling the D-Bus method (methods defined inside the `interface` macro's impl block) from outside? You should not need to do that and should not do. You should instead add a normal method that handle the SignalReceiver creation as a detail (or use #898 when it's implemented).

Indeed [#898](https://github.com/dbus2/zbus/issues/898) was what prompted this, I wrote a custom #interface macro that does a bunch of specific stuff for, generated types based on dbus xml annotations etc that are used in different parts of the code aswell as generates a custom setter function that then in turn calls the dbus set function followed but a sending the model updated signal almost exactly like what's described in that issues. 
it looks a bit like this, maybe something that could be added to the proper zbus macro in the future? 
```Rust
#[interface(
    idl_path = "example-service/src/libexampleservice/example.idl2",
    type_overwrite = "set_sound>CustomSound"
)]
impl ExampleTrait for ExampleService<ExampleType> {
    /// Sound property
    #[zbus(property)]
    async fn sound(&self) -> Sound {
        self.lock().sound
    }
    #[zbus(property(internal = true))]
    async fn set_sound(&mut self, value: CustomSound) -> Result<(), zbus::fdo::Error> {
        println!("SetSound: {value:?}");
        self.lock().sound = value.get_sound();
        Ok(())
    }
}
```
Then I have some code gen that implements a function on the struct that looks like this.
First it will call the dbus property externally and internally. But because it's marked as internal you know it's an actual dbus method (it worked for us). then it does a tokio::spawn to get around this deadlocking issue. With this patch it works without the tokio::spawn.
```Rust
pub async fn #pt_set_name(&mut self, value: #overwritten_ty, path: &str) -> zbus::fdo::Result<()> {
    self.#set_name(value).await?;
    let object_server = self.dbus_server.conn.object_server().clone();
    let path = String::from(path.clone());
    tokio::spawn(async move {
        let iface_ref = object_server
        .interface::<_, Self>(path.clone())
        .await.expect(format!("Could not get interface at {}", path).as_str());
        let iface = iface_ref.get_mut().await;
        iface.#signal_name(iface_ref.signal_emitter()).await.expect("Could not emit signal");
    });
    Ok(())
}
```

I agree that just assuming oh the tests pass is not a good enough argument but maybe this way of working could be interesting for zbus as a whole? I also added support for generating a trait based on a dbus xml file to my interface macro but that's mostly because we have thousands of dbus interfaces as it's what we use for almost all IPC.

zbus/src/object_server/mod.rs
sander-machado: You only take a clone of the Arc<RwLock<>> and you can access the signal emitter reference without needing to take the lock. I think that's why it works.

zbus/src/object_server/mod.rs
zeenix: > I agree that just assuming oh the tests pass is not a good enough argument but maybe this way of working could be interesting for zbus as a whole?

You're contradicting yourself. :) This is not how good engineering works. You need to understand what your code is doing.

sander-machado: > > I agree that just assuming oh the tests pass is not a good enough argument but maybe this way of working could be interesting for zbus as a whole?
> 
> You're contradicting yourself. :) This is not how good engineering works. You need to understand what your code is doing.

I meant once we are sure how it works of course with the intention of properly solving  #898 But from some further testing I've found my flaw when you go to actually send the signal you end up in a deadlock again. I do really want to tackle this issue, and I agree this wasn't the correct approach but now that I noticed you can actually get the SignalEmitter without ever needing to take a lock that might be an avenue worth exploring?

sander-machado: I'm thinking exploring a path where during the code gen, a new function is generated set_property_with_notify(value: PropertyValue, path: P). This function would call the set_property function defined on the interface and then through a tokio::spawn call the signal handler basically like i've done in my code now? Getting the SignalEmitter can be achieved quit easily with just the connection and path through the object_server, not sure how to best get the object server in this case though? I don't know if you are open to exploring something along these lines?

zeenix: > I meant once we are sure how it works of course with the intention of properly solving #898

Ah ok. Sorry it wasn't clear from your sentence. :)

> But from some further testing I've found my flaw when you go to actually send the signal you end up in a deadlock again. I do really want to tackle this issue, and I agree this wasn't the correct approach but now that 

I'm not at all surprised. :)

> I noticed you can actually get the SignalEmitter without ever needing to take a lock that might be an avenue worth exploring?

That's already very much possible through the 2 [constructor methods](https://docs.rs/zbus/latest/zbus/object_server/struct.SignalEmitter.html#method.new) of `SignalEmitter`.

> Getting the SignalEmitter can be achieved quit easily with just the connection and path through the object_server, not sure how to best get the object server in this case though? I don't know if you are open to exploring something along these lines?

You don't need an ObjectServer at all. Did you read the `SignalEmitter` docs. I have feeling that you missed the constructors. In fact you don't even need access to the `Interface` or (even its type) to construct the emitter instance and sending signals through it (the latter needs you to use the low-level/untyped API though).

zeenix: Oh and btw the interface can keep a signal emitter in it and then use that for any signal emission.

sander-machado: > > Getting the SignalEmitter can be achieved quit easily with just the connection and path through the object_server, not sure how to best get the object server in this case though? I don't know if you are open to exploring something along these lines?
> 
> You don't need an ObjectServer at all. Did you read the `SignalEmitter` docs. I have feeling that you missed the constructors. In fact you don't even need access to the `Interface` or (even its type) to construct the emitter instance and sending signals through it (the latter needs you to use the low-level/untyped API though).

Is there a way to get the dbus connection through the InterfaceRef? Otherwise indeed with those constructors you can probably just code gen the proper signal update, you lose the type safety but it shouldn't matter because the code will be generated right? It would then go set_property_with_notify -> { calls set_property.await/ sends the signal through the untyped api}. Even if you have to manually pass the connection I feel like this would already greatly improve useability?

sander-machado: > Oh and btw the interface can keep a signal emitter in it and then use that for any signal emission.

this would solve the need to pass the connection issue indeed

zeenix: > Is there a way to get the dbus connection through the InterfaceRef?

I don't know but it's not at all a problem for 99.99% of the users since inside the methods, they can get a connection instance passed to them and from outside, you can have a connection instance around easily (since they're cheaply cloneable).

> Otherwise indeed with those constructors you can probably just code gen the proper signal update, you lose the type safety but it shouldn't matter because the code will be generated right?

It's only a tiny bit of type-safety lost and yes if you generate it, then that's not a problem at all.

> It would then go set_property_with_notify -> { calls set_property.await/ sends the signal through the untyped api}.

That's what the setter already do (#898 is only about adding a non-DBus exposed equivalent). The thing is that you need the value of the property to send out the signal. The generated `<property>_changed` method, calls the property getter to get the value. You can of course make use of the `invalidates` attribute to change that.

sander-machado: > > It would then go set_property_with_notify -> { calls set_property.await/ sends the signal through the untyped api}.
> 
> That's what the setter already do (#898 is only about adding a non-DBus exposed equivalent). The thing is that you need the value of the property to send out the signal. The generated `<property>_changed` method, calls the property getter to get the value. You can of course make use of the `invalidates` attribute to change that.

If you call the setter directly on the struct that implements the Interface trait it doesn't right? Does it do that if you call it through the InterfaceRef?

PR #1439, opened 2025-07-20
🎨 zb: Remove unsafe type casting code

sander-machado:
Rustc 1.86 stabilizes a new features called trait_upcasting. As quoted from the rust docs "Enable upcasts from dyn Trait1 to dyn Trait2 if Trait1 is a subtrait of Trait2". Which is exactly what this unsafe code was doing. I added the new syntax and increased the compiler version to 1.86 to enable this feature and remove these unsafe blocks. There were also multiple places where this check was called without any reason as it's checked at construction (and would just panic either way).

Related to: https://github.com/rust-lang/rust/issues/65991

### DIFF
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -1,6 +1,6 @@
 //! The object server API.
 
-use std::{collections::HashMap, marker::PhantomData, sync::Arc};
+use std::{any::Any, collections::HashMap, marker::PhantomData, sync::Arc};
 use tracing::{Instrument, debug, instrument, trace, trace_span};
 
 use zbus_names::InterfaceName;
@@ -274,11 +274,11 @@ impl ObjectServer {
             .instance
             .clone();
 
-        // Ensure what we return can later be dowcasted safely.
-        lock.read()
-            .await
-            .downcast_ref::<I>()
-            .ok_or(Error::InterfaceNotFound)?;
+        {
+            // Ensure what we return can later be downcasted safely.
+            let iface: &dyn Any = &*lock.read().await;
+            iface.downcast_ref::<I>().ok_or(Error::InterfaceNotFound)?;
+        }
 
         let conn = self.connection();
         // SAFETY: We know that there is a valid path on the node as we already converted w/o error.
--- a/zbus/src/object_server/interface/interface_deref.rs
+++ b/zbus/src/object_server/interface/interface_deref.rs
@@ -20,7 +20,10 @@ where
     type Target = I;
 
     fn deref(&self) -> &I {
-        self.iface.downcast_ref::<I>().unwrap()
+        let any_ref: &dyn std::any::Any = &*self.iface;
+        any_ref
+            .downcast_ref::<I>()
+            .expect("Unexpected interface type")
     }
 }
 
@@ -37,7 +40,10 @@ where
     type Target = I;
 
     fn deref(&self) -> &I {
-        self.iface.downcast_ref::<I>().unwrap()
+        let any_ref: &dyn std::any::Any = &*self.iface;
+        any_ref
+            .downcast_ref::<I>()
+            .expect("Unexpected interface type")
     }
 }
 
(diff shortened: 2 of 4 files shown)


zeenix:
Cool. I'll look at it in more detail later but in the meantime, please split the commit to keep commits atomic. Here, I see at least 3 logical changes: removal of unneeded checks (which I'm not too sure about), bumping of MSRV and removal of unsafe code.

zeenix:
@sander-machado ping?

zeenix:
@sander-machado I think Rust 1.86 is not too new anymore. Let's bump the MSRV as part of this PR? You'll want to rebase though.

sander-machado:
> @sander-machado I think Rust 1.86 is not too new anymore. Let's bump the MSRV as part of this PR? You'll want to rebase though.

Great I'll update my PR once that is done!

zeenix:
> > @sander-machado I think Rust 1.86 is not too new anymore. Let's bump the MSRV as part of this PR? You'll want to rebase though.
> 
> Great I'll update my PR once that is done!

I meant, you bump the MSRV as part of this PR but it's about to happen with another PR that's about to be merged: #1620. Please rebase on that.

sander-machado:
> > > @sander-machado I think Rust 1.86 is not too new anymore. Let's bump the MSRV as part of this PR? You'll want to rebase though.
> > 
> > 
> > Great I'll update my PR once that is done!
> 
> I meant, you bump the MSRV as part of this PR but it's about to happen with another PR that's about to be merged: #1620. Please rebase on that.

Yeah I was referring to that one (had to bump it 1 more anyway), as that's a coworker who made it. But yeah that's not obvious for you ofcourse haha. I have updated my commits to hopefully be atomic and tried to use to correct emojis. Btw I noticed that if you click the name of the emoji you copy the emoji string which almost made me use that instead before I noticed maybe that's why you keep getting people use the string?

zbus/src/object_server/mod.rs
```diff
@@ -272,11 +272,11 @@ impl ObjectServer {
...
 
-        // Ensure what we return can later be dowcasted safely.
-        lock.read()
-            .await
-            .downcast_ref::<I>()
-            .ok_or(Error::InterfaceNotFound)?;
+        {
+            // Ensure what we return can later be dowcasted safely.
+            let lock: &dyn Any = &*lock.read().await;
+            let _lock: &I = lock.downcast_ref::<I>().ok_or(Error::InterfaceNotFound)?;
```
zeenix: the `lock` isn't the lock but rather the ref. Why is this change needed btw? It looks identical to the one we already have. 🤔

zeenix (changes requested):
LGTM apart from one thing and the usual nitpicks on commit messages: 😄 

* They're missing the `zb: ` prefix.
* second one could be shorter to be close to the guideline, something like`Replace unsafe downcast API w/ std's`.
* A description would be nice, even if a short one. For the second commit, you could use the same description as the PR's since this is the main commit.
* typo: build ins -> built-ins. However this becomes irrelevant if you change the title.

zbus/src/object_server/mod.rs
sander-machado: you need to explicitly make the detour with the &dyn Any otherwise the compiler doesn't like it

zeenix: Ah ok. Would be good to change the variable names still. 😉

zeenix: @sander-machado could you please take this to the finish line? :pray:

zeenix (changes requested):
The changes look good to me, but please use 🦺 as the emoji prefix and ensure that lines in description of git commit message < 75 chars (as per our and git conventions).

zeenix:
@sander-machado Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

PR #1460, opened 2025-08-13
👷CI: use vcpkg to install win32 deps

elmarco:
Testing this on github CI, see if it helps building win32 version more reliably.


zeenix:
@elmarco Thanks for the quick action! Seems it works now? Should we merge?

BTW, "Draft: " prefix doesn't work like on gitlab here. You need to explicitly mark/convert it to draft status.

elmarco:
seems fine to me, hopefully we will have less random failure with vcpkg!

PR #1472, opened 2025-08-21
🐛 zb: Fall back to no groups rather than erroring out for peer creds

swick:
It turns out that looking up the name of a user id requires some files which can be absent in some sandboxes/namespaces. The dbus spec allows us to send either no group or all groups. So instead of erroring out, we can just do what we did before and send no groups at all in those cases.

Discovered and fix proposed by Sophie Herold.

Fixes: c30dcdd7 ("♻️ zb: Restructure get_unix_peer_creds for the next commits")

### DIFF
--- a/zbus/src/connection/socket/unix.rs
+++ b/zbus/src/connection/socket/unix.rs
@@ -340,6 +340,7 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
             unistd::{getgrouplist, Gid, Uid, User},
         };
         use std::ffi::CString;
+        use tracing::debug;
 
         let (uid, gid, pid) = {
             let unix_creds = getsockopt(&fd, PeerCredentials)?;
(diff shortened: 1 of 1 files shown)


zeenix:
> Fixes: [c30dcdd](https://github.com/dbus2/zbus/commit/c30dcdd776401354541c7362b3bcc49e4b7c992f) ("♻️ zb: Restructure get_unix_peer_creds for the next commits")

ahaha, that Sebastian guy! Can't trust him. :laughing: I'll review later but probably a good idea to log errors instead of completely ignoring them.

swick:
Adding some debug logging to the two places where a runtime error might occur. `CString::new` also returns a `Result` but I'm fairly certain that `user.name` will never contain NULL bytes, so leaving the error there.

swick:
(I apparently also struggle to write normal English sentences)

zbus/src/connection/socket/unix.rs
```diff
@@ -353,11 +354,15 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
 
         // the dbus spec requires groups to be either absent or complete (primary + secondary
         // groups)
-        let mut groups = User::from_uid(uid)?
+        let mut groups = User::from_uid(uid)
+            .map_err(|e| debug!("Not sending peer groups: User lookup failed: {}", e))
```
zeenix: This function is not necessarily about sending:

```suggestion
            .map_err(|e| debug!("User lookup failed: {}", e))
```

is good enough.

zbus/src/connection/socket/unix.rs
```diff
@@ -353,11 +354,15 @@ fn get_unix_peer_creds_blocking(fd: RawFd) -> io::Result<crate::fdo::ConnectionC
...
+            .ok()
+            .flatten()
             .map(|user| CString::new(user.name))
             .transpose()?
             .map(|user| getgrouplist(&user, gid))
             .transpose()
+            .map_err(|e| debug!("Not sending peer groups: Group lookup failed: {}", e))
```
zeenix: Same here:

```suggestion
            .map_err(|e| debug!("Group lookup failed: {}", e))
```

zeenix:
I took the liberty of adding the changes in. Just squash them and force push. I'll then merge it..

PR #1477, opened 2025-08-25
Add call_method timeouts

sergeyfd:
Add timeouts for `call_method`. The timeout is an optional feature and can be used to handle the issues with non-answering DBus broker.


zbus/src/abstractions/async_timer.rs
zeenix, line 1: * The function is called `timeout` so we should probably also name the file `timeout` (there is no need for `async-` prefix here).
* Please a bit more info in the commit log, and make it clear that it's internal.
* 🧑‍💻 is perhaps a better emoji since it's internal and only helpful for us as zbus devs.

zbus/src/abstractions/async_timer.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{Error, Result};
+use futures_core::Future;
+use std::time::Duration;
+
+#[cfg(not(feature = "tokio"))]
+use futures_lite::FutureExt;
```
zeenix: hint: if you import this in the function itself, you won't need the `cfg`.

zbus/src/abstractions/async_timer.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{Error, Result};
+use futures_core::Future;
```
zeenix: `Future` is part of `std` so let's import from there.

zbus/src/abstractions/async_timer.rs
```diff
@@ -0,0 +1,32 @@
...
+    F: Future<Output = Result<T>>,
+{
+    match tokio::time::timeout(timeout, fut).await {
+        Ok(r) => r,
+        Err(_) => Err(Error::CallMethodTimedOut),
+    }
+}
+
+#[cfg(not(feature = "tokio"))]
+pub(crate) async fn await_with_timeout<F, T>(fut: F, timeout: Duration) -> Result<T>
```
zeenix: you'll need to repeat the documentation here but just c&p that line and that's enough.

zbus/src/abstractions/async_timer.rs
```diff
@@ -0,0 +1,32 @@
...
+
+/// Awaits a future with a provided timeout
+#[cfg(feature = "tokio")]
+pub(crate) async fn await_with_timeout<F, T>(fut: F, timeout: Duration) -> Result<T>
+where
+    F: Future<Output = Result<T>>,
+{
+    match tokio::time::timeout(timeout, fut).await {
+        Ok(r) => r,
+        Err(_) => Err(Error::CallMethodTimedOut),
```
zeenix: * You're assuming here that it's a method timeout. I think we don't need a new Error variant here since we already have [`std::io::ErrorKind::TimedOut`](https://doc.rust-lang.org/stable/std/io/enum.ErrorKind.html#variant.TimedOut) so these functions should either return `std::io::Error` or `zbus::Error::InputOutput` variant.
* The error variant used here is being introduced in the commit that comes after this so this commit won't even build. That's why splitting the introduction of the new Error variant should be a separate commit. The smaller the commits (as long as they build and don't introduce any regressions/bugs), the better.

Cargo.toml
```diff
@@ -93,7 +86,6 @@ proc-macro2 = "1.0.81"
 proc-macro-crate = "3.2.0"
 syn = { version = "2.0.64", features = ["extra-traits", "full"] }
 quote = "1.0.36"
-static_assertions = "1.1.0"
```
zeenix: What the hell is going on here? Why are you removing deps? If this change is somehow good, it's completely irrelevant to this commit and hence must be split and given a justification.

zbus/src/connection/builder.rs
```diff
@@ -356,6 +357,13 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Set the timeout for [`Connection::call_method`] calls.
+    pub fn call_method_timeout(mut self, timeout: std::time::Duration) -> Self {
```
zeenix: * Let's call this `method_timeout`.
* Despite my reminder, you still forgot to add the blocking wrapper for this. :laughing:

zbus/src/connection/builder.rs
```diff
@@ -356,6 +357,13 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Set the timeout for [`Connection::call_method`] calls.
```
zeenix: * Could you please add a bit of explanation here. Users might not know (for sure) that all method calls (especially ones from the generated proxy) go through `Connection::call_method` so I think I'll only mention that in the second paragraph (details). Also need to inform users that by default method calls don't timeout.
* `the timeout` -> `a timeout`.

zbus/src/connection/mod.rs
```diff
@@ -1118,10 +1128,15 @@ impl Connection {
         self.inner.executor.spawn(remove_match, &task_name).detach()
     }
 
+    pub(crate) fn call_method_timeout(&self) -> Option<Duration> {
```
zeenix: * I think this should be public.
* A oneliner doc comment please.

zbus/src/connection/mod.rs
```diff
@@ -74,6 +76,8 @@ pub(crate) struct ConnectionInner {
     object_server_dispatch_task: OnceLock<Task<()>>,
 
     drop_event: Event,
+
+    call_method_timeout: Option<Duration>,
```
zeenix: nit: `method_timeout`.

zbus/src/connection/mod.rs
```diff
@@ -322,17 +326,23 @@ impl Connection {
...
+                body,
+            )
+            .await?
+            .expect("no reply");
+
+        if let Some(timeout) = self.call_method_timeout() {
+            await_with_timeout(method, timeout).await
+        } else {
+            method.await
+        }
```
zeenix: How about:

```rust
let duration = self.method_timeout().unwrap_or(Duration::MAX);
timeout(method, duration).await
```
instead? I think we can safely assume that no method should ever take 584,942,417,355 years. :)

zbus/src/abstractions/async_timer.rs
```diff
@@ -0,0 +1,32 @@
+use crate::{Error, Result};
+use futures_core::Future;
+use std::time::Duration;
+
+#[cfg(not(feature = "tokio"))]
+use futures_lite::FutureExt;
+
+/// Awaits a future with a provided timeout
+#[cfg(feature = "tokio")]
+pub(crate) async fn await_with_timeout<F, T>(fut: F, timeout: Duration) -> Result<T>
```
zeenix: I think `timeout` is a good enough name here and same as what tokio calls it (IIRC embassy does the same).

zbus/src/connection/mod.rs
sergeyfd: Do we need it to be an `Option` then? We can have as `Duration` with default set to Duration::MAX.

zbus/src/connection/builder.rs
sergeyfd: Do you mean `[cfg(feature = "blocking-api")]` ? Why does it need it?

zbus/src/connection/mod.rs
sergeyfd: Also with this change `test_graceful_shutdown` becomes flaky and times out  on building service session connection. It doesn't make much sense to me because timeout is definitely set to Duration::MAX, but probably it's due to `#[tokio::test(start_paused = true)]` used in that test.

zbus/src/connection/builder.rs
zeenix: We have blocking wrapper types in `blocking` module. The whole module is feature gated so you don't need cfg there. You need wrapper for pub methods you're adding here in `blocking::connection::{Connection, Builder}`.

> Why does it need it?

Because when someone uses the blocking API, they can also use the timeout. In fact, this is a lot more useful for blocking API, since unlike the async API, user can't simply wrap the calls in a timeout. They'll need to launch threads etc to make it work.

zbus/src/connection/mod.rs
zeenix: > Also with this change `test_graceful_shutdown` becomes flaky and times out on building service session connection. It doesn't make much sense to me because timeout is definitely set to Duration::MAX, but probably it's due to `#[tokio::test(start_paused = true)]` used in that test.

Oh right, that's actually the reason. Then let's go with Option still.

zbus/src/abstractions/timer.rs
```diff
@@ -0,0 +1,31 @@
...
+/// Awaits a future with a provided timeout
+#[cfg(feature = "tokio")]
+pub(crate) async fn timeout<F, T>(fut: F, timeout: Duration) -> Result<T>
+where
+    F: Future<Output = Result<T>>,
+{
+    match tokio::time::timeout(timeout, fut).await {
+        Ok(r) => r,
+        Err(_) => Err(Error::InputOutput::from(std::io::ErrorKind::TimedOut)),
+    }
```
zeenix: nit: this would be simpler with `.map_err(|_| Error::InputOutput::from(std::io::ErrorKind::TimedOut))`. I think I hinted to that in the previous review but I guess I didn't make it very explicit. :)

zbus/src/abstractions/timer.rs
zeenix, line 1: I'm confused. Why do we have `timeout` and `timer` modules providing the same API? Left over from the renaming? Please review your commits before pushing. :pray:

zbus/src/connection/mod.rs
```diff
@@ -1118,10 +1128,15 @@ impl Connection {
         self.inner.executor.spawn(remove_match, &task_name).detach()
     }
 
+    pub fn method_timeout(&self) -> Option<Duration> {
```
zeenix, line 1132: * No docs.
* You forgot the blocking wrapper. Remember I said you'll need to edit `blocking::connection::{Connection, Builder}`, not only the `blocking::connection::Builder`

zbus/src/blocking/connection/builder.rs
```diff
@@ -202,6 +202,16 @@ impl<'a> Builder<'a> {
         self.0.unique_name(unique_name).map(Self)
     }
 
+    /// Set a timeout for [`Connection::call_method`] calls.
+    /// `Connection::call_method` will return
+    /// `Connection::Error::InputOutput(std::io::ErrorKind::TimedOut)`
+    /// if a client does not receive an answer from a server in time.
+    /// Note that this includes all methods for genrated proxies.
+    /// No timeouts enforced by default.
```
zeenix: Thanks so much for adding docs with all the details. :+1:

* Please follow the general Rust conventions (see other docs in this module and/or crate):
  * Empty line between the title and description (similar to a git commit).
  * Break the lines at 100 chars at word boundary and **not** before.
*  I think we should not even mention `call_method` here. `call_method` is used by all (not only the generated proxy methods). Also `call_method` is more an implementation detail that need to be mentioned.

zbus/src/connection/builder.rs
```diff
@@ -356,6 +357,18 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Set a timeout for [`Connection::call_method`] calls.
+    /// `Connection::call_method` will return
+    /// `Connection::Error::InputOutput(std::io::ErrorKind::TimedOut)`
+    /// if a client does not receive an answer from a server in time.
+    /// Note that this includes all methods for genrated proxies.
+    /// No timeouts enforced by default.
```
zeenix: Same comment about docs here.

zbus/src/abstractions/timer.rs
zeenix: Oh, I know. You ended up putting the removal of this file to the wrong commit. Please rearrange. If you need help with this, please let me know and I'll do it for you.

zbus/src/blocking/connection/builder.rs
sergeyfd: Does this look right?

```
    /// Set a timeout for [`Connection::call_method`] calls.
    /// 
    /// Calls over `Connection` will return
    /// `Connection::Error::InputOutput(std::io::ErrorKind::TimedOut)` if a client does not receive
    /// an answer from a server in time. Note that this includes all methods for generated proxies.
    /// No timeouts enforced by default.
```

zbus/src/abstractions/timeout.rs
```diff
@@ -0,0 +1,36 @@
...
+/// Awaits a future with a provided timeout
+#[cfg(feature = "tokio")]
+pub(crate) async fn timeout<F, T>(fut: F, timeout: Duration) -> Result<T>
+where
+    F: Future<Output = Result<T>>,
+{
+    tokio::time::timeout(timeout, fut).await.map_err(|_| {
+        Error::from(std::io::Error::new(
+            ErrorKind::TimedOut,
+            format!("timed out {:?}", timeout),
```
zeenix: let's not do allocation here and the exact value isn't important here and also the error message is then incompatible with the async-io case.

zbus/src/abstractions/mod.rs
```diff
@@ -7,6 +7,7 @@ pub use executor::*;
 mod async_drop;
 pub(crate) mod async_lock;
 pub use async_drop::*;
+pub(crate) mod timeout;
```
zeenix, line 10: wrong commit.

zbus/src/connection/mod.rs
zeenix: Still no docs. Before pushing, please read through the review comments and ensure that each commit addresses all points raised. :pray:

zbus/src/blocking/connection/builder.rs
zeenix: @sergeyfd you fixed the formatting, thanks but you seem to have ignored other pointed raised here. :cry:

zbus/src/connection/mod.rs
zeenix: * "Set `method_timeout`" -> "The method timeout (if any)".
* Looks like you actually didn't address the other bullet point either in here. :cry:

zbus/src/connection/mod.rs
```diff
@@ -1863,4 +1882,41 @@ mod p2p_tests {
 
         (conn1, conn2)
     }
+
+    #[test]
+    #[timeout(15000)]
+    fn test_method_timeout() {
```
zeenix: * I just realized that you forgot that I told you on the Matrix channel that these tests don't need to be specific to p2p at all. Come to think of it, it's probably best to add a method to the e2e test (in `zbus/tests/e2e.rs`) that tests this, instead of a completely separate test.
* `test_` prefixes are redundant and inconsistent with naming of other tests (even in this same module here).

zbus/src/connection/mod.rs
```diff
@@ -1863,4 +1882,41 @@ mod p2p_tests {
...
+        #[cfg(feature = "tokio")]
+        use tokio::net::UnixStream;
+        #[cfg(all(windows, not(feature = "tokio")))]
+        use uds_windows::UnixStream;
+
+        let guid = Guid::generate();
+
+        let (p0, p1) = UnixStream::pair().unwrap();
+
+        #[allow(unused)]
```
zeenix: Please prefix names with `_` instead. In the future the code might change and an unused warning could actually be pointing to an issue and we'll miss it with this around. Hence the preference for `_` prefixes.

zbus/src/connection/builder.rs
```diff
@@ -356,6 +357,17 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Set a timeout for method calls.
+    ///
+    /// Method calls will return
+    /// `zbus::Error::InputOutput(std::io::Error(kind: ErrorKind::TimedOut))`` if a client does not
+    /// receive an answer from a server in time.
```
zeenix: server -> service. Unless it's a p2p connection, there is a big diff.

zbus/src/blocking/connection/builder.rs
```diff
@@ -202,6 +202,15 @@ impl<'a> Builder<'a> {
         self.0.unique_name(unique_name).map(Self)
     }
 
+    /// Set a timeout for method calls.
+    ///
+    /// Method calls will return
+    /// `zbus::Error::InputOutput(std::io::Error(kind: ErrorKind::TimedOut))`` if a client does not
+    /// receive an answer from a server in time.
```
zeenix: `server` -> `service` here too.

zeenix (changes requested):
Getting close to a merge, despite all the nitpicks (most of them are trivial to address). :) Apart from the inline comments, one nit about the last commit log:

> The timeout is an optional feature

This makes it sound like a cargo feature. How about something like the following?

```
Add a way to specify an timeout for method calls. If set, the method calls
will timeout after the specified duration, returning an error.

This can be used to handle the issues with non-answering D-Bus services.
```

zbus/src/connection/mod.rs
sergeyfd: Is it Ok to use `#[cfg(feature = "p2p")]` there? Without it I can't build Connections with no Authentication. I see there's a bunch of code there wrapped in `p2p`.

zbus/src/connection/mod.rs
sergeyfd: I don't think `blocking ::Connection` needs any changes, unless I am mistaken. `blocking::Builder` sets timeout  directly on inner `Connection` there. But I can add a getter there, if you think that it needs one. Note that there're no setter for `Connection::method_timeout`.

zbus/src/connection/mod.rs
zeenix: > Is it Ok to use `#[cfg(feature = "p2p")]` there? Without it I can't build Connections with no Authentication. I see there's a bunch of code there wrapped in `p2p`.

As I wrote before, there is no reason for this to be specific to p2p. The method should timeout when it doesn't get a response, regardless of whether it's directly talking to the service or through the bus.

sergeyfd: Ok. I THINK I did it they way you want it, but I am not 100% sure.

zbus/src/blocking/connection/builder.rs
```diff
@@ -202,6 +202,15 @@ impl<'a> Builder<'a> {
         self.0.unique_name(unique_name).map(Self)
     }
 
+    /// Set a timeout for method calls.
+    ///
+    /// Method calls will return
+    /// `zbus::Error::InputOutput(std::io::Error(kind: ErrorKind::TimedOut))`` if a client does not
```
zeenix: typo: there is an extra backtick.

zbus/src/connection/builder.rs
```diff
@@ -356,6 +357,17 @@ impl<'a> Builder<'a> {
         Ok(self)
     }
 
+    /// Set a timeout for method calls.
+    ///
+    /// Method calls will return
+    /// `zbus::Error::InputOutput(std::io::Error(kind: ErrorKind::TimedOut))`` if a client does not
```
zeenix: same here.

zbus/tests/e2e.rs
```diff
@@ -1118,3 +1118,45 @@ async fn iface_and_proxy_(#[allow(unused)] p2p: bool) {
     );
     debug!("Bus confirmed that all names were definitely released.");
 }
+
+struct TimerTest {}
```
zeenix: Just to remind of my exact wording:

> it's probably best to add a method to the e2e test (in zbus/tests/e2e.rs) that tests this, instead of a completely separate test.

You're adding a completely separate test still.

zbus/src/abstractions/timeout.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{Error, Result};
+use std::{future::Future, io::ErrorKind, time::Duration};
+
+/// Awaits a future with a provided timeout
```
zeenix: I wasn't going to mention this but since you're going to modify the PR again anyway, please add the missing `.` at the end. (Same for the duplicate below).

zeenix (changes requested):
Sorry but I think you misunderstood.

zbus/tests/e2e.rs
sergeyfd: I tried my best there, still probably not exactly how you want it.

zeenix: > I tried my best there

And thanks for that. I didn't doubt it. I just feel like you sometimes miss my comments.

> still probably not exactly how you want it.

I want like to believe that it's less about what I want and more about the best solution. If/when you disagree with me on that, please do say so and we can try to come up with the best solution together.

I realize that the interface here is quite bloated already and I would love contributions on refactoring it. Simply adding a new test case for each feature, is not a solution either as that would add to the compilation and test run time, and add duplication.

zbus/src/connection/mod.rs
zeenix: Oh, I had missed this reply because you had marked this conversation as resolved. Please only do that when you've addressed the comment. If you disagree, then please leave it unresolved so we can discuss. Anyway:

> I don't think `blocking ::Connection` needs any changes, unless I am mistaken.

I was referring to this getter method. When people use `blocking::Connection`, they should be able to directly get the timeout value from it.  It's the same as other getter methods in  `blocking::Connection`.

Now, you also didn't address my other bullet point (which you didn't disagree with afaict :)).

zbus/tests/e2e.rs
```diff
@@ -894,10 +900,35 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
 
+    #[cfg(not(feature = "p2p"))]
+    test_timeout(&conn).await?;
+
     proxy.quit().await?;
+
     Ok(val)
 }
 
+#[cfg(not(feature = "p2p"))]
```
zeenix: why? As I wrote before, this is irrelevant to p2p or bus. We don't need any cfgs to be added here.

zbus/tests/e2e.rs
```diff
@@ -894,10 +900,35 @@ async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> {
...
+async fn test_timeout(conn: &Connection) -> zbus::Result<()> {
+    assert!(conn
+        .method_timeout()
+        .is_some_and(|d| d == Duration::from_secs(1)));
+
+    let proxy = MyIfaceProxy::builder(&conn)
+        .destination("org.freedesktop.MyService")?
+        .path("/org/freedesktop/MyService")?
+        .build()
+        .await?;
```
zeenix: We already have a proxy so we don't need to do this. Let me be more specific about my suggestion: In the existing interface and proxy here:

* modify the connection creation to also set a timeout.
* add a method that never replies.
* in the existing proxy side of the test, add the call to the new method (basically what you're doing below) but best at the end, so everything else doesn't have to wait on the timeout.

zbus/src/connection/mod.rs
zeenix: > > I don't think `blocking ::Connection` needs any changes, unless I am mistaken.
> 
> I was referring to this getter method. When people use `blocking::Connection`, they should be able to directly get the timeout value from it. It's the same as other getter methods in `blocking::Connection`.

Oh it's me who misread this time, sorry! You did indeed add the `blocking::Connection::timeout`. This comment from you just confused me.

So there is only the bullet point you missed left now:

> * "Set method_timeout" -> "The method timeout (if any)".

zeenix (changes requested):

zbus/src/connection/mod.rs
sergeyfd: I can do that, but it requires wrapping `method_timeout` filed into a Mutex, unless you have a better idea.

zeenix: I'm confused. I had two points:

* ~~You forgot the blocking wrapper. Remember I said you'll need to edit blocking::connection::{Connection, Builder}, not only the blocking::connection::Builder.~~ You addressed this already by adding the wrapper.
* "Set method_timeout" -> "The method timeout (if any)".

Here, I'm only asking for a slight change in wording of the docs. Currently, it's referring to itself to explain itself (user doesn't see the field).

zeenix: Thanks but I didn't mean you should remove the "See [`Builder::method_timeout`] for details." part. That's why I tried to be explicit in what exactly to replace with what. I guess my `"<copy-pasted-existing-code>" -> "<my-suggestion-for-replacment>" isn't as universally understood as I thought.

Anyway, please bring it back as I think it's useful.

zbus/tests/e2e.rs
```diff
@@ -493,6 +493,16 @@ impl MyIface {
...
         Ok(())
     }
+
+    #[instrument]
+    async fn timeout(&self) -> String {
+        debug!("`Timeout` called.");
+        #[cfg(feature = "tokio")]
+        tokio::time::sleep(Duration::from_secs(10)).await;
+        #[cfg(not(feature = "tokio"))]
+        async_io::Timer::after(Duration::from_secs(10)).await;
```
zeenix: I'm pretty sure I explicitly suggested use of [`std::future::pending`](https://doc.rust-lang.org/stable/std/future/fn.pending.html) but maybe I didn't. :shrug: Anyway, that is more reliable and simpler so please use that.

zbus/tests/e2e.rs
```diff
@@ -493,6 +493,16 @@ impl MyIface {
         self.emits_changed_false = val;
         Ok(())
     }
+
+    #[instrument]
+    async fn timeout(&self) -> String {
```
zeenix: * We don't need to return a value here. This method is not supposed to actually return.
* While you update the return value, perhaps also update the name to `never` or `never_rerturn`.

zbus/tests/e2e.rs
```diff
@@ -978,7 +1001,9 @@ async fn iface_and_proxy_(#[allow(unused)] p2p: bool) {
                         .server(guid)
                         .unwrap()
                         .p2p(),
-                    connection::Builder::tcp_stream(p1).p2p(),
+                    connection::Builder::tcp_stream(p1)
+                        .method_timeout(Duration::from_secs(2))
+                        .p2p(),
```
zeenix: We're setting the timeout later (just before finalizing the build) so this and the change at line 988 is redundant, no? :thinking:

zeenix (changes requested):
I'm very hopeful the next push will be merged (assuming the CI is green too). :) Please let me know if you are tired and would like me to finish this. Thank you for your patience and hard work! :pray:

zbus/tests/e2e.rs
sergeyfd: I disagree with this. Having it returning something was very helpful to troubleshoot test in e2e when I was getting `Ok` instead of Error for incorrect version of `sleep`. I can convert it, but I'd rather have it as is to have a more complete test case.

zbus/tests/e2e.rs
sergeyfd: It's not. These are all different branches: `p2p` vs `non p2p`, then `windows` vs `unix` and `tokio` `not tokio`  in the `windows` branch.

zbus/tests/e2e.rs
sergeyfd: You didn't but see my answer to the other comment. If this method never returns then we won't have a coverage for a case when `a service doesn't answer in given time` which might be useful.

zbus/tests/e2e.rs
zeenix: > getting `Ok` instead of Error for incorrect version of `sleep`

With my suggestion of using `pending`, this won't be an issue and the method will actually not return. What we need to keep in mind here is that what is being tested. We are testing that the client timesout if the method doesn't reply in time. From the service side, the "no reply" is guaranteed with `pending`.

zbus/tests/e2e.rs
zeenix: Yes but all of them converging to one place where we call `build` to finalize it. None of those branches are finalizing the build until that one point on line 1009 (or 1033 after you change here).

zbus/tests/e2e.rs
zeenix: > You didn't but see my answer to the other comment.

Oh, I'm very sorry then. 

> If this method never returns then we won't have a coverage for a case when `a service doesn't answer in given time` which might be useful.

* The important bit here (which is what we need to test) is that the client timesout if it doesnt' receive a reply within a specified amount of time. It's not relevant how long the service takes to actually reply. As long as the time it takes to reply is higher than the timeout, we have means to test the timeout on the client-side. Choosing forever, makes it 100% reliable as it's guaranteed to be higher than whatever timeout the client-side uses.
* We don't have a realtime clock and scheduler involved here and it will all depend on the system and load on the system so any comparisons of timing here, would be unreliable and we'll have the test failing randomly at times.

zeenix:
@sergeyfd Thanks again for your contribution and your patience on this. :+1:

sergeyfd:
Thanks for all your help.

PR #1480, opened 2025-08-26
zb Provide a better error message for a failed connection that uses an address

### LINKED ISSUE #1478: Improve the connection error with Address
Hello,

When the connection fails for some reasons, it can be hard to trouble shoot for an inexperienced zbus developer. I'm raising this issue because I passed few days to troubleshoot a __system__ connection failure. I have been able to find the issue by stepping in the code. I found that the **system** connection is derived from the **DBUS_SYSTEM_BUS_ADDRESS**  environnemental variable and if the variable is not set, it's default to **unix:path=/var/run/dbus/system_bus_socket**

I propose that the connection error returns the Address as well the current IOError to know to which address zbus tries to connect to ease the troubleshooting.

Thank you

plrigaux:
Fixes #1478

I created a new error, but I have got  no choice to clone the address to send the error with the address information. Because the value is moved with the call of connect(). 

_Maybe some refactoring can remove the necessity to use clone()._

### DIFF
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -525,7 +525,16 @@ impl<'a> Builder<'a> {
             Target::VsockStream(stream) => stream.into(),
             Target::Address(address) => {
                 guid = address.guid().map(|g| g.to_owned().into());
-                match address.connect().await? {
+
+                match address
+                    .clone() //This clone is necessary to capture on error the address because the connect
+                    // function consume the address object.
+                    .connect()
+                    .await
+                    .map_err(|error| match error {
+                        Error::InputOutput(error) => Error::InputOutputAddress(error, address),
+                        _ => error,
+                    })? {
                     #[cfg(any(unix, not(feature = "tokio")))]
                     address::transport::Stream::Unix(stream) => stream.into(),
                     #[cfg(unix)]
--- a/zbus/src/error.rs
+++ b/zbus/src/error.rs
@@ -5,6 +5,7 @@ use zvariant::{Error as VariantError, ObjectPath};
 use crate::{
     fdo,
     message::{Message, Type},
+    Address,
 };
 
 /// The error type for `zbus`.
@@ -20,6 +21,8 @@ pub enum Error {
     Address(String),
     /// An I/O error.
     InputOutput(Arc<io::Error>),
+    /// An I/O error with the related URI.
+    InputOutputAddress(Arc<io::Error>, Address),
     /// Invalid message field.
     InvalidField,
     /// Data too large.
@@ -96,6 +99,7 @@ impl error::Error for Error {
             Error::InterfaceNotFound => None,
             Error::Address(_) => None,
             Error::InputOutput(e) => Some(e),
+            Error::InputOutputAddress(e, _) => Some(e),
             Error::ExcessData => None,
             Error::Handshake(_) => None,
             Error::IncorrectEndian => None,
@@ -125,6 +129,7 @@ impl fmt::Display for Error {
             Error::Address(e) => write!(f, "address error: {e}"),
             Error::ExcessData => write!(f, "excess data"),
             Error::InputOutput(e) => write!(f, "I/O error: {e}"),
+            Error::InputOutputAddress(e, address) => write!(f, "I/O error: {e} for {address}"),
             Error::Handshake(e) => write!(f, "D-Bus handshake failed: {e}"),
             Error::IncorrectEndian => write!(f, "incorrect endian"),
             Error::InvalidField => write!(f, "invalid message field"),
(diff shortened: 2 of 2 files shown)


zeenix:
Thank you but please follow our contributing guideline.

plrigaux:
I apology for not having followed the guidelines.

I will follow them and resubmit

zeenix:
> I apology for not having followed the guidelines.

No worries. You're not the first one to miss them and you'll certainly not be the last one. :)

> I will follow them and resubmit

Why are you closing the PR then. Just (force) push to the same branch.

zeenix:
@plrigaux ping?

plrigaux:
Sorry busy , maybe for this weekend

plrigaux:
I tried to follow the guidelines and push force it. Hope it is ok.

zeenix:
> I tried to follow the guidelines and push force it. Hope it is ok.

Thanks for trying. You created 2 additional commits instead of modifying your original commit. Please rebase and squash them and add the emoji prefix (I've made that as easy as possible these days so no reason to skip it) while you're at it. 

Forced push is not a requirement on its own, you just need to do that when you modify the commits, which you have yet.

zbus/src/error.rs
```diff
@@ -20,6 +21,8 @@ pub enum Error {
     Address(String),
     /// An I/O error.
     InputOutput(Arc<io::Error>),
+    /// An I/O error with the related URI.
+    InputOutputAddress(Arc<io::Error>, Address),
```
zeenix, line 25: This a very strange name IMO that will confuse people. I actually suggested a specific name in #14778: `zbus::Error::Connection`. This error is about connection failures.

zbus/src/error.rs
```diff
@@ -20,6 +21,8 @@ pub enum Error {
     Address(String),
     /// An I/O error.
     InputOutput(Arc<io::Error>),
+    /// An I/O error with the related URI.
```
zeenix, line 24: Let's be a bit more descriptive here and document the arguments (especially since we don't name them in here, which I know is not ideal).

zbus/src/connection/builder.rs
```diff
@@ -525,7 +525,16 @@ impl<'a> Builder<'a> {
             Target::VsockStream(stream) => stream.into(),
             Target::Address(address) => {
                 guid = address.guid().map(|g| g.to_owned().into());
-                match address.connect().await? {
+
+                match address
+                    .clone() //This clone is necessary to capture on error the address because the connect
+                    // function consume the address object.
```
zeenix, line 531: We could have the `Address::connect` do the mapping instead, or return `Result<Stream, (io::Error, Self)` (open to other options too) or maybe even have the `Address::connect` just not take ownership of address instance (it's internal after all).

Since this should not require a major effort and there is no urgency here, I think it makes sense to already avoid the cloning here, instead of leaving it for the future, which is likely never going to happen. :)

P.S. do remember the atomic commits rule, so please put the needed refactoring in a separate commit that comes before this one. :pray:

zbus/src/error.rs
```diff
@@ -125,6 +129,7 @@ impl fmt::Display for Error {
             Error::Address(e) => write!(f, "address error: {e}"),
             Error::ExcessData => write!(f, "excess data"),
             Error::InputOutput(e) => write!(f, "I/O error: {e}"),
+            Error::InputOutputAddress(e, address) => write!(f, "I/O error: {e} for {address}"),
```
zeenix, line 132: This will need to be update too:

```suggestion
            Error::InputOutputAddress(e, address) => write!(f, "failed to connect to `{address}`: {e}"),
```

plrigaux:
Good I will make the changes. Also I will look how I modify a commit (I'm not fluent in git)

zbus/src/connection/builder.rs
zeenix: > We could have the `Address::connect` do the mapping instead, or return `Result<Stream, (io::Error, Self)` (open to other options too) or maybe even have the `Address::connect` just not take ownership of address instance (it's internal after all).

Just FYI, I would very much prefer the first option here: have the `Address::connect` (or `Transport::connect`) do the mapping.

zeenix:
> Good I will make the changes. Also I will look how I modify a commit (I'm not fluent in git)

Ping? If you need help with git, please let me know and I can try to help out. We do have hints in [our contribution guidelines](https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests) how to modify commits btw. :)

zeenix:
@plrigaux hello? If you don't have time for this anymore, please let me know. It'd be shame though since it's only some very minor things that are needed to get this merged.

plrigaux:
Hello,

the change is minor, I tried to submit a pr, but I'm not savvy in open
source multiple pr and it will take me time to learn how to do it.

I don't mind if you reject my pr

thank you



Pierre-Luc Rigaux
514 808 6241

zeenix:
> the change is minor, I tried to submit a pr, but I'm not savvy in open source multiple pr and it will take me time to learn how to do it.

That's quite a contradiction to your previous comment here, where you showed a positive attitude. I'm not sure what you mean even. You mostly had to make some very minor improvements to your work and it would have been good to go. Not much to learn here except maybe some essential of git (that are useful to learn in general) and I offered you help with that so you could have taken my offer and you could have focused only on the parts that require absolutely no learning from you.
 
> I don't mind if you reject my pr

If I wanted to do that, I wouldn't have spent time in reviewing the PR. Next time if you don't intend to address  review comments, please help reviewer save time by being upfront about this. :pray:

zeenix:
@plrigaux Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

PR #1485, opened 2025-08-28
Add new method to Split

### LINKED ISSUE #1484: Impossible to implement Socket trait
Hi, I've been using this crate for a while and am now trying to upgrade from 3.15.2 to 5.10.0, but I'm not able to migrate my implementation of the `Socket` trait because there is no way to instantiate a `Split` value, since there's no `new()` method and the fields are crate-private.

This would be fixed by adding a `new()` method to `Split`.

bouk:
This allows implementing the Socket trait outside this crate.

Closes #1484


bouk:
done

zeenix (approved):
Oops. This was waiting on my approval.

Issue #689, opened 2024-03-26
"{}" as span names


nekto0n:
Hi!
Noticed strangely named spans, which are literally `{}`. Examples include:

https://github.com/dbus2/zbus/blob/759a71759f9b3ff908cdebd71ddbb73d81c836ec/zbus/src/connection/mod.rs#L683

https://github.com/dbus2/zbus/blob/759a71759f9b3ff908cdebd71ddbb73d81c836ec/zbus/src/connection/mod.rs#L727

https://github.com/dbus2/zbus/blob/759a71759f9b3ff908cdebd71ddbb73d81c836ec/zbus/src/proxy/mod.rs#L312

Looking at the history it looks like the intention was to assign a literal string to a variable to be able to reuse for other purposes.

If I am reading this correctly, then the proper solution would look like:
```
.instrument(info_span!(name: task_name))
```

zeenix:
> Looking at the history it looks like the intention was to assign a literal string to a variable to be able to reuse for other purposes.

Actually, the intention was just to assign a name/string to the task so traces include the context.

> If I am reading this correctly, then the proper solution would look like:
> 
> ```
> .instrument(info_span!(name: task_name))
> ```

hmm.. can it just be:

```rust
.instrument(info_span!(task_name))
```

In either case, a PR would be most welcome.

PR #1495, opened 2025-09-09
🐛 zb: Fix tracing span names showing as "{}"

hadihaider055:
Replaced `info_span!("{}", task_name)` with descriptive span names and structured fields. Spans now have meaningful static names like `monitor_name_lost`, and `monitor_name_acquired` with `name = %well_known_name`.

The issue is that `info_span!("{}", task_name)` doesn't work like a format string. In tracing, the first argument to info_span! is always the literal span name, not a format string.

**Output before this PR:**

2025-09-16T14:19:15.862Z  INFO {}{task_name="monitor name org.example.Service acquired"}: zbus::connection: Processing connection in task
2025-09-16T14:19:15.862Z  INFO {}: Processing connection in task
2025-09-16T14:19:15.862Z  INFO {}: Handling lost connection

**Output after this PR:**

2025-09-16T14:23:19.341Z  INFO monitor_name_acquired{name=org.example.Service}: zbus::connection: Processing connection in task
2025-09-16T14:23:19.341Z  INFO monitor_name_acquired{name=org.example.Service}: Processing connection in task
2025-09-16T14:23:19.341Z  INFO monitor_name_lost{name=org.example.Service}: Handling lost connection

Fixes: #689

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -661,7 +661,8 @@ impl Connection {
             .await?
             .body()
             .deserialize::<RequestNameReply>()?;
-        let lost_task_name = format!("monitor name {well_known_name} lost");
+        let lost_task_name = format!("monitor_name_lost{{name={well_known_name}}}");
+        let lost_task_name_span = info_span!("monitor_name_lost", name = %well_known_name);
         let name_lost_fut = if flags.contains(RequestNameFlags::AllowReplacement) {
             let weak_conn = WeakConnection::from(self);
             let well_known_name = well_known_name.to_owned();
@@ -705,7 +706,7 @@ impl Connection {
                         }
                     }
                 }
-                .instrument(info_span!("{}", lost_task_name)),
+                .instrument(lost_task_name_span),
             )
         } else {
             None
@@ -714,7 +715,8 @@ impl Connection {
             RequestNameReply::InQueue => {
                 let weak_conn = WeakConnection::from(self);
                 let well_known_name = well_known_name.to_owned();
-                let task_name = format!("monitor name {well_known_name} acquired");
+                let task_name = format!("monitor_name_acquired{{name={well_known_name}}}");
+                let task_name_span = info_span!("monitor_name_acquired", name = %well_known_name);
                 let task = self.executor().spawn(
                     async move {
                         loop {
@@ -748,7 +750,7 @@ impl Connection {
                             }
                         }
                     }
-                    .instrument(info_span!("{}", task_name)),
+                    .instrument(task_name_span),
                     &task_name,
                 );
 
@@ -949,7 +951,6 @@ impl Connection {
             trace!("starting ObjectServer task");
             let weak_conn = WeakConnection::from(self);
 
-            let obj_server_task_name = "ObjectServer task";
             self.inner.executor.spawn(
                 async move {
                     let mut stream = match weak_conn.upgrade() {
@@ -1025,8 +1026,8 @@ impl Connection {
                         }
                     }
                 }
-                .instrument(info_span!("{}", obj_server_task_name)),
-                obj_server_task_name,
+                .instrument(info_span!("obj_server_task")),
+                "obj_server_task",
             )
         });
     }


zbus/src/connection/mod.rs
```diff
@@ -705,7 +705,7 @@ impl Connection {
                         }
                     }
                 }
-                .instrument(info_span!("{}", lost_task_name)),
+                .instrument(info_span!("lost_task", name = %lost_task_name)),
```
zeenix: I'm a bit confused. According to [the docs](https://docs.rs/tracing/latest/tracing/macro.info_span.html), we're already doing the right thing and the `lost_task_name` (unrelated to your change) is more a title here. Perhaps you could add how the log output looks like before and after you changes in the commit+PR description and that will help understanding the need for this change?

hadihaider055: The issue is that `info_span!("{}", task_name)` doesn't work like a format string. In tracing, the first argument to `info_span!` is always the literal span name, not a format string.

**Before:**
`.instrument(info_span!("{}", task_name))` this creates a span literally named "{}". The span name is the literal string "{}", not the value of task_name.

**Outputs:** 
2025-09-11T13:34:08.905083Z  INFO {}{task_name="Hello world!"}: lecture_13: Inside the span!
2024-09-11T10:30:45.123456Z  INFO {}: Processing connection in task 
2024-09-11T10:30:45.234567Z  INFO {}: Handling lost connection

**After:**
`.instrument(info_span!("task", name = %task_name))` this creates a span named "task" with a field name containing the value of task_name.

**Outputs:** 
2025-09-11T13:34:44.425203Z  INFO task_name{name=Hello world!}: lecture_13: Inside the span!
2024-09-11T10:30:45.123456Z  INFO task{name=my_connection_task}: Processing connection in task
2024-09-11T10:30:45.234567Z  INFO lost_task{name=lost_connection_task}: Handling lost connection

We can also use this syntax: `.instrument(info_span!("task", task_name))`

The alternative syntax `info_span!("task", task_name)` would create a field named task_name (using the variable name), while `info_span!("task", name = %task_name)` explicitly names the field name. Both are valid, but the explicit naming might be clearer for readers.

zeenix: @hadihaider055 Thanks so much for explaining in great detail. 

> `.instrument(info_span!("task", name = %task_name))` this creates a span named "task" with a field name containing the value of task_name.

Right. However, the task_name is not exactly a field really, more a value and hence the desire to use a format string. I think we should split the dynamic parts from the static part. So here we want to print: `monitor name lost(name=<well_known_name>)`.

hadihaider055: Makes sense, @zeenix. 

> So here we want to print: monitor name lost(name=<well_known_name>).

So, what I understood so far, to replace the current implementation with the `.instrument(info_span!("monitor_name_lost", name = %well_known_name))`? Does that sound good to replace with all over the place where we have the static and dynamic implementation?

zeenix: > So, what I understood so far, to replace the current implementation with the `.instrument(info_span!("monitor_name_lost", name = %well_known_name))`?

Yeah, sounds good.

> Does that sound good to replace with all over the place where we have the static and dynamic implementation?

It's the same with other parts you're modifying here, so yeah **but** just with different task names (static part) and field (dynamic part). If there is no dynamic part, just use the static part directly.

zbus/src/connection/mod.rs
```diff
@@ -1025,7 +1025,7 @@ impl Connection {
                         }
                     }
                 }
-                .instrument(info_span!("{}", obj_server_task_name)),
+                .instrument(info_span!("obj_server_task", name = %obj_server_task_name)),
```
zeenix: This one doesn't seem to have a dynamic part so it'd just be:

```suggestion
                .instrument(info_span!("obj_server_task")),
```

zbus/src/connection/mod.rs
```diff
@@ -705,14 +706,15 @@ impl Connection {
...
-                .instrument(info_span!("{}", lost_task_name)),
+                .instrument(info_span!("monitor_name_lost", name = %well_known_name_info_span)),
             )
         } else {
             None
         };
         let status = match reply {
             RequestNameReply::InQueue => {
                 let weak_conn = WeakConnection::from(self);
+                let well_known_name_info_span = well_known_name.as_str();
```
zeenix: No need for this. Just clone the `well_known_name` as needed and use `well_known_name` directly.

zbus/src/connection/mod.rs
```diff
@@ -1025,7 +1027,7 @@ impl Connection {
                         }
                     }
                 }
-                .instrument(info_span!("{}", obj_server_task_name)),
+                .instrument(info_span!("obj_server_task")),
                 obj_server_task_name,
```
zeenix: There is nothing dynamic here so you don't need `obj_server_task_name`. It's only repeating the same exact info.

zeenix: Actually, I thought you might make this mistake so I had given [specific instructions about this](https://github.com/dbus2/zbus/pull/1495#discussion_r2341709113). :laughing: In the future, please ensure that you don't resolve conversations that you have not addressed.

hadihaider055: Ah, sorry about that. I totally forgot that.

zbus/src/connection/mod.rs
hadihaider055: Btw, I'm having issues implementing your suggestion to: 

> Just clone the well_known_name as needed and use well_known_name directly.

The value well_known_name can't be used directly like `info_span!("monitor_name_acquired", name = %well_known_name.clone())` because it gets moved into the async move closure. I get this compiler error:

```
borrow of moved value: `well-known_name`
value borrowed here after move

move occurs because `well_known_name` has type `WellKnownName<'_>`, which does not implement the `Copy` trait
```

To resolve this, I had to extract the string value before the async move and store it in a separate variable. I used .as_str() instead of .clone() in line [line 717 ](https://github.com/dbus2/zbus/pull/1495/files#diff-651c1796657982de94b9bafb1c89c58f9b46507f26682581c471e58eb14a5258R717), since we only need the string representation for tracing

zbus/src/connection/mod.rs
hadihaider055: Just an addition to the previous comment, I just added the static part, same as suggested by you [here](https://github.com/dbus2/zbus/pull/1495/files#diff-651c1796657982de94b9bafb1c89c58f9b46507f26682581c471e58eb14a5258R1030). 

What would you suggest using this as a format string `.instrument(info_span!("{}", obj_server_task_name))`, same as the previous implementation, or the current implementation with the static part only `.instrument(info_span!("obj_server_task"))`.

zbus/src/connection/mod.rs
zeenix: > The value well_known_name can't be used directly like `info_span!("monitor_name_acquired", name = %well_known_name.clone())` because it gets moved into the async move closure. I get this compiler error:

Well, I did say that you'll need to clone it. :)

zeenix: > To resolve this, I had to extract the string value before the async move and store it in a separate variable. I used .as_str() instead of .clone() in line [line 717 ](https://github.com/dbus2/zbus/pull/1495/files#diff-651c1796657982de94b9bafb1c89c58f9b46507f26682581c471e58eb14a5258R717), since we only need the string representation for tracing

That's complicating the code a tiny bit more than needed. Let's just clone it (which in most cases would be a zero-cost operation and in others, a cheap operation).

zbus/src/connection/mod.rs
zeenix: > .instrument(info_span!("obj_server_task"))

This is more concise and does the same, so this one.

zbus/src/proxy/mod.rs
```diff
@@ -316,7 +317,7 @@ impl PropertiesCache {
                 debug!("Error keeping properties cache updated: {e}");
             }
         }
-        .instrument(info_span!("{}", task_name));
+        .instrument(info_span!("monitor_proxy_caching", name = %interface_clone));
         let task = executor.spawn(proxy_caching, &task_name);
```
zeenix, line 320: * we want to have these two to result in the same string. The only reason to add the `.instrument` here is that currently tokio doesn't use the task name unless `tokio_unstable` cfg is enabled.
* Perhaps you can avoid the clone if you create the span after task_name is created? Then you can also do `info_span!("{}", task_name)` ?

zeenix:
BTW, don't forget to update the descriptions after your recent changes. Thank you!

zbus/src/connection/mod.rs
```diff
@@ -748,7 +750,7 @@ impl Connection {
                             }
                         }
                     }
-                    .instrument(info_span!("{}", task_name)),
+                    .instrument(info_span!("monitor_name_acquired", name = %well_known_name_clone)),
```
zeenix: You can't avoid the cloning by creating the span earlier instead (where `task_name` is being created)? :thinking: You could name it `task_name_span` (for example). Same for the other clone above.

zeenix (changes requested):
Oh and btw, each line in git commit description is supposed to be max 74 chars.

Getting close to a merge. Thanks for your patience and persistence. :+1:

zbus/src/connection/mod.rs
hadihaider055: Alright, so you mean I should do this rather? 

```
let task_name_span = info_span!("monitor_name_acquired", name = %well_known_name);
```

and use it directly like this:

`.instrument(task_name_span),`

Is this what you expect, right @zeenix?

zeenix: yes

zeenix:
> Before: info_span!({}, task_name) creates a span literally named {}
> After: info_span!(monitor_task_name, name = %well_known_name) creates a span named monitor_task_name with field

nitpic: The point here would be to show the output/results of the change, not the change itself, like you did [here](https://github.com/dbus2/zbus/pull/1495#discussion_r2340922940).

hadihaider055:
Updated @zeenix, I hope this explains the changes made.

zeenix:
> Updated @zeenix, I hope this explains the changes made.

Thanks but you forgot to update the commit part of "commit+PR description". :) BTW, you don't need to include the changes themselves into the description (it only duplicates information). I'll remove that part from the PR description and you can then just c&P (but don't forget the 74 chars rule :wink:).

hadihaider055:
Hi @zeenix, are there any other changes required?

zbus/src/connection/mod.rs
```diff
@@ -662,6 +662,7 @@ impl Connection {
             .body()
             .deserialize::<RequestNameReply>()?;
         let lost_task_name = format!("monitor name {well_known_name} lost");
+        let lost_task_name_span = info_span!("monitor_name_lost", name = %well_known_name);
```
zeenix, line 665: Oh I missed this before but as I wrote before (for object server task name), the task name and span should be the same, so these 2 strings should be the same.

zeenix (changes requested):
Apart from that, I see you removed the whole description from the commit message. I think you misundstood me. I only asked that the details of the code changes be removed. I edited the PR description to make it easy for you so you can just copy&paste it.

zbus/src/connection/mod.rs
hadihaider055: Ah alright @zeenix, so you mean remove the dynamic stuff from the format string and make it like: 
`let task_name = "monitor_name_acquired";`

zeenix: > Ah alright @zeenix, so you mean remove the dynamic stuff from the format string and make it like:
> `let task_name = "monitor_name_acquired";`

Not at all. :) `lost_task_name` and `lost_task_name_span` should end up having the same value/output (just like it is already now of object server case).

zeenix: hmm.. just realized object server isn't a good example as it doesn't have any dynamic parts. I mean this:

```rust
let lost_task_name = format!("monitor_name_lost{{{name}}}", name = well_known_name);
```

Please do the same for the other name acquired task below.

hadihaider055: @zeenix, I got you, so basically you want the output value to be the same; however, the code you suggested was giving output like: `monitor_name_lost{org.example.Service}`

So, I made some changes to make it like the output of lost_task_name_span:
```
let task_name = format!("monitor_name_acquired{{{{name={well_known_name}}}}}");
```

Now, the output looks like this:
`monitor_name_lost{{name=org.example.Service}}`

The output from `info_span`:
`2025-09-16T14:23:19.341Z INFO monitor_name_acquired{name=org.example.Service}: Processing connection in task`


Is it fine, or should I follow the approach you suggested?

zeenix: > owever, the code you suggested was giving output like: `monitor_name_lost{org.example.Service}`

Correct and that was based on the output you pasted in the PR description.

> So, I made some changes to make it like the output of lost_task_name_span:

Not sure how I ended up giving the impression that I want extra `{}` in the output but that's not what I meant. :)

> The output from `info_span`:

Yes, that was the reason to only pring one pair of `{}` so they're the same output.

hadihaider055:
Thanks @zeenix bhai! Learned a lot from this review process 😀. Looking forward to contributing more to the project. Happy to connect on Matrix for guidance/mentorship if you're open to it.

zeenix:
> Thanks @zeenix bhai! Learned a lot from this review process 😀.

You're welcome! :)

> Looking forward to contributing more to the project.

Great to hear. That's a very positive attitude.

> Happy to connect on Matrix for guidance/mentorship if you're open to it.

I see you're already on our channel. Feel free to ask questions there. :+1:

Issue #1523, opened 2025-09-24
`NameTaken` error's message is `org.freedesktop.zbus.Error: no description`


arun-mani-j:
We are using zbus indirectly through ashpd in [XDG Desktop Portal Phrosh](https://gitlab.gnome.org/guidog/xdg-desktop-portal-phosh/).

Whenever we launch the D-Bus server when there is already one, it fails with `NameTaken` error (as expected), but when we print the error, it is always `org.freedesktop.zbus.Error: no description`.

The flow is from ashpd's `error.rs` https://github.com/bilelmoussaoui/ashpd/blob/0.12/src/error.rs#L9 to zbus_macros' `error` https://github.com/dbus2/zbus/blob/d641460202416dec9437d6231412c74d54463b18/zbus_macros/src/error.rs#L166.

@zeenix helped by saying this looks like a bug in `zbus_macros` and to file an issue. This is their full reply:

> yeah, this looks like a bug in zbus_macros. The issue is here I think: https://github.com/dbus2/zbus/blob/d641460202416dec9437d6231412c74d54463b18/zbus_macros/src/error.rs#L89
[zbus/zbus_macros/src/error.rs at d641460202416dec9437d6231412c74d54463b18 · dbus2/zbus](https://github.com/dbus2/zbus/blob/d641460202416dec9437d6231412c74d54463b18/zbus_macros/src/error.rs#L89) 
I think we should add special handling for #zbus::Error::Fdo there
should be trivial actually since fdo::Error implements DBusError trait

> probably as easy as adding:

> Self::#ident(#zbus::Error::FDO(e)) => e.description(),

I tried my best to explain what is happening, please let me know if you need more details. Happy to edit the title or description to keep it more clear.

PR #1526, opened 2025-09-29
💬 zbus_macros: Add description for `NameTaken` error

arun-mani-j:
We add the error message for the `NameTaken` error. Without this, whenever the error is printed, it says `org.freedesktop.zbus.Error: no description`.

Closes: https://github.com/dbus2/zbus/issues/1523

Ideally we should extend this to all errors but I'm not sure what is the best way to do it. Happy to update the PR if someone could explain the process :sweat_smile:.


zeenix:
@arun-mani-j Thanks so much. I'll look soon but please keep in mind that emojis prefixes are not random but from a curated list and have specific meanings. :) Please use the link provided in the contributors guide to pick one. I am not sure if I can make it any easier. :laughing:

arun-mani-j:
> @arun-mani-j Thanks so much. I'll look soon but please keep in mind that emojis prefixes are not random but from a curated list and have specific meanings. :) Please use the link provided in the contributors guide to pick one. I am not sure if I can make it any easier. 😆

Hey thanks for the comment. I did use your site https://zeenix.github.io/gimoji/ as reference when picking the emoji. And the text-balloon's description looked accurate of what the PR does (adding text literal?):

image

But I'm quite new to the emoji-based PRs, so happy to reword it as per your suggestion :see_no_evil:.

zeenix:
> I did use your site https://zeenix.github.io/gimoji/ as reference when picking the emoji.

Cool! Oh, indeed. Sorry about that.

>  And the text-balloon's description looked accurate of what the PR does (adding text literal?):

Hmm.. yeah, although I feel like it doesn't really represent what the change is about. 🥅 (Catch errors) seems more appropriate, IMO.

> But I'm quite new to the emoji-based PRs, so happy to reword it as per your suggestion 🙈.

No worries, most contributors are new to it since it's not (yet) a widely adopted practice. I really hope it catches on because not only the git history and release notes look so much cooler, it's very helpful for highly-visual folks like myself.

zbus_macros/src/error.rs
```diff
@@ -86,6 +86,7 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
                 if error {
                     quote! {
                         Self::#ident(#zbus::Error::MethodError(_, desc, _)) => desc.as_deref(),
+                        Self::#ident(#zbus::Error::NameTaken) => Some("name already taken on the bus"),
```
zeenix: Oh, I had for some reason assumed that this was about `zbus::Error::FDO`. Indeed, you're right that if we add this, then we should also be adding other variants here and that seems wrong and seem to be doing what `zbus::Error` should be doing.

I recall that we have contemplated `zbus::Error` implementing `DBusError` trait but I can't seem to recall why we didn't go through with that. 🤔 Perhaps now would be the time to do it. If we go that way, we can then just have a `Self::#ident(e) => e.description()` here.

zeenix:
We'll also want a test here.

zbus_macros/src/error.rs
arun-mani-j: Thanks for the review!

So that means we add a new method to `DBusError` which will take care of return an `Option<&str>`? Do we have an example of some similar pattern in `zbus`, so I can use that as reference?

Thanks again :smile:.

zeenix: > So that means we add a new method to `DBusError` which will take care of return an `Option<&str>`?

No. :) `DBusError::description` is already the method that we need. We need to implement `DBusError` trait for `zbus::Errror` and then making use of that, here. The implementation should ensure to always return `Some`.

> Do we have an example of some similar pattern in `zbus`, so I can use that as reference?

I'm not sure what you mean exactly but see the existing trait impls of `zbus::Error`, especially `Display`.

arun-mani-j: I think we can't use `DBusError` macro directly in `Error` because:
```
error[E0119]: conflicting implementations of trait `std::fmt::Display` for type `error::Error`
   --> /home/arun-mani-j/Projects/Phosh/zbus/zbus/src/error.rs:14:10
    |
14  | #[derive(DBusError, Debug)]
    |          ^^^^^^^^^ conflicting implementation for `error::Error`
...
122 | impl fmt::Display for Error {
    | --------------------------- first implementation here
    |
    = note: this error originates in the derive macro `DBusError` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0119]: conflicting implementations of trait `StdError` for type `error::Error`
  --> /home/arun-mani-j/Projects/Phosh/zbus/zbus/src/error.rs:14:10
   |
14 | #[derive(DBusError, Debug)]
   |          ^^^^^^^^^ conflicting implementation for `error::Error`
...
94 | impl error::Error for Error {
   | --------------------------- first implementation here
   |
   = note: this error originates in the derive macro `DBusError` (in Nightly builds, run with -Z macro-backtrace for more info)

   --> /home/arun-mani-j/Projects/Phosh/zbus/zbus/src/error.rs:14:10
    |
14  | #[derive(DBusError, Debug)]
    |          ^^^^^^^^^
    |          |
    |          expected `&str`, found `&&Arc<Error>`
    |          arguments to this enum variant are incorrect
    |
    = note: expected reference `&str`
               found reference `&&Arc<std::io::Error>`
help: the type constructed contains `&&Arc<std::io::Error>` due to the type of the argument passed
   --> /home/arun-mani-j/Projects/Phosh/zbus/zbus/src/error.rs:14:10
    |
14  | #[derive(DBusError, Debug)]
    |          ^^^^^^^^^ this argument influences the type of `Some`
...
```

Also I'm confused if this is right because won't this mean we are treating every error as a DBus error? 

Assuming that we are going to not use the macro but implement the methods manually, then in `description` method, how do we convert the `String` description to an `Option<&str>`? This is what is confusing me. If I understand it right, we need to store the `String` somewhere to be able to return a reference to it.

Sorry I'm not that proficient in Rust, that's why need some little hand-holding :sweat_smile:.

zeenix: > I think we can't use `DBusError` macro directly in `Error` because:

Those seem like the derive macro limitations/bugs. I was actually assuming a manual implementation anways.

> Also I'm confused if this is right because won't this mean we are treating every error as a DBus error?

Not 100% sure what you mean but it basically would just mean that `zbus::Error` **can** then be used as a `DBusError` wherever needed. All code won't automatically just start treating it as a `DBusError`, it'll be up to the using code.

> Assuming that we are going to not use the macro but implement the methods manually, then in `description` method, how do we convert the `String` description to an `Option<&str>`?

We don't need to be very precise here. We just provide a generic description w/o any specific parameters.

arun-mani-j: Thanks! 

I took a little deviation from your above approach. Instead of implementing `DBusError` trait, I add a `description` method to `zbus::Error` that returns `Options<&str>`. I felt this to be simple.

What do you think? :sweat_smile:

zeenix: That's fine but could you please describe your reasoning? 😊

arun-mani-j: Sure. So to implement `DBusError` trait, we need to implement `create_reply` and `name` methods. This felt unnecessary for the problem we are trying to solve. 

Then each of these methods requires additional context that is not immediately clear. For example what error name should we use for `name` method etc.

So I felt just providing `description` method is better (at least for now).

zeenix: Right, makes sense. Thanks. We could just use `org.zbus.<VariantName>` for names that's not a problem, really but `create_reply` while we don't (currently) need `zbus::Error` to be used as a `DBusError`, does seem like an overkill.

zbus/src/error.rs
```diff
@@ -153,6 +153,34 @@ impl fmt::Display for Error {
     }
 }
 
+impl Error {
+    pub fn description(&self) -> Option<&str> {
```
zeenix, line 162: Missing docs. We should specify how it's different from the `Display::fmt` here (and in the commit message). You're probably going to do that anyway but just a reminder to follow the Rust conventions:

```
Short summary of the method.

Any other details (here, that would be the Display::fmt thing).
```

zbus_macros/src/error.rs
```diff
@@ -85,8 +85,7 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
             Fields::Unnamed(_) => {
                 if error {
                     quote! {
-                        Self::#ident(#zbus::Error::MethodError(_, desc, _)) => desc.as_deref(),
```
zeenix, line 88: We don't want this. The method errors need to be handles the same way.

zbus/src/error.rs
arun-mani-j: Thanks for the review! Do you have some suggestion on what the doc should be? 

I thought of putting it as:
```
A simple description of the error.

One should always prefer the `Display::fmt` for the detailed error message. But for the circumstance where an `Option<&str>` is required, this method can be used.
```

But it doesn't feel neat to me. Like I don't see a place where someone would want to use this method directly, so I'm not sure how to describe its usecase..

zbus_macros/src/error.rs
arun-mani-j: You mean we should bring in that line? But why, that would lead to duplication right? :thinking:

zeenix: > But why, that would lead to duplication right? 🤔

Ah, the `Error::description` does the same? Then it's not needed indeed.

zbus/src/error.rs
zeenix: The use case is when you need a string reference instead of a owned string. You actually don't need to get into "why" but rather just stick to "what":

```
A description of the error.

This is a generic description of the error (if any). For a more detailed description, make use of the [`std::fmt::Display`] implementation (for example, through [`std::string::ToString`]).
```

arun-mani-j: Thanks a lot! Done.

zeenix (changes requested):
LGTM. Just one last nitpick on the first commit message: "description method" -> "Error::description method".

arun-mani-j:
> LGTM. Just one last nitpick on the first commit message: "description method" -> "Error::description method".

Thanks done!

zeenix:
@arun-mani-j Thanks. Please pay attention to the CI. :)

arun-mani-j:
> @arun-mani-j Thanks. Please pay attention to the CI. :)

Thanks, sorry missed it. Fixed in the new commit (wish GitHub allowed to see old CI runs).

zeenix:
Thanks so much for your perseverance on this.

arun-mani-j:
> Thanks so much for your perseverance on this.

Thanks a lot for zbus and helping in every stage of this PR! :smile:

PR #1530, opened 2025-10-03
➖ zb,zv: Switch from `rand` to `fastrand`

Cheong-Lau:
Removes `rand` as a dependency, but since `fastrand` is already a transient dependency this doesn't build any more code.

This removes multiple other dependencies as well, notably `zerocopy` which is quite large.
```
ppv-lite86 0.2.21
rand 0.9.2
rand_chacha 0.9.0
rand_core 0.9.3
zerocopy 0.8.25
zerocopy-derive 0.8.25
```

Currently, `zbus` generates random `u32`s and `zvariant` generates random strings, both of which are trivial enough to be handled by `fastrand`. `rand` provides more utility, more rngs and is reproducible across platforms, so if those are needed then `fastrand` probably isn't appropriate. Otherwise, this change would let downstream crates have fewer dependencies.


zeenix:
@Cheong-Lau Thanks so much for this PR. Actually, if you look at the git history, you'll find that we used to use fastrand and switched to rand on purpose because we needed a cryptographically secure randomness for SHA1 (d4c7666fad0ade913ae5dcde99fbb4ede4400652). However, we have since dropped support for SHA1 (#1039) and now this is only needed for GUID generation and the question is if that requires a secure random? In fact, I wonder if it would be a good idea to use the ubiquitous `uuid` crate here (which is already our optional dep). :thinking: 

@elmarco thoughts?

elmarco:
yes, I suppose uuid could be enough

zeenix:
@elmarco Thanks!

@Cheong-Lau Could you please modify the PR to switch to uuid in zbus. For zvariant's dev dep, we can still switch to fastrand. These would then have to be (at least) 2 separate commits.

Cheong-Lau:
No problem, I can do that once I’m awake tomorrow

Cheong-Lau:
@zeenix I have a question about the `validate_guid()` function in `zbus`, right now it's a custom parser using `winnow`:
```rust
fn validate_guid(value: &str) -> crate::Result<()> {
    use winnow::{stream::AsChar, token::take_while, Parser};

    take_while::<_, _, ()>(32, AsChar::is_hex_digit)
        .map(|_| ())
        .parse(value.as_bytes())
        .map_err(|_| crate::Error::InvalidGUID)
}
```

But `uuid` provides `uuid::Uuid::try_parse(&str)`, so I could replace this function with something simpler, like:
```rust
const fn validate_guid(value: &str) -> crate::Result<()> {
    match uuid::Uuid::try_parse(value) {
        Ok(_) => Ok(()),
        Err(_) => Err(crate::Error::InvalidGUID),
    }
}
```

Keeping the old code lets `uuid` be an optional dependency, but this does mean less "optimal" code to do the parsing.
What are your thoughts?

zeenix:
> Keeping the old code lets `uuid` be an optional dependency, but this does mean less "optimal" code to do the parsing.

1. You mean the `uuid::Uuid::try_parse` is slower than our impl? How come?
2. This is certainly not in the hot path so being a bit slower is fine, especially if that means the validation is more thorough.

zeenix:
> These would then have to be (at least) 2 separate commits.

You still made it the same commit, despite my explicitly informing you how these need to be **at least** 2 commits. :disappointed:

Cheong-Lau:
> 1. You mean the `uuid::Uuid::try_parse` is slower than our impl? How come?

I think it might be the other way around, the `uuid` function checks for specific sentinel values, so I would think that might be the better choice, even if it would make `uuid` required.

> 2. This is certainly not in the hot path so being a bit slower is fine, especially if that means the validation is more thorough.

No problem, I'll push what I have and you can have a look.

> > These would then have to be (at least) 2 separate commits.
> 
> You still made it the same commit, despite my explicitly informing you how these need to be **at least** 2 commits. 😞

Oops, my bad, I should have made it clearer that I hadn't pushed my other commit yet! The only commit that I've pushed right now is the `zvariant` commit, nothing's changed for `zbus` yet. ~~I'll push the other commit soon.~~ Pushed!

zeenix:
> > 1. You mean the `uuid::Uuid::try_parse` is slower than our impl? How come?
> 
> I think it might be the other way around, the `uuid` function checks for specific sentinel values, so I would think that might be the better choice, even if it would make `uuid` required.

Yeah.

> > You still made it the same commit, despite my explicitly informing you how these need to be **at least** 2 commits. 😞
> 
> Oops, my bad, I should have made it clearer that I hadn't pushed my other commit yet!

Ah ok, sorry.

zbus/src/guid.rs
```diff
@@ -130,13 +122,11 @@ impl<'de> Deserialize<'de> for Guid<'de> {
...
 }
 
-fn validate_guid(value: &str) -> crate::Result<()> {
-    use winnow::{stream::AsChar, token::take_while, Parser};
-
-    take_while::<_, _, ()>(32, AsChar::is_hex_digit)
-        .map(|_| ())
-        .parse(value.as_bytes())
-        .map_err(|_| crate::Error::InvalidGUID)
+const fn validate_guid(value: &str) -> crate::Result<()> {
```
zeenix, line 125: Wonder if we can just make `Guid` type a wrapper around `uuid::Uuid`. Then we won't need this separate validation function.

zeenix (commented):
Other than that, looks good. Nitpicks on the commit messages though:

* The first commit should make it clearer that we already indirectly depend on fastrand.
* The commit lines have a 74 chars limit for the description part. Your editor should warning you about that. :wink:

zbus/src/guid.rs
Cheong-Lau: I tried for a bit, and it looks like it would have to be a massive breaking change. `uuid::Uuid` internally is a 16-byte array `[u8; 16]`, so I can't really implement things like `as_str()` very easily. Also, `OwnedGuid` would be a really awkward and confusing type to have around when `Guid` is already owned...

Maybe in a future release, `Guid` could be removed and everything refactored to use `uuid::Uuid` instead? Wouldn't be too hard if I could break everything 🙃

Cheong-Lau:
> * The commit lines have a 74 chars limit for the description part. Your editor should warning you about that. 😉

I used the git cli, maybe I should properly learn git before I do any more PRs 🫠

zbus/src/guid.rs
zeenix: Oh, ok. This API is very rarely used so it won't be a huge breaking change but a breaking change nonetheless and not worth the benefits. So yeah, let's not do that yet. I'll create a new issue for this, which can be addressed in zbus 6.0.

PR ready otherwise?

zeenix:
> > * The commit lines have a 74 chars limit for the description part. Your editor should warning you about that. 😉
> 
> I used the git cli, 

I use `git` too. It typically (always?) launches your editor for writing the commit message and that's typically configured to break the lines automatically and warn you about this but maybe that doesn't work in all editors. :shrug: 

> maybe I should properly learn git before I do any more PRs 🫠

Learning in general is always good. :) The importance of learning `git` essentials in specific, is often very understated. :shrug:

Cheong-Lau:
> It typically (always?) launches your editor for writing the commit message

Good to know, thanks! I had it configured to launch nano, but I just changed it to my IDE and that is much better.

> Learning in general is always good. :) The importance of learning `git` essentials in specific, is often very understated. 🤷

I agree, I just like to learn by doing :)

Anyways, if there aren't any other concerns I think this is ready to merge. Thanks for your help!

zbus/Cargo.toml
```diff
@@ -25,7 +25,7 @@ camino = ["zvariant/camino"]
 # Enables API that is only needed for bus implementations (enables `p2p`).
 bus-impl = ["p2p"]
 # Enables API that is only needed for peer-to-peer (p2p) connections.
-p2p = ["dep:rand"]
+p2p = ["uuid/v4"]
```
zeenix, line 28: hmm.. actually we could still keep `uuid` optional but required by `p2p` feature and always enable `v4` feature of `uuid` in zbus (only, zvariant shouldn't need to require `v4` I think).

zeenix (changes requested):
LGTM otherwise.

zeenix:
@Cheong-Lau Hi. Please don't merge main (that creates redundant charges in the PR), only rebase on it. 🙏

zbus/Cargo.toml
Cheong-Lau: we could, but then that would mean I can't use uuid's `try_parse()` in `validate_guid()` :face_with_diagonal_mouth: 

(maybe you misunderstood my comment a few days ago? I was wondering which approach to use myself.)

Cheong-Lau:
> @Cheong-Lau Hi. Please don't merge main (that creates redundant charges in the PR), only rebase on it. 🙏

Got it, I was just trying out some changes. I'll rebase once I figure out how to handle `try_parse()`

zbus/Cargo.toml
Cheong-Lau: Yeah, no matter what this doesn't look like it's a very clean change, both ways are pretty hacky. I think it might be best to just use the old code, then just replace it all later in a breaking release.

Any other way around this? :man_shrugging:

zeenix: > we could, but then that would mean I can't use uuid's `try_parse()` in `validate_guid()` 🫤

I'm sorry, I misunderstood and thought that `validdate_guid` is needed for `p2p` only. Let's keep this as is.

zbus/Cargo.toml
```diff
@@ -62,7 +62,7 @@ futures-lite.workspace = true
 async-broadcast.workspace = true
 hex.workspace = true
 ordered-stream.workspace = true
-rand = { workspace = true, optional = true }
+uuid.workspace = true
```
zeenix, line 65: commit log is not metioning that uuid is now a mandatory dep.

Cheong-Lau: Oops, forgot I overwrote the commit message. Fixed now

PR #1536, opened 2025-10-09
✨ zv: Added a signature! macro

### LINKED ISSUE #984: Provide a `signature!` macro that constructs a const singature
This was discussed before but at the time, we didn't have the means to do it. Now it should quite easy, with the parsed signature type. In fact, the `Type` macro is already doing it for its `signature` attribute. This would be mostly about moving that logic out into a separate macro. The folks can do things like:

```rust
let singature = signature!("a{sv}");
const SIGNATURE: Signature = signature!("a{sv}");
```

with the string checked at compile-time.

hadihaider055:
Add a `signature!` macro that creates const signatures with compile-time validation. This makes it easier to work with D-Bus signatures while catching invalid ones early.

Fixes #984

### DIFF
--- a/zbus/src/message/header.rs
+++ b/zbus/src/message/header.rs
@@ -321,7 +321,7 @@ mod tests {
     use std::{borrow::Cow, error::Error};
     use test_log::test;
     use zbus_names::{InterfaceName, MemberName};
-    use zvariant::{ObjectPath, Signature};
+    use zvariant::{signature, ObjectPath, Signature};
 
     #[test]
     fn header() -> Result<(), Box<dyn Error>> {
@@ -362,7 +362,7 @@ mod tests {
         assert_eq!(h.destination().unwrap(), ":1.11");
         assert_eq!(h.reply_serial().map(Into::into), Some(88));
         assert_eq!(h.sender(), None);
-        assert_eq!(h.signature(), &Signature::try_from("say").unwrap());
+        assert_eq!(h.signature(), &signature!("say"));
         assert_eq!(h.unix_fds(), Some(12));
 
         Ok(())
--- a/zvariant/src/value.rs
+++ b/zvariant/src/value.rs
@@ -1058,10 +1058,7 @@ mod tests {
 
         assert_eq!(
             Value::new((
-                vec![
-                    Signature::try_from("").unwrap(),
-                    Signature::try_from("(ysa{sd})").unwrap(),
-                ],
+                vec![crate::signature!(""), crate::signature!("(ysa{sd})"),],
                 vec![
                     ObjectPath::from_static_str("/").unwrap(),
                     ObjectPath::from_static_str("/a/very/looooooooooooooooooooooooo0000o0ng/path")
--- a/zvariant/tests/signature.rs
+++ b/zvariant/tests/signature.rs
@@ -1,5 +1,4 @@
-use std::str::FromStr;
-use zvariant::{Signature, Value, LE};
+use zvariant::{signature, Signature, Value, LE};
 
 #[macro_use]
 mod common {
@@ -8,7 +7,8 @@ mod common {
 
 #[test]
 fn signature() {
-    let sig = Signature::from_str("yys").unwrap();
+    let sig: Signature = signature!("yys");
+
     // Structure will always add () around the signature if it's a struct.
     basic_type_test!(LE, DBus, sig, 7, Signature, 1);
 
@@ -20,13 +20,13 @@ fn signature() {
     assert_eq!(v.value_signature(), "g");
     let encoded = value_test!(LE, DBus, v, 10);
     let v = encoded.deserialize::<Value<'_>>().unwrap().0;
-    assert_eq!(v, Value::Signature(Signature::try_from("yys").unwrap()));
+    assert_eq!(v, Value::Signature(signature!("yys")));
 
     // GVariant format now
     #[cfg(feature = "gvariant")]
     {
         let encoded = value_test!(LE, GVariant, v, 8);
         let v = encoded.deserialize::<Value<'_>>().unwrap().0;
-        assert_eq!(v, Value::Signature(Signature::try_from("yys").unwrap()));
+        assert_eq!(v, Value::Signature(signature!("yys")));
     }
 }


zeenix:
@hadihaider055 thanks! I'll have a look later but please remember to use emoji rather than the emoji code. 😉

hadihaider055:
Ah sure @zeenix, I just saw this on mobile and it seems like the emoji code isn't working. Thanks for pointing it out!

zeenix:
> Ah sure @zeenix, I just saw this on mobile and it seems like the emoji code isn't working. Thanks for pointing it out

It won't work in many software (including git cli tools). That's why it's best to follow the advice in the contribution guidelines exactly. 😉

zvariant_derive/src/type.rs
```diff
@@ -192,7 +192,7 @@ fn signature_for_variant(
     }
 }
 
-fn signature_to_tokens(signature: &Signature, zv: &TokenStream) -> TokenStream {
+pub fn signature_to_tokens(signature: &Signature, zv: &TokenStream) -> TokenStream {
```
zeenix: `pub(crate)` no? Also if it's going to be shared between modules, best to either put it in separate module or `utils.rs`. Actually, I think it can simply be moved to the new module (perhaps named `signature`) you'll create for the macro impl (see my other comment in `lib.rs`).

zvariant_derive/src/lib.rs
```diff
@@ -489,3 +492,33 @@ pub fn owned_value_macro_derive(input: TokenStream) -> TokenStream {
...
+            s => Signature::from_str(s).map_err(|e| syn::Error::new(signature_str.span(), e))?,
+        };
+
+        let zv = utils::zvariant_path();
+        let signature_tokens = r#type::signature_to_tokens(&signature, &zv);
+
+        Ok(signature_tokens)
+    };
+
+    result().unwrap_or_else(|err| err.to_compile_error()).into()
```
zeenix: This seems like a very strange pattern to me and also inconsistent with how we do things in the other macros. Please split the impl into a separate function in seperate module.

hadihaider055:
@zeenix, I have moved the implementation to the `zvariant_utils` and re-exported it from the `zvariant`. Fixed all the issues related to `pub(crate)`, and the impl is split and moved to a separate [file](https://github.com/z-galaxy/zbus/pull/1536/files#diff-17bc425b1898f55730e4427419eab0521d5ec9ce723f9353ff2acbb3bafc7376) in `zvariant_utils`.

zeenix:
> @zeenix, I have moved the implementation to the `zvariant_utils` and re-exported it from the `zvariant`. Fixed all the issues related to `pub(crate)`, and the impl is split and moved to a separate [file](https://github.com/z-galaxy/zbus/pull/1536/files#diff-17bc425b1898f55730e4427419eab0521d5ec9ce723f9353ff2acbb3bafc7376) in `zvariant_utils`.

Thanks for doing that. :+1: However, the macro still remains in zvariant_derive. The idea was that it doesn't belong there, so it should only be zvariant that rexports it and the whole macro (not just the impl) should be in zvariant_utils. Having said that, there are some strange rules about where macros could be and can't be, so it's possible that those rules don't allow a macro to be in zvariant_utils. **If** so, we'll have to live with it being in zvariant_derive (even if it's strange). You should be able to very quickly figure that out when you try to move the macro there.

BTW, the PR title and commit title say "Siganture", while the macro is called "signature".

hadihaider055:
> Thanks for doing that. 👍 However, the macro still remains in zvariant_derive. The idea was that it doesn't belong there, so it should only be zvariant that rexports it and the whole macro (not just the impl) should be in zvariant_utils. Having said that, there are some strange rules about where macros could be and can't be, so it's possible that those rules don't allow a macro to be in zvariant_utils. If so, we'll have to live with it being in zvariant_derive (even if it's strange). You should be able to very quickly figure that out when you try to move the macro there.

Thanks @zeenix! You're right on moving the entire macro to `zvariant_utils`, but there's a Rust compiler [limitation](https://doc.rust-lang.org/reference/procedural-macros.html#r-macro.proc.def): procedural macros can only be defined in crates with `proc-macro = true` in their `Cargo.toml`.

Currently, zvariant_derive has `proc-macro = true`; however, the zvariant_utils does not (it's a regular library crate). I've already moved the implementation logic to [macro_impl.rs](https://github.com/z-galaxy/zbus/pull/1536/files#diff-17bc425b1898f55730e4427419eab0521d5ec9ce723f9353ff2acbb3bafc7376) and the macro in [lib.rs](https://github.com/z-galaxy/zbus/pull/1536/files#diff-da8065d324f079f677dc3b6dee4e662f42e8b44a080e433a3a18ba897c8b1b3c) just has a thin wrapper that calls it.

For now, I think we have 2 options:
1. Keep the macro wrapper in zvariant_derive (current approach).
2. Convert zvariant_utils to a proc-macro crate, but this would prevent it from being used as a regular library
dependency.

What do you think is the best approach? I believe option 1 is the most practical since most of the logic is already in `zvariant_utils`.


> BTW, the PR title and commit title say "Siganture", while the macro is called "signature".

Thanks for pointing it out! I'll fix this.

zeenix:
> Thanks @zeenix! You're right on moving the entire macro to `zvariant_utils`, but there's a Rust compiler [limitation](https://doc.rust-lang.org/reference/procedural-macros.html#r-macro.proc.def): procedural macros can only be defined in crates with `proc-macro = true` in their `Cargo.toml`.

Yeah, that's what I feared. :cry: 

> 1. Keep the macro wrapper in zvariant_derive (current approach).

As I wrote in my previous comment, that's what we have to do then. Then there is also no need for the impl to be in _utils since that crate is mainly to keep common code between zvariant_derive and zbus_macros and zvariant API that we need in the macro code itself but can't be in zvariant_derive itself.

> What do you think is the best approach? I believe option 1 is the most practical since most of the logic is already in `zvariant_utils`.

Yeah but we don't want the impl in _utils for the reason I mentioned.

zvariant_derive/src/lib.rs
```diff
@@ -489,3 +490,18 @@ pub fn owned_value_macro_derive(input: TokenStream) -> TokenStream {
         .unwrap_or_else(|err| err.to_compile_error())
         .into()
 }
+
+/// Constructs a const [`Signature`] with compile-time validation.
+///
+/// Allows to create a `Signature` from a string literal at compile time,
+/// with validation to ensure the signature string is valid.
+///
+/// The macro validates the signature string at compile time:
```
zeenix: We need examples here.

zvariant_derive/src/signature.rs
```diff
@@ -0,0 +1,107 @@
...
+    literal: Literal,
+}
+
+impl Parse for SignatureInput {
+    fn parse(input: syn::parse::ParseStream<'_>) -> syn::Result<Self> {
+        Ok(SignatureInput {
+            literal: input.parse()?,
+        })
+    }
+}
```
zeenix: nitpick: We prefer top-down code layout: public API before private and usage before declaration/definition.

zvariant_derive/src/signature.rs
```diff
@@ -0,0 +1,107 @@
...
+///
+/// Takes a string literal signature and converts it to compile-time tokens
+/// representing a const `Signature`.
+pub fn expand_signature_macro(input: TokenStream) -> Result<TokenStream, Error> {
+    let SignatureInput {
+        literal: signature_str,
+    } = syn::parse2(input)?;
+
+    let signature_string = signature_str.to_string();
+    let signature_string = signature_string.trim_matches('"');
```
zeenix, line 18: What's this about? :thinking:

zvariant_derive/src/lib.rs
zeenix: This could just be a copy of the examples in the `Signature` docs itself (after you've converted them to use this macro) + a const example (just copy from the commit log).

zvariant_derive/src/signature.rs
hadihaider055: Just removing the quotes from the string literals. Because we have to convert it into signature tokens with quotes. Let's say `"h" to h`. 

`let signature_string = signature_str.to_string();` This will convert the Literal token to a String which might have `"\"h\""` and if by doing `trimming` it will return the String "h" and strip the escaped quotes `\"\"`.

zeenix:
> Please keep the atomic commits preference in mind, while at it. ;)

:point_up:

hadihaider055:
@zeenix, I've added the documentation with proper examples in the Signature macro and updated the Signature type docs to recommend it. Let me know if there's anything unclear for the readers.

hadihaider055:
> Please keep the atomic commits preference in mind, while at it. ;)

@zeenix, on it! Should I separate it into 3 commits with the feature, docs, and testing? Or rather, the tests be committed with the feature commit `✨ zv: Added a signature macro`.

zeenix:
> @zeenix, on it! Should I separate it into 3 commits with the feature, docs, and testing? Or rather, the tests be committed with the feature commit ✨ zv: Added a signature macro.

Both options are fine but 3 commits would be better indeed.

hadihaider055:
@zeenix, I've split the changes into 3 commits (feature, doc, and test). Is it fine?

hadihaider055:
Sure, thanks @zeenix, I'll do it within the next few days.

PR #1539, opened 2025-10-13
💣 zb: Fix a possible crash when executor poll panic

### LINKED ISSUE #1531: zbus panic on Connection creation when executed during a system reboot
I recently got a crash when constructing a zbus system Connection during a reboot sequence of the host, here is the backtrace extract (I use zbus 5.5.0)
```
#10 0x0000563e82fc8362 in core::result::unwrap_failed ()
#11 0x0000563e830a499e in <zbus::abstractions::executor::Task<T> as core::future::future::Future>::poll ()
#12 0x0000563e83095c19 in zbus::connection::builder::Builder::build_::{{closure}} ()
#13 0x0000563e830948c3 in <core::pin::Pin<P> as core::future::future::Future>::poll ()
#14 0x0000563e830bf24e in zbus::connection::builder::Builder::build::{{closure}} ()
#15 0x0000563e830b50bd in zbus::connection::Connection::system::{{closure}} ()
```

In the log of my app I have in addition these two info:
```
Oct 04 10:34:39.417595: thread 'tokio-runtime-worker' panicked at /usr/src/debug/bmond-rs/v1.0.0+gitAUTOINC+6f7b6c62e2-r0/cargo_home/bitbake/zbus-5.5.0/src/abstractions/executor.rs:225:24:
Oct 04 10:34:39.417595: tokio::task::JoinHandle error: JoinError::Cancelled(Id(124))
```

I didn't manage to reproduce it  (timing dependant ?) but this is surely due to a particular system state during reboot sequence.

fmartinsons-una:
Crash have been seen in particular case when we create a Connection
during a shutdown sequence and when task is cancelled.

Fixes #1531

### DIFF
--- a/zbus/src/abstractions/executor.rs
+++ b/zbus/src/abstractions/executor.rs
@@ -5,9 +5,14 @@ use async_task::Task as AsyncTask;
 #[cfg(not(feature = "tokio"))]
 use std::sync::Arc;
 #[cfg(feature = "tokio")]
-use std::{future::pending, marker::PhantomData};
+use std::{
+    future::pending,
+    io::{Error, ErrorKind},
+    marker::PhantomData,
+};
 use std::{
     future::Future,
+    io::Result,
     pin::Pin,
     task::{Context, Poll},
 };
@@ -204,12 +209,14 @@ impl<T> Drop for Task<T> {
 }
 
 impl<T> Future for Task<T> {
-    type Output = T;
+    type Output = Result<T>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         #[cfg(not(feature = "tokio"))]
         {
-            Pin::new(&mut self.get_mut().0.as_mut().expect("async_task::Task is none")).poll(cx)
+            Pin::new(&mut self.get_mut().0.as_mut().expect("async_task::Task is none"))
+                .poll(cx)
+                .map(|r| Ok(r))
         }
 
         #[cfg(feature = "tokio")]
@@ -222,7 +229,16 @@ impl<T> Future for Task<T> {
                     .expect("tokio::task::JoinHandle is none"),
             )
             .poll(cx)
-            .map(|r| r.expect("tokio::task::JoinHandle error"))
+            .map(|r| match r {
+                Ok(v) => Ok(v),
+                Err(e) => {
+                    if e.is_cancelled() {
+                        Err(Error::new(ErrorKind::Other, "tokio::task cancelled"))
+                    } else {
+                        panic!("tokio::task::JoinHandle error: {e}")
+                    }
+                }
+            })
         }
     }
 }
--- a/zbus/src/address/transport/tcp.rs
+++ b/zbus/src/address/transport/tcp.rs
@@ -147,7 +147,7 @@ impl Tcp {
             "connect tcp",
         )
         .await
-        .map_err(|e| Error::Address(format!("Failed to receive TCP addresses: {e}")))?;
+        .map_err(|e| Error::Address(format!("Failed to receive TCP addresses: {e}")))??;
 
         // we could attempt connections in parallel?
         let mut last_err = Error::Address("Failed to connect".into());
--- a/zbus/src/connection/socket/tcp.rs
+++ b/zbus/src/connection/socket/tcp.rs
@@ -44,7 +44,7 @@ impl ReadHalf for Arc<Async<TcpStream>> {
             },
             "peer credentials",
         )
-        .await
+        .await?
     }
 
     #[cfg(not(windows))]
@@ -78,7 +78,7 @@ impl WriteHalf for Arc<Async<TcpStream>> {
             move || stream.get_ref().shutdown(std::net::Shutdown::Both),
             "close socket",
         )
-        .await
+        .await?
     }
 
     async fn peer_credentials(&mut self) -> io::Result<crate::fdo::ConnectionCredentials> {
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -121,7 +121,7 @@ impl Transport {
                     },
                     "unix stream connection",
                 )
-                .await?;
+                .await??;
                 #[cfg(not(feature = "tokio"))]
                 {
                     Async::new(stream)
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -1766,7 +1766,8 @@ mod p2p_tests {
 
         let server = async {
             #[cfg(all(feature = "vsock", not(feature = "tokio")))]
-            let server = crate::Task::spawn_blocking(move || listener.incoming().next(), "").await;
+            let server =
+                crate::Task::spawn_blocking(move || listener.incoming().next(), "").await?;
             #[cfg(feature = "tokio-vsock")]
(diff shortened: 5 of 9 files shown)


zbus/src/abstractions/executor.rs
```diff
@@ -222,7 +226,16 @@ impl<T> Future for Task<T> {
                     .expect("tokio::task::JoinHandle is none"),
             )
             .poll(cx)
-            .map(|r| r.expect("tokio::task::JoinHandle error"))
+            .map(|r| match r {
+                Ok(v) => Ok(v),
+                Err(e) => {
+                    if e.is_panic() {
+                        Err(io::Error::new(io::ErrorKind::Other, "tokio::task panic"))
```
zeenix: Use of `io::Error` would be appropriate if there was already a specific ErrorKind for this case but given that you had to go for `Other`, means that's not the case. We should use `crate::Error::Failure` here IMO. This would imply the output result type changing to `crate::Result`.

zbus/src/abstractions/executor.rs
```diff
@@ -222,7 +226,16 @@ impl<T> Future for Task<T> {
                     .expect("tokio::task::JoinHandle is none"),
             )
             .poll(cx)
-            .map(|r| r.expect("tokio::task::JoinHandle error"))
+            .map(|r| match r {
+                Ok(v) => Ok(v),
+                Err(e) => {
+                    if e.is_panic() {
```
zeenix: I think I made a mistake and wrote it the other way around on the issue. We should be returning an error in case of `e.is_cancelled`.

zbus/src/abstractions/executor.rs
```diff
@@ -14,6 +14,8 @@ use std::{
 #[cfg(feature = "tokio")]
 use tokio::task::JoinHandle;
 
+use std::io;
```
zeenix: I'm pretty sure fmt will fail on this since we have `imports_granularity = "Crate"` config.

zbus/src/address/transport/tcp.rs
```diff
@@ -151,7 +151,7 @@ impl Tcp {
 
         // we could attempt connections in parallel?
         let mut last_err = Error::Address("Failed to connect".into());
-        for addr in addrs {
+        for addr in addrs? {
```
zeenix: nitpick: best to `?` at `addrs` init on line 150.

zbus/src/abstractions/executor.rs
fmartinsons-una: Sure, that makes more sense

zbus/src/abstractions/executor.rs
fmartinsons-una: There are a lot of place which already use this kind of import

zbus/src/abstractions/executor.rs
fmartinsons-una: Sure, but a lot of caller of this internal API use io::Result as ouput , for example  [here](https://github.com/dbus2/zbus/blob/main/zbus/src/connection/socket/tcp.rs#L75) or [there](https://github.com/dbus2/zbus/blob/main/zbus/src/connection/socket/unix.rs#L322), if I use crate::Result , on which kind of IO error should I map for ?

zbus/src/abstractions/executor.rs
zeenix: Impossible! CI would catch that. You probably missed the `cfg` part. :) E.g See 2 separate imports from `std` on line 8 and 9 here. One is with cfg and the other w/o.

zbus/src/abstractions/executor.rs
zeenix: Good point. If we're going to remap to io::Result, anyway might as well just use io::Result here.

zbus/src/abstractions/executor.rs
fmartinsons-una: Sorry if I misunderstood but there are occurences like [here](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/tcp.rs#L3), [here](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/unix.rs#L3) or [here](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/channel.rs#L1)

But fine, I'll not use raw std::io import for this module

zeenix: > Sorry if I misunderstood but there are occurences like [here](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/tcp.rs#L3), [here]([https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/unix.rs#L3]](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/unix.rs#L3%5D) 

Please read again. These are examples of the different `cfg` I mentioned. You can't do:

```rust
use std::{
     #[cfg(whatever)]
     some_module,
     another_module,
}
```

so you have to put the 2 imports separately and that's what is going on in the code you linked to. As I wrote, the CI would have caught this if we were being inconsistent here.

> or [here](https://github.com/z-galaxy/zbus/blob/main/zbus/src/connection/socket/channel.rs#L1)

I only see a single import here.

> But fine, I'll not use raw std::io import for this module

Not sure what "raw" means in this context but all you should need to do is `cargo +nightly fmt`.

zeenix: Oh, looking at your recently changes, I think you totally misunderstood and now I get what you meant by "raw". I wasn't saying we can't import non-leaf paths (modules), I was saying that all imports from a single crate in a file need to be put in a single `use` statement. That's what `imports_granularity = "Crate"` config is about.

fmartinsons-una: Ok I should have read the doc of rustfmt (https://rust-lang.github.io/rustfmt/?version=v1.8.0&search=group#Crate%5C%3A)  before answering, thanks for the info

fmartinsons-una:
Sorry, an error in vsock.rs have been detected, not sure why I didn't see this on previous run, I pushed force the correction

fmartinsons-una:
> Sorry, an error in vsock.rs have been detected, not sure why I didn't see this on previous run, I pushed force the correction

And again, locally I run 
```
cargo test --all-features
cargo +nightly fmt --all
cargo clippy -- -D warnings
```

and all is ok.

Looking at vsock, I think I'm tricked by  the condition like `#[cfg(all(feature = "vsock", not(feature = "tokio")))]` 
I'll try to fix all errors in vsock.rs and push force again

zeenix:
> Looking at vsock, I think I'm tricked by the condition like `#[cfg(all(feature = "vsock", not(feature = "tokio")))]`

Yeah, `--all-features` doesn't exactly work well for us, unfortunately. It's very annoying that Cargo only has additive features. :shrug: 

Anyway, you can easily reproduce the issue locally by running the same exact command the GH workflow is running and failing.

zbus/src/connection/socket/tcp.rs
zeenix, line 124: @fmartinsons-una You forgot to update here. See #1686.

fmartinsons-una: argh sorry, I'll open a new MR to fix #1686 , do you know if something is feasible on CI side to catch that in the future ?

fmartinsons-una: I've open #1687 , let' continue the discussion there

PR #1543, opened 2025-10-15
docs: Fix broken intra-doc links

wlcx:
Fixes most broken intra-doc links (identified with cargo-deadlinks).

Mostly this replaces manual path links with rustdoc intra-doc links - I _think_ I've identified the correct thing to link to in each case but in some cases I've had to infer a bit. Please do check the links make sense!

There are still some broken links:
```
Found invalid urls in proxy/enum.MethodFlags.html:
	Fragment #tymethod.bits at proxy/enum.MethodFlags.html does not exist!
Found invalid urls in fdo/enum.RequestNameFlags.html:
	Fragment #tymethod.bits at fdo/enum.RequestNameFlags.html does not exist!
Found invalid urls in message/enum.Flags.html:
	Fragment #tymethod.bits at message/enum.Flags.html does not exist!
Found invalid urls in struct.Optional.html:
	Linked file at path /Users/samw/prog/tmp/zbus/target/std/string/struct.String.html does not exist!
Found invalid urls in serialized/index.html:
	Fragment #functions at serialized/index.html does not exist!
Found invalid urls in serialized/struct.Context.html:
	Fragment #functions at serialized/index.html does not exist!
```

but I ran out of time to really dig into them/couldn't work out what they were intended to link to.

Would definitely suggest running cargo-deadlinks as part of CI, if that sounds good I could probably be convinced to add it...

### DIFF
--- a/zbus/src/blocking/proxy/mod.rs
+++ b/zbus/src/blocking/proxy/mod.rs
@@ -50,14 +50,13 @@ pub use builder::Builder;
 ///
 /// # Note
 ///
-/// It is recommended to use the [`proxy`] macro, which provides a more convenient and
-/// type-safe *façade* `Proxy` derived from a Rust trait.
+/// It is recommended to use the [`macro@crate::proxy`] macro, which provides a more
+/// convenient and type-safe *façade* `Proxy` derived from a Rust trait.
 ///
 /// ## Current limitations:
 ///
 /// At the moment, `Proxy` doesn't prevent [auto-launching][al].
 ///
-/// [`proxy`]: attr.proxy.html
 /// [al]: https://github.com/z-galaxy/zbus/issues/54
 #[derive(Clone)]
 pub struct Proxy<'a> {
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -102,7 +102,8 @@ pub(crate) type MsgBroadcaster = Broadcaster<Result<Message>>;
 ///
 /// For higher-level message handling (typed functions, introspection, documentation reasons etc),
 /// it is recommended to wrap the low-level D-Bus messages into Rust functions with the
-/// [`proxy`] and [`interface`] macros instead of doing it directly on a `Connection`.
+/// [`macro@crate::proxy`] and [`macro@crate::interface`] macros instead of doing it directly on a
+/// `Connection`.
 ///
 /// Typically, a connection is made to the session bus with [`Connection::session`], or to the
 /// system bus with [`Connection::system`]. Then the connection is used with [`crate::Proxy`]
@@ -132,8 +133,6 @@ pub(crate) type MsgBroadcaster = Broadcaster<Result<Message>>;
 ///
 /// [method calls]: struct.Connection.html#method.call_method
 /// [signals]: struct.Connection.html#method.emit_signal
-/// [`proxy`]: attr.proxy.html
-/// [`interface`]: attr.interface.html
 /// [`Clone`]: https://doc.rust-lang.org/std/clone/trait.Clone.html
 /// [`set_max_queued`]: struct.Connection.html#method.set_max_queued
 ///
@@ -823,7 +822,7 @@ impl Connection {
 
     /// The unique name of the connection, if set/applicable.
     ///
-    /// The unique name is assigned by the message bus or set manually using
+    /// The unique name is assigned by the message bus, or set manually using
     /// [`Connection::set_unique_name`].
     pub fn unique_name(&self) -> Option<&OwnedUniqueName> {
         self.inner.unique_name.get()
--- a/zbus/src/guid.rs
+++ b/zbus/src/guid.rs
@@ -12,10 +12,9 @@ use zvariant::{Str, Type};
 ///
 /// See the D-Bus specification [UUIDs chapter] for details.
 ///
-/// You can create a `Guid` from an existing string with [`Guid::try_from::<&str>`][TryFrom].
+/// You can create a `Guid` from an existing string with [`Guid::try_from::<&str>`].
 ///
 /// [UUIDs chapter]: https://dbus.freedesktop.org/doc/dbus-specification.html#uuids
-/// [TryFrom]: #impl-TryFrom%3C%26%27_%20str%3E
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Type, Serialize)]
 pub struct Guid<'g>(Str<'g>);
 
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -62,11 +62,8 @@ pub use defaults::Defaults;
 ///
 /// # Note
 ///
-/// It is recommended to use the [`proxy`] macro, which provides a more convenient and
-/// type-safe *façade* `Proxy` derived from a Rust trait.
-///
-/// [`futures` crate]: https://crates.io/crates/futures
-/// [`proxy`]: attr.proxy.html
+/// It is recommended to use the [`macro@crate::proxy`] macro, which provides a more
+/// convenient and type-safe *façade* `Proxy` derived from a Rust trait.
 #[derive(Clone, Debug)]
 pub struct Proxy<'a> {
     pub(crate) inner: Arc<ProxyInner<'a>>,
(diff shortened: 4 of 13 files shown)


zvariant/src/ser.rs
```diff
@@ -208,11 +208,10 @@ where
 /// Serialize `T` that has the given signature, to a new byte vector.
 ///
 /// Use this function instead of [`to_bytes`] if the value being serialized does not implement
-/// [`DynamicType`]. See [`from_slice_for_signature`] documentation for an example of how to use
+/// [`DynamicType`]. See [`deserialize_for_signature`] documentation for an example of how to use
```
zeenix: We could just make this [`Data::deserialize_for_signature`] since `Data` is already in the scope. A lot of what you're updating is from the time before you could directly link like this and all the new doc comments make heavy use of this directly linking so might be good to make things consistent and simpler while you're at it?

zeenix (commented):
Thanks so much! Apart the nitpick on docs style, please also add details to the commit log (so all information is always available in the repo offline) as per our contributions guide (which you followed very precisely otherwise).

zvariant/src/ser.rs
wlcx: Yeah sounds good to me, I suspected that it wasn't supported when the docs were written but didn't want to change in case there was some other reason to have it that way.

zeenix:
> Would definitely suggest running cargo-deadlinks as part of CI, if that sounds good I could probably be convinced to add it...

Sure thing. But even w/o it, [our CI already tests doc build with `-D warnings`](https://github.com/z-galaxy/zbus/blob/d4254d63cce0be0d6ef762644e97cb5aa293d36a/.github/workflows/rust.yml#L187) so if we would convert all the html links to direct links (I really don't know what's the correct term for that but hopefully you know what I'm talking about :laughing:), the CI will catch this.

wlcx:
Heh, just noticed that all the `#tymethod` errors above are a rustdoc bug https://github.com/rust-lang/rust/issues/141115

wlcx:
Could you advise where "serialization and deserialization API" is suppose to link now? (https://docs.rs/zvariant/latest/zvariant/serialized/index.html#functions)

That's the cause for 2 of the 4 remaining dead links

zeenix:
> Could you advise where "serialization and deserialization API" is suppose to link now? (https://docs.rs/zvariant/latest/zvariant/serialized/index.html#functions)

For serialization, you can use [this](https://docs.rs/zvariant/latest/zvariant/index.html#functions) and for and deserialization, [this](https://docs.rs/zvariant/latest/zvariant/serialized/struct.Data.html#method.deserialize).

wlcx:
Ok, fixed them (and replaced a few surrounding path-based links with intra-doc links). I think this is good to go now?

zbus/src/proxy/mod.rs
```diff
@@ -62,8 +62,8 @@ pub use defaults::Defaults;
 ///
 /// # Note
 ///
-/// It is recommended to use the [`proxy`] macro, which provides a more convenient and
-/// type-safe *façade* `Proxy` derived from a Rust trait.
+/// It is recommended to use the [`proxy!`](macro@crate::proxy) macro, which provides a more
+/// convenient and type-safe *façade* `Proxy` derived from a Rust trait.
```
zeenix: What I meant by direct link is [`crate::proxy`] here then we don't need to have the `proxy` footnote below. Similarly for other links you're fixing in here: they could just be direct link.

zeenix: Actually you're not using the footnoted link anymore here and you probably just forgot to remove it. The `!` here is also wrong, since this is an attribute macro.

zeenix:
> I think this is good to go now?

Soon, very soon. :)

zbus/src/proxy/mod.rs
```diff
@@ -62,11 +62,8 @@ pub use defaults::Defaults;
 ///
 /// # Note
 ///
-/// It is recommended to use the [`proxy`] macro, which provides a more convenient and
-/// type-safe *façade* `Proxy` derived from a Rust trait.
-///
-/// [`futures` crate]: https://crates.io/crates/futures
-/// [`proxy`]: attr.proxy.html
+/// It is recommended to use the [`proxy`](macro@crate::proxy) macro, which provides a more
```
zeenix: almost there but it's still not a direct link. Perhaps it wasn't clear what I meant because of GH's formatting. This should clarify:


```suggestion
/// It is recommended to use the [`crate::proxy`] macro, which provides a more
```

As I wrote before, the same goes for many links you're fixing/updating in here.

zbus/src/blocking/proxy/mod.rs
```diff
@@ -50,8 +50,8 @@ pub use builder::Builder;
 ///
 /// # Note
 ///
-/// It is recommended to use the [`proxy`] macro, which provides a more convenient and
-/// type-safe *façade* `Proxy` derived from a Rust trait.
+/// It is recommended to use the [`proxy!`](macro@crate::proxy) macro, which provides a more
```
zeenix: Here, the `!` is still being added.

zbus/src/proxy/mod.rs
wlcx: Ah sorry, I hadn't quite twigged you meant that rather than just avoiding footnote style links.

I mean sure, I was trying to keep the link text identical to avoid `crate::` etc appearing everywhere (it seems that `macro@` gets stripped at least). If you prefer that over using `[]()` I can change it.

wlcx:
How's that?

zbus/src/guid.rs
```diff
@@ -12,10 +12,9 @@ use zvariant::{Str, Type};
 ///
 /// See the D-Bus specification [UUIDs chapter] for details.
 ///
-/// You can create a `Guid` from an existing string with [`Guid::try_from::<&str>`][TryFrom].
+/// You can create a `Guid` from an existing string with [`Guid::try_from::<&str>`].
```
zeenix, line 15: hmm.. this works? 🤔

zvariant/src/type/mod.rs
```diff
@@ -16,24 +16,23 @@ use crate::Signature;
...
-/// [`Serialize`]: https://docs.serde.rs/serde/trait.Serialize.html
-/// [`Deserialize`]: https://docs.serde.rs/serde/de/trait.Deserialize.html
-/// [basic types]: trait.Basic.html
-/// [`Vec`]: https://doc.rust-lang.org/std/vec/struct.Vec.html
-/// [`HashMap`]: https://doc.rust-lang.org/std/collections/struct.HashMap.html
+/// [serialization]: zvariant#functions
+/// [deserialization]: zvariant::serialized::Data::deserialize
+/// [`Serialize`]: trait@serde::Serialize
+/// [`Deserialize`]: serde::de::Deserialize
+/// [`HashMap`]: std::collections::HashMap
```
zeenix: These could also be direct links, no?

zeenix (changes requested):
One question and one fix and it's good to go. :)

zbus/src/guid.rs
wlcx: seems to?

zeenix: I'll trust you checked by building the docs and seeing the results. :)

wlcx: Clicked the resulting link in a `cargo doc` build, yep

zbus/src/connection/mod.rs
```diff
@@ -823,8 +822,8 @@ impl Connection {
 
     /// The unique name of the connection, if set/applicable.
     ///
-    /// The unique name is assigned by the message bus or set manually using
-    /// [`Connection::set_unique_name`].
+    /// The unique name is assigned by the message bus, or set manually using
+    /// [`Connection::set_unique_name`] when the `bus-impl` feature is enabled.
```
zeenix: I had missed this one before. This is not needed. The docs for `Connection::set_unique_name` informs the user that this method is only available when `bus-impl` feature is enabled.

wlcx: yeah but if that feature isn't enabled and you build docs this is a broken link - so thought it was worth mentioning

wlcx: But I guess you're always going to be building with all features, so maybe it's just extra noise. Have removed

zeenix: > But I guess you're always going to be building with all features

Yeah. Otherwise, people won't be able to find API that is feature-dependent. :)

PR #1546, opened 2025-10-17
🚧 Use `sysinfo` for getting user's groups on unix

zeenix:
➕ Use `sysinfo` for getting user's groups in unix
    
`sysinfo` allows us to get the user's groups in a portable and safe way.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -D warnings
-      MSRV: 1.77.0
+      MSRV: 1.88.0
     steps:
       - uses: actions/checkout@v5
       - uses: dtolnay/rust-toolchain@master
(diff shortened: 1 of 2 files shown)


zeenix:
Turns out this has a major performance regression. We don't have a benchmark for the credentials API but I wrote a simple code to see how much time on average we're adding:

```rust
        ...
        let now = std::time::Instant::now();
        client.peer_credentials().await.unwrap();
        let mut elapsed = now.elapsed();

        for _ in 0..1000 {
            let now = std::time::Instant::now();
            client.peer_credentials().await.unwrap();
            // Keep track of the average elapsed time.
            elapsed = (elapsed + now.elapsed()) / 2;
        }

        println!("{:?}", elapsed);
```

I'm sad to report that this branch makes the output go from around 346 µs to 3 ms. I think the issue here is the refreshing of the system. We're interested in only a single user here (for a call) but sysinfo API is strangely designed around fetching all the infos (and then manually refreshing it) and to make things worse, it's not even async.

So in the light of all that, I'm abandoning this.

PR #1553, opened 2025-10-20
zb: make object destruction more robust

Schmarni-Dev:
the current implementation of object destruction causes issues if all interfaces are removed on an object that has a child with interfaces, this pr fixes that by only destroying objects if they don't have children, and cleaning up any parents that no longer have children

### DIFF
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -191,24 +191,46 @@ impl ObjectServer {
         let path = path.try_into().map_err(Into::into)?;
         let mut root = self.root.write().await;
         let (node, manager_path) = root.get_child_mut(&path, false);
+        let manager_path = manager_path.map(|p| p.to_owned());
         let node = node.ok_or(Error::InterfaceNotFound)?;
         if !node.remove_interface(I::name()) {
             return Err(Error::InterfaceNotFound);
         }
-        if let Some(manager_path) = manager_path {
+        if let Some(manager_path) = manager_path.clone() {
             let ctxt = SignalEmitter::new(&self.connection(), manager_path.clone())?;
             ObjectManager::interfaces_removed(&ctxt, path.clone(), (&[I::name()]).into()).await?;
         }
         if node.is_empty() {
-            let mut path_parts = path.rsplit('/').filter(|i| !i.is_empty());
-            let last_part = path_parts.next().unwrap();
-            let ppath = ObjectPath::from_string_unchecked(
-                path_parts.fold(String::new(), |a, p| format!("/{p}{a}")),
-            );
-            root.get_child_mut(&ppath, false)
-                .0
-                .unwrap()
-                .remove_node(last_part);
+            fn gen_paths(paths: &mut Vec<(ObjectPath<'_>, String)>, path: ObjectPath<'_>) {
+                if path.is_empty() {
+                    return;
+                }
+                let mut path_parts = path.rsplit('/').filter(|i| !i.is_empty());
+                let last_part = path_parts.next().unwrap();
+                let ppath = ObjectPath::from_string_unchecked(
+                    path_parts.fold(String::new(), |a, p| format!("/{p}{a}")),
+                );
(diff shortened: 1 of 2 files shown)


zeenix:
Thanks for the PR. I'm going on vacation for a week so I will likely only be able to look into this next week.

zbus/src/object_server/node.rs
```diff
@@ -101,12 +101,16 @@ impl Node {
...
     pub(super) fn is_empty(&self) -> bool {
-        !self.interfaces.keys().any(|k| {
+        (!self.interfaces.keys().any(|k| {
             *k != Peer::name()
                 && *k != Introspectable::name()
                 && *k != Properties::name()
                 && *k != ObjectManager::name()
-        })
+        })) && self.children.is_empty()
+    }
```
zeenix, line 110: Looks like this change is about fixing the logic of "is this node empty" while the other changes aren't related to this fix? If so, could we please split the commit? It will also make it easier for me to review the changes.

zeenix: > Looks like this change is about fixing the logic of "is this node empty" while the other changes aren't related to this fix?

Right, reading up the PR description, the other changes seems to be related but still a separate change: removing all empty nodes above the node being removed.

zbus/src/object_server/mod.rs
```diff
@@ -191,24 +191,46 @@ impl ObjectServer {
...
+            gen_paths(&mut paths, path);
+            for (ppath, last_part) in paths {
+                let path = ObjectPath::from_string_unchecked(format!("{ppath}/{last_part}"));
+                let node = root.get_child_mut(&path, false).0.unwrap();
+                if !node.is_empty() {
+                    break;
+                }
+                let interfaces = node.interfaces().cloned().collect::<Vec<_>>();
+                if let Some(manager_path) = manager_path.clone() {
+                    let ctxt = SignalEmitter::new(&self.connection(), manager_path.clone())?;
```
zeenix, line 226: Why do we keep cloning the `manager_path`? 🤔 Pretty sure we don't need to do that for `if let ...` statements at least.

zbus/src/object_server/mod.rs
```diff
@@ -191,24 +191,46 @@ impl ObjectServer {
...
+                    return;
+                }
+                let mut path_parts = path.rsplit('/').filter(|i| !i.is_empty());
+                let last_part = path_parts.next().unwrap();
+                let ppath = ObjectPath::from_string_unchecked(
+                    path_parts.fold(String::new(), |a, p| format!("/{p}{a}")),
+                );
+                paths.push((ppath.clone(), last_part.to_owned()));
+                gen_paths(paths, ppath);
+            }
```
zeenix, line 215: That's potentially a lot of allocations. I'm confident we don't need them. We can parse the path as we traverse the hierarchy.

zbus/src/object_server/mod.rs
```diff
@@ -191,24 +191,46 @@ impl ObjectServer {
...
+            let mut paths = Vec::new();
+            gen_paths(&mut paths, path);
+            for (ppath, last_part) in paths {
+                let path = ObjectPath::from_string_unchecked(format!("{ppath}/{last_part}"));
+                let node = root.get_child_mut(&path, false).0.unwrap();
+                if !node.is_empty() {
+                    break;
+                }
+                let interfaces = node.interfaces().cloned().collect::<Vec<_>>();
```
zeenix, line 224: why are we cloning here?

zeenix:
@Schmarni-Dev so you think you'll be able to finish this?

zeenix:
> @Schmarni-Dev so you think you'll be able to finish this?

Hello?

zeenix:
@Schmarni-Dev Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

PR #1563, opened 2025-11-04
WIP Implement `OrderedStream` for `PropertyStream`

ids1024:
This seems like it could be useful. The behavior of an ordered stream may bit a little odd in what it does and does not guarantee with the behavior of dropping intermediate values that `PropertyStream` has, but I believe this can still be useful. This looks like it also applies to `NameOwnerChangedStream`, which is already an `OwnedStream`.

I think this approach should work, but there are a few details that still need fixing, and I need to test it. But it seems good to create a PR rather than just leave this sitting in my git stash.

### DIFF
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -16,7 +16,7 @@ use std::{
 use tracing::{debug, info_span, instrument, trace, Instrument};
 
 use zbus_names::{BusName, InterfaceName, MemberName, UniqueName};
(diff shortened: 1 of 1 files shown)


zeenix:
@danieldg This would be something you'd be the best person to review.

zeenix:
@ids1024 would you be finishing this?

zeenix:
@ids1024 it's been several months since my last ping. Please close this if you don't intend to finish it, at least. :pray:

Issue #158, opened 2021-04-18
r#keyword args get the 'r#' in introspection


zeenix:
In GitLab by @valpackett on Apr 18, 2021, 23:47

When an argument name is the same as a reserved keyword in Rust, one has to write the `r#` thing to be able to use it as an identifier:

```rust
    fn create_session(
        &self,
        uid: u32,
        pid: u32,
        service: String,
        r#type: String,
        …
```

Currently the zbus introspection generator includes the `r#` thing in the output though:

![Screen_2021-04-19-00_46_29](https://gitlab.freedesktop.org/dbus/zbus/uploads/b8c06f727c38a95c74832ddde537e058/Screen_2021-04-19-00_46_29.png)

slightlyfaulty:
Just ran into this. Is there no fix planned?

A `type` argument name is pretty common, but not possible with zbus currently it seems.

zeenix:
> Just ran into this. Is there no fix planned?

No. Just a reminder that this is a community project and hence if something is important to you, it's best to provide a PR for this (especially if there is an issue already and maintainers have expressed interest in fixing the issue).

PR #1565, opened 2025-11-05
🐛 macro: ignore r# prefix in method names

### LINKED ISSUE #214: proxy macro panics when one of the methods is `fn r#type(&self)`
In GitLab by @MaxVerevkin on Oct 13, 2021, 17:43

```rust
#[dbus_proxy(
    interface = "org.freedesktop.UPower.Device",
    default_service = "org.freedesktop.UPower"
)]
trait Device {
    /// Type property
    #[dbus_proxy(property)]
    fn r#type(&self) -> zbus::Result<u32>;
}
```
Fails:
```
error: custom attribute panicked
  --> src/blocks/battery/zbus_upower.rs:24:1
   |
24 | / #[dbus_proxy(
25 | |     interface = "org.freedesktop.UPower.Device",
26 | |     default_service = "org.freedesktop.UPower"
27 | | )]
   | |__^
   |
   = help: message: `"cached_r#type"` is not a valid identifier
```

hepptho:
The `r#` prefix required for identifiers that would otherwise be rust keywords is no longer carried over into the dbus method / property names when used in interface methods.

Users no longer have to resort to the `name` attribute in these cases.

There is a related open issue regarding `r#` in parameter names (#158).
That is not addressed here, but might be in a follow-up PR.

<del>This is technically a BREAKING CHANGE,
though it seems unlikely anyone wants an `r#` prefix in their dbus names or relies upon the previous behavior.</del>

Closes #214

### DIFF
--- a/zbus/tests/iface_and_proxy/client.rs
+++ b/zbus/tests/iface_and_proxy/client.rs
@@ -65,10 +65,15 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
             .find(|&n| n.name().is_some_and(|n| n == name))
             .expect("Child node not exist");
     }
-    assert!(node
+    let my_iface = node
         .interfaces()
         .iter()
-        .any(|i| i.name() == "org.freedesktop.MyIface"));
+        .find(|i| i.name() == "org.freedesktop.MyIface")
+        .unwrap();
+    // Test if the r# prefix for the keyword was removed
+    assert!(my_iface.methods().iter().any(|m| m.name() == "Type"));
+    assert!(my_iface.properties().iter().any(|p| p.name() == "Let"));
+    assert!(my_iface.signals().iter().any(|s| s.name() == "Match"));
 
     let proxy = MyIfaceProxy::builder(&conn)
         .destination("org.freedesktop.MyService")?
@@ -152,6 +157,10 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
     proxy.test_no_reply().await?;
     proxy.test_no_autostart().await?;
     proxy.test_interactive_auth().await?;
+
+    assert_eq!(proxy.r#let().await?, 0);
+    proxy.set_let(1).await?;
+    assert_eq!(proxy.r#let().await?, 1);
 
     let err = proxy.fail_property().await;
     assert_eq!(
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -256,6 +256,8 @@ impl MethodInfo {
             if is_property && has_inputs {
                 assert!(name.starts_with("set_"));
                 name = name[4..].to_string();
+            } else if name.starts_with("r#") {
+                name = name[2..].to_string();
             }
             pascal_case(&name)
         });
--- a/zbus_macros/src/proxy.rs
+++ b/zbus_macros/src/proxy.rs
@@ -209,6 +209,9 @@ pub fn create_proxy(
             let property = method_attrs.property.as_ref();
 
             let rust_method_name = m.sig.ident.to_string();
+            let r#_stripped_rust_method_name = rust_method_name
+                .strip_prefix("r#")
+                .unwrap_or(rust_method_name.as_str());
 
             let is_signal = method_attrs.signal;
             let is_property = property.is_some();
@@ -217,10 +220,10 @@ pub fn create_proxy(
             let dbus_member_name = method_attrs.name.clone().unwrap_or_else(|| {
                 case::pascal_or_camel_case(
                     if is_property && has_inputs {
-                        assert!(rust_method_name.starts_with("set_"));
-                        &rust_method_name[4..]
+                        assert!(r#_stripped_rust_method_name.starts_with("set_"));
+                        &r#_stripped_rust_method_name[4..]
                     } else {
-                        &rust_method_name
+                        r#_stripped_rust_method_name
                     },
                     true,
                 )
@@ -241,7 +244,7 @@ pub fn create_proxy(
 
                 gen_proxy_property(
                     &dbus_member_name,
-                    &rust_method_name,
+                    r#_stripped_rust_method_name,
                     m,
                     &async_opts,
                     emits_changed_signal,
@@ -251,7 +254,7 @@ pub fn create_proxy(
                     &proxy_name,
                     &iface_name,
                     &dbus_member_name,
-                    &rust_method_name,
+                    r#_stripped_rust_method_name,
                     m,
                     &async_opts,
                     visibility,
@@ -548,7 +551,8 @@ fn gen_proxy_method_call(
         _ => None,
     };
 
-    let method = Ident::new(rust_method_name, Span::call_site());
+    let mut method = parse_str::<Ident>(rust_method_name)?;
+    method.set_span(Span::call_site());
     let inputs = &m.sig.inputs;
     let mut generics = m.sig.generics.clone();
     let where_clause = generics.where_clause.get_or_insert(parse_quote!(where));
--- a/zbus/tests/iface_and_proxy/iface.rs
+++ b/zbus/tests/iface_and_proxy/iface.rs
@@ -20,6 +20,7 @@ pub struct MyIface {
     emits_changed_invalidates: u32,
     emits_changed_const: u32,
     emits_changed_false: u32,
+    r#let: u32,
 }
 
 impl MyIface {
@@ -32,6 +33,7 @@ impl MyIface {
             emits_changed_invalidates: 0,
             emits_changed_const: 0,
             emits_changed_false: 0,
+            r#let: 0,
         }
     }
 }
@@ -179,6 +181,12 @@ impl MyIface {
             .send(NextAction::DestroyObj(key.into()))
             .await
             .unwrap();
+    }
+
+    #[instrument]
+    async fn r#type(&self) -> zbus::fdo::Result<String> {
+        debug!("`r#type` called.");
+        Ok("r# prefix method works".into())
     }
 
     #[cfg(feature = "option-as-array")]
@@ -363,6 +371,9 @@ impl MyIface {
     #[zbus(signal)]
     pub async fn alert_count(emitter: &SignalEmitter<'_>, val: u32) -> zbus::Result<()>;
 
+    #[zbus(signal)]
+    pub async fn r#match(emitter: &SignalEmitter<'_>) -> zbus::Result<()>;
+
     #[instrument]
     #[zbus(property)]
     fn emits_changed_default(&self) -> u32 {
@@ -435,6 +446,21 @@ impl MyIface {
     fn set_emits_changed_false(&mut self, val: u32) -> zbus::fdo::Result<()> {
         debug!("`EmitsChangedFalse` setter called.");
         self.emits_changed_false = val;
+        Ok(())
+    }
+
+    #[instrument]
+    #[zbus(property)]
+    fn r#let(&self) -> u32 {
+        debug!("`Let` getter called.");
+        self.r#let
+    }
+
+    #[instrument]
+    #[zbus(property)]
+    fn set_let(&mut self, val: u32) -> zbus::fdo::Result<()> {
+        debug!("`Let` setter called.");
+        self.r#let = val;
         Ok(())
     }
 


hepptho:
Disclaimer: I am not a macro expert at all. I hope this change is small enough not to hide any stupid mistakes, but if it does, I apologize.

zbus_macros/src/iface.rs
```diff
@@ -257,6 +257,9 @@ impl MethodInfo {
                 assert!(name.starts_with("set_"));
                 name = name[4..].to_string();
             }
+            if name.starts_with("r#") {
```
zeenix: nit: This should be an `else if` I think, since this won't happen for the property setters that the previous `if` is for?

zbus_macros/src/proxy.rs
```diff
@@ -542,7 +547,19 @@ fn gen_proxy_method_call(
...
+            ident
+        }
+        Err(_) => {
+            // Ident::new("r#keyword, ...) panics, even though it's a valid identifier.
+            // syn::parse_str works, so we use that instead.
+            let mut ident = parse_str::<Ident>(format!("r#{snake_case_name}").as_str())?;
+            ident.set_span(Span::call_site());
+            ident
+        }
+    };
```
zeenix: I'm puzzled. Isn't the `snake_case` name passed to this method already has the `r#` removed at line 212?

zeenix (changes requested):
Thanks so much for looking into this. Apart from the nit and a question, could we also add a small test case for this? Just add a method using `r#` in the e2e testcase (file `zbus/tests/iface_and_proxy/iface.rs`).

zbus_macros/src/iface.rs
hepptho: Yes, they're exclusive. I've fixed it.

zbus_macros/src/proxy.rs
hepptho: Yes, the `r#` was removed. That is good when the string is later used as part of another identifier (like `cached_{name}`, which caused the error in #214), because `r#` is not valid in the middle.
And mostly it is used like that, but here it is used as-is, meaning for it to be valid, we need to add the `r#`  back if it's a keyword, and that is what happens.
If `snake_case_name` is not a valid identifier, the first `parse_str` will return an Err and in that case we  try again with the `r#` prefix.

I've added a comment that hopefully makes it clearer.

hepptho:
Changes in the last force-push:

- used `if else` according to comment above
- added clippy fixes
- added comment
- added tests (which was pleasant to do, that test setup is quite nice :+1: )

zeenix:
Thanks so much. I'll have a look soon.

> added clippy fixes

If the fixes were to code you didn't touch, please split them into a separate commit as per our practice of atomic commits. 🙏

hepptho:
Some clippy fixes were in lines I didn't touch, but necessary as a direct consequence of my changes: `method_name` in `proxy.rs` is now `&str` instead of `String`, and clippy complained about now unnecessary `&` when passed as a parameter down the line ("reference immediately dereferenced by the compiler").
Should they still be separated out?

zeenix:
> Some clippy fixes were in lines I didn't touch, but necessary as a direct consequence of my changes: `method_name` in `proxy.rs` is now `&str` instead of `String`, and clippy complained about now unnecessary `&` when passed as a parameter down the line ("reference immediately dereferenced by the compiler").
> Should they still be separated out?

Ah ok. No, that's why started with "if". I just hadn't had a chance to look more closely to the changes.

zbus_macros/src/proxy.rs
zeenix: Thanks for explaining. That makes sense now.

However,  instead of recreating the identifier with `r#`, why don't we just pass the original to this method as well? Then we don't need this complication.

hepptho: That's an option, I just hesitated because there's already a lot of parameters and I wasn't sure how to name them.
But you're right, it's not great as it is either.
What do you think of a helper struct to keep them together, like this:

```rust
struct MethodName {
    original: String,
    r#_stripped: String,
}
```

hepptho: Or maybe `RustMethodName`, to avoid confusion with dbus methods.

zeenix: > What do you think of a helper struct to keep them together, like this:
> Or maybe `RustMethodName`, to avoid confusion with dbus methods.

Sounds good to me. We could actually just make a `Method` struct, add all params into it and then make this function a method of that, called just as `generate`. However, let's not mix this refactor into this change. I'd suggest you first add a new parameter to this function and then do the refactor on top in separate commit.

hepptho: Done. Turns out we don't need both variants in `gen_proxy_method_call`, only the one with `r#`.
The struct isn't necessary, I think.
I would still like to rename parameters in `gen_proxy_property`, `gen_proxy_signal`, and `gen_proxy_method_call` to be clear as to what they get. The latter in particular is confusing, because the parameter names are not consistent with the names of the variables passed to it (`member_name` / `method_name` are switched) .
New commit, or squash into the same one?

hepptho:
Change in the last force-push:

Pass the original rust method name to `gen_proxy_method_call`, to avoid having to add back the `r#` prefix in case of keyword.

zbus_macros/src/proxy.rs
zeenix: > Done. Turns out we don't need both variants in `gen_proxy_method_call`, only the one with `r#`.
> The struct isn't necessary, I think.

Sure.

> I would still like to rename parameters in `gen_proxy_property`, `gen_proxy_signal`, and `gen_proxy_method_call` to be clear as to what they get. The latter in particular is confusing, because the parameter names are not consistent with the names of the variables passed to it (`member_name` / `method_name` are switched) .
> New commit, or squash into the same one?

If you're going to do a general refactor like this, I'd suggest doing it in separate commit as it's unrelated to the `r#` fix. Ideally, refactors should come before other changes as it's more likely that you'd need to revert a functional change than a refactor, but I don't want to make things difficult for you, in case you're not too familiar with git rebasing etc.

---

P.S. Could you please do me a huge personal favour and not start every sentence on a new line? Thanks. :pray:  I know it's a big trend now but not everyone prefers it (I find the text hard to read) and it's grammatically incorrect anyway.

zeenix:
> This is technically a BREAKING CHANGE,
though it seems unlikely anyone wants an r# prefix in their
dbus names or relies upon the previous behavior.

BTW, you can't have an `r#` in D-Bus member names so I don't think this is a breaking change.

zbus_macros/src/proxy.rs
hepptho: No worries, I can rebase with the best of them. (It's because I screw up so much.) RE line breaks: Sure thing, though I did keep the line breaks in the commit message, because the [guidelines](https://handbook.gnome.org/development/commit-messages.html) linked in CONTRIBUTING.md have a character limit per line.

hepptho:
Changes in the last force-push: Put a refactoring commit with name changes before the main commit.

zbus_macros/src/proxy.rs
zeenix: > No worries, I can rebase with the best of them. (It's because I screw up so much.) RE line breaks: Sure thing,

:+1: 

> though I did keep the line breaks in the commit message, because the [guidelines](https://handbook.gnome.org/development/commit-messages.html) linked in CONTRIBUTING.md have a character limit per line.

That's very different. Breaking the line for the line limit on a word boundary is very normal and grammatically correct as well. Breaking the line for a new paragraph is also fine but then you need an empty line in between (which you seem to have done in the commit log, so all good).

zbus_macros/src/proxy.rs
```diff
@@ -800,7 +804,7 @@ fn gen_proxy_signal(
     proxy_name: &Ident,
     iface_name: &str,
     signal_name: &str,
-    rust_method_name: &str,
+    r_hash_stripped_rust_method_name: &str,
```
zeenix: Don't think we need to rename this one. The `r#` being removed, isn't super important for this method.

zbus_macros/src/proxy.rs
```diff
@@ -668,7 +672,7 @@ fn gen_proxy_method_call(
 
 fn gen_proxy_property(
     property_name: &str,
-    rust_method_name: &str,
+    r_hash_stripped_rust_method_name: &str,
```
zeenix: Same here.

zeenix (changes requested):
LGTM otherwise. We're merging soon. :)

zbus_macros/src/proxy.rs
hepptho: Done :)

zbus_macros/src/proxy.rs
hepptho: Done :)

zbus_macros/src/proxy.rs
```diff
@@ -209,6 +209,9 @@ pub fn create_proxy(
             let property = method_attrs.property.as_ref();
 
             let rust_method_name = m.sig.ident.to_string();
+            let r#_stripped_rust_method_name = rust_method_name
```
zeenix, line 212: I think we can just shadow `rust_method_name` here instead of using a new name (just like for the function arguments). So this can just be:

```rust
let rust_method_name = rust_method_name
````

File: zbus_macros/src/proxy.rs [hunk 6]
```diff
@@ -548,7 +551,8 @@ fn gen_proxy_method_call(
         _ => None,
     };
 
-    let method = Ident::new(rust_method_name, Span::call_site());
+    let mut method = parse_str::<Ident>(rust_method_name)?;
+    method.set_span(Span::call_site());
```
zeenix: Why do we need this change? :thinking:

zbus_macros/src/proxy.rs
hepptho: That's how it was: https://github.com/z-galaxy/zbus/compare/840292f3520ca224440e8b42959403db5e3c71d7..0252e9948ecb48ea98ec2d6fc97617b398e264dd

But we need the original with r# to pass to gen_proxy_method_call, otherwise we need to add it back again there if it's a keyword.

zbus_macros/src/proxy.rs
hepptho: `Ident::new` panics when it gets `r#keyword`, even though that's a valid identifier. I'm not sure why, but `parse_str` works.

zbus_macros/src/proxy.rs
hepptho: More generally, we need both because there are two cases: We need the original when the identifier is reused as-is, like in proxy methods, because without `r#` it might not be valid there. And we need it without any `r#` when it is used as part of a derived identifier like `receive_{ident}` (in signals) or `cached_{ident}` (in properties), because there it's not valid _with_ `r#`.

zbus_macros/src/proxy.rs
zeenix: > `Ident::new` panics when it gets `r#keyword`,

Hmm.. that sounds like a bug in `syn` then. It would be great to have an issue link in a comment above this: `

```rust
// FIXME: We should be using `Ident::new` here but currently we can't because of a `syn` bug:
// URL
```
Can you check if there is already one and if not, file one? :pray:

zbus_macros/src/proxy.rs
zeenix: Right, thanks for explaining. Yeah, I guess it's better to be explicit here. :)

zbus_macros/src/proxy.rs
hepptho: I found dtolnay/syn#669, and learned that [`Ident::new_raw`](https://docs.rs/proc-macro2/1.0.103/proc_macro2/struct.Ident.html#method.new_raw) exists. That one takes the passed string and prepends `r#` whether it's a keyword or not (though it also panics if you pass it a string that already starts with `r#`). But when I use it, the `r#` doesn't actually end up everywhere, only where needed. I suspect `quote!` quietly removes it when it's not necessary, but couldn't find any definitive word on that.

Using `Ident::new_raw` would work for us, with the added benefit that we wouldn't need to keep the original method name anymore. We can strip any `r#` so it's safe to use in derived identifiers, and `Ident::new_raw` can handle it when reused as-is even if we're left with a keyword. I am a bit uncertain if it's ok to pass everything through `Ident::new_raw`. It seems to work, there are no problems I can see, but I'm sure I'm not seeing everything. [Here's](https://github.com/hepptho/zbus/commit/9714f6cdb267d4c7b142e0512c74f0ea749ca820) what would change if we use `Ident::new_raw`. Should we go ahead with that?

zeenix: Thanks for looking into this in great detail. Looks like even `Ident::new_raw` can't handle all keywords so let's go with your solution.

zbus/tests/iface_and_proxy/client.rs
```diff
@@ -65,10 +65,16 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32>
...
             .expect("Child node not exist");
     }
-    assert!(node
+    let my_iface = node
         .interfaces()
         .iter()
-        .any(|i| i.name() == "org.freedesktop.MyIface"));
+        .find(|i| i.name() == "org.freedesktop.MyIface");
+    assert!(my_iface.is_some());
+    let my_iface = my_iface.unwrap();
```
zeenix: Great that you're being so thorough here. :+1: `unwrap` covers the assert before so it's unneccessary. In fact you can just append `.unwrap()` to the previous call chain.

zeenix (commented):
Super close! Since you'll be addressing the last nit pick anyway, could you please edit the last commit log so you don't break the line before 74 chars:

```
The r# prefix required for identifiers that would otherwise be rust
keywords is no longer carried over into the dbus method / property /
signal names when used in interface or proxy methods.
```

zbus_macros/src/proxy.rs
hepptho: To be clear, the restrictions regarding `self` and `super` mentioned in the docs of [`Ident::new_raw`](https://docs.rs/proc-macro2/1.0.103/proc_macro2/struct.Ident.html#method.new_raw) apply not only there. `r#self` / `r#super` are not allowed _anywhere_. Some keywords, it turns out, are more key than others, and not accepted in raw identifiers. The [reference](https://doc.rust-lang.org/reference/identifiers.html) lists them as `crate`, `self`, `super`, `Self`.

zbus/tests/iface_and_proxy/client.rs
hepptho: Done, and the commit message as well :)

zbus_macros/src/proxy.rs
zeenix: Ah ok. Thanks. TIL.

hepptho:
Thank you for your guidance :)

zeenix:
> Thank you for your guidance :)

No, thank you for your patience. 😊

PR #1576, opened 2025-11-16
🐛 macros: ignore r# prefix in parameter names

### LINKED ISSUE #158: r#keyword args get the 'r#' in introspection

hepptho:
The r# prefix of raw identifiers in parameter names is no longer carried over into the dbus introspection XML.

This is a follow-up to #1565, which dealt with method names.

Closes #158

### DIFF
--- a/zbus/tests/iface_and_proxy/client.rs
+++ b/zbus/tests/iface_and_proxy/client.rs
@@ -74,6 +74,20 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
     assert!(my_iface.methods().iter().any(|m| m.name() == "Type"));
     assert!(my_iface.properties().iter().any(|p| p.name() == "Let"));
     assert!(my_iface.signals().iter().any(|s| s.name() == "Match"));
+    assert_eq!(
+        my_iface
+            .methods()
+            .iter()
+            .find(|m| m.name() == "RawIdentifierParameter")
+            .unwrap()
+            .args()
+            .iter()
+            .next()
+            .unwrap()
+            .name()
+            .unwrap(),
+        "type"
+    );
 
     let proxy = MyIfaceProxy::builder(&conn)
         .destination("org.freedesktop.MyService")?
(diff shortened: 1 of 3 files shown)


zbus/tests/iface_and_proxy/client.rs
```diff
@@ -162,6 +170,11 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32>
     proxy.set_let(1).await?;
     assert_eq!(proxy.r#let().await?, 1);
 
+    assert_eq!(
+        proxy.raw_identifier_parameter("param").await?,
+        "type: param".to_string()
```
zeenix: Nit: don't think an allocation is needed for a comparison. :)

zbus/tests/iface_and_proxy/client.rs
```diff
@@ -74,6 +74,14 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32>
...
     assert!(my_iface.properties().iter().any(|p| p.name() == "Let"));
     assert!(my_iface.signals().iter().any(|s| s.name() == "Match"));
+    assert!(my_iface
+        .methods()
+        .iter()
+        .find(|m| m.name() == "RawIdentifierParameter")
+        .unwrap()
+        .args()
+        .iter()
+        .any(|a| a.name() == Some("type")));
```
zeenix: We only expect 1 param so let's be specific:


```suggestion
    assert_eq!(my_iface
        .methods()
        .iter()
        .find(|m| m.name() == "RawIdentifierParameter")
        .unwrap()
        .args()
        .iter()
        .next()
        .unwrap()
        .name(),
        "type");
```

zeenix (commented):
LGTM otherwise. Thanks.

zbus/tests/iface_and_proxy/client.rs
hepptho: Yes, good catch.

zbus/tests/iface_and_proxy/client.rs
hepptho: Done :)

PR #1585, opened 2025-11-28
🩹 make cli optional

HigherOrderLogic:
Avoid forcing to compile the big `clap` crate when consumed downstream as library.


zeenix:
> Avoid forcing to compile the big `clap` crate when consumed downstream as library.

That's a great idea. Thanks!

However, I wanted to avoid too many changes for this so I went with a different approach in #1589. Please have a look at that and let me know if you see any issues with that.

I'll close this in favour of my PR but we can always re-open it if it turns out this approach is better.

PR #1589, opened 2025-11-29
➖ xmlgen: Make the library part not depend on clap

zeenix:
By making clap an optional dependency and requiring it as a feature by
the binary, we ensure that by default the library doesn't depend on clap
but the binary does.

This supersedes PR #1585 and solves the same issue but in a simpler/less
intrusive way.

### DIFF
--- a/zbus_xmlgen/Cargo.toml
+++ b/zbus_xmlgen/Cargo.toml
@@ -20,17 +20,21 @@ license = "MIT"
 categories = ["os::unix-apis", "development-tools"]
 readme = "README.md"
 
+[features]
+default = ["cli"]
+cli = ["dep:clap"]
+
 [[bin]]
 name = "zbus-xmlgen"
 path = "src/main.rs"
+required-features = ["cli"]
 
 [dependencies]
 zbus = { path = "../zbus", features = ["blocking-api"], version = "5.5.0" }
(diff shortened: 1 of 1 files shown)


HigherOrderLogic:
I'm generally fine with this as a solution, but seems like `required-features` is not on by default (see rust-lang/cargo#4663) so there maybe some friction introduced for CLI consumer.

zeenix:
> I'm generally fine with this as a solution, but seems like `required-features` is not on by default (see [rust-lang/cargo#4663](https://github.com/rust-lang/cargo/issues/4663)) so there maybe some friction introduced for CLI consumer.

Oh, that's a good point. I should have tested this better:

```
❯ cargo b --bin zbus-xmlgen 
error: target `zbus-xmlgen` in package `zbus_xmlgen` requires the features: `clap`
Consider enabling them by passing, e.g., `--features="clap"`
```

I got a better idea. I'll push it in a second..

zeenix:
> I got a better idea. I'll push it in a second..

Done! It's not really a perfect solution at all, I'm afraid but still something I'd prefer over creating a dummy binary (when cli feature is disabled).

HigherOrderLogic:
Yep looks good to me now.

PR #1590, opened 2025-12-01
➖ remove `proc-macro-crate`

### LINKED ISSUE #1365: Drop proc-macro-crate dependency
This crate brings in quite a few deps and it's a bit a hack anyway. I put a lot of effort into reducing the deps but my efforts weren't [too well accepted](https://github.com/bkchr/proc-macro-crate/pull/56#issuecomment-2588223279). From the extremely rare comments on issues and PRs, I don't think it's even well maintained.

So let's just drop this. It's an API break though so we'll have to wait till 6.0.

cachebag:
closes #1365

Drops `proc-macro-crate`.

The macros no longer detect a renamed `zbus` crate; they default to the hardcoded `::zbus` path (the wire derives to `::zbus::wire`). This is a breaking change for crates that rename `zbus` in their `Cargo.toml`: they must now point the macros at the right path with the `crate` attribute, e.g. `#[interface(crate = "my_zbus")]` for the zbus macros or `#[zbus(crate = "my_zbus::wire")]` for the wire derives. Since the zvariant merge (#1919), crates using the derives through a plain `zbus` dependency need no attribute at all.

For that escape hatch to actually work everywhere, this also:

* forwards the `crate` path from `#[interface]` to the nested `proxy` macro invocation it generates,
* fixes code paths that emitted hardcoded `zbus` paths regardless of the `crate` attribute (`proxy::MethodFlags` and `message::Flags`),
* adds two compile-only fixture crates: one whose `zbus` dependency is renamed and one that uses the wire derives through a plain `zbus` dependency without any `crate` attribute.


zvariant_derive/src/utils.rs
```diff
@@ -1,15 +1,11 @@
 use proc_macro2::TokenStream;
-use proc_macro_crate::{crate_name, FoundCrate};
 use quote::{format_ident, quote};
 use zvariant_utils::{case, def_attrs};
 
-pub fn zvariant_path() -> TokenStream {
-    if let Ok(FoundCrate::Name(name)) = crate_name("zvariant") {
-        let ident = format_ident!("{}", name);
+pub fn zvariant_path(crate_path: Option<&str>) -> TokenStream {
```
zeenix: but nothing passes `Some` to this? 🤔

zeenix:
> If I understood this issue correctly, I think this should cleanly drop the crate.

Yeah, that's the idea.

> Users who rename the crate will need to use the original name or a future crate attribute could be added if needed.

Correct and I think we should add the `crate` attribute **before** dropping `proc-macro-crate`. Unfortunately it's still a breaking change though so we'll likely have to wait till 6.0 to get it merged. We seems to have collected [enough breaking changes in the pipeline](https://github.com/z-galaxy/zbus/milestone/7) though, so I think I'll have to look into that soon. Many issues still need doing though so it could still take some time unfortunately but the more ready PRs are available, the less time I'll have to find and more likely that 6.0 happens sooner so PRs like these really help. So thanks for this contribution.

> P.s. I love the work you've all done. `zbus` is awesome!

Thanks!

zvariant_derive/src/utils.rs
cachebag: I think I was thinking ahead to potentially adding the `crate` attribute. For now I just took the param out. Thanks for catching that

zeenix:
> Correct and I think we should add the crate attribute before dropping proc-macro-crate.

Any chance you could add that? We can already merge and advertise that in the docs then. Best to do that in a separate PR and then rebase this one on top of main when that's merged.

cachebag:
Accidentally deleted this branch on my fork.

cachebag:
Revisiting this once we get to that finish line!

zeenix (approved):
Awesome. LGTM but just to be sure I'll also get a machine review it, in case I missed something. :)

zbus_macros/src/utils.rs
```diff
@@ -13,13 +12,10 @@ pub fn parse_crate_path(crate_attr: Option<&str>) -> Result<Option<syn::Path>, s
 /// Returns the path to the zbus crate.
 ///
 /// If a custom crate path is provided via the `crate` attribute, it will be used.
-/// Otherwise, uses `proc-macro-crate` to detect the crate name.
+/// Otherwise, defaults to `::zbus`.
```
zeenix, line 20: 🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: **Blocking:** `#[interface(crate = "foo", proxy)]` resolves `foo` for the interface expansion, but `Proxy::gen` emits the nested `#[foo::proxy(...)]` invocation without forwarding `crate = "foo"`. That nested proxy expansion therefore reaches this default and generates `::zbus` paths. Once `proc-macro-crate` is removed, a downstream crate that renamed or re-exported zbus has no `zbus` extern and fails despite using the documented escape hatch. Please forward the outer crate path into the generated proxy invocation and cover this with a small downstream compile fixture whose Cargo dependency is actually renamed.

zvariant_derive/src/utils.rs
```diff
@@ -1,20 +1,9 @@
...
-    if let Ok(FoundCrate::Name(name)) = crate_name("zvariant") {
-        let ident = format_ident!("{}", name);
-        quote! { ::#ident }
-    } else if let Ok(FoundCrate::Name(name)) = crate_name("zbus") {
-        let ident = format_ident!("{}", name);
-        quote! { ::#ident::zvariant }
-    } else {
-        quote! { ::zvariant }
-    }
+    quote! { ::zvariant }
```
zeenix: 🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: This removes not only renamed-dependency detection, but also the old `::zbus::zvariant` fallback. A zbus-only consumer following the book examples (`use zbus::zvariant::Type; #[derive(Type)]`) will now expand to `::zvariant` without having a direct `zvariant` dependency, so it fails unless each type supplies `#[zvariant(crate = "zbus::zvariant")]` (or the crate adds a direct dependency). This may be an intentional 6.0 break, but please document that migration, update the affected examples, and add a zbus-only downstream compile fixture so the full scope is deliberate and tested.

zeenix (commented):
🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: I found one blocking macro-composition issue and one broader 6.0 migration/documentation gap; the implementation and lockfile updates otherwise look coherent, and the current CI is fully green.

Separately, please update the PR description: #1605 has already added the `crate` attribute, so it is no longer a future possibility. Please also fix `pro-macro-crate` → `proc-macro-crate` in the PR/commit title and bring the commit message in line with CONTRIBUTING (for example, a `zm,zd:` package prefix and a short explanatory body). The commit is otherwise atomic.

No need to rebase the current two-commit drift from `main` now; that can wait until `main` reaches 6.0.

zvariant_derive/src/utils.rs
zeenix: > but please document that migration

Please ignore this part. We only need to document in the commit message.

book/src/client.md
```diff
@@ -577,6 +577,7 @@ use serde::{Serialize, Deserialize};
 use zbus::{zvariant::Type, proxy};
 
 #[derive(Debug, Type, Serialize, Deserialize)]
+#[zvariant(crate = "zbus::zvariant")]
```
zeenix: Hmm.. now that I see it, I think this would be quite annoying for users actually. Wonder what can be done here. 🤔

Perhaps zbus should just `pub use zvariant::*;` in the root instead? I have been wondering about doing that anyway. One issue with that would be that the attributes will be inconsistent to the user: they'll use `zbus::Type` derive but write `zvariant(signature = "a{sv}")`. We could have zvariant macros take `zbus` as well though, which although would be a workaround but it's not too bad IMO and will effectively solve the problem completely. WDYT?

zeenix: With zbus re-exporting the zvariant API, we'll also most abstract away zvariant for zbus users, who don't really need to care about this split anyway.

zeenix: 🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: The facade direction makes sense, but `pub use zvariant::*;` alone will not fix the macro expansions. A re-exported proc macro cannot tell whether it was invoked as `zbus::Type` or `zvariant::Type`, so the current derive entry points would still generate their default `::zvariant` paths. The derives already accept both helper namespaces (`attributes(zbus, zvariant)`), so accepting `#[zbus(...)]` is not the missing piece either—plain `#[derive(zbus::Type)]` has no helper attribute to switch the path.

The clean solution seems to be zbus-specific proc-macro entry points in `zbus_macros` for the zvariant derives, reusing the shared codegen but defaulting to `::zbus::zvariant`. zbus can explicitly re-export those under the normal derive names, while zvariant's own entry points retain `::zvariant`; for the derives, only an actually renamed zbus dependency then needs the `crate = ...` escape hatch. The same zbus-default entry point is needed for `signature!`, whose current expansion hardcodes `::zvariant`; unlike the derives it has no crate-path escape hatch, so rename support there would need a separate mechanism or remain a documented exception. I would make the wrapper-macro work a prerequisite change and then rebase this PR, so the book does not gain a `crate = "zbus::zvariant"` annotation on every type.

Flattening more of zvariant into zbus can still be considered as the public-API decision, but I would avoid a blind glob: `Error` and `Result` already conflict with zbus's root API, and a glob couples zbus to every future zvariant export. Curated re-exports (plus the explicit wrapper macros) keep the abstraction without that ambiguity.

zeenix (commented):
🤖 ChatGPT GPT-5.6 on behalf of Zeeshan: Rechecked `ddba1be`. The two earlier implementation findings are fixed and covered by the new downstream fixtures, and I found no additional code defect in the current explicit-`crate` approach. CodSpeed, Lint/Build/Test, and commitlint all pass.

The open facade/macro thread is now the substantive design blocker; I replied there. Two housekeeping items remain for the eventual 6.0 rebase:

- `ddba1be` is unrelated to this PR and duplicates the exact Clippy fix already carried by #1917. It is reasonable if this old branch needs it to stay green for now, but the fix should land once and this commit should be dropped when #1590 rebases.
- The PR title still says `pro-macro-crate`, and the description still presents the `crate` attribute as a future possibility even though #1605 and this update already use it.

book/src/client.md
cachebag: `zbus::zvariant::Type` is `zvariant::Type`, so it'll still default to `::zvariant` even after the wrapper. I also think this means the book examples need migrating either way. 

I can do a follow up PR for this.

zeenix: Hmm.. good point from the machine. 

Another (rather ambitious idea) could to be completely merge zvariant into zbus, now that gvariant support is split into a separate crate. If we do this, we'll want to feature-gate all the existing API so that zbus could be used as a drop-in replacement for anyone needing zvariant directly.

zeenix: For the merge idea, I created [another separate issue](https://github.com/z-galaxy/zbus/issues/1919). Let's discuss that one there from now.

zeenix: @cachebag Just fyi, I'm working on #1919. Since it's mostly boring mechanical changes, I gave the task to Claude.

zeenix: @cachebag this is all in main now. Please rebase and enjoy. :)

zbus_macros/src/iface.rs
```diff
@@ -816,7 +817,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                             #args_from_msg
                             let reply = self.#ident(#args_names)#method_await;
                             let hdr = __zbus__message.header();
-                            if hdr.primary().flags().contains(zbus::message::Flags::NoReplyExpected) {
+                            if hdr.primary().flags().contains(#zbus::message::Flags::NoReplyExpected) {
```
zeenix, line 820: I think these changes should go into their own commit. Also the test fixtures.

zeenix (changes requested):
LGTM otherwise.

zeenix (approved):
Awesome!

PR #1593, opened 2025-12-02
🚸 Warn when user requests name before object server setup (#1575)

### LINKED ISSUE #1575: Warn or panic on name registration before object server setup
In your service, if you register your name before setting up the object server, any method calls that come between these 2 events, can get lost and it could cause other interesting issues too. This is why we document this caveat but looks like people still end up making this mistake.

This is a follow-up from the discussion [here](https://github.com/z-galaxy/zbus/issues/1109#issuecomment-3536461547) but also from seeing people making this mistake all the time, despite it being well-documented. If we can be 100% certain there is no legitimate reason to do this, we can even make it a panic so people never make this mistake and waste their time. Otherwise, we could just make it a warning on the tracing log.

cachebag:
closes #1575

I think panicking might be okay but I've had instances where I request a name without needing the `ObjectServer` (e.g., using the low-level `MessageStream` API or just claiming a name for presence on the bus).

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -625,6 +625,16 @@ impl Connection {
         W::Error: Into<Error>,
     {
         let well_known_name = well_known_name.try_into().map_err(Into::into)?;
+
+        // Warn if requesting a name before setting up the object server, as this can cause
+        // method calls to be lost.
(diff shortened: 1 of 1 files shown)


zeenix (approved):
Thanks! LGTM.

PR #1594, opened 2025-12-03
🐛 Fix `Peer::get_machine_id` on Windows

### LINKED ISSUE #373: Peer::get_machine_id is unix-specific
It doesn't need to be. The D-Bus spec even recommends which windows API to use.

Not sure what to do on Mac though. 🤔

cachebag:
Use `GetCurrentHwProfile` API as seen in D-Bus spec (#373)

closes #373

--

So macOS doesn't have a `/var/lib/dbus/machine-id` or `/etc/machine-id`, so a proper implementation would use `IOPlatformUUID` via [IOKit](https://developer.apple.com/documentation/iokit) or `gethostuuid()`. 

A `gethostuuid()` syscall would be would be the simplest path forward as it avoids extra deps. Perhaps even simpler than that would be to shell out out to `ioreg` but that seems quite crude.

It technically will compile still via `#[cfg(unix)]` but spit a runtime error: `IOError` (since the files don't exist). I wanted to add a TODO comment for this but not sure how welcome those are.

### DIFF
--- a/zbus/src/fdo/peer.rs
+++ b/zbus/src/fdo/peer.rs
@@ -25,22 +25,60 @@ impl Peer {
     /// single system at least until that system next reboots. It should be the same across reboots
     /// if possible, but this is not always possible to implement and is not guaranteed. It does not
     /// matter which object path a GetMachineId is sent to.
+    ///
+    /// Note: Currently only implemented for Linux and Windows. On other platforms (macOS, *BSD),
+    /// this method returns a `NotSupported` error.
     fn get_machine_id(&self) -> Result<String> {
-        let mut id = match std::fs::read_to_string("/var/lib/dbus/machine-id") {
-            Ok(id) => id,
-            Err(e) => {
-                if let Ok(id) = std::fs::read_to_string("/etc/machine-id") {
-                    id
-                } else {
-                    return Err(Error::IOError(format!(
-                        "Failed to read from /var/lib/dbus/machine-id or /etc/machine-id: {e}"
-                    )));
-                }
-            }
-        };
+        get_machine_id()
+    }
+}
 
-        let len = id.trim_end().len();
-        id.truncate(len);
-        Ok(id)
+#[cfg(target_os = "linux")]
+fn get_machine_id() -> Result<String> {
+    let mut id = match std::fs::read_to_string("/var/lib/dbus/machine-id") {
+        Ok(id) => id,
+        Err(e) => {
+            if let Ok(id) = std::fs::read_to_string("/etc/machine-id") {
+                id
+            } else {
+                return Err(Error::IOError(format!(
+                    "Failed to read from /var/lib/dbus/machine-id or /etc/machine-id: {e}"
+                )));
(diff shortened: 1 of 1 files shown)


zbus/src/fdo/peer.rs
```diff
@@ -26,21 +26,77 @@ impl Peer {
...
+        }
+    };
+
+    let len = id.trim_end().len();
+    id.truncate(len);
+    Ok(id)
+}
+
+#[cfg(windows)]
+fn get_machine_id() -> Result<String> {
```
zeenix, line 58: we've been keeping windows-specific code in `win32` module so best the impl goes there.

zeenix:
> It technically will compile still via `#[cfg(unix)]` but spit a runtime error: `IOError` (since the files don't exist). I wanted to add a TODO comment for this but not sure how welcome those are.

TODO is more than welcome but `#[cfg(unix)]` is not appropriate if it's Linux-specific currently. I think it'd be best to have this (and the test) under `#[cfg(target_os = "linux")]` and since we know this won't work on non-Linux, we add a dummy impl that errors out for `#[cfg(all(unix, not(target_os = "linux")))]`. The `TODO:` comment can go on top of that then. :)

zbus/src/fdo/peer.rs
cachebag: Makes sense. Should be fixed now

zbus/src/win32.rs
```diff
@@ -332,4 +352,14 @@ mod tests {
         let sid = process_token.sid().unwrap();
         assert!(!sid.is_empty());
     }
+
+    #[test]
+    fn test_machine_id() {
```
zeenix: nitpick: `test_` prefix in Rust is redundant, especially if the containing module is called `tests`.

zbus/src/fdo/peer.rs
```diff
@@ -26,21 +26,56 @@ impl Peer {
...
-        };
+        }
+    };
+
+    let len = id.trim_end().len();
+    id.truncate(len);
+    Ok(id)
+}
+
+// TODO: Use `IOPlatformUUID` via IOKit or `gethostuuid()` on macOS.
```
zeenix:
```suggestion
// TODO: Use `IOPlatformUUID` via IOKit or `gethostuuid()` on macOS and *BSD.
```

zbus/src/fdo/peer.rs
```diff
@@ -26,21 +26,56 @@ impl Peer {
     /// if possible, but this is not always possible to implement and is not guaranteed. It does not
     /// matter which object path a GetMachineId is sent to.
     fn get_machine_id(&self) -> Result<String> {
```
zeenix, line 31: It might be best to update the docs while you're at this, to reflect that currently this method errors out on all platforms except Linux and Windows (in a separate commit, since the limitations were already there).

cachebag:
Thanks for the feedback. Lmk if anything else is needed :D

PR #1597, opened 2025-12-05
⚡️ Store object interfaces sorted by their interface names

mulkieran:
Within each interface, the introspection information for methods and
properties is sorted syntactically.

Combining non-random order of interfaces with syntactic order for their
methods and properties makes the introspection information yielded by
two invocations of the org.freedesktop.DBus.Introspectable interface's
Introspect method readily comparable via simple textual methods or the
informed glance of a human being.

There's no real benefit to the Introspect method returning the
introspection information for each interface in random order.

Usually, an object path doesn't have all that many interfaces. Since a
the order of the BTreeMap is 11, it is likely that all the interfaces
will fit into a single root node (or there might be two child nodes).
The HashMap size is usually doubled as it reaches capacity, as far as I
have observed, so there may not be much to choose between the two
implementations in terms of space usage.

If the usual case is assumed to be about five interfaces, the standard
freedesktop interfaces and then one or two more, HashMap may do slightly
better in terms of space allocation.

Interfaces can be added or removed, but it doesn't happen with
extraordinary frequency. So the insertion cost, generally higher for the
BTreeMap is likely not worth worrying about in this context.

Using a BTreeMap is likely to be more efficient when iterating over
all elements, since a BTreeMap is designed with keys packed rather tightly
so that all may fit in a single cache line.

### DIFF
--- a/zbus/src/object_server/node.rs
+++ b/zbus/src/object_server/node.rs
@@ -1,7 +1,7 @@
 //! The object server API.
 
 use std::{
-    collections::{hash_map::Entry, HashMap},
+    collections::{btree_map, hash_map, BTreeMap, HashMap},
     fmt::Write,
 };
 
@@ -20,7 +20,7 @@ use super::{ArcInterface, Interface};
 pub(crate) struct Node {
     path: OwnedObjectPath,
     children: HashMap<String, Node>,
-    interfaces: HashMap<InterfaceName<'static>, ArcInterface>,
+    interfaces: BTreeMap<InterfaceName<'static>, ArcInterface>,
 }
 
 impl Node {
@@ -77,7 +77,7 @@ impl Node {
 
             write!(&mut node_path, "/{i}").unwrap();
             match node.children.entry(i.into()) {
-                Entry::Vacant(e) => {
+                hash_map::Entry::Vacant(e) => {
                     if create {
                         let path = node_path.as_str().try_into().expect("Invalid Object Path");
                         node = e.insert(Node::new(path));
@@ -85,7 +85,7 @@ impl Node {
                         return (None, obj_manager_path);
                     }
                 }
-                Entry::Occupied(e) => node = e.into_mut(),
+                hash_map::Entry::Occupied(e) => node = e.into_mut(),
             }
         }
 
@@ -119,11 +119,11 @@ impl Node {
         arc_iface: ArcInterface,
     ) -> bool {
         match self.interfaces.entry(name) {
-            Entry::Vacant(e) => {
+            btree_map::Entry::Vacant(e) => {
                 e.insert(arc_iface);
                 true
             }
-            Entry::Occupied(_) => false,
+            btree_map::Entry::Occupied(_) => false,
         }
     }
 


zeenix:
@mulkieran thanks for adding super detailed rationale. Much appreciated! 🙏 However, the most important bit (for me) is missing: why do we/clients need this order? 

Also I didn't quite understand why we can't switch to a BtreeMap. 🤔 The order of the entries in this map doesn't depend on when you inserted them, but by the Ord impl of the keys AFAIK but I'm probably missing something here?

mulkieran:
> @mulkieran thanks for adding super detailed rationale. Much appreciated! 🙏 However, the most important bit (for me) is missing: why do we/clients need this order?
> 
> Also I didn't quite understand why we can't switch to a BtreeMap. 🤔 The order of the entries in this map doesn't depend on when you inserted them, but by the Ord impl of the keys AFAIK but I'm probably missing something here?

I think ```BTreeMap``` would also be good. It turns out that the Rust B-Tree has very small nodes, apparently has far as I can tell it is order 11. So most times, the Btree holding interfaces would be a single root node with some kind of sorted array within it. (I hadn't considered the BTree because I had misremembered its order and consequently had thought there would be a lot of space allocation for the single root node and hence a lot of empty space typically for this usage case.)

mulkieran:
> @mulkieran thanks for adding super detailed rationale. Much appreciated! 🙏 However, the most important bit (for me) is missing: why do we/clients need this order?
> 
That is the most philosophical question. I think that some order is better than no order for clients, generally. However, https://github.com/z-galaxy/zbus/issues/992 is a much more compelling concern for me. You can feel free to retire this; I'll revive it if it becomes more compelling at some later date.

zeenix:
> That is the most philosophical question.

It's not :) Changes need justification based on tangible benefits, especially if they have a cost. :) 

> You can feel free to retire this; I'll revive it if it becomes more compelling at some later date.

Could you please check if BTreeMap does what you need? If it does, then I've no problem adding it.

mulkieran:
Based on my arguments for using the BTReeMap for interfaces, using a BTreeMap for nodes is also a good idea. That way they will be ordered by their object path instead of randomly ordered and the space usage will be much less. But those arguments could be faulty...depends on initial capacity of HashMap. EDIT: Yeah, turns out comparing the two implementations according to space usage is not an obvious win for either. I amended commit message.

mulkieran:
@zeenix Ok. I think this is ready again.

zeenix:
> @zeenix Ok. I think this is ready again.

Cool. Thanks so much! I preferred the older title btw, since use of BTreeMap is just an implementation detail.

zbus/src/fdo/object_manager.rs
```diff
@@ -12,7 +15,7 @@ use crate::{interface, message::Header, object_server::SignalEmitter, Connection
 
 /// The type returned by the [`ObjectManagerProxy::get_managed_objects`] method.
 pub type ManagedObjects =
-    HashMap<OwnedObjectPath, HashMap<OwnedInterfaceName, HashMap<String, OwnedValue>>>;
+    HashMap<OwnedObjectPath, BTreeMap<OwnedInterfaceName, HashMap<String, OwnedValue>>>;
```
zeenix: Oh, I just realised that this part is an API break since this is exposed in the public API. Could you please either remove this commit or split it into a separate PR (probably best) so we can merge that one on the next break?

zbus/src/object_server/node.rs
```diff
@@ -222,7 +222,7 @@ impl Node {
         // Recursively get all properties of all interfaces of descendants.
         let mut node_list: Vec<_> = self.children.values().collect();
         while let Some(node) = node_list.pop() {
-            let mut interfaces = HashMap::new();
+            let mut interfaces = BTreeMap::new();
```
zeenix: I believe this change ended up in the wrong commit.

zeenix (changes requested):
LGTM otherwise. :+1:

zbus/src/fdo/object_manager.rs
mulkieran: Ok. I dropped the second commit, which contains that change, and will re-post it in a separate PR after this one is merged.

zbus/src/object_server/node.rs
mulkieran: It belonged in the second commit, since it ends up inserted into the ManagedObjects struct.

zbus/src/object_server/node.rs
```diff
@@ -119,11 +119,11 @@ impl Node {
         arc_iface: ArcInterface,
     ) -> bool {
         match self.interfaces.entry(name) {
-            Entry::Vacant(e) => {
+            btree_map::Entry::Vacant(e) => {
```
zeenix, line 122: This and the following hunk becomes unnecessary if you keep the import of the `Entry` instead of the module? 😉

zeenix (changes requested):
Last nit and I promise next we merge (as long as CI passes, which will need a rebase on the latest `main` because of the new rust release). :)

zbus/src/object_server/node.rs
mulkieran: I tried rewriting the imports so that btree_map and hash_map modules are treated the same. So I use ```hash_map::Entry```, ```btree_map::Entry``` to disambiguate the different Entries.

zeenix: I'm confused. You removed the import of `hash_map::Entry`, so what's the ambiguity? 🤔

zeenix: Oh, sorry. I'm so blind. :(

mulkieran:
Rebased...

PR #1601, opened 2025-12-05
🗑 Deprecate `vec_to_cstr` (#1561)

### LINKED ISSUE #1561: Hide & deprecate accidentally exposed internal vec_to_cstr
In #1010 we not only ended up exposing this internal function but also putting it on the top-level. As I wrote [here](https://github.com/z-galaxy/zbus/pull/1010/files#r2439331798), we should hide and deprecate it.

cachebag:
closes #1561

### DIFF
--- a/zvariant/src/file_path.rs
+++ b/zvariant/src/file_path.rs
@@ -69,6 +69,7 @@ impl<'f> From<&'f OsString> for FilePath<'f> {
 
 impl From<OsString> for FilePath<'_> {
     fn from(value: OsString) -> Self {
+        #[allow(deprecated)]
         FilePath(vec_to_cstr(value.as_encoded_bytes().to_vec()))
     }
 }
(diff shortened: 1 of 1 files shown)


zeenix:
LGTM. I'm not sure why cargo-semver errors out on deprecation. It doesn't make any sense to me.

zvariant/src/file_path.rs
```diff
@@ -141,6 +142,11 @@ impl<'f> From<FilePath<'f>> for PathBuf {
 /// # Returns
 ///
 /// A `Cow<'_, CStr>` containing a *guaranteed* null-terminated string.
+#[doc(hidden)]
+#[deprecated(
+    since = "5.9.0",
+    note = "This function was never meant to be public and will be removed in a future release."
+)]
 pub fn vec_to_cstr(mut bytes: Vec<u8>) -> Cow<'static, CStr> {
```
cachebag, line 150: > I'm not sure why cargo-semver errors out on deprecation. It doesn't make any sense to me.

@zeenix

Probably best to note that changing this to `pub (crate)` would satisfy the lint but obviously break API.

zeenix: Note where? Changing a pub to pub(crate) is very much an API break, while deprecation is not. Semver will and should break on the former, but not on the latter.

cachebag: Thinking about this more because the failing job is annoying me- did we need to bump `Cargo.toml` minor? I guess I'm lost on the order of these checks

zeenix: > Thinking about this more because the failing job is annoying me- did we need to bump `Cargo.toml` minor?

Oh, that might be it.

cachebag: > > Thinking about this more because the failing job is annoying me- did we need to bump `Cargo.toml` minor?
> 
> Oh, that might be it.

Fixed in #1608

PR #1604, opened 2025-12-06
✨ Implement `TryFrom<&Value>` for tuples

HigherOrderLogic:
Make tuples compatible with `.downcast_ref()`.

### DIFF
--- a/zvariant/src/structure.rs
+++ b/zvariant/src/structure.rs
@@ -337,6 +337,19 @@ macro_rules! tuple_impls {
                 }
             }
 
+            impl<'a, E, $($name),+> TryFrom<&Value<'a>> for ($($name),+,)
+            where
+                $($name: TryFrom<Value<'a>, Error = E>,)+
+                crate::Error: From<E>,
+
+            {
+                type Error = crate::Error;
+
+                fn try_from(v: &Value<'a>) -> core::result::Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


zeenix:
Thank you! I'll take a look soon but as pointed out in the contribution guide, please use emoji and not its code (since most git clients don't support them). You just need to click on it (or hit enter after selection) [here](https://zeenix.github.io/gimoji/) and then simply paste in the commit message and PR title.

zeenix:
Also this is more a new feature addition, so ✨ is more appropriate categorization. I took the liberty of adding it to the pr title.

zeenix (approved):

HigherOrderLogic:
Seems like CI failure is not caused by this PR.

Also the perf gain, that's kinda wild...

zeenix:
>Seems like CI failure is not caused by this PR

Need to bump the minor version. Can you please do that in a separate commit (not necessarily a different PR).

zeenix:
I think you need to rebase, at least for some of the issues.

zeenix:
> I think you need to rebase, at least for some of the issues.

Actually, rebase will fix all issues (no need to bump that minor version, it's already done).

zeenix:
> Also the perf gain, that's kinda wild...

That benchmark is super fast (it's in nanoseconds) so even a small variation is declared a huge gain, which it is in terms of percentage. 😊

PR #1605, opened 2025-12-06
✨ zb: Add `crate` attribute for custom crate paths

cachebag:
Allows users to specify custom crate paths via the crate attribute when zbus or zvariant is renamed or re-exported in Cargo.toml.

As discussed in #1590

### DIFF
--- a/zbus_macros/src/lib.rs
+++ b/zbus_macros/src/lib.rs
@@ -53,6 +53,8 @@ mod utils;
 ///   attribute nor one of the default values are specified. Please make sure to explicitly set
 ///   either this attribute or the default values, according to your needs.
 ///
+/// * `crate` - specify the path to the `zbus` crate if it's renamed or re-exported.
+///
 /// Each trait method will be expanded to call to the associated D-Bus remote interface.
 ///
 /// Trait methods accept `proxy` attributes:
@@ -115,7 +117,7 @@ mod utils;
 ///
 /// ```no_run
 /// # use std::error::Error;
-/// use zbus_macros::proxy;
+/// use zbus::proxy;
 /// use zbus::{blocking::Connection, Result, fdo, zvariant::Value};
 /// use futures_util::stream::StreamExt;
 /// use async_io::block_on;
@@ -240,6 +242,8 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///   (Default: `true`). If your interface is well-known or well-documented, you may want to set
 ///   this to `false` to reduce the the size of your binary and D-Bus traffic.
 ///
+/// * `crate` - specify the path to the `zbus` crate if it's renamed or re-exported.
+///
 /// The methods accepts the `interface` attributes:
 ///
 /// * `name` - override the D-Bus name (pascal case form of the method by default)
@@ -312,7 +316,7 @@ pub fn proxy(attr: TokenStream, item: TokenStream) -> TokenStream {
 ///
 /// ```
 /// # use std::error::Error;
-/// use zbus_macros::interface;
+/// use zbus::interface;
 /// use zbus::{ObjectServer, object_server::SignalEmitter, message::Header};
 ///
 /// struct Example {
@@ -400,10 +404,15 @@ pub fn interface(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// Each variant (except for the special `zbus` one) can optionally have a (named or unnamed)
 /// `String` field (which is used as the human-readable error description).
 ///
+/// The following type-level attributes are supported:
+///
+/// * `prefix` - the D-Bus error name prefix.
+/// * `crate` - specify the path to the `zbus` crate if it's renamed or re-exported.
+///
 /// # Example
 ///
 /// ```
-/// use zbus_macros::DBusError;
+/// use zbus::DBusError;
 ///
 /// #[derive(DBusError, Debug)]
 /// #[zbus(prefix = "org.myservice.App")]
--- a/zbus_macros/tests/tests.rs
+++ b/zbus_macros/tests/tests.rs
@@ -312,6 +312,48 @@ fn test_interface() {
     }
 }
 
+// Test that the `crate` attribute works for custom crate paths.
+mod crate_attr_test {
+    #[zbus_macros::proxy(
+        interface = "org.freedesktop.zbus_macros.CrateAttrTest",
+        default_service = "org.freedesktop.zbus_macros",
+        default_path = "/org/freedesktop/zbus_macros/crate_attr_test",
+        crate = "zbus"
+    )]
+    trait CrateAttrTest {
+        fn test_method(&self) -> zbus::Result<String>;
+    }
+}
+
+#[test]
+fn test_interface_with_crate_attr() {
+    use zbus::object_server::Interface;
+
+    struct CrateAttrInterface;
+
+    #[interface(name = "org.freedesktop.zbus.CrateAttrTest", crate = "zbus")]
+    impl CrateAttrInterface {
+        fn test_method(&self) -> String {
+            "test".to_string()
+        }
+    }
+
+    assert_eq!(
+        CrateAttrInterface::name(),
+        "org.freedesktop.zbus.CrateAttrTest"
+    );
+}
+
+#[test]
+fn derive_error_with_crate_attr() {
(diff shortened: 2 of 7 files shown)


zeenix:
That was fast. :) The name should be `crate`. I know you were trying to avoid conflict with the keyword but it's not necessary. See zlink and serde macros' code, which support this very sane attribute for the same reason.

cachebag:
> I know you were trying to avoid conflict with the keyword but it's not necessary. See zlink and serde macros' code, which support this very sane attribute for the same reason.

Ah! Okay perfect. One moment

cachebag:
> See zlink and serde macros' code, which support this very sane attribute for the same reason.
@zeenix 

Okay I'm sadly running out of time today as my fiance has plans for us but maybe you could point me in the right direction..

From the little snooping I did, serde manually parses attributes in `Container::from_ast`. When it sees `meta.path == CRATE`, it extracts the value. It stores this as `Option<syn::Path>` in a field that's NOT named crate at all (if I understood [this ](https://github.com/serde-rs/serde/blob/f7fc3e0d676486236edf3bc3ebe0b9e9a552f81a/serde_derive/src/internals/attr.rs#L483-L487)correctly...)

My solution (for now) is admittedly hackish. Do you mind pointing me to some refs on how exactly serde or zlink support the attribute? I wanna make sure I get this right.

zeenix:
> From the little snooping I did, serde manually parses attributes in `Container::from_ast`. When it sees `meta.path == CRATE`, it extracts the value.

Yes, with `CRATE` being just the "crate" symbol.

 > It stores this as `Option<syn::Path>` in a field that's NOT named crate at all (if I understood [this ](https://github.com/serde-rs/serde/blob/f7fc3e0d676486236edf3bc3ebe0b9e9a552f81a/serde_derive/src/internals/attr.rs#L483-L487)correctly...)

Yeah, and we don't need to name it the same either.

I see two options:

* modify the zvariant_utils API/macro to be able to support keywords too.
* Use a different attribute name, which is what you went for but w/o misspelling it. :) How about `crate_name`?

cachebag:
> modify the zvariant_utils API/macro to be able to support keywords too.

Perfect! I got started on this exact approach but was worried it was out of scope. 

I've got an exam right now and then I'll update this PR when I get home. Thank you!

cachebag:
Did not mean to close, but this should be good now. The macro automatically maps the `crate_path` field to the "crate" attribute name.

zeenix:
> Did not mean to close, but this should be good now. The macro automatically maps the `crate_path` field to the "crate" attribute name.

Awesome! Reminder: atomic commits: every logical change in its own commit with its own log and (optional) rationale. :)

zeenix (changes requested):
LGTM, thanks for going the extra mile! We're missing tests and docs though.

zbus_macros/src/lib.rs
```diff
@@ -189,6 +189,23 @@ mod utils;
 ///
 /// [`zbus_polkit`] is a good example of how to bind a real D-Bus API.
 ///
+/// # Custom crate path
+///
+/// If you've renamed `zbus` in your `Cargo.toml` or are using it through a re-export,
+/// you can specify the crate path using the `crate` attribute:
+///
+/// ```
+/// use zbus_macros::proxy;
```
zeenix: We don't want people to use the macros and derive crates directly so the docs shouldn't use them either. This is why zbus is the dev-dep of zbus_macros and similarly for zvariant and zm.

zvariant_derive/src/lib.rs
```diff
@@ -162,6 +162,21 @@ mod value;
...
+/// ```
+/// use zvariant::Type;
+///
+/// #[derive(Type)]
+/// #[zvariant(crate = "zvariant")]
+/// struct MyStruct {
+///     field: String,
+/// }
+/// ```
+///
```
zeenix, line 179: This ended up in the wrong commit. Please go through all commits and ensure, they only contain changes that belong to them. :pray:

zbus_macros/src/lib.rs
```diff
@@ -403,7 +434,7 @@ pub fn interface(attr: TokenStream, item: TokenStream) -> TokenStream {
 /// # Example
 ///
 /// ```
-/// use zbus_macros::DBusError;
+/// use zbus::DBusError;
```
zeenix, line 415: Not a biggie but for future reference, split unrelated fixes into their own commits, especially trivial improvements like these that we'd not want to revert if we had to revert this commit (e.g if it causes some regression we didn't think about).

zbus_macros/tests/tests.rs
```diff
@@ -312,6 +312,48 @@ fn test_interface() {
...
+        default_path = "/org/freedesktop/zbus_macros/crate_attr_test",
+        crate = "zbus"
+    )]
+    trait CrateAttrTest {
+        fn test_method(&self) -> zbus::Result<String>;
+    }
+}
+
+#[test]
+fn test_interface_with_crate_attr() {
```
zeenix, line 329: Thanks for adding tests and also docs. However, I think they only test that the attribute is accepted and not the actual functionality. Not sure how we can easily test it though. 🤔 

One thing we could do to test the functionality in zvariant, is to use the `zvariant(crate = "zbus::zvariant")` attribute in zbus tests you're adding. If that works, it's unlikely that the zbus macros doesn't, since they share the code through -utils crate. What do you think?

zbus_macros/src/lib.rs
zeenix, line 314: Thanks for adding a separate section but what I meant was listing the new attributes in the lists of attributes here. Just a single line bullet point is sufficient IMO. The example can't be run as real test anyway.

Same goes for all other macors. If you see a list of attributes, please just add it there.

zeenix (changes requested):
Getting there.  Thanks for your patience with my perfectionism. :) Talking of which, don't forget to update  the PR description. 😉

zbus_macros/tests/tests.rs
cachebag: I think so..I think what I can do is add tests in `types.rs` that use `#[zvariant(crate = "zbus::zvariant")]` and `#[zbus(crate = "zbus")]` to actually test the crate path redirection works. I believe that covers this well?

>  If that works, it's unlikely that the zbus macros doesn't, since they share the code through -utils crate. 

The tests I originally added can serve as compile-time acceptance tests, but if they serve no use, you can let me know if you'd prefer I remove those.

zeenix: > I think so..I think what I can do is add tests in `types.rs` that use `#[zvariant(crate = "zbus::zvariant")]` and `#[zbus(crate = "zbus")]` to actually test the crate path redirection works. I believe that covers this well?

The former (`#[zvariant(crate = "zbus::zvariant")]`) yes but the latter is not really testing anything since the value is the same as the default. But as I wrote, just testing the zvariant part through zbus, is sufficient IMO.

> The tests I originally added can serve as compile-time acceptance tests, but if they serve no use, you can let me know if you'd prefer I remove those.

We can keep them. I just want one test that actually tests the functionality.

cachebag: @zeenix 👍🏻 Okay done. Sorry about that merge from main, I actually did think I rebased...not sure where I went wrong.

zeenix: Actually the merge is still there. The first commit is "Merge pull request z-galaxy#1608 from cachebag/ver-bump". It's not `main` you merged but another branch, it seems, which is merged into main. I think you may have forgotten to update your main (`git pull`) so the rebase didn't think this commit is already on the main branch. :)

zeenix: Pro tip: quickly look at the git tree/log before pushing. `tig` command is very useful for that.

zbus_macros/tests/tests.rs
```diff
@@ -312,6 +312,48 @@ fn test_interface() {
...
+    }
+
+    assert_eq!(
+        CrateAttrInterface::name(),
+        "org.freedesktop.zbus.CrateAttrTest"
+    );
+}
+
+#[test]
+fn test_derive_error_with_crate_attr() {
```
zeenix: big nitpick: `test_` is redundant IMO due to `#[test]` just above it and during the run, it's also obvious that it's a run. You don't necessarily have to fix this here but just FYI.

PR #1607, opened 2025-12-09
✨ Support `object`attribute in proxy properties

### LINKED ISSUE #356: Cannot use object attribute on properties
The `object` attribute in the `dbus_proxy` macro is a nice quality-of-life enhancement, but only seems to work on DBus methods. When trying to use it on a DBus property, it produces errors like this:

```
  --> src/wifi/iwd.rs:60:5
   |
60 |     fn adapter(&self);
   |     ^^^^^^^^^^^^^^^^^- help: consider using a semicolon here: `;`
   |     |
   |     expected `()`, found `Result<_, _>`
   |
   = note: expected unit type `()`
                   found enum `std::result::Result<_, _>`
```

This is from `Device` in the following code:

```rust
#[dbus_proxy(
    interface = "net.connman.iwd.Adapter",
    default_service = "net.connman.iwd",
    assume_defaults = false,
)]
pub trait Adapter {
    #[dbus_proxy(property)]
    fn powered(&self) -> Result<bool>;

    #[dbus_proxy(property)]
    fn set_powered(&mut self, powered: bool) -> Result<()>;

    #[dbus_proxy(property)]
    fn name(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn model(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn vendor(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn supported_modes(&self) -> Result<Vec<String>>;
}

#[dbus_proxy(
    interface = "net.connman.iwd.Device",
    default_service = "net.connman.iwd",
    assume_defaults = false,
)]
pub trait Device {
    #[dbus_proxy(property)]
    fn name(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn address(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn powered(&self) -> Result<bool>;

    #[dbus_proxy(property)]
    fn set_powered(&mut self, powered: bool) -> Result<()>;

    #[dbus_proxy(property, object = "Adapter")]
    fn adapter(&self);

    #[dbus_proxy(property)]
    fn mode(&self) -> Result<String>;

    #[dbus_proxy(property)]
    fn set_mode(&mut self, mode: &str) -> Result<()>;
}
```

Adding `-> Result<OwnedObjectPath>;` fixes the error and allows this code to compile, but it seems the `object` attribute is ignored entirely.

Based on all this, I'm guessing the `object` attribute simply isn't implemented for properties. If possible, it would be a nice enhancement to have.

cachebag:
closes #356

### DIFF
--- a/zbus/tests/issue/issue_356.rs
+++ b/zbus/tests/issue/issue_356.rs
@@ -1,0 +1,79 @@
+use test_log::test;
+use tracing::instrument;
+use zbus::zvariant::OwnedObjectPath;
+
+/// The "child" interface that will be referenced by the property.
+struct Adapter {
+    name: String,
+}
+
+#[zbus::interface(interface = "org.test.Adapter", proxy(assume_defaults = true))]
+impl Adapter {
+    #[zbus(property)]
+    fn name(&self) -> String {
+        self.name.clone()
+    }
+}
+
+/// The "parent" interface with a property that returns an ObjectPath to an Adapter.
+struct Device {
+    adapter_path: OwnedObjectPath,
+}
+
+#[zbus::interface(interface = "org.test.Device", proxy(assume_defaults = true))]
+impl Device {
+    /// This property returns an ObjectPath, but the generated proxy should return an AdapterProxy.
+    #[zbus(property, proxy(object = "Adapter"))]
+    fn adapter(&self) -> OwnedObjectPath {
+        self.adapter_path.clone()
+    }
+}
+
+#[test(tokio::test(flavor = "multi_thread"))]
+#[instrument]
+async fn issue_356() {
+    use zbus::connection::Builder;
+
+    let adapter_path: OwnedObjectPath = "/org/test/adapter0".try_into().unwrap();
+    let device_path: OwnedObjectPath = "/org/test/device0".try_into().unwrap();
+
+    // Create the service connection with both interfaces registered
+    let connection = Builder::session()
+        .unwrap()
+        .serve_at(
+            &adapter_path,
+            Adapter {
+                name: "TestAdapter".to_string(),
+            },
+        )
+        .unwrap()
+        .serve_at(
+            &device_path,
+            Device {
+                adapter_path: adapter_path.clone(),
+            },
+        )
+        .unwrap()
+        .name("org.test.Issue356")
+        .unwrap()
+        .build()
+        .await
+        .unwrap();
+
+    // Create a proxy for the Device interface using the auto-generated DeviceProxy
+    let device_proxy = DeviceProxy::builder(&connection)
+        .destination("org.test.Issue356")
+        .unwrap()
+        .path(&device_path)
+        .unwrap()
+        .build()
+        .await
+        .unwrap();
+
+    // Call the adapter() property - this should return an AdapterProxy, not OwnedObjectPath
+    let adapter_proxy = device_proxy.adapter().await.unwrap();
+
+    // Verify we can use the returned proxy to call methods on the Adapter interface
+    let adapter_name = adapter_proxy.name().await.unwrap();
+    assert_eq!(adapter_name, "TestAdapter");
+}
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -1511,7 +1511,7 @@ impl Proxy {
         let member_name = method_info.member_name;
         let mut proxy_method_attrs = quote! { name = #member_name, };
         proxy_method_attrs.extend(match method_info.method_type {
-            MethodType::Signal => quote!(signal),
+            MethodType::Signal => quote!(signal,),
             MethodType::Property(_) => {
                 let emits_changed_signal = properties
                     .get(&member_name)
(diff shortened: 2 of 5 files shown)


zbus_macros/tests/tests.rs
```diff
@@ -18,6 +18,30 @@ mod param {
     }
 }
 
+// Test that the `object` attribute works on properties (issue #356)
```
zeenix: Kudos for adding a test. 👍 However it doesn't actually test the functionality. For proper-test, you'll want to use the `interface` macro here to generate both the interface and proxy, start the interface and retrieve the property.

There are some p2p tests in `zbus::connection` module you can copy the boilerplate from.

cachebag: Fixing now. And thanks for the note on the PR hygiene, will heed that from this point on!

zbus/tests/issue/issue_356.rs
```diff
@@ -0,0 +1,93 @@
...
+#[zbus::interface(name = "org.test.Device")]
+impl Device {
+    #[zbus(property)]
+    fn adapter(&self) -> OwnedObjectPath {
+        self.adapter_path.clone()
+    }
+}
+
+/// Proxy for the Adapter interface.
+#[zbus::proxy(interface = "org.test.Adapter", assume_defaults = false)]
```
zeenix: If you recall, I mentioned that it'd be better test to generate the proxy from the interface directly so it could be a complete end2end test

zbus_macros/tests/tests.rs
zeenix: With the other test, we can just ditch this one reallyl

zeenix:
BTW, best to use the short form for commit titles: `zm` instead of `zbus_macros`.

zbus/tests/issue/issue_356.rs
zeenix, line 1: Some of the changes to this file, ended up in the wrong commit. Since the docs are for the feature you add in the previous commit, feel free to squash the second commit into the first one since they can be thought of one logical change.

zbus_macros/src/iface.rs
zeenix, line 1: same about some changes in this file.

zbus_macros/src/proxy.rs
zeenix, line 1: and this file. :) I believe it was rebase gone wrong? :wink:

zbus/tests/issue/issue_356.rs
cachebag: Hopefully I understood that correctly. I definitely admit I was mistakenly mashing `git commit --amend` and force pushing everytime I revised a change 😅

zeenix: Well we all learn. :) `--amend` is only good for amending the last commit (`HEAD` in git terms). You want to use `--fix COMMIT_TO_FIX` option (or try to use the tools recommended in the contributing guide) and then `git rebase main -i --autosquash` (you can also make autosquash the default in rebase through git config).

zbus_macros/src/iface.rs
zeenix, line 1: Changes to this file are needed? 🤔

zbus_macros/src/proxy.rs
```diff
@@ -771,16 +814,28 @@ fn gen_proxy_property(
...
+        } else {
+            quote! {
+                #(#other_attrs)*
+                #[allow(clippy::needless_question_mark)]
+                pub #usage #signature {
+                    #body
+                }
 
-            #cached_getter_method
+                #cached_getter_method
```
zeenix: Both the if and else are sharing quite a bit. Best to take out the different bits and create them in the if/else blocks and then just include them as `#variable` .

zbus_macros/src/iface.rs
cachebag: When using `#[interface(proxy)]` with `#[zbus(property, proxy(object = "..."))]`, the macro generates proxy code like `property(...), object = "..."`. Without the trailing comma, it would generate `property(...)object = "..."` which is invalid syntax.

cachebag:
> Sorry, I always find new things on each look. 🤦

That's not an apology _you_ should be making 😅

zeenix:
@cachebag please don't merge main, rebase. Otherwise, PR gets redundant commits.

zbus_macros/src/proxy.rs
```diff
@@ -608,6 +609,7 @@ fn gen_proxy_method_call(
                     )
                     #wait?;
                 #proxy_path::builder(&self.0.connection())
+                    .destination(self.0.destination().to_owned())?
```
zeenix: Why is this change needed? This looks like an unrelated fix.

zeenix (changes requested):
Almost ready for merge. :)

zbus_macros/src/proxy.rs
cachebag: Yeah, that's my bad. I thought that even though the method version worked before with proxies that have `default_service` set, it would fail for proxies _without_ defaults so I figured I'd add it to both for consistency.

Just removed it though, feel free to lmk what else I missed :D

PR #1608, opened 2025-12-09
🔖 zv: bump minor version to satisfy lint

cachebag:
Should fix my messing up of the lint in #1601 (hopefully that's the right emoji for this, it's just a version bump technically)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2269,7 +2269,7 @@ dependencies = [
 
 [[package]]
 name = "zvariant"
-version = "5.8.0"
+version = "5.9.0"
 dependencies = [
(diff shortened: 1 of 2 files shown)


zeenix:
> Should fix my messing up of the lint in #1601

We all make mistakes, the best amongst us try to fix them after realising. :+1: 

> (hopefully that's the right emoji for this, it's just a version bump technically)

Actually, no and neither is the mod prefix. :wink: We're not modifying the CI here. `🔖 zv:` would be more appropriate. You can then remove the `zvariant ` from the rest of the log as well.

zeenix (approved):

PR #1610, opened 2025-12-10
👷 ci: introduce `commitlint` workflow

jokeyrhyme:
e.g. 🐛 zbus_macros,zbus_xml: did something

We've added some enforcement of use of `gimoji`-style emoji,
and enforcement of a maximum line length for commit subject,
and a warning for maximum line length for commit body,
to make the contribution process more efficient.


.commitlintrc.mjs
```diff
@@ -0,0 +1,119 @@
...
+export default {
+    defaultIgnores: true,
+    helpUrl: "https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md",
+    ignores: [
+        (commit) =>
+            commit.includes(
+                "This PR was generated by [Mend Renovate](https://mend.io/renovate/).",
+            ),
+        (commit) =>
+            commit.startsWith("🔊 zb: lower trace/instrument verbosity\n"),
```
jokeyrhyme: this commit has a body with a line that is longer than 75 characters: https://github.com/z-galaxy/zbus/commit/1cd3e7d8d29c9da69e87b88b59e8779a6f623eb8

we can exempt it as-is, or increase the body line length beyond what the GNOME rules suggest, or rebase the commit

zeenix: > we can exempt it as-is, or increase the body line length beyond what the GNOME rules suggest, or rebase the commit

The limit doesn't come from GNOME but rather directly from git itself (see `man git commit`) and there is a reason for it. While the value is too low for modern days, many cli tools still assume this limit.

Anyway, I think this job should be optional anyway. We can also split the job, into mandatory part and non-mandatory one.

.commitlintrc.mjs
```diff
@@ -0,0 +1,119 @@
+// nushell> http get https://raw.githubusercontent.com/zeenix/gimoji/refs/heads/main/emojis.json | get gitmojis.emoji | to json
+const gitmojis = [
```
zeenix: As I wrote before:

* we don't want to hardcode the emojis. We need to get them on the fly if possible; otherwise find a way to keep it in sync. I'd very much prefer the former. As I also mentioned, gimoji has logic to detect the lack of emojis in a commit message so one idea could be to add a CLI arg to it that make it, given a file/string, exit with either success or failure status. Then we can use it directly here.
* We don't use gitmoji DB but rather [gimoji DB](https://github.com/zeenix/gimoji/blob/main/emojis.json), which is a superset of the former.

.commitlintrc.mjs
jokeyrhyme: i've made this specific rule warning-only and removed the exemption: https://github.com/z-galaxy/zbus/pull/1610/commits/abc236f44bbf5c768357647bb38f414645495b89

.commitlintrc.mjs
jokeyrhyme: huh, this turned out to be fairly simple: https://github.com/z-galaxy/zbus/pull/1610/commits/21cd15dd963f1f09880fda0c2e491419c9059404

but i do also like the idea of raising a `gimoji --lint` PR over there ...

jokeyrhyme:
haha, i just ran afoul my own commit message linting work ... :P

jokeyrhyme:
oh, fyi, you can review that actual outcome of the proposed pipeline here: https://github.com/jokeyrhyme/zbus/actions/workflows/commitlint.yml

.commitlintrc.mjs
zeenix: > i've made this specific rule warning-only and removed the exemption: [abc236f](https://github.com/z-galaxy/zbus/commit/abc236f44bbf5c768357647bb38f414645495b89)

Is the warning visible on the PR? If it is only visible in the CI logs, nobody would see it. :)

.commitlintrc.mjs
```diff
@@ -0,0 +1,48 @@
...
+    );
+    assert.ok(res.ok);
+    const json = await res.json();
+    gimojis = json.gitmojis.map(({ emoji }) => emoji);
+} catch (cause) {
+    throw new Error("unable to fetch gimoji emojis.json", { cause });
+}
+
+// extra symbols from neither gitmoji nor gimoji
+const extras = ["🗑"];
```
zeenix: Don't think we need this. If anyone used a custom emoji in the past, we don't need to care about it. We'll catch them using it in the future and that's the important bit.

.commitlintrc.mjs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            rules: {
+                "start-with-gitmoji": ({ header }) => {
+                    return [
+                        [gimojis, extras].some((collection) =>
+                            collection.some((symbol) =>
+                                header.startsWith(symbol),
+                            ),
+                        ),
+                        "Your subject should start with an appropriate Unicode emoji from https://gitmoji.dev/",
```
zeenix: We still point to gitmoji and variables/names also need to be updated.

zeenix:
> ### Merging #1610 will **improve performances by 55.11%**

I wish I could specify the threshold for each benchmark separately, so we don't get incorrect assessments like these. :cry:

.commitlintrc.mjs
jokeyrhyme: now points at `gimoji` GitHub repository

.commitlintrc.mjs
jokeyrhyme: dropped this exemption

jokeyrhyme:
okay, resolved all feedback and squashed to a single commit

.commitlintrc.mjs
zeenix: Cool, I think it should point to https://zeenix.github.io/gimoji/ though so users can't directly pick from there.

jokeyrhyme: okay, fixed: https://github.com/z-galaxy/zbus/pull/1610/commits/14e72f5efcc5398fea66cf9640760f2a1606cdce

jokeyrhyme:
once we're happy with how this works out in the wild, another option we have could be to extract this configuration into a separate npmjs.com package (published from a new z-galaxy project?) and then wire it up in all the other z-galaxy repositories

unless it's not likely to change much over time, in which case copy-pasting it into the other projects isn't a big deal, either

zeenix (changes requested):
You forgot to update the PR description. :) Also please always add details to commit logs too so everything is locally available in git repos. 🙏

zeenix:
> once we're happy with how this works out in the wild, another option we have could be to extract this configuration into a separate npmjs.com package (published from a new z-galaxy project?) and then wire it up in all the other z-galaxy repositories

Yeah, sounds great. Although I've to admit that I'm not really a JS guy but I'm happy to create the repo and give you write-access if you could maintain that?

jokeyrhyme:
> create the repo and give you write-access if you could maintain that?

i wonder if this probably needs to be a "github.com/z-galaxy/commitlint-plugin-gimoji" (there's a gitmoji one already, but no gimoji one): https://commitlint.js.org/reference/plugins.html

and then the actual shareable config in "github.com/z-galaxy/commitlint-config": https://commitlint.js.org/concepts/shareable-config.html

we can have a "z-galaxy" scope at npmjs.com

i'll use these myself in all of my projects and hopefully some projects at work :)

jokeyrhyme:
silly me, i just remembered that a single `git` repository could publish both npmjs.com packages, so this doesn't need to be multiple `git` repositories for any particular reason

zeenix:
> silly me, i just remembered that a single `git` repository could publish both npmjs.com packages, so this doesn't need to be multiple `git` repositories for any particular reason

Ah ok. So we can keep this all in gimji repo?

jokeyrhyme:
yeah, that makes the most sense

I'll raise some PRs there to publish the shared config and the rule plugin, and then modify this PR to adopt that

zeenix:
> I'll raise some PRs there to publish the shared config and the rule plugin, and then modify this PR to adopt that

I'd rather we get this merged first and then do the splitting. I keep getting new PRs that make the same mistake about the emoji (if they included it). :)

jokeyrhyme:
okay, hopefully i can get some time tonight to fix:

- [x] PR description: https://github.com/z-galaxy/zbus/pull/1610#pullrequestreview-3575687858
- [x] git commit message body

.commitlintrc.mjs
```diff
@@ -0,0 +1,41 @@
...
+    throw new Error("unable to fetch gimoji emojis.json", { cause });
+}
+
+export default {
+    defaultIgnores: true,
+    helpUrl: "https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md",
+    ignores: [
+        (commit) =>
+            commit.includes(
+                "This PR was generated by [Mend Renovate](https://mend.io/renovate/).",
```
zeenix, line 21: This reminds me, we need to also skip merge commits. Or, we could make the checks only run on PRs and not every push. That probably makes sense anyway since there is not much point in checking the commits/pushes to main branch (or any branch really).

jokeyrhyme: that's what the default ignores setting does: https://commitlint.js.org/reference/configuration.html#configuration-object-example

https://github.com/conventional-changelog/commitlint/blob/dddcc2834a24727ccb4cc63cdc4e7b7c06322e9b/%40commitlint/is-ignored/src/defaults.ts#L17

PR #1616, opened 2025-12-15
🐛 Make out_args attribute work w/ single-output methods in introspection

### LINKED ISSUE #1599: Apply out_args interface attribute also for single output methods
It seems that annotating a method with a single output does not add the `name="SomeOutput"` field in the introspection XML. 

`#[zbus(out_args("SomeOutput"))]`

I wonder if that could be enabled. I am aware that the name of a function with a single output probably indicates what the output is, but sometimes it does not.

```xml
<node>
  <interface name="com.example.server">
    <method name="SomeFunction">
      <arg name="SomeInput" type="s" direction="in"/>
      <arg type="s" direction="out"/>
    </method>
</node>
```

As soon as a second output parameter is added it works.

```xml
<node>
  <interface name="com.example.server">
    <method name="SomeFunction">
      <arg name="SomeInput" type="s" direction="in"/>
      <arg name="SomeOutput1" type="s" direction="out"/>
      <arg name="SomeOutput2" type="s" direction="out"/>
    </method>
</node>
```

cachebag:
closes #1599 

This PR enables the `out_args` attribute to work with single-output methods in introspection XML generation.

Originally, the attribute was only being applied when generating introspection XML for methods returning tuples with multiple values. For single-output methods, the attribute was silently ignored.

I modified `introspect_add_output_args()` in the macros to check for single-element `arg_names` and apply it to the XML `name` attribute for non-tuple returns.

Also added a test; `test_single_ret_with_name()` that verifies the introspection XML correctly includes `name="SomeOutput"` for single-output methods.

Edit:
Also replaced `panic!()` with proper `Error::new_spanned()` for mismatched `out_args` count in tuple returns (same test)

### DIFF
--- a/zbus/tests/iface_and_proxy/client.rs
+++ b/zbus/tests/iface_and_proxy/client.rs
@@ -202,7 +202,10 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
         .unwrap();
     let methods = iface.methods();
     for method in methods {
-        if method.name() != "TestSingleStructRet" && method.name() != "TestMultiRet" {
+        if method.name() != "TestSingleStructRet"
+            && method.name() != "TestMultiRet"
+            && method.name() != "TestSingleRetWithName"
+        {
             continue;
         }
         let args = method.args();
@@ -214,17 +217,25 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
             assert_eq!(args.len(), 1);
             assert_eq!(out_args.next().unwrap().ty(), "(is)");
             assert!(out_args.next().is_none());
-        } else {
+        } else if method.name() == "TestMultiRet" {
             assert_eq!(args.len(), 2);
             let foo = out_args.find(|a| a.name() == Some("foo")).unwrap();
             assert_eq!(foo.ty(), "i");
             let bar = out_args.find(|a| a.name() == Some("bar")).unwrap();
             assert_eq!(bar.ty(), "s");
+        } else if method.name() == "TestSingleRetWithName" {
+            // Test that single output with out_args attribute gets the name
+            assert_eq!(args.len(), 1);
+            let out_arg = out_args.next().unwrap();
+            assert_eq!(out_arg.name(), Some("SomeOutput"));
+            assert_eq!(out_arg.ty(), "s");
+            assert!(out_args.next().is_none());
         }
     }
     // build-time check to see if macro is doing the right thing.
     let _ = proxy.test_single_struct_ret().await?.foo;
     let _ = proxy.test_multi_ret().await?.1;
+    let _ = proxy.test_single_ret_with_name().await?;
 
     let val = proxy.ping().await?;
 
--- a/zbus/tests/iface_and_proxy/iface.rs
+++ b/zbus/tests/iface_and_proxy/iface.rs
@@ -114,6 +114,13 @@ impl MyIface {
     }
 
     #[instrument]
+    #[zbus(out_args("SomeOutput"))]
+    fn test_single_ret_with_name(&self) -> zbus::fdo::Result<String> {
+        debug!("`TestSingleRetWithName` called.");
+        Ok(String::from("test output"))
+    }
+
+    #[instrument]
     fn test_response_notify(
         &self,
         #[zbus(connection)] conn: &Connection,


zeenix:
Thanks! Some nits before I look at the actual code changes. :)

* The commit logs' summaries and title of the PR doesn't make it clear that this is about introspection.
* "fix" is way too vague.
* The PR itself is missing a description.

zbus_macros/src/iface.rs
```diff
@@ -1302,7 +1302,13 @@ fn introspect_add_output_args(
                 args.extend(introspect_output_arg(&t.elems[i], name, cfg_attrs));
             }
         } else {
-            args.extend(introspect_output_arg(ty, None, cfg_attrs));
+            // Handle single output case - check if out_args was specified
+            let name = arg_names.and_then(|names| match names.len() {
+                1 => Some(&names[0]),
+                0 => None,
+                _ => panic!("Number of out arg names different from out args specified"),
```
zeenix: This should be an error (like on line 1285 above for example).

zbus_macros/src/iface.rs
```diff
@@ -1293,16 +1293,43 @@ fn introspect_add_output_args(
...
-                    // Turn into error
-                    panic!("Number of out arg names different from out args specified")
+                    return Err(Error::new_spanned(
+                        ty,
+                        format!(
+                            "Number of out arg names ({}) different from out args specified ({})",
+                            arg_names.len(),
+                            t.elems.len()
+                        ),
+                    ));
```
zeenix, line 1303: Kudos for fixing the existing code too but no good deed goes unpunished :-) so you gotta split this. ;)

zbus_macros/src/iface.rs
```diff
@@ -1293,16 +1293,43 @@ fn introspect_add_output_args(
...
+                    }
+                    0 => {
+                        args.extend(introspect_output_arg(ty, None, cfg_attrs));
+                    }
+                    _ => {
+                        return Err(Error::new_spanned(
+                            ty,
+                            format!(
+                                "Number of out arg names ({}) different from out args specified (1)",
+                                names.len()
```
zeenix, line 1321: Are we sure the error corresponds to what the problem is? 🤔

zbus_macros/src/iface.rs
```diff
@@ -1293,16 +1293,43 @@ fn introspect_add_output_args(
...
+                    _ => {
+                        return Err(Error::new_spanned(
+                            ty,
+                            format!(
+                                "out_args specifies {} names but method has a single output",
+                                names.len()
+                            ),
+                        ));
+                    }
+                }
```
zeenix: How about this simplication to improve readability?

```suggestion
                let names = match names.len() 
                    1 => Some(&names[0]),
                    0 => None,
                    _ => {
                        return Err(Error::new_spanned(
                            ty,
                            format!(
                                "out_args specifies {} names but method has a single output",
                                names.len()
                            ),
                        ));
                    }
                };
                args.extend(introspect_output_arg(ty, names, cfg_attrs));
```

cachebag:
> 99% there. :)

Thanks for the push up the hill. I promise one of these days I'll submit a non-problematic PR 😅

zeenix:
> I promise one of these days I'll submit a non-problematic PR 😅

None of your PRs have been problematic and we all make mistakes. I'm sure you'd find a lot of stuff to raise in my PRs too. :)

zbus_macros/src/iface.rs
```diff
@@ -1293,16 +1293,40 @@ fn introspect_add_output_args(
...
+            } else {
+                None
+            };
+            args.extend(introspect_output_arg(ty, names, cfg_attrs));
```
zeenix: hmm.. you move this outside the the if block, which changes the logic. I guess the new logic is correct/better? 🤔

zbus_macros/src/iface.rs
```diff
@@ -1293,16 +1293,40 @@ fn introspect_add_output_args(
...
                 }
             }
             for i in 0..t.elems.len() {
                 let name = arg_names.map(|names| &names[i]);
                 args.extend(introspect_output_arg(&t.elems[i], name, cfg_attrs));
             }
         } else {
-            args.extend(introspect_output_arg(ty, None, cfg_attrs));
+            // Handle single output case - check if out_args was specified
+            let names = if let Some(names) = arg_names {
```
zeenix: since now you're calling the  `args.extend` in both case, you can further improve the readability here by make use of `Option::map` so you'll just have one match statement then and less indentation.

zeenix: I see 2 `match` statements still. You're not making use of `Option::map` method to have only 1 level.

cachebag: Sorry about that- maybe I misunderstood; `Option::map` transforms the value inside an Option, so `arg_names.map(|names| names.first())` would return `Option<Option<&String>>`..Isn't that a nested Option? because `first()` itself returns an Option

zeenix: I meant that you can do a `arg_names.map(Vec::len)` but I didn't think that you need the unwrapped `names` in one of the inner match branches, so nm.

PR #1617, opened 2025-12-15
⚡️ zb: Sort interfaces by name in managed objects

mulkieran:
For GetManagedObjects result, it would be more convenient if the interfaces on each object path represented are sorted by interface name instead of sent in random order.

This is not achieved by this change. While the interfaces are sorted in the ManagedObjects structure, the marshalling procedure, which is defined by the D-Bus spec, removes the order. The interfaces sent on the D-Bus are in random order.

This change likely results in slightly more efficent processing when building the temporary interfaces BTreeMap to create the ObjectManager struct.

The interfaces, being ordered in Node's interfaces attribute, are inserted in order into the temporary interfaces BTreeMap, so there is no cost involved in shuffling the keys for sorting during insertion.

The space allocation cost is probably a toss up.

### DIFF
--- a/zbus/src/fdo/object_manager.rs
+++ b/zbus/src/fdo/object_manager.rs
@@ -3,7 +3,10 @@
 //! The D-Bus specification defines the message bus messages and some standard interfaces that may
 //! be useful across various D-Bus applications. This module provides their proxy.
 
-use std::{borrow::Cow, collections::HashMap};
+use std::{
+    borrow::Cow,
+    collections::{BTreeMap, HashMap},
+};
 use zbus_names::{InterfaceName, OwnedInterfaceName};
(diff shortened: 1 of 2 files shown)


zeenix (approved):
Thanks. We should merge it when the `main` branch is open of API breaks.

zeenix:
> Thanks. We should merge it when the `main` branch is open of API breaks.

@mulkieran that's now the case. Are you available for rebasing this so we can merge it?

mulkieran:
> > Thanks. We should merge it when the `main` branch is open of API breaks.
> 
> @mulkieran that's now the case. Are you available for rebasing this so we can merge it?

@zeenix Yup, done.

PR #1619, opened 2025-12-17
🥅 zm: Handle non-String fields in error variant matching

Hendrik-Hamerlinck:
This PR updates the `DBusError` derive macro to support enum variants where the first field is not a String.
Previously, such cases resulted in an error when using custom errors. Now, the macro attempts to parse the first field as a String. If parsing fails, it returns None instead.

## Example

```rust
#[derive(Debug, DBusError)]
pub enum ExampleDBusError {
   UnitVariant,
   UnnamedVariant(u32, String),
   UnnamedVariantString(String),
   NamedVariant { code: u32, message: String },
   NamedVariantString { message: String },
}
```

This now compiles and expands the `description` as following
```rust
fn description(&self) -> Option<&str> {
    match self {
        Self::UnitVariant => None,
        Self::UnnamedVariant(..) => None,
        Self::UnnamedVariantString(desc, ..) => Some(desc),
        Self::NamedVariant { .. } => None,
        Self::NamedVariantString { message } => Some(message),
    }
}
```

### DIFF
--- a/zbus_macros/src/error.rs
+++ b/zbus_macros/src/error.rs
@@ -1,6 +1,6 @@
 use proc_macro2::TokenStream;
-use quote::{ToTokens, quote};
-use syn::{Data, DeriveInput, Error, Fields, Ident, Variant, spanned::Spanned};
+use quote::{quote, ToTokens};
+use syn::{Data, DeriveInput, Error, Fields, Ident, Type, Variant, spanned::Spanned};
 use zvariant_utils::def_attrs;
 
 def_attrs! {
@@ -79,31 +79,41 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
             zbus_error_variant = Some(quote! { #ident });
         }
 
-        // FIXME: this will error if the first field is not a string as per the dbus spec, but we
-        // may support other cases?
         let e = match &variant.fields {
             Fields::Unit => quote! {
                 Self::#ident => None,
             },
-            Fields::Unnamed(_) => {
+            Fields::Unnamed(fields) => {
                 if error {
                     quote! {
                         Self::#ident(e) => e.description(),
                     }
+                } else if let Some(first_field) = fields.unnamed.first()
+                    && let Type::Path(type_path) = &first_field.ty
+                    && type_path.path.is_ident("String")
+                {
+                    quote! {
+                        Self::#ident(desc, ..) => Some(desc),
+                    }
                 } else {
                     quote! {
-                        Self::#ident(desc, ..) => Some(&desc),
+                        Self::#ident(..) => None,
                     }
                 }
             }
             Fields::Named(n) => {
-                let f = &n
-                    .named
-                    .first()
-                    .ok_or_else(|| Error::new(n.span(), "expected at least one field"))?
-                    .ident;
-                quote! {
-                    Self::#ident { #f, } => Some(#f),
+                if let Some(first_field) = n.named.first()
+                    && let Type::Path(type_path) = &first_field.ty
+                    && type_path.path.is_ident("String")
+                {
+                    let f = &first_field.ident;
+                    quote! {
+                        Self::#ident { #f, } => Some(#f),
+                    }
+                } else {
+                    quote! {
+                        Self::#ident { .. } => None,
+                    }
                 }
             }
         };
--- a/zbus_macros/tests/tests.rs
+++ b/zbus_macros/tests/tests.rs
@@ -122,6 +122,7 @@ fn test_derive_error() {
         LetItBe {
             desc: String,
         },
+        Timeout(String, u64),
     }
 }
 


zeenix:
@Hendrik-Hamerlinck Thanks so much for this contribution. :+1: I'll have a look at it soon but let me start with a nitpick about commit emoji. :laughing: This is not a bug fix since I don't think we have advertised anywhere that this already works, nor it has every worked AFAIK.

Also I have to ask, how come you ended up using the emoji code rather than the emoji character? You are far from the first one to make this mistake so this is not on you but more like me failing in communicating this in the [contribution guide](https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). I'm only asking so I can try to make this clearer.

>  You can also pick an emoji direcitly from here (please remember to copy the emoji itself and not the :emoji-code: string, by just clicking on it). NOTE: This is a curated list of emojis that have specific meanings. Please use one of the methods recommended here to select/fetch the most appropriate one. 🙏

zeenix:
I took the liberty of adding the most appropriate emoji to the PR title (please copy&paste that on the commit title).

zbus_macros/src/error.rs
```diff
@@ -79,33 +79,44 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
...
         }
 
-        // FIXME: this will error if the first field is not a string as per the dbus spec, but we
-        // may support other cases?
         let e = match &variant.fields {
             Fields::Unit => quote! {
                 Self::#ident => None,
             },
-            Fields::Unnamed(_) => {
+            Fields::Unnamed(fields) => (|| {
```
zeenix: Why switch to anonymous functions, just to call them immediately? This makes this change more intrusive than it would otherwise need to be and it's best to keep changes as small as possible.

Hendrik-Hamerlinck:
> @Hendrik-Hamerlinck Thanks so much for this contribution. 👍 I'll have a look at it soon but let me start with a nitpick about commit emoji. 😆 This is not a bug fix since I don't think we have advertised anywhere that this already works, nor it has every worked AFAIK.
> 
> Also I have to ask, how come you ended up using the emoji code rather than the emoji character? You are far from the first one to make this mistake so this is not on you but more like me failing in communicating this in the [contribution guide](https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md#submitting-pull-requests). I'm only asking so I can try to make this clearer.
> 
> > You can also pick an emoji direcitly from here (please remember to copy the emoji itself and not the :emoji-code: string, by just clicking on it). NOTE: This is a curated list of emojis that have specific meanings. Please use one of the methods recommended here to select/fetch the most appropriate one. 🙏

Thanks for the quick review. 

Missing on the emoji is on me, I just noticed I didn't read the entire emoji part through. After clicking on the first link I assumed you just needed to copy the code. I will edit it as requested.

zbus_macros/src/error.rs
Hendrik-Hamerlinck: Yeah I was not entirely happy about it. In my own version where we use rust 2024 I did something like this: 
```rust
Fields::Named(n) => {
    if let Some(first_field) = n.named.first()
        && let Type::Path(type_path) = &first_field.ty
        && type_path.path.is_ident("String")
    {
        let f = &first_field.ident;
        quote! {
            Self::#ident { #f, } => Some(#f),
        }
    } else {
        quote! {
            Self::#ident { .. } => None,
        }
    }
}
```

But the project is on 2021 and you cannot chain let pattern matching. So to prevent having to return the same value in each else branch I switched to an inline function, so that I could return directly. Perhaps there are better ways to do it?

zeenix: To avoid nesting, I tend to use multiple if let .. else statements in a row combined with early return/break (in the else blocks). I think you can use that here.

---

2024 edition was stabilized in 1.85, which is almost a year old now so I'd be happy to bump the MSRV and switch to that edition. However, that should happen in a separate PR, if you'd like to contribute that as well.

Hendrik-Hamerlinck: Great, that would simplify this change. While updating the tests I noticed that my change only works under the 2024 edition. I'll send a PR for bumping the MSRV.

zeenix:
@Hendrik-Hamerlinck Hi, could you please rebase this branch?

Hendrik-Hamerlinck:
> @Hendrik-Hamerlinck Hi, could you please rebase this branch?

Hey, sorry for the delay. I rebased it and updated it, but I noticed there is an issue with the logic when I was updating the tests. As it cannot be deserialized if it is not always a String as a field.

Not to sure what the best approach is, we could deserialize and serialize using json, when the first field is not a string? Or would you rather drop this change? What do you think?

zeenix:
> > @Hendrik-Hamerlinck Hi, could you please rebase this branch?
> 
> Hey, sorry for the delay. I rebased it and updated it, but I noticed there is an issue with the logic when I was updating the tests. As it cannot be deserialized if it is not always a String as a field.

I'm a bit confused. What deserialization? I don't see any. If you could fix the failures in the CI, I could perhaps see the actual failure in the tests and know what you mean exactly.

zeenix:
> I'm a bit confused. What deserialization? I don't see any. If you could fix the failures in the CI, I could perhaps see the actual failure in the tests and know what you mean exactly.

Hello?

zeenix:
@Hendrik-Hamerlinck Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

PR #1626, opened 2025-12-21
♻️ zbn: Consolidate zbus name type impls with `define_name_type_impls!` macro

cachebag:
Introduce a declarative macro that generates all boilerplate code for D-Bus name types and their owned variants.

The macro generates all trait impls (`TryFrom`, `From`, `Deref`, `Borrow`, `Display`, `Debug`, `PartialEq`, `Deserialize`, `NoneValue`, `Basic`) and methods (`as_str`, `as_ref`, `from_str_unchecked`, `from_static_str`, `to_owned`, `into_owned`, `into_inner`, `inner`).

Figured this might be easier on the eyes for the future.


zbus_names/src/utils.rs
```diff
@@ -7,29 +7,363 @@ macro_rules! impl_str_basic {
...
-            impl<'s> TryFrom<$from> for $type {
-                type Error = Error;
+/// Generates all boilerplate code for a D-Bus name type and its owned variant.
+///
+/// This macro generates all the boilerplate code for a D-Bus name type and its owned variant.
+///
+/// # Parameters
+/// - `$name`: The name of the borrowed type (e.g., `InterfaceName`)
+/// - `$owned_name`: The name of the owned type (e.g., `OwnedInterfaceName`)
+/// - `$validate_fn`: The validation function to use
```
zeenix: * Huge nitpick but please add `.` at the end of each sentence.
* missing `inner_vis`.

zbus_names/src/utils.rs
```diff
@@ -7,29 +7,363 @@ macro_rules! impl_str_basic {
...
+macro_rules! define_name_type_impls {
+    (
+        name: $name:ident,
+        owned: $owned_name:ident,
+        validate: $validate_fn:ident,
+        inner_vis: $($inner_vis:tt)*
```
zeenix: where is `inner_vis` being used? 🤔

zbus_names/src/member_name.rs
```diff
@@ -40,117 +31,15 @@ use zvariant::{NoneValue, OwnedValue, Str, Type, Value};
...
-impl_try_from! {
-    ty: MemberName<'s>,
-    owned_ty: OwnedMemberName,
-    validate_fn: validate,
-    try_from: [&'s str, String, Arc<str>, Cow<'s, str>, Str<'s>],
+define_name_type_impls! {
+    name: MemberName,
+    owned: OwnedMemberName,
+    validate: validate,
+    inner_vis:
```
zeenix: empty value looks strange. Can it not be just skipped in such a case?

zbus_names/src/utils.rs
cachebag: When writing the macor, I was initially thinking it might need it to control field visibility, but since the struct definitions are outside of it, I see _now_ that it was never actually used. Removed it.

zeenix:
Somehow this change ended up breaking things for folks:

```rust
use zbus_names::{OwnedUniqueName, UniqueName};
fn main() {
    let x : OwnedUniqueName = todo!();
    let foo: UniqueName = x.as_ref();
}
```

This code worked with zbus_names 4.2.0  but doesn't with zbus_names 4.3.0 (released today with this change). I looked into it but I am not sure how it breaks things. :thinking:

marc2332:
> Somehow this change ended up breaking things for folks:
> 
> ```rust
> use zbus_names::{OwnedUniqueName, UniqueName};
> fn main() {
>     let x : OwnedUniqueName = todo!();
>     let foo: UniqueName = x.as_ref();
> }
> ```
> 
> This code worked with zbus_names 4.2.0 but doesn't with zbus_names 4.3.0 (released today with this change). I looked into it but I am not sure how it breaks things. 🤔

Doesn't [this](https://github.com/z-galaxy/zbus/pull/1626/changes#diff-f33b30a1d719177b9caceff4f7356a0e7dee674bf41e99703b9e721ed94d5312R46) explain one of the errors I get?

```
error[E0433]: failed to resolve: could not find `zvariant` in the list of imported crates
  --> /home/marc/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/ashpd-0.12.0/src/file_path.rs:11:10
   |
11 | #[derive(Type, Debug, Default, PartialEq)]
   |          ^^^^ could not find `zvariant` in the list of imported crates
   |
   = note: this error originates in the derive macro `Type` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing this enum
   |
 1 + use crate::zvariant::Signature;
 ```
 
 It should be something like `$crate::zvariant` I think? 
 
 Edit: Maybe its completely unrelated.

zeenix:
> Edit: Maybe its completely unrelated.

Yeah, I think it's unrelated.

cachebag:
Looking at this hash b925961c98bc03183af8334063b2296f42240803...

The old code did not implement AsRef for owned types:
https://github.com/z-galaxy/zbus/blob/f2dbe6e82e3a2e495aa9a6cba48af5e18b55ab13/zbus_names/src/unique_name.rs#L216-L226

But now we do...
https://github.com/z-galaxy/zbus/blob/a4d6ec541e83dddc529123a480fb8af2dc1cfa3e/zbus_names/src/utils.rs#L288-L293

Could this be why? @zeenix I found out that[ Rust's method resolution prefers trait methods over `Deref` coercion](https://doc.rust-lang.org/reference/expressions/method-call-expr.html). Idk I may be wrong but for instance, `let foo: UniqueName = x.as_ref()` fails here because `&str` cannot be assigned to `UniqueName`.

If this _is_ the issue, I have to say the macro looked complete to me, adding `AsRef<str>` seems like a reasonable convenience impl. I thought it was a common pattern for string wrapper types. I am super sorry for this blunder

-- 

> Rust's method resolution prefers trait methods over `Deref` coercion

Ref:
image

marc2332:
> Somehow this change ended up breaking things for folks:
> 
> ```rust
> use zbus_names::{OwnedUniqueName, UniqueName};
> fn main() {
>     let x : OwnedUniqueName = todo!();
>     let foo: UniqueName = x.as_ref();
> }
> ```
> 
> This code worked with zbus_names 4.2.0 but doesn't with zbus_names 4.3.0 (released today with this change). I looked into it but I am not sure how it breaks things. 🤔

With this PR https://github.com/cachebag/zbus/blob/b925961c98bc03183af8334063b2296f42240803/zbus_names/src/utils.rs#L288-L292 was added.

marc2332:
Oh, no way @cachebag ! We replied at the same time, I swear I just saw it too

zeenix:
> I am super sorry for this blunder

Don't sweat it! We have all been there. :) Besides, as the reviewer, I'm also responsible. :)

PR #1632, opened 2025-12-23
⚡️ zbn: Add structured error types and early-exit validation for D-Bus names

cachebag:
Currently, when a D-Bus name fails validation, you get a generic `Error::InvalidName(&'static str)` with a message like "Invalid interface name". I never found that to be super helpful when trying to figure out *why* it failed. Was it too long? Did an element start with a digit? Was there an invalid character somewhere?

This adds two new enums: `NameType` (which kind of name failed like `WellKnown`, `Unique`, `Interface`, `Member`, `Property`, `Error`) and `InvalidNameReason` (WHY it failed, `Empty`, `TooLong`, `MissingColonPrefix`, `StartsWithDot`, `MissingDotSeparator`, `ElementStartsWithDigit`, `EmptyElement`, `InvalidCharacter`, `ParseError`). 

These are combined in a new `Error::InvalidNameDetail { name_type, reason }` variant.

The `InvalidNameReason::InvalidCharacter` and `ElementStartsWithDigit` variants include position info, so you can pinpoint exactly where validation failed. `TooLong` includes both `actual` and `max` lengths. I marked them both `#[non_exhaustive]` so we can extend them later without breaking downstream.

I also added `validate_detailed()` functions that do cheap early-exit checks before invoking the full winnow parser. So a length check (<=255 bytes), empty check, and first-character validation. 

For `BusName`, there's also now a first-byte dispatch: if it starts with `:`, go straight to unique name validation; otherwise try well-known. This avoids the previous behavior of attempting one parse, failing, then trying the other.

### DIFF
--- a/zbus_names/src/interface_name.rs
+++ b/zbus_names/src/interface_name.rs
@@ -1,6 +1,9 @@
-use crate::{Error, Result, utils::define_name_type_impls};
+use crate::{Error, InvalidNameReason, NameType, Result, utils::define_name_type_impls};
 use serde::Serialize;
 use zvariant::{OwnedValue, Str, Type, Value};
+
+/// The maximum length of a D-Bus name in bytes.
+const MAX_NAME_LENGTH: usize = 255;
 
 /// String that identifies an [interface name][in] on the bus.
 ///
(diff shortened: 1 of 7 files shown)


zbus_names/src/interface_name.rs
```diff
@@ -44,46 +47,270 @@ define_name_type_impls! {
...
+/// * Only ASCII alphanumeric and `_`
+/// * Must not begin with a `.`
+/// * Must contain at least one `.`
+/// * Each element must:
+///   * not begin with a digit
+///   * be at least 1 character (so name must be minimum 3 characters long)
+/// * <= 255 characters
+///
+/// Note: A `-` is not allowed, which is why we can't use the same parser as for `WellKnownName`.
+pub(crate) fn validate_detailed(bytes: &[u8], name_type: NameType) -> Result<()> {
```
cachebag, line 65: I am kind of iffy on this. The complexity increases here but I would still want to use `InvalidNameDetail` even if you wanted to revert to `winnow` for the actual parsing; just because the structured errors I think is worthwhile addition. Again, I defer to you

zbus_names/src/error.rs
```diff
@@ -58,8 +161,18 @@ pub enum Error {
                 Use `Error::InvalidName` instead."
     )]
     InvalidErrorName(String),
-    /// An invalid name.
+    /// An invalid name (legacy variant with static message).
     InvalidName(&'static str),
```
cachebag, line 165: We should deprecate this but obviously it requires a minor bump. I defer to you on that

zbus_names/src/interface_name.rs
zeenix: Yeah, I would very much like to stick to `winnow` usage and keep things as simple as possible. That doesn't necessarily mean you can't accomplish your goals here still though. 

May I also ask what is the exact use case here? Typically the names are copy&pasted and often statis part of the code. Some (maybe even most) D-Bus libraries don't even check these strings for conformance, let alone do a thorough checks and give very specific errors AFAIK.

zeenix: Having said that, the performance improvements are nice so if you could make this work with winnow and not complicate the code too much, I'd be very happy to accept this contribution (if the peformance gains are still there, that'd be a nice cherry on top but as long as the performance doesn't degrade, it'd be still acceptable).

cachebag: Admittedly, looking at this a few hours later it is largely just a micro-optimization I got tunnel vision on because I fat fingered an interface name on my own project. 

So actually, if you don't see it useful I would be comfortable closing, especially considering the error messages will probably touch only 3 people on earth that would need it 😂

zeenix: That's fair enough. :) I wouldn't be too sure of the error only hitting 3 people on the planet actually but given that these names have very simple rules that are documented and we link to them in the docs, I think there are easy (at least not so difficult) ways to figure this out.

PR #1635, opened 2025-12-24
🐛 zb: Correct Peer interface to work on any arbitrary object path

### LINKED ISSUE #1400: Service interfaces do not reply to pings in arbitrary paths
From the [DBus spec](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer):

>  On receipt of the METHOD_CALL message org.freedesktop.DBus.Peer.Ping, an application should do nothing other than reply with a METHOD_RETURN as usual. It does not matter which object path a ping is sent to. The reference implementation handles this method automatically. 

At the moment pinging in a path not added via `ObjectServer::at` results in 

```
$ busctl --user call $MY_SERVER /org/freedesktop/SomeRandomPath org.freedesktop.DBus.Peer Ping
Call failed: Unknown object '/org/freedesktop/SomeRandomPath'
```

cachebag:
Peer interface methods (`Ping`, `GetMachineId`) failed with "Unknown object" on unregistered paths, violating the D-Bus spec.

This PR handles `org.freedesktop.DBus.Peer` in `ObjectServer::dispatch_method_call_try` before the path lookup, responding directly without requiring a registered node.

closes #1400

merry christmas to you and your loved ones! @zeenix 🎅🏻 🎄

### DIFF
--- a/zbus/src/fdo/mod.rs
+++ b/zbus/src/fdo/mod.rs
@@ -43,6 +43,8 @@ mod tests {
     use tokio::runtime;
     use zbus_names::WellKnownName;
 
+    use super::PeerProxy;
+
     #[test]
     fn error_from_zerror() {
         let m = Message::method_call("/", "foo")
@@ -197,5 +199,106 @@ mod tests {
             .into_iter()
             .collect()
         );
+    }
+
+    #[test]
+    #[timeout(15000)]
+    fn peer_on_arbitrary_path() {
+        crate::block_on(peer_on_arbitrary_path_async());
+    }
+
+    /// Test that org.freedesktop.DBus.Peer works on arbitrary paths.
+    ///
+    /// According to the D-Bus specification:
+    /// "On receipt of the METHOD_CALL message org.freedesktop.DBus.Peer.Ping, an application
+    /// should do nothing other than reply with a METHOD_RETURN as usual. It does not matter
+    /// which object path a ping is sent to."
+    ///
+    /// This test verifies that Ping and GetMachineId work on paths that haven't been
+    /// registered via ObjectServer::at.
+    async fn peer_on_arbitrary_path_async() {
+        // Create a service with only a registered path at /registered
+        struct TestObj;
+        #[interface(name = "org.zbus.TestObj")]
+        impl TestObj {}
+
+        let _service_conn = zbus::conn::Builder::session()
+            .unwrap()
+            .name("org.zbus.PeerArbitraryPathTest")
+            .unwrap()
+            .serve_at("/registered", TestObj)
+            .unwrap()
+            .build()
+            .await
+            .unwrap();
+
+        // Create a client connection
+        let client_conn = zbus::Connection::session().await.unwrap();
+
+        // Ping on an unregistered path should succeed
+        let proxy = PeerProxy::new(
+            &client_conn,
+            "org.zbus.PeerArbitraryPathTest",
+            "/this/path/does/not/exist",
+        )
+        .await
+        .expect("Failed to create PeerProxy");
+        let result = proxy.ping().await;
+        assert!(result.is_ok(), "Ping on unregistered path should succeed");
+
+        // GetMachineId on an unregistered path should succeed
+        let proxy = PeerProxy::new(
+            &client_conn,
+            "org.zbus.PeerArbitraryPathTest",
+            "/another/unregistered/path",
+        )
+        .await
+        .expect("Failed to create PeerProxy");
+        let result = proxy.get_machine_id().await;
+        // GetMachineId may fail on some platforms (like *BSD), but should not fail
+        // with "Unknown object" error
+        match &result {
+            Ok(id) => {
+                assert!(!id.is_empty(), "Machine ID should not be empty");
+            }
+            #[cfg(not(any(
+                target_os = "linux",
+                target_os = "macos",
+                target_os = "freebsd",
+                target_os = "dragonfly",
+                target_os = "openbsd",
+                target_os = "netbsd",
+                windows,
+            )))]
+            Err(fdo::Error::NotSupported(_)) => {
+                // NotSupported is acceptable on platforms where it's not implemented
+            }
+            Err(e) => {
+                panic!("GetMachineId failed unexpectedly: {:?}", e);
+            }
+        }
+
+        // Ping on registered path should still work
+        let proxy = PeerProxy::new(
+            &client_conn,
+            "org.zbus.PeerArbitraryPathTest",
+            "/registered",
+        )
+        .await
+        .expect("Failed to create PeerProxy");
+        let result = proxy.ping().await;
+        assert!(result.is_ok(), "Ping on registered path should succeed");
+
+        // Unknown method on Peer interface should fail properly
+        let result = client_conn
+            .call_method(
+                Some("org.zbus.PeerArbitraryPathTest"),
+                "/unregistered",
+                Some("org.freedesktop.DBus.Peer"),
+                "UnknownMethod",
+                &(),
+            )
+            .await;
+        assert!(result.is_err(), "Unknown method should fail");
     }
 }
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -369,8 +369,18 @@ impl ObjectServer {
         // way, the object server can be mutated during that time.
         let (iface, with_spawn) = {
             let root = self.root.read().await;
+
+            // D-Bus spec: org.freedesktop.DBus.Peer interface works on ANY path, even unregistered
+            // ones. See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+            // Switch the path to "/" for Peer interface calls.
+            let path = if *iface_name == fdo::Peer::name() {
+                ObjectPath::from_static_str_unchecked("/")
+            } else {
+                path.clone()
+            };
+
             let node = root
-                .get_child(path)
+                .get_child(&path)
                 .ok_or_else(|| fdo::Error::UnknownObject(format!("Unknown object '{path}'")))?;
 
             let iface = node.interface_lock(iface_name.as_ref()).ok_or_else(|| {


cachebag:
Also, so long as the PR looks okay, #1636 might need to be merged first and this PR will need to be rebased so I can adjust this for BSD as well

zeenix:
> merry christmas to you and your loved ones! @zeenix 🎅🏻 🎄

Thank you. I hope you are having relaxing time with your loved ones too. :+1:

zbus/src/object_server/mod.rs
```diff
@@ -361,10 +388,18 @@ impl ObjectServer {
...
         // all the details for the call before spawning.
         // See also https://github.com/z-galaxy/zbus/issues/674 for future of Interface.
-        let _ = hdr
+        let member = hdr
             .member()
             .ok_or_else(|| fdo::Error::Failed("Missing member".into()))?;
 
+        // D-Bus spec: org.freedesktop.DBus.Peer interface works on ANY path, even unregistered
+        // ones. See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+        if iface_name.as_str() == "org.freedesktop.DBus.Peer" {
```
zeenix: A bit too hard-coded. Why not:


```suggestion
        if iface_name == Peer::name() {
```

zbus/src/object_server/mod.rs
```diff
@@ -289,6 +289,33 @@ impl ObjectServer {
...
+    /// Handle org.freedesktop.DBus.Peer interface methods on any path.
+    ///
+    /// According to the D-Bus specification, the Peer interface methods (Ping and GetMachineId)
+    /// must work on any object path, not just registered ones.
+    /// See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+    async fn handle_peer_interface(
+        &self,
+        connection: &Connection,
+        hdr: &Header<'_>,
+        member: &str,
```
zeenix: Why can't we use the more specific `Member` type here? We don't impl `PartialEq<&str> for Member`? We should still use the specific type even if that's not implemented and add a FIXME for implementing that so we don't need to do `as_str()` internally in this method then.

zbus/src/object_server/mod.rs
```diff
@@ -289,6 +289,33 @@ impl ObjectServer {
         })
     }
 
+    /// Handle org.freedesktop.DBus.Peer interface methods on any path.
+    ///
+    /// According to the D-Bus specification, the Peer interface methods (Ping and GetMachineId)
+    /// must work on any object path, not just registered ones.
+    /// See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+    async fn handle_peer_interface(
```
zeenix: Remember the "usage before declaration/definition" preference.

zeenix (changes requested):
Thanks.

zeenix:
Oh I didn't realise this wasn't merged yet.

zbus/src/fdo/mod.rs
```diff
@@ -198,4 +198,107 @@ mod tests {
...
+        let result = client_conn
+            .call_method(
+                Some("org.zbus.PeerArbitraryPathTest"),
+                "/this/path/does/not/exist",
+                Some("org.freedesktop.DBus.Peer"),
+                "Ping",
+                &(),
+            )
+            .await;
+        assert!(result.is_ok(), "Ping on unregistered path should succeed");
```
zeenix: You could use the [high-level API](https://docs.rs/zbus/latest/zbus/fdo/struct.PeerProxy.html) here. It's quite ok to create a temporary proxy, just to quickly call a method on it (in fact we do it in our code base):

```rust
let result = PeerProxy::new(&conn, "org.zbus.PeerArbitraryPathTest", "/this/path/does/not/exist")?
    .ping()
    .await;
```

zbus/src/object_server/mod.rs
```diff
@@ -361,10 +361,16 @@ impl ObjectServer {
...
         // all the details for the call before spawning.
         // See also https://github.com/z-galaxy/zbus/issues/674 for future of Interface.
-        let _ = hdr
+        let member = hdr
             .member()
             .ok_or_else(|| fdo::Error::Failed("Missing member".into()))?;
 
+        // D-Bus spec: org.freedesktop.DBus.Peer interface works on ANY path, even unregistered
+        // ones. See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+        if *iface_name == fdo::Peer::name() {
```
zeenix: hmm.. I feel a bit uneasy about special-casing this actually. This also means that us automatically adding `Peer` interface to all paths becomes useless. I think a proper fix would be to add support for interfaces that exists on all paths (even non-existent ones) but that might be way too much work for just one interface since I don't really see this being useful for anything else.

Perhaps, we can go for a middle-ground and instead of manually handling `Peer` methods, we modify the path to `/` here? 🤔 Then the request would be handled by `Peer` interface, as always.

zeenix (changes requested):
Actually still a few concerns.

zbus/src/fdo/mod.rs
cachebag: I kept the last test case (unknown method) using `call_method`. Is there no high-level API method for calling an unknown/non-existent method?

zbus/src/object_server/mod.rs
cachebag: Makes sense to me. 

Additionally, as I mentioned maybe we should merge #1636 first?

zbus/src/fdo/mod.rs
zeenix: > Is there no high-level API method for calling an unknown/non-existent method?

No but here you don't need unknown method but rather unknown object path and the proxy will happily create that for you. So yeah, the high-level API I recommended should work and makes sense IMO. :)

zbus/src/object_server/mod.rs
zeenix: > Additionally, as I mentioned maybe we should merge #1636 first?

What are you talking about, that was merged already awhile back. I totally did not forget about it to only merge it today. :stuck_out_tongue_winking_eye:

zbus/src/object_server/mod.rs
```diff
@@ -361,17 +361,25 @@ impl ObjectServer {
...
-            let node = root
-                .get_child(path)
-                .ok_or_else(|| fdo::Error::UnknownObject(format!("Unknown object '{path}'")))?;
+            let mut node = root.get_child(path);
+
+            // D-Bus spec: org.freedesktop.DBus.Peer interface works on ANY path, even unregistered
+            // ones. See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+            // If the path doesn't exist and it's a Peer interface call, fall back to "/"
+            if node.is_none() && *iface_name == fdo::Peer::name() {
+                node = root.get_child(&ObjectPath::from_static_str_unchecked("/"));
```
zeenix: I meant something even simpler: Before `root.get_child` call above, we switch the object path so the rest of the code works exactly as before. I'm not sure if `root.get_child` can handle `/` though (since that's its own path) but we can easily modify it to do that.

cachebag: Oooohh okay okay

cachebag:
Oops, messed up cleaning out my branches again.

zbus/src/object_server/mod.rs
```diff
@@ -361,16 +361,26 @@ impl ObjectServer {
         // Note that an unknown member will still spawn a task. We should instead gather
         // all the details for the call before spawning.
         // See also https://github.com/z-galaxy/zbus/issues/674 for future of Interface.
-        let _ = hdr
+        let _member = hdr
```
zeenix: unrelated change. Please split or remove. :)

zbus/src/object_server/mod.rs
```diff
@@ -361,16 +361,26 @@ impl ObjectServer {
...
             .ok_or_else(|| fdo::Error::Failed("Missing member".into()))?;
 
         // Ensure the root lock isn't held while dispatching the message. That
         // way, the object server can be mutated during that time.
         let (iface, with_spawn) = {
             let root = self.root.read().await;
+
+            // D-Bus spec: org.freedesktop.DBus.Peer interface works on ANY path, even unregistered
+            // ones. See: https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-peer
+            // Switch the path to "/" for Peer interface calls
```
zeenix: master nitpicker at work: missing `.`.

zbus/src/object_server/mod.rs
```diff
@@ -361,16 +361,26 @@ impl ObjectServer {
...
+            let lookup_path = if *iface_name == fdo::Peer::name() {
```
zeenix: just overshadow the `path` binding.

zbus/src/fdo/mod.rs
```diff
@@ -198,4 +200,98 @@ mod tests {
...
+                Some("org.zbus.PeerArbitraryPathTest"),
+                "/unregistered",
+                Some("org.freedesktop.DBus.Peer"),
+                "UnknownMethod",
+                &(),
+            )
+            .await;
+        assert!(result.is_err(), "Unknown method should fail");
+
+        drop(service_conn);
```
zeenix: explicit drop is unnecessary when we're at the end of the block.

zbus/src/fdo/mod.rs
```diff
@@ -198,4 +200,98 @@ mod tests {
...
+        .expect("Failed to create PeerProxy");
+        let result = proxy.get_machine_id().await;
+        // GetMachineId may fail on some platforms (like *BSD), but should not fail
+        // with "Unknown object" error
+        match &result {
+            Ok(id) => {
+                assert!(!id.is_empty(), "Machine ID should not be empty");
+            }
+            Err(fdo::Error::NotSupported(_)) => {
+                // NotSupported is acceptable on platforms where it's not implemented
```
zeenix, line 274: even better if we define this arm only for platforms where this we don't support it:


```suggestion
            #[cfg(not(any(target_os = "linux", target_os = "macos", ...)))]
            Err(fdo::Error::NotSupported(_)) => {
                // NotSupported is acceptable on platforms where it's not implemented
```

zbus/src/object_server/mod.rs
```diff
@@ -357,20 +357,23 @@ impl ObjectServer {
             // error, or deliver the message as though it had an arbitrary one of those
             // interfaces.
             .ok_or_else(|| fdo::Error::Failed("Missing interface".into()))?;
-        // Check that the message has a member before spawning.
-        // Note that an unknown member will still spawn a task. We should instead gather
-        // all the details for the call before spawning.
-        // See also https://github.com/z-galaxy/zbus/issues/674 for future of Interface.
-        let _ = hdr
-            .member()
-            .ok_or_else(|| fdo::Error::Failed("Missing member".into()))?;
```
zeenix, line 366: Why are we removing this? 🤔

zbus/src/object_server/mod.rs
zeenix: I meant that `_` -> `_member` is unrelated. Not the existing code here. :)

zbus/src/fdo/mod.rs
```diff
@@ -198,4 +200,108 @@ mod tests {
...
+        .expect("Failed to create PeerProxy");
+        let result = proxy.get_machine_id().await;
+        // GetMachineId may fail on some platforms (like *BSD), but should not fail
+        // with "Unknown object" error
+        match &result {
+            Ok(id) => {
+                assert!(!id.is_empty(), "Machine ID should not be empty");
+            }
+            #[cfg(all(
+                unix,
```
zeenix: why is `unix` needed? It even conflicts with `windows` in the `not` part. :)

zeenix (approved):

PR #1603, opened 2025-12-05
🐛 Fix `get_machine_id` for macOS

cachebag:
closes #1595. Was going to do BSD but I assume you want to keep these things contained.  Additionally, I think it will take me a bit longer (particularly for OpenBSD and NetBSD).

Also, please tell me if I'm being overbearing with the PR's. I promise I don't mean to bother.

PR #1636, opened 2025-12-24
🐛 zb: Implement `get_machine_id()` for *BSD platforms

cachebag:
This PR extends upon work completed in #1594 and #1603 respectively.

First tries standard D-Bus machine-id locations
(`/var/lib/dbus/machine-id`, `/etc/machine-id`, `/var/db/dbus/machine-id`)

Falls back to `kern.hostuuid sysctl`, which provides a persistent UUID on these platforms.

It then reads from standard D-Bus machine-id locations and returns a helpful error message if no machine-id file exists. (these platforms do not have a built-in machine UUID mechanism)

### DIFF
--- a/zbus/src/fdo/peer.rs
+++ b/zbus/src/fdo/peer.rs
@@ -26,15 +26,30 @@ impl Peer {
     /// if possible, but this is not always possible to implement and is not guaranteed. It does not
     /// matter which object path a GetMachineId is sent to.
     ///
-    /// Note: Currently only implemented for Linux, macOS, and Windows. On other Unix platforms
-    /// (*BSD), this method returns a `NotSupported` error.
+    /// This method is implemented for:
+    /// - Linux: Reads from `/var/lib/dbus/machine-id` or `/etc/machine-id`
+    /// - macOS: Uses `gethostuuid()` system call
+    /// - FreeBSD/DragonFlyBSD: Reads from standard D-Bus locations, falls back to `kern.hostuuid`
+    /// - OpenBSD/NetBSD: Reads from standard D-Bus locations (`/var/db/dbus/machine-id`, etc.)
+    /// - Windows: Uses Windows hardware profile GUID
     fn get_machine_id(&self) -> Result<String> {
-        get_machine_id()
+        // On *BSD platforms, first try standard D-Bus machine-id locations
+        #[cfg(any(
+            target_os = "freebsd",
+            target_os = "dragonfly",
+            target_os = "openbsd",
+            target_os = "netbsd"
+        ))]
+        if let Some(id) = read_dbus_machine_id() {
+            return Ok(id);
+        }
+
+        get_platform_machine_id()
     }
 }
 
 #[cfg(target_os = "linux")]
-fn get_machine_id() -> Result<String> {
+fn get_platform_machine_id() -> Result<String> {
     let mut id = match std::fs::read_to_string("/var/lib/dbus/machine-id") {
         Ok(id) => id,
         Err(e) => {
@@ -54,7 +69,7 @@ fn get_machine_id() -> Result<String> {
 }
 
 #[cfg(target_os = "macos")]
-fn get_machine_id() -> Result<String> {
+fn get_platform_machine_id() -> Result<String> {
     unsafe extern "C" {
         fn gethostuuid(id: *mut u8, wait: *const libc::timespec) -> libc::c_int;
     }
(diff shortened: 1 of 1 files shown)


zbus/src/fdo/peer.rs
```diff
@@ -76,8 +80,120 @@ fn get_machine_id() -> Result<String> {
...
+            }
+        }
+    }
+    None
+}
+
+/// Get the machine ID on FreeBSD or DragonFlyBSD using the kern.hostuuid sysctl.
+/// This returns a UUID that is typically generated at install time and persists across reboots.
+#[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+fn get_hostuuid_sysctl() -> Result<String> {
```
zeenix: Prefer top-down code layout: usage before declarate/definition. Same for the new `try_read_dbus_machine_id`

zbus/src/fdo/peer.rs
```diff
@@ -76,8 +80,120 @@ fn get_machine_id() -> Result<String> {
...
-#[cfg(all(unix, not(any(target_os = "linux", target_os = "macos"))))]
+/// Try to read machine ID from standard D-Bus locations.
+/// Returns the machine ID if found at any of the standard paths.
+#[cfg(any(
+    target_os = "freebsd",
+    target_os = "dragonfly",
+    target_os = "openbsd",
+    target_os = "netbsd"
+))]
+fn try_read_dbus_machine_id() -> Option<String> {
```
zeenix: `try_` prefix is typically used for methods returning a Result so probably best to skip it.

zbus/src/fdo/peer.rs
```diff
@@ -76,8 +80,120 @@ fn get_machine_id() -> Result<String> {
...
+    Ok(machine_id)
+}
+
+#[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+fn get_machine_id() -> Result<String> {
+    // First try standard D-Bus machine-id locations
+    if let Some(id) = try_read_dbus_machine_id() {
+        return Ok(id);
+    }
+
```
zeenix: This is getting repeated now in multiple impls so perhaps it'd be best to move it to the `Peer::get_machine_id` and rename this to `get_platform_machine_id`?

zbus/src/fdo/peer.rs
```diff
@@ -115,4 +231,31 @@ mod tests {
             "machine ID should only contain hex characters"
         );
     }
+
+    #[test]
+    #[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
```
zeenix, line 230: Do you happen to have access to machines to check if the tests actually work? :)

cachebag: > Do you happen to have access to machines to check if the tests actually work? :)

I do back at home however I'm currently in Mexico. I'll be back in a few days and will run all tests.

In the meantime I'll try to spin up some VM's but my WiFi is quite bad down here so no promises 🫤

zeenix: Ah ok. No hurry on this. Enjoy your vacation instead. :)

cachebag:
# FreeBSD (15.0)

image

Command run was `cargo test -p zbus peer::tests::freebsd_machine_id -- --nocapture`

## Verification of `sysctl kern.hostuuid`
image

#

Sorry for the delay on this. Will report back with other variants if you'd like, hopefully tomorrow. But honestly, if FreeBSD worked, DragonFly should work too since they share the same `kern.hostuuid` mechanism. 

Also if there's anything specific you'd like me to test or show, please lmk @zeenix

zeenix:
> Command run was `cargo test -p zbus peer::tests::freebsd_machine_id -- --nocapture`

The screenshot doesn't show that test being run. Does it run and succeed?

> Sorry for the delay on this.

No worries. :)

zeenix:
@cachebag btw, you might want to use GNOME Boxes for such things. I may have had a hand in creation of that. :wink:

cachebag:
Here's a (really slow and shitty, might have to open in a new tab, sorry) screen cap.

> I may have had a hand in creation of that. 😉

No way! We used that in one of my Cybersecurity courses at school!

zeenix:
> Here's a (really slow and shitty, might have to open in a new tab, sorry) screen cap.

Dude! It's not only shitty, the important bit goes away super fast so not super useful. :) It's ok, I trust you that you tested it. You could just copy&paste the text. :laughing:

cachebag:
https://github.com/user-attachments/assets/0445863c-09fa-4070-bfd1-5bb45a8a16fa



Sorry about that 😅 Here's the run written to a log too.
[test.log](https://github.com/user-attachments/files/24528947/test.log)

zbus/src/fdo/peer.rs
```diff
@@ -76,16 +91,117 @@ fn get_machine_id() -> Result<String> {
     Ok(uuid.iter().map(|b| format!("{b:02x}")).collect())
 }
 
-// TODO: Implement for *BSD platforms.
-#[cfg(all(unix, not(any(target_os = "linux", target_os = "macos"))))]
-fn get_machine_id() -> Result<String> {
+#[cfg(any(target_os = "freebsd", target_os = "dragonfly"))]
+fn get_platform_machine_id() -> Result<String> {
+    // Fall back to kern.hostuuid sysctl
+    get_hostuuid_sysctl()
```
zeenix: I don't think we need a separate `get_hostuuid_sysctl` now. Just merge that into here?

PR #1638, opened 2025-12-25
✨ zm: add special handling for ao DBus signatures

### LINKED ISSUE #332: dbug_proxy! binding for method which returns an array of object paths
In GitLab by @danburkert on May 6, 2023, 20:06

The `dbug_proxy!` macro provides the `object` annotation generating a proxy return type for methods which return an object path - it would be nice if there were an equivalent annotation for handling methods that return an array of object paths.

farnz:
Extend the support for the `object` attribute to return a `Vec<_>` if the `object_vec` attribute is also present.

This simplifies use of APIs like [NetworkManager's `GetDevices`](https://networkmanager.dev/docs/api/latest/gdbus-org.freedesktop.NetworkManager.html#gdbus-method-org-freedesktop-NetworkManager.GetDevices) by allowing you to ask the macro to do the conversion from `Vec<OwnedObjectPath>` to `Vec<SuitableProxyType>`.

Fixes #332.

### DIFF
--- a/zbus_macros/src/proxy.rs
+++ b/zbus_macros/src/proxy.rs
@@ -36,6 +36,7 @@ def_attrs! {
         object str,
         async_object str,
         blocking_object str,
+        object_vec none,
         no_reply none,
         no_autostart none,
         allow_interactive_auth none
@@ -525,6 +526,7 @@ fn gen_proxy_method_call(
                 .unwrap_or_else(|| format!("{o}Proxy"))
         }
     });
+    let proxy_vec = method_attrs.object_vec;
 
     let method_flags = match (
         method_attrs.no_reply,
@@ -602,24 +604,51 @@ fn gen_proxy_method_call(
 
     if let Some(proxy_path) = proxy_object {
         let proxy_path = parse_str::<Path>(&proxy_path)?;
+        let ok_type = if proxy_vec {
+            quote!(Vec<#proxy_path<'p>>)
+        } else {
+            quote!(#proxy_path<'p>)
+        };
         let signature = quote! {
-            fn #method #ty_generics(#inputs) -> #zbus::Result<#proxy_path<'p>>
+            fn #method #ty_generics(#inputs) -> #zbus::Result<#ok_type>
             #where_clause
         };
 
+        let proxy_build = quote! {
+            #proxy_path::builder(&self.0.connection())
+                .path(object_path)?
+                .build()
+                #wait
+        };
+        let method_call = quote! {
+            self.0.call(
+                #dbus_member_name,
+                &#zbus::zvariant::DynamicTuple((#(#args,)*)),
+            )
+            #wait?
+        };
+        let body = if proxy_vec {
+            quote! {
+                let object_paths: Vec<#zbus::zvariant::OwnedObjectPath> = #method_call;
+
+                let mut proxies = Vec::with_capacity(object_paths.len());
+                for object_path in object_paths {
+                    let proxy = #proxy_build?;
+                    proxies.push(proxy);
+                }
+
+                Ok(proxies)
+            }
+        } else {
+            quote! {
+                let object_path: #zbus::zvariant::OwnedObjectPath = #method_call;
+                #proxy_build
+            }
+        };
         Ok(quote! {
             #(#other_attrs)*
             pub #usage #signature {
-                let object_path: #zbus::zvariant::OwnedObjectPath =
-                    self.0.call(
-                        #dbus_member_name,
-                        &#zbus::zvariant::DynamicTuple((#(#args,)*)),
-                    )
-                    #wait?;
-                #proxy_path::builder(&self.0.connection())
-                    .path(object_path)?
-                    .build()
-                    #wait
+                #body
             }
         })
     } else {
@@ -726,6 +755,8 @@ fn gen_proxy_property(
             }
         });
 
+        let proxy_vec = method_attrs.object_vec;
+
         // This should fail to compile only if the return type is wrong,
         // so use that as the span.
         let body_span = if let ReturnType::Type(_, ty) = &signature.output {
@@ -745,17 +776,44 @@ fn gen_proxy_property(
             let proxy_path: Path = parse_str(&proxy_path_str).unwrap();
             let method_name = Ident::new(rust_method_name, m.sig.ident.span());
             let inputs = &signature.inputs;
-            let custom_sig = quote! {
-                fn #method_name(#inputs) -> #zbus::Result<#proxy_path<'p>>
+            let custom_sig = if proxy_vec {
+                quote! {
+                    fn #method_name(#inputs) -> #zbus::Result<Vec<#proxy_path<'p>>>
+                }
+            } else {
+                quote! {
+                    fn #method_name(#inputs) -> #zbus::Result<#proxy_path<'p>>
+                }
             };
-            let body = quote_spanned! {body_span =>
-                let object_path: #zbus::zvariant::OwnedObjectPath =
-                    self.0.get_property(#property_name)#wait?;
+            let property_get = quote! {
+                self.0.get_property(#property_name)#wait?
+            };
+            let proxy_build = quote! {
                 #proxy_path::builder(&self.0.connection())
                     .destination(self.0.destination().to_owned())?
                     .path(object_path)?
                     .build()
                     #wait
+            };
+            let body = if proxy_vec {
+                quote_spanned! {body_span =>
+                    let object_paths: Vec<#zbus::zvariant::OwnedObjectPath> =
+                        #property_get;
+
+                    let mut proxies = Vec::with_capacity(object_paths.len());
+                    for object_path in object_paths {
+                        let proxy = #proxy_build?;
+                        proxies.push(proxy);
+                    }
+
+                    Ok(proxies)
+                }
+            } else {
+                quote_spanned! {body_span =>
+                    let object_path: #zbus::zvariant::OwnedObjectPath =
+                        #property_get;
+                    #proxy_build
+                }
             };
             (body, Some(custom_sig))
         } else {
--- a/zbus_macros/tests/tests.rs
+++ b/zbus_macros/tests/tests.rs
@@ -430,3 +430,63 @@ mod signal_from_message {
             .expect_err("Message does not have correct data");
     }
 }
+
+#[test]
+fn test_proxy_object_list() {
+    #[derive(Clone)]
+    struct ObjectList {
+        paths: [zbus::zvariant::ObjectPath<'static>; 2],
+    }
+
+    #[zbus_macros::interface(
+        name = "org.freedesktop.zbus_macros.ObjectList",
+        proxy(default_service = "org.freedesktop.zbus_macros")
+    )]
+    impl ObjectList {
+        #[zbus(proxy(object = "ObjectList", object_vec))]
+        async fn get_test_objects(&self) -> Vec<zbus::zvariant::ObjectPath<'static>> {
+            self.paths.to_vec()
+        }
+
+        #[zbus(property, proxy(object = "ObjectList", object_vec))]
+        fn objects(&self) -> Vec<zbus::zvariant::ObjectPath<'static>> {
+            self.paths.to_vec()
+        }
+    }
+
+    static OBJECT_LIST: ObjectList = ObjectList {
+        paths: [
+            zbus::zvariant::ObjectPath::from_static_str_unchecked(
+                "/org/freedesktop/zbus_macros/object_list/0",
+            ),
+            zbus::zvariant::ObjectPath::from_static_str_unchecked(
+                "/org/freedesktop/zbus_macros/object_list/1",
+            ),
+        ],
+    };
+
+    fn check_return(list: Vec<ObjectListProxyBlocking<'_>>) {
+        for (correct, returned) in OBJECT_LIST.paths.iter().zip(list.into_iter()) {
+            assert!(returned.inner().path() == correct);
+        }
+    }
+
+    let connection = zbus::blocking::connection::Builder::session()
+        .unwrap()
+        .serve_at(OBJECT_LIST.paths[1].as_ref(), OBJECT_LIST.clone())
+        .unwrap()
+        .build()
+        .unwrap();
+    let destination = connection.unique_name().unwrap().clone();
+
+    let proxy = ObjectListProxyBlocking::builder(&connection)
+        .path(OBJECT_LIST.paths[1].as_ref())
+        .unwrap()
+        .destination(&destination)
+        .unwrap()
+        .build()
+        .unwrap();
+
+    check_return(proxy.get_test_objects().unwrap());
+    check_return(proxy.objects().unwrap());
+}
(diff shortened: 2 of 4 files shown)


zbus_macros/src/proxy.rs
```diff
@@ -602,26 +604,56 @@ fn gen_proxy_method_call(
...
+            quote! {
+            fn #method #ty_generics(#inputs) -> #zbus::Result<Vec<#proxy_path<'p>>>
             #where_clause
+            }
+        } else {
+            quote! {
+                fn #method #ty_generics(#inputs) -> #zbus::Result<#proxy_path<'p>>
+                #where_clause
+            }
         };
```
zeenix: A bit of code-duplication both blocks, so better only generate the return type in each block and then use that in the `quote!` block that was already there;

```rust
let ret = if proxy_vec {
    quote! { Vec<#proxy_path<'p>> }
} else {
    quote! { #proxy_path<'p> }
};
let signature = quote! { 
    fn #method #ty_generics(#inputs) -> #zbus::Result<#ret>
};
```

zbus_macros/src/proxy.rs
```diff
@@ -602,26 +604,56 @@ fn gen_proxy_method_call(
...
+                            &#zbus::zvariant::DynamicTuple((#(#args,)*)),
+                        )
+                        #wait?;
+                    #proxy_path::builder(&self.0.connection())
+                        .path(object_path)?
+                        .build()
+                        #wait
+                }
+            })
+        }
```
zeenix: Same comment about code-duplication here. Let's avoid that.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,95 @@ mod signal_from_message {
             .expect_err("Message does not have correct data");
     }
 }
+
+mod proxy_object_list {
```
zeenix: If it's going to be just one test, no need for a mod. Let's just put the types in the test function itself please.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,95 @@ mod signal_from_message {
...
+            }
+            if !connection
+                .object_server()
+                .at(object_list.paths[1].as_ref(), object_list.clone())
+                .await
+                .unwrap()
+            {
+                panic!("path 1 already registered");
+            }
+            let objects_count = std::thread::spawn({
```
zeenix: * `object_count` is not the best name for a thread handle.
* We tend to spawn tasks instead: `connection.executor.spawn` would do.
* In fact, given that zbus serves the interfaces automatically in a task, I don't think there is any need for spawning anything here. Just register the interfaces first and then call the methods through the proxy and check the values to ensure everything works as expected.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,95 @@ mod signal_from_message {
...
+                ],
+            };
+            let destination = connection.unique_name().unwrap().clone();
+            if !connection
+                .object_server()
+                .at(object_list.paths[0].as_ref(), object_list.clone())
+                .await
+                .unwrap()
+            {
+                panic!("path 0 already registered");
```
zeenix: * Just an `assert!` is more than enough IMO.
* You can eliminate the need for this entirely and simplify the code by making use of the `connection::Buiilder` and register all the interfaces at build time.

zbus_macros/tests/tests.rs
farnz: I can't use a blocking connection inside the runtime spawned by `zbus::block_on`:

```
Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.
```

zeenix: Sure but you don't necessarily need to use the blocking API here.

farnz: Had a look at the existing dependencies, and I can't find a good runtime-agnostic solution here. I also realise I'm racing the thread against the async work.

I need an equivalent of [`tokio::sync::Notify`](https://docs.rs/tokio/latest/tokio/sync/struct.Notify.html) to address the race condition - I can open-code it (use `futures_util::lock::Mutex`, pass an owned lock into the thread, before calling `join()`, wait until I can asynchronously own the lock), but I'm open to other suggestions.

zbus_macros/tests/tests.rs
farnz: Can you tell I only normally use the `proxy` side of ZBus? :-)

I should have realised there was a simpler way to do things, though - the crates are very well-written.

zbus_macros/tests/tests.rs
zeenix: We depend on event-listener quite heavily and in fact use it for the same reason in many tests. Grep for `event_listener` and you'll find at least one example to learn from. :)

zeenix: Although I am pretty sure you don't really need any notifications here. All that's needed for testing this new attribute is adding an interface with methods that return a vec of objects and an interface(s) returned by this method, and then use the proxy side to make the calls, check the results and you're done. This can all be the same exact thread, connection and task.

zbus_macros/tests/tests.rs
zeenix: > Can you tell I only normally use the `proxy` side of ZBus? :-)

Most people do. :) BTW, it's always written as "zbus" (all small letters). :wink:

zbus_macros/tests/tests.rs
farnz: Would you be OK with me simply stripping out the test that the blocking API I generate works, and relying on hope and code inspection to confirm that it's fine?

I can do that if you're OK with it - the point of the thread is simply to check that the blocking API codegen does the right thing for `Vec<_>`, and generates working code (in part because I don't use the blocking API at all, and thus wouldn't notice it coincidentally if it breaks).

If I keep the blocking API test in place, then I need a separate thread, and a notification that the thread has finished to avoid an annoying race condition in the test where it usually succeeds (because the blocking thread finishes before the async API test is finished), except when you're unlucky - and retrying the flaky test appears to "fix" it.

farnz: For now, I've pushed something cleaning up the test (and checking more things, and removing the blocking API check. I can re-add the blocking API check if you'd like me to.

zeenix: > Would you be OK with me simply stripping out the test that the blocking API I generate works,

Even though the blocking API is just a thin wrapper(s) around the async API, I am always open to more complete testing so I don't have any objections on keeping the test blocking.

> If I keep the blocking API test in place, then I need a separate thread

This is what I don't quite understand.

> the thread has finished to avoid an annoying race condition in the test where it usually succeeds (because the blocking thread finishes before the async API test is finished)

Given that the client side in the main thread would block (or await) on getting replies and then check the response before test function returns, I don't think there is any possibility of things going wrong but just not noticed.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,95 @@ mod signal_from_message {
...
+
+        #[zbus(property, proxy(object = "ObjectList", object_vec))]
+        fn objects(&self) -> Vec<zbus::zvariant::OwnedObjectPath> {
+            self.paths.clone()
+        }
+    }
+
+    #[test]
+    fn test_proxy_object_list() {
+        zbus::block_on(async {
```
zeenix: Let's follow the convention in the existing code to avoid one level of indentation:


```suggestion
    fn test_proxy_object_list() {
        zbus::block_on(test_proxy_object_list_async());
    }

    async fn test_proxy_object_list_async() {
```

zeenix (changes requested):
All good except for one nitpick. I think we can merge after that. I'm sorry for the very late review but I've been off for obvious reasons. :)

zbus_macros/tests/tests.rs
farnz: The blocking API test has to be outside the `block_on`, otherwise it can't create its runtime; this is easiest done by starting a separate thread (so that I don't have to think too hard about the server-side interface, just the proxy).

Once I have two threads in play in the test case, I want to avoid the risk of the async task reaching the `join()` before the blocking thread has done its communications with the task spawned by `serve_at`. If I'm unlucky, the OS thread scheduler will not start executing the blocking thread until I reach `.join()` at the end of the `block_on`. But once the async block calls `.join()`, it will not return to the async scheduler (to drive the server side forward) until the blocking thread finishes.

We're now locked up - the blocking thread is waiting for the async scheduler to run the task spawned by `serve_at`, but the async scheduler doesn't have any threads to run on until the blocking thread completes.

This won't ever happen if the async scheduler creates at least 2 threads, or if the blocking thread completes before we get to the `join()`, which is why it's an unusual case - it's very likely that the blocking thread will complete before we get to the join on any system with 2 or more CPU cores.

zeenix: > The blocking API test has to be outside the `block_on`

I'm well-aware of that. You use block_on if you want to use the async API and not otherwise.

> Once I have two threads in play in the test case

That's what I'm saying: you don't need more than 1 thread here. You only need to do basically this:

```rust
// Assuming blocking API.
let conn = connection::Builder::session()?.serve_at(path1, interface1)?.serve_at(path2, interface2)?.build()?;
let proxy = ObjectListProxy::new(&conn, ..);
let proxies = proxy.objects()?;
for p in proxies {
    // Assertions and method calls on `p` goes here.
}
```

zeenix: I noticed that the test is still spawning threads so I unresolved this conversation. I'd suggest looking around at other test cases for inspiration to make this test super simple, in case my pseudo-code above isn't very readable/helpful. :)

farnz: That works, too, but ran smack into my lack of understanding of both the serving-side and the blocking API of zbus :see_no_evil:

It was obvious to me that the async interface must spawn a task to handle `serve_at`, but not obvious that the hints that the blocking interface is a thin wrapper around the async interface implied that `serve_at` doesn't block the running thread. Mostly a matter of my ignorance - but it's really hard to fix ignorance when you don't know what assumptions I'm making, and thus have to keep digging until you find the false assumption.

Thanks for taking the time to explain things to me!

zeenix: > Thanks for taking the time to explain things to me!

NP.

zbus_macros/tests/tests.rs
zeenix: This isn't resolved.

zbus_macros/tests/tests.rs
zeenix: This isn't resolved either afaict. 🤔

farnz:
User error on my part - made the changes, forgot about git's staging area so they didn't get in. Should look better now, and I'll add the blocking API test momentarily.

zbus_macros/src/proxy.rs
zeenix: This is also still not addressed.

zbus_macros/src/proxy.rs
zeenix: neither is this. Please don't mark discussions as resolved w/o verifying in the code on the PR/branch.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,91 @@ mod signal_from_message {
...
+            fn $func_name(list: Vec<$proxy_type>) {
+                for (correct, returned) in OBJECT_LIST.paths.iter().zip(list.into_iter()) {
+                    assert!(returned.inner().path().as_str() == correct.as_str());
+                }
+            }
+        };
+    }
+
+    check_return!(check_async_return, ObjectListProxy<'_>);
+    check_return!(check_blocking_return, ObjectListProxyBlocking<'_>);
```
zeenix: * IMO with macro only containing 4 LOC and creating a function that then needs to be called, is not helping the readability here. Can we remove the use of macro_rules here. This is not that complex a test.
* Also there is no reason for both async and blocking tests. As I pointed out before, the blocking API is just a wrapper around the async one so if you only use blocking API, it will indirectly test async API too.

farnz:
> I hate to go back on my words about this being ready very soon but I had trusted that you had actually address on my review comments but on careful review (which I should have done before too, sorry) I find that you hadn't address many comments but marked them as resolved. Kindly address them and let's simplify the test even more (see inline comments).

I do apologise - I fouled up driving git.

And please don't worry about asking for more changes as desired - I understand that your priority is "if the contributor disappears, is this code still maintainable?", and thus it's not personal, but about maximum quality.

farnz:
I believe I've done the right things with git this time, but I'm leaving the comments unresolved for you to double-check.

zbus_macros/src/proxy.rs
```diff
@@ -602,24 +604,51 @@ fn gen_proxy_method_call(
...
             #where_clause
         };
 
+        let build_proxy = quote! {
+            #proxy_path::builder(&self.0.connection())
+                .path(object_path)?
+                .build()
+                #wait
+        };
+        let call_method = quote! {
```
zeenix: Pretty big nitpick but since you welcome those (😉): The rest of the macro code, we've used nouns rather than verbs so best to do the same for consistency: `proxy_build` and `method_call`.

zbus_macros/tests/tests.rs
```diff
@@ -430,3 +430,63 @@ mod signal_from_message {
...
+            ),
+            zbus::zvariant::ObjectPath::from_static_str_unchecked(
+                "/org/freedesktop/zbus_macros/object_list/1",
+            ),
+        ],
+    };
+
+    fn check_return(list: Vec<ObjectListProxyBlocking<'_>>) {
+        for (correct, returned) in OBJECT_LIST.paths.iter().zip(list.into_iter()) {
+            assert!(returned.inner().path().as_str() == correct.as_str());
```
zeenix: `.as_str()` shouldn't be needed, perhaps you just need to deref one or both of them?

zeenix (changes requested):
Getting there. :)

zbus_macros/tests/tests.rs
farnz: That looks to be left-over from a previous attempt where I had an `OwnedObjectPath` on one side, not a `&ObjectPath<'_>`. I needed `as_ref()` or a deref to make it work, and chose `as_ref` because I have Perl4 trauma around sigils.

Removed.

zbus_macros/src/proxy.rs
farnz: I am more than happy to fix up even the tiniest of nits - compare the amount of work I've done to make this PR mergeable to the amount I've avoided doing by reusing zbus instead of writing my own DBus client implementation, and I think I still come out *way* ahead on the deal by any sensible measure :smile:

farnz:
Thank you very much, @zeenix , for all your patience with me - both explaining things clearly when I wasn't getting them, and putting up with me failing to drive git properly.

zeenix:
> Thank you very much, @zeenix , for all your patience with me - both explaining things clearly when I wasn't getting them, and putting up with me failing to drive git properly.

No worries at all. I really like your very positive attitude. Most contributors just drop the ball and disappear when asked to improve their contributions. We need more folks like you!

PR #1670, opened 2026-01-10
🐛 zvd: Fix hard-coded zvariant path in signature generation

cachebag:
`signature_to_tokens()` now respects crate path detection instead of hard-coding `::zvariant`, fixing compilation errors in downstream crates that access zvariant through zbus re-exports.

# 

[Fixes the issue seen here](https://github.com/bilelmoussaoui/ashpd/issues/323#issuecomment-3730798061)...

`signature_to_tokens()` in was hard-coded to use `::zvariant` instead of respecting the crate path detection from `zvariant_path()`. This caused compilation failures in downstream crates like [ashpd](https://github.com/bilelmoussaoui/ashpd/issues/323#issuecomment-3730798061) that access zvariant through zbus re-exports rather than as a direct dependency.

Added `signature_to_tokens_with_crate()` that accepts a crate path parameter and updated `type.rs` to pass the detected path. All recursive calls within the function now use the parameterized version to ensure consistency.

#1605 didn't change the default behavior, but this bug existed in `signature_to_tokens()` since it was introduced and is now exposed by increased usage.

Edit: Bring in test from https://github.com/z-galaxy/zbus/pull/1671 @smazurov

### DIFF
--- a/zvariant_derive/src/signature.rs
+++ b/zvariant_derive/src/signature.rs
@@ -111,3 +111,54 @@ pub fn signature_to_tokens_with_crate(signature: &Signature, zv: &TokenStream) -
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn signature_to_tokens_with_crate_uses_custom_path() {
+        let custom_path = quote! { ::zbus::zvariant };
+        let sig = Signature::Str;
+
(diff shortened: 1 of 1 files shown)


zeenix:
@cachebag @smazurov thanks so much for your quick actions! I'll roll out a release as soon as I'm home..

PR #1671, opened 2026-01-10
🐛 zv_derive: use dynamic path in signature_to_tokens

smazurov:
## Summary

Fixes compilation for crates that depend on zbus but not zvariant directly (e.g., udisks2).

## Problem

The `signature_to_tokens` function hardcodes `::zvariant` path, ignoring the dynamically resolved path from `zvariant_path()`. This causes the Type derive macro to generate incorrect code:

```rust
impl ::zbus::zvariant::Type for Foo {
    const SIGNATURE: ... = &::zvariant::Signature::Str;
    //                       ^^^^^^^^^ wrong - should be ::zbus::zvariant
}
```

## Solution

Add `signature_to_tokens_with_path()` that accepts the zvariant path as a parameter, and use it in the Type derive macro.

## Test plan

- Added unit tests for the new function
- All existing tests pass

### DIFF
--- a/zvariant_derive/src/signature.rs
+++ b/zvariant_derive/src/signature.rs
@@ -43,10 +43,18 @@ impl Parse for SignatureInput {
 /// Converts a parsed `Signature` to compile-time token representation.
 ///
 /// This function generates the Rust tokens that will construct the signature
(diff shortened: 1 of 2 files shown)


smazurov:
same fix as #1670 plus a test

PR #1673, opened 2026-01-11
🐛 zm: Allow `out_args` with multiple names for non-tuple types

cachebag:
The validation added in 53d053e was too strict, rejecting `out_args` with multiple names for non-tuple Rust types. This broke types with custom DBus tuple signatures (e.g., `#[zvariant(signature = "(ua{sv})")])`.

Instead of erroring, we now gracefully handle multiple names by using the types full signature when we cant extract individual field types.

@zeenix Idk how you want to move from here; this unbreaks the release which is good, but introspection is technically incomplete. The individual struct fields aren't named in the XML and I'm pretty sure `out_args` are silently ignored. We'd need to do something like runtime signature parsing during macro expansion. Lmk what you think?

### DIFF
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -1307,23 +1307,22 @@ fn introspect_add_output_args(
                 args.extend(introspect_output_arg(&t.elems[i], name, cfg_attrs));
             }
         } else {
-            // Handle single output case - check if out_args was specified
-            let name = match arg_names {
-                Some(names) => match names.len() {
-                    1 => Some(&names[0]),
-                    0 => None,
-                    _ => {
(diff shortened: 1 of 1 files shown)


zeenix:
@cachebag Thanks so much for the quick actions here, especially on a Sat. :pray: 

> Lmk what you think?

Actually, given that we know this is not a good solution, I think for now I'll simply revert 53d053edf92d68f23c3397079fe6e8bd2e65081a for now and make a micro release. Then we're not in a huge hurry and we can resolve this properly later.

zeenix (approved):
Actually this is superior to my simple naive solution of reverting (#1674) so I'll go with this one. 👍

PR #1674, opened 2026-01-11
Revert "🐛 zm: Apply out_args to single outputs in introspection XML"

zeenix:
This reverts commit 53d053edf92d68f23c3397079fe6e8bd2e65081a, which
broke the build for types with custom D-Bus struct signatures (e.g.,
`#[zvariant(signature = "(ua{sv})")]`).

We'll try to bring back this fix later in a way that it doesn't break this
use case.


zeenix:
Closing in favor of #1673.

PR #1682, opened 2026-01-13
✨ zv: Implement Basic for more types

### LINKED ISSUE #1681: Can't use basic type as map key wrapped in some container (Cow)
I had used `HashMap` with `Cow<'a, str>` key and it worked well, until  I've updated version and it broke compilation.

In docs I don't see that `Basic` trait is implemented for containers such as Box, Rc, Arc and Cow.

I see solution like this:

```rust
impl<T: Basic> for Box<T> {
    const SIGNATURE_CHAR: char = T::SIGNATURE_CHAR;
    /// The type signature, as a string.
    const SIGNATURE_STR: &'static str = T::SIGNATURE_STR;
}

// and impls for other containers
```

kusstas:
Implement `Basic` trait for:

- heapless::String
- isize/usize
- atomic types
- deref types
- wrapper types

Fixes #1681.

### DIFF
--- a/zvariant/src/type/libstd.rs
+++ b/zvariant/src/type/libstd.rs
@@ -6,18 +6,9 @@ use std::{
     num::{Saturating, Wrapping},
     ops::{Range, RangeFrom, RangeInclusive, RangeTo},
     rc::{Rc, Weak as RcWeak},
-    sync::{
-        Arc, Mutex, RwLock, Weak as ArcWeak,
-        atomic::{
-            AtomicBool, AtomicI8, AtomicI16, AtomicI32, AtomicIsize, AtomicU8, AtomicU16,
-            AtomicU32, AtomicUsize,
-        },
-    },
+    sync::{Arc, Mutex, RwLock, Weak as ArcWeak},
     time::Duration,
 };
-
-#[cfg(target_has_atomic = "64")]
-use std::sync::atomic::{AtomicI64, AtomicU64};
 
 impl<T> Type for PhantomData<T>
 where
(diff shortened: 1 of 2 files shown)


zvariant/src/type/libstd.rs
```diff
@@ -106,18 +112,43 @@ macro_rules! deref_impl {
...
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for ArcWeak<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for Mutex<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for RwLock<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for Box<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for Rc<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for RcWeak<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for Cell<T>);
+impl_type_for_deref!(T, <T: ?Sized + Type> Type for RefCell<T>);
+
+macro_rules! impl_basic_for_deref {
```
zeenix: Instead of moving `Basic` impl under `type` module, I'd prefer keeping them in `basic` module, which is what I'm doing in my WIP branch for this. If you think that module is getting a bit too big, feel free to split that into a hierarchy too (just like `type` is) but please do so in a separate commit that comes before the actual changes.

zvariant/src/type/libstd.rs
```diff
@@ -90,12 +90,18 @@ impl<const CAP: usize> Type for heapless::String<CAP> {
...
+    const SIGNATURE_STR: &'static str = <&str as Basic>::SIGNATURE_STR;
+}
+
 // Empty type deserves empty signature
 impl Type for () {
     const SIGNATURE: &'static Signature = &Signature::Unit;
 }
 
-macro_rules! deref_impl {
+macro_rules! impl_type_for_deref {
```
zeenix, line 89: This is a good change but not exactly part of implementing Basic trait so please split into a separate commit that comes before. Use ♻️ emoji for it (to help you avoid the trouble of choosing one 🙂).

zeenix (changes requested):
Overall, really good job! As I wrote, I was also working on this not realizing how fast you'd act on my request for a PR. 😆 Since your implementation is complete and covers a lot more types, I'd ditch my WIP work and take your PR but it needs some changes. Aside from the inline comments, some notes (nitpicks) on the commit message and PR title+description:

* Please keep title small and make use of the desciption to add details. Here you could have "zv: Implement Basic for more types" and (if you want) add names of types in the description part.
* It's missing the "Fixes #1681." line in the description of the commit and PR.
* This change is itself not a bug fix. ✨ would be more appropriate.

zeenix:
While this is no longer the highest-priority (see #1685), it's still a good change and will be very much needed when we re-introduce that change back again for 6.0. @kusstas please let me know if you won't be able to finish this so I can try to find some time to finish the work for you. :pray:

kusstas:
Hi, at this moment I'm not able to finish this PR, sorry about that. I'll able after few weeks, so..

zeenix:
> Hi, at this moment I'm not able to finish this PR, sorry about that. I'll able after few weeks, so..

Thanks for the honesty. :+1: I did everything now.

PR #1684, opened 2026-01-15
🐛 xmlgen: avoid generating duplicate `receive_*_changed` methods

antonok-edm:
Resolves `Case 2` of #231

I ran into this issue when trying to generate `zbus` bindings for ModemManager, which has four such naming collisions. This resolves the issue there and should help for others who have commented on #231.

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -1,5 +1,6 @@
 use snakecase::ascii::to_snakecase;
 use std::{
+    collections::HashMap,
     error::Error,
     fmt::{Display, Formatter, Write},
     process::{Command, Stdio},
@@ -185,6 +186,11 @@ impl GenTrait<'_> {
 
         let mut signals = iface.signals().to_vec();
         signals.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
+
+        // collect signal method names for collision detection with implicit property change
+        // receivers
+        let mut changed_suffix_signals = HashMap::new();
+
         for signal in &signals {
             let args = parse_signal_args(signal.args());
             let name = to_identifier(&to_snakecase(signal.name().as_str()));
@@ -196,20 +202,41 @@ impl GenTrait<'_> {
                 writeln!(w, "    #[zbus(signal)]")?;
             }
             writeln!(w, "    fn {name}({args}) -> zbus::Result<()>;",)?;
+
+            if name.ends_with("_changed") {
+                changed_suffix_signals.insert(name, signal.name());
+            }
         }
 
         let mut props = iface.properties().to_vec();
         props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for p in props {
             let name = to_identifier(&to_snakecase(p.name().as_str()));
-            let fn_attribute = if pascal_case(&name) != p.name().as_str() {
-                format!("    #[zbus(property, name = \"{}\")]", p.name())
+
+            let prop_change_receiver_name = format!("{name}_changed");
+            let signal_collision = changed_suffix_signals.get(&prop_change_receiver_name);
+
+            let (property_args, comment) = if let Some(signal_collision) = signal_collision {
+                (
+                    "(emits_changed_signal = \"false\")",
+                    format!(
+                        "    // Note: change signal is shadowed by `{}` signal.\n",
+                        signal_collision.as_str()
+                    ),
+                )
             } else {
-                "    #[zbus(property)]".to_string()
+                ("", "".to_string())
             };
+            let name_arg = if pascal_case(&name) != p.name().as_str() {
+                format!(", name = \"{}\"", p.name())
+            } else {
+                "".to_string()
+            };
+            let fn_attribute = format!("    #[zbus(property{property_args}{name_arg})]");
 
             writeln!(w)?;
             writeln!(w, "    /// {} property", p.name())?;
+            write!(w, "{comment}")?;
             if p.access().read() {
                 writeln!(w, "{fn_attribute}")?;
                 let output = to_rust_type(p.ty(), false, false);
--- a/zbus_xmlgen/tests/data/sample_object0.rs
+++ b/zbus_xmlgen/tests/data/sample_object0.rs
@@ -68,6 +68,10 @@ pub trait SampleInterface0 {
     #[zbus(signal)]
     fn signal_value(&self, value: zbus::zvariant::Value<'_>) -> zbus::Result<()>;
 
+    /// StateChanged signal
+    #[zbus(signal)]
+    fn state_changed(&self, state: u32, something_else: u32) -> zbus::Result<()>;
+
     /// Bar property
     #[zbus(property)]
     fn bar(&self) -> zbus::Result<u8>;
@@ -94,4 +98,9 @@ pub trait SampleInterface0 {
             std::collections::HashMap<String, zbus::zvariant::OwnedValue>,
         )>,
     >;
+
+    /// State property
+    // Note: change signal is shadowed by `StateChanged` signal.
+    #[zbus(property(emits_changed_signal = "false"))]
+    fn state(&self) -> zbus::Result<u32>;
 }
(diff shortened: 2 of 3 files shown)


zbus_xmlgen/tests/data/sample_object0.rs
```diff
@@ -94,4 +98,8 @@ pub trait SampleInterface0 {
             std::collections::HashMap<String, zbus::zvariant::OwnedValue>,
         )>,
     >;
+
+    /// State property
+    #[zbus(property(emits_changed_signal = "false"))]
```
zeenix, line 104: I don't think this is desirable. The name collision doesn't necessarily mean that there is a relationship between the signal and property. I mean, it most likely does but we have no way to be sure about it so let's not assume.

antonok-edm: From my perspective, `zbus_xmlgen` is a great bootstrapping tool for getting a best-effort initial configuration up and running. In that context, if there's a tradeoff between "may need tweaking in rare cases" versus "doesn't even compile", I'd strongly prefer the former. But, ultimately up to you :sweat_smile: 

Is there another solution you had in mind?

zeenix: I agree actually and I'm not suggesting that we don't solve this. 

The solution I had in mind was to simply change the name of the Rust method (using the `name` attribut to keep the D-Bus name remain the same) if it's going to conflict with a property method. However it's hard to come up with names 😂

So perhaps your solution isn't bad at all. We should add a non-doc comment above the property saying why the attribute was added though so users don't get confused. 

Once someone adds annotations (https://github.com/z-galaxy/zbus/issues/789) to xmlgen, they can control this from the command line.

antonok-edm: > We should add a non-doc comment above the property saying why the attribute was added though so users don't get confused.

Done! Feel free to tweak the format if you'd like

zbus_xmlgen/src/lib.rs
```diff
@@ -202,14 +216,30 @@ impl GenTrait<'_> {
...
-                format!("    #[zbus(property, name = \"{}\")]", p.name())
+
+            let prop_change_receiver_name = format!("{name}_changed");
+            let signal_collision = signal_method_names.get(&prop_change_receiver_name);
+
+            let property_args = if signal_collision.is_some() {
+                "(emits_changed_signal = \"false\")"
+            } else {
+                ""
+            };
```
zeenix: We could simplify the code by initializizing the comment here and then writting to `w` unconditionally:

```suggestion
            let (property_args, comment) = if signal_collision.is_some() {
                (
                    "(emits_changed_signal = \"false\")",
                    "    // Note: change signal is shadowed by imported `{signal_collision}` signal.",
            } else {
                ("", "")
            };
```

zbus_xmlgen/src/lib.rs
```diff
@@ -202,14 +216,30 @@ impl GenTrait<'_> {
...
+                "".to_string()
             };
+            let fn_attribute = format!("    #[zbus(property{property_args}{name_arg})]");
 
             writeln!(w)?;
             writeln!(w, "    /// {} property", p.name())?;
+            if let Some(signal_collision) = signal_collision {
+                writeln!(
+                    w,
+                    "    // Note: change signal is shadowed by imported `{signal_collision}` signal"
```
zeenix: This is ok but: 
* remove ` imported`. I'm not sure what it means and if I'm not sure, user is unlikely to either. :)
* missing `.` a the end. Actually most of the phrases and sentences you're adding here, are missing it.

zbus_xmlgen/src/lib.rs
antonok-edm: had to include the newline in the comment and switch it to `write!` instead of `writeln!`, but this is done

zbus_xmlgen/src/lib.rs
antonok-edm: I guess what I meant is that `xmlgen` read it from the interface XML file as opposed to somebody writing it by hand. But, sure, it doesn't add much value :sweat_smile: 

done both

zbus_xmlgen/src/lib.rs
```diff
@@ -185,6 +186,19 @@ impl GenTrait<'_> {
...
+        // receivers
+        let signal_method_names: HashMap<String, String> = signals
+            .iter()
+            .map(|s| {
+                (
+                    to_identifier(&to_snakecase(s.name().as_str())),
+                    s.name().as_str().to_owned(),
+                )
+            })
+            .collect();
```
zeenix: Didn't notice this before, sorry:

* this seems to be duplicating the iteration over signals just below
* unnecessarily allocating the signal name (value).

antonok-edm: Done, although this does also imply recalculating the snakecase version of the names again in the inner loop, so it might not strictly be an improvement... lmk what you think :sweat_smile:

zeenix: > Done

:+1: 

> although this does also imply recalculating the snakecase version of the names again in the inner loop

Isn't this because you swtich from HashMap to HashSet? (I didn't suggest that btw :wink:).

antonok-edm: you're right :sweat_smile: 

done!

zbus_xmlgen/src/lib.rs
```diff
@@ -185,6 +186,11 @@ impl GenTrait<'_> {
 
         let mut signals = iface.signals().to_vec();
         signals.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
+
+        // collect signal method names for collision detection with implicit property change
+        // receivers
```
zeenix, line 191: Missing `.` again. Sorry! I just like consistency too much. Although in these files I've allowed changes from contributors that missed `.` but it's inconsistent with rest of the code and even code here, so let's not add to that. 🙏 I'll add `.` to existing comments/docs in these files myself later.

zbus_xmlgen/src/lib.rs
```diff
@@ -196,20 +202,41 @@ impl GenTrait<'_> {
...
 
         let mut props = iface.properties().to_vec();
         props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
         for p in props {
             let name = to_identifier(&to_snakecase(p.name().as_str()));
-            let fn_attribute = if pascal_case(&name) != p.name().as_str() {
-                format!("    #[zbus(property, name = \"{}\")]", p.name())
+
+            let prop_change_receiver_name = format!("{name}_changed");
+            let signal_collision = changed_suffix_signals.get(&prop_change_receiver_name);
```
zeenix, line 217: I liked the previous naming better. The variable isn't holding a collision but rather the `Signal`. The name makes it sound like it's a boolean so makes code slightly harder to follow/read.

zbus_xmlgen/tests/data/sample_object0.rs
```diff
@@ -94,4 +98,9 @@ pub trait SampleInterface0 {
             std::collections::HashMap<String, zbus::zvariant::OwnedValue>,
         )>,
     >;
+
+    /// State property
+    // Note: change signal is shadowed by `StateChanged` signal.
```
zeenix, line 103: * `change signal` -> `changed signal`.
* I just realized btw: wouldn't it be better if this was instead:

```suggestion
    // Note: change signal is shadowed by `state_changed` signal method.
```

? That would also make the xmlgen part simpler and you can also go back to HashSet but this time w/o the caveat of reallocation. 😉

zbus_xmlgen/src/lib.rs
```diff
@@ -185,6 +186,11 @@ impl GenTrait<'_> {
 
         let mut signals = iface.signals().to_vec();
```
zeenix, line 187: btw, can't we just switch to be a HashMap instead? Wouldn't it make the code even simpler? 🤔

zeenix (changes requested):
Breaking my promise again by finding more nitpicks. 😆 Sorry! If you don't have the time and motivation anymore, just let me know and I can help out with remaining stuff.

zeenix:
@antonok-edm Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

PR #1691, opened 2026-01-20
🐛 zx: Wrap single struct returns in tuple to distinguish from multiple returns

### LINKED ISSUE #1241: zbus_xmlgen can't distinguish between "ss" and "(ss)"
consider the following two introspections.

`
<method name="test">
    <arg name="testmethod" type="(ss)" direction="out" />
</method>
`

`
<method name="test">
    <arg name="testmethod" type="ss" direction="out" />
</method>
`

zbus_xmlgen will generate the same rust code
fn test_method(&self) -> zbus::Result<(String, String)>;

I would expect the first result to be
fn test_method(&self) -> zbus::Result<((String, String),)>;

cachebag:
When a method has a single output argument that is a struct type (e.g., type="(ss)"), we now wrap it in a tuple to distinguish it from multiple return values. This ensures type="(ss)" generates Result<((String, String),)> while two separate string outputs generate Result<(String, String)>.

Fixes #1241

### DIFF
--- a/zbus_xmlgen/src/lib.rs
+++ b/zbus_xmlgen/src/lib.rs
@@ -259,7 +259,7 @@ fn hide_clippy_type_complexity_lint<W: Write>(
 
 fn inputs_output_from_args(args: &[Arg]) -> (String, String) {
     let mut inputs = vec!["&self".to_string()];
-    let mut output = vec![];
+    let mut output: Vec<OutputArg> = vec![];
     let mut n = 0;
     let mut gen_name = || {
         n += 1;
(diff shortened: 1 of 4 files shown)


zbus_xmlgen/src/lib.rs
```diff
@@ -279,18 +279,30 @@ fn inputs_output_from_args(args: &[Arg]) -> (String, String) {
...
 
-    let output = match output.len() {
-        0 => "()".to_string(),
-        1 => output[0].to_string(),
-        _ => format!("({})", output.join(", ")),
+    let output_str = match &output[..] {
+        [] => "()".to_string(),
+        [(ty, true)] => {
+            // If there's a single output argument that is a struct type, we need to wrap it
+            // in a tuple to distinguish it from multiple return values
```
zeenix: Clearer wording:

```suggestion
            // If there's a single output argument and it is a struct type, we need to wrap it
            // in a tuple to distinguish it from multiple return values
```

zbus_xmlgen/src/lib.rs
```diff
@@ -259,7 +259,7 @@ fn hide_clippy_type_complexity_lint<W: Write>(
 
 fn inputs_output_from_args(args: &[Arg]) -> (String, String) {
     let mut inputs = vec!["&self".to_string()];
-    let mut output = vec![];
+    let mut output: Vec<(String, bool)> = vec![];
```
zeenix: Can we create a named struct for this? It'd be more readable IMO.

zeenix (changes requested):
LGTM otherwise. 👍

zbus_xmlgen/src/lib.rs
```diff
@@ -257,9 +257,14 @@ fn hide_clippy_type_complexity_lint<W: Write>(
     Ok(())
 }
 
+struct OutputArg {
```
zeenix: * Remember, top-down code layout: use before definition.
* Please add a `#[derive(Debug)]` even if we don't currently need it.

zeenix:
> Last thing. :)

Oh and the prefixes on the PR and commit are different and the commit one is wrong. Please either use the the full name of the crate or abbreviation: `zbus_xmlgen` or `zx`.

zbus_xmlgen/src/lib.rs
cachebag: > Remember, top-down code layout: use before definition

Right, yes sorry about that 🤦🏻

zeenix: No need to be. It's not yet addressed btw. :wink:

cachebag: > No need to be. It's not yet addressed btw. 😉

Yep, just noticed that a minute ago

zeenix: and now you did the opposite. You moved it even further up. :laughing:

cachebag: > and now you did the opposite. You moved it even further up. 😆

I'm confused. Move it before `inputs_output_from_args` ?

zeenix: Use **before** definition, so no.

cachebag: Lol okay I'm sorry, it's early over here. That should do it.

PR #1692, opened 2026-01-20
CI: fail early in windows test

elmarco:


zbus/src/connection/socket/unix.rs
```diff
@@ -1,5 +1,6 @@
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
+#[allow(unused)]
```
zeenix: That's a workaround. I was looking into this yesterday and IMO the cleanest solution would be to remove this import and use the full path everywhere.

PR #1696, opened 2026-01-22
🧑‍💻 cargo: Set resolver to 3

A6GibKm:
This is the default since rust 2024. See https://doc.rust-lang.org/edition-guide/rust-2024/cargo-resolver.html for more details.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       RUSTFLAGS: -D warnings
-      MSRV: 1.85.0
+      MSRV: 1.87.0
     steps:
       - uses: actions/checkout@v6
       - uses: dtolnay/rust-toolchain@master
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -109,7 +109,7 @@ zbus/src/
 
 ## Development Guidelines
 
-- **MSRV**: 1.85.0
+- **MSRV**: 1.87.0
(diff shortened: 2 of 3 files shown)


A6GibKm:
`cargo semver-checks check-release` with version 0.45 (same as CI) passes locally. I will do some research.

EDIT: See https://github.com/obi1kenobi/cargo-semver-checks/issues/1553 from 2 hours ago.

Cargo.lock
```diff
@@ -790,12 +790,12 @@ checksum = "841d1cc9bed7f9236f321df977030373f4a4163ae1a7dbfe1a51a2c1a51d9100"
 
 [[package]]
 name = "heapless"
-version = "0.9.2"
+version = "0.9.1"
```
zeenix: What's the downgrade about?

A6GibKm: As per the commit message, .2 has a msrv greater than ours. The idea of bumping the resolver is that it will detect these issues.

zeenix: Ah ok, sorry I missed that. I'd prefer bumping the MSRV.

A6GibKm: Done.

zeenix: Reused the same emoji, eh? 😂 ⬆️ Is more appropriate. These emojis are used in our automated release process so it's kinds important.

zeenix: Reused the same emoji, eh? 😂 ⬆️ Is more appropriate. These emojis are used in our automated release process so it's kinds important.

A6GibKm: I copied it from https://github.com/z-galaxy/zbus/commit/4512e9413835d7b30fdc5dde1d49938d1bdb637e, but sure.

PR #1697, opened 2026-01-22
✨ zb: Add helper for IBus connection creation

### LINKED ISSUE #964: Helper for ibus connection creation

cachebag:
This PR implements `connection::Builder::ibus()` for connecting with IBus. The IBus address is fetched from the `ibus address` command, and is handled similarly to how we handle the `launchd` address on macOS. 

Fixes #964

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -92,7 +92,8 @@ jobs:
           sudo mkdir -p /run/user/$UID |
           sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
           sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
-          sudo apt-get install -y dbus
+          sudo apt-get update
+          sudo apt-get install -y dbus ibus
       - uses: dtolnay/rust-toolchain@master
         with:
           toolchain: stable
@@ -101,6 +102,9 @@ jobs:
       - name: Build and Test
         run: |
           dbus-run-session --config-file /tmp/dbus-session-abstract.conf -- cargo --locked test --release --verbose -- basic_connection
+          # Start IBus daemon for IBus test.
+          ibus-daemon -d --xim
+          sleep 2
           # All features except tokio.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
             cargo --locked test --release --verbose --features uuid,url,time,chrono,option-as-array,vsock,bus-impl \


zbus/src/address/transport/ibus.rs
```diff
@@ -0,0 +1,137 @@
+use crate::{Address, Result, process::run};
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+#[non_exhaustive]
```
zeenix: I dind't know you can use this attribute on a struct. Does it make it so that people can't assume the fields? 🤔

zbus/src/connection/builder.rs
```diff
@@ -98,6 +98,49 @@ impl<'a> Builder<'a> {
...
+    /// # Errors
+    ///
+    /// Returns an error if:
+    /// - The `ibus` command is not found or fails to execute
+    /// - The IBus daemon is not running
+    /// - The command output cannot be parsed as a valid D-Bus address
+    ///
+    /// # Example
+    ///
+    /// ```no_run
```
zeenix, line 119: I think we should run the test, if possible. We can make it linux-specific and install/launch ibus in the linux test job in the CI (or we can mock the server side too, if we want it to be reliable and runable anywhere).

zbus/src/address/transport/ibus.rs
cachebag: Yep and the user must use the provided constructor but thinking about this more, I think I will just remove it..it's a zero-sized unit struct with no fields, so there's nothing to be "exhaustive" about. I think I just instintively matched launchd's struct (which actually has fields)

zbus/src/connection/builder.rs
cachebag: How would we integrate this into CI? Just getting the deps for ibus sounds like a pain unless you're familiar with methods to do so

zeenix: > Just getting the deps for ibus sounds like a pain unless you're familiar with methods to do so

`apt-get install package-names..` like [we already do](https://github.com/z-galaxy/zbus/blob/main/.github/workflows/rust.yml#L95)? :wink:

cachebag: Whattt lol. I didn't even know you were running dbus in CI. I've been trying to get NetworkManager to run in my CI forever..though maybe they don't work the same since nm needs a network device..

zbus/src/abstractions/process.rs
```diff
@@ -63,7 +63,7 @@ impl Command {
 
     /// Executes the command as a child process, waiting for it to finish and
     /// collecting all of its output.
-    #[cfg(target_os = "macos")]
+    #[cfg(unix)]
```
zeenix, line 66: These internal changes could be split into a separate commit. If we had to revert the ibus changes later for any reason but by that time, these are used by other non-macos code, we'll want to keep these changes.

zbus/src/blocking/connection/builder.rs
```diff
@@ -35,6 +35,40 @@ impl<'a> Builder<'a> {
...
+    ///
+    /// ```no_run
+    /// # use std::error::Error;
+    /// # use zbus::blocking::connection;
+    /// #
+    /// let conn = connection::Builder::ibus()?
+    ///     .build()?;
+    ///
+    /// // Use the connection to interact with IBus services
+    /// # drop(conn);
```
zeenix: why the explicit drop? We're at the end of the block.

zbus/tests/basic.rs
```diff
@@ -363,3 +363,65 @@ async fn test_freedesktop_credentials() -> Result<()> {
...
+    // Prepend temp directory to PATH so our mock ibus is found first
+    let original_path = env::var("PATH").unwrap_or_default();
+    let new_path = format!("{}:{}", temp_dir.display(), original_path);
+    unsafe {
+        env::set_var("PATH", &new_path);
+    }
+
+    let result = std::panic::catch_unwind(|| {
+        block_on(test_ibus_connection()).unwrap();
+    });
```
zeenix: Best to just have the whole thing async. `tokio` is a dev-dep so you can just use the same IO API from it as you're using from the std.

zbus/tests/basic.rs
```diff
@@ -363,3 +363,65 @@ async fn test_freedesktop_credentials() -> Result<()> {
...
+
+    // Restore PATH and remove temp directory
```
zeenix: all comments missing a `.`. 😆

zbus/Cargo.toml
```diff
@@ -98,10 +98,11 @@ uds_windows.workspace = true
 [target.'cfg(unix)'.dependencies]
 rustix.workspace = true
 libc.workspace = true
-
-[target.'cfg(any(target_os = "macos", windows))'.dependencies]
 async-recursion.workspace = true
 
+[target.'cfg(windows)'.dependencies.async-recursion]
+workspace = true
+
```
zeenix, line 105: This could be split into a separate commit too.

zbus/tests/basic.rs
```diff
@@ -363,3 +363,60 @@ async fn test_freedesktop_credentials() -> Result<()> {
 
     Ok(())
 }
+
+#[cfg(unix)]
+#[tokio::test]
+#[timeout(15000)]
+async fn ibus_connection() {
+    use std::{env, fs, os::unix::fs::PermissionsExt};
```
zeenix: `fs` is what I was expecting you'd use from tokio.

zbus/tests/basic.rs
```diff
@@ -363,3 +363,60 @@ async fn test_freedesktop_credentials() -> Result<()> {
...
+    fs::set_permissions(&mock_ibus, perms).unwrap();
+
+    // Prepend temp directory to PATH so our mock ibus is found first.
+    let original_path = env::var("PATH").unwrap_or_default();
+    let new_path = format!("{}:{}", temp_dir.display(), original_path);
+    unsafe {
+        env::set_var("PATH", &new_path);
+    }
+
+    let result = test_ibus_connection().await;
```
zeenix, line 416: you know, it would be even better if we first try this w/o setting up the mock service and if it fails (if we can check for the specific error, even better) we then launch the mock service and try again?

.github/workflows/rust.yml
```diff
@@ -92,7 +92,8 @@ jobs:
           sudo mkdir -p /run/user/$UID |
           sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
           sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
-          sudo apt-get install -y dbus
+          sudo apt-get update
+          sudo apt-get install -y dbus ibus
```
zeenix, line 96: hmm.. but none of the tests that are run actually need it? 🤔 If you act on my suggestion about only running mock service when needed, this can then be useful though.

zeenix:
> BTW, one of your commit messages is breaking lines too early. 50 chars suggestion is only for the title/shortlog. :)

Not yet addressed btw. :)

zbus/tests/basic.rs
```diff
@@ -363,3 +363,69 @@ async fn test_freedesktop_credentials() -> Result<()> {
...
+#[tokio::test]
+#[timeout(15000)]
+async fn ibus_connection() {
+    use std::env;
+    use tokio::fs;
+
+    // First try with real IBus if available.
+    let result = test_ibus_connection().await;
+
+    if result.is_ok() {
```
zeenix: could you check for a specific error? 🤔

.github/workflows/rust.yml
```diff
@@ -101,16 +102,23 @@ jobs:
...
+            
+            # All features except tokio.
             cargo --locked test --release --verbose --features uuid,url,time,chrono,option-as-array,vsock,bus-impl \
               -- --skip fdpass_systemd
-          # Test tokio support.
-          dbus-run-session --config-file /tmp/dbus-session.conf -- \
+            # Test tokio support.
             cargo --locked test --release --verbose --tests -p zbus --no-default-features \
               --features tokio-vsock -- --skip fdpass_systemd
-          dbus-run-session --config-file /tmp/dbus-session.conf -- \
```
zeenix, line 112: how come these lines are changing? 🤔

.github/workflows/rust.yml
```diff
@@ -92,7 +92,8 @@ jobs:
           sudo mkdir -p /run/user/$UID |
           sed -e s/UID/$UID/ -e s/PATH/path/ CI/dbus-session.conf > /tmp/dbus-session.conf
           sed -e s/UID/$UID/ -e s/PATH/abstract/ CI/dbus-session.conf > /tmp/dbus-session-abstract.conf
-          sudo apt-get install -y dbus
+          sudo apt-get update
```
zeenix: why is this needed now? 🤔

cachebag: https://github.com/cachebag/zbus/actions/runs/21538835286

```log
Get:46 http://azure.archive.ubuntu.com/ubuntu noble-updates/main amd64 libgtk2.0-bin amd64 2.24.33-4ubuntu1.1 [7954 B]
E: Failed to fetch https://security.ubuntu.com/ubuntu/pool/main/m/mesa/libegl-mesa0_25.0.7-0ubuntu0.24.04.2_amd64.deb  404  Not Found [IP: 20.106.104.242 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
Fetched 30.2 MB in 3s (10.2 MB/s)
```

🤷🏻‍♂️ apparently `ibus` includes packages that require a fresh index?

bilelmoussaoui:
Here is the helper I am using to get the ibus address without relying on the ibus address command 
```

async fn get_ibus_address() -> Result<String> {
    if let Ok(addr) = std::env::var("IBUS_ADDRESS") {
        return Ok(addr);
    }

    let machine_id = tokio::fs::read_to_string("/etc/machine-id")
        .await?
        .trim()
        .to_string();

    let display_part = if let Ok(wayland_display) = std::env::var("WAYLAND_DISPLAY") {
        format!("unix-{}", wayland_display)
    } else if let Ok(display) = std::env::var("DISPLAY") {
        let display_num = display
            .trim_start_matches(':')
            .split('.')
            .next()
            .ok_or_else(|| anyhow::anyhow!("Invalid DISPLAY format"))?;
        format!("unix-{}", display_num)
    } else {
        return Err(anyhow::anyhow!(
            "Neither WAYLAND_DISPLAY nor DISPLAY environment variable is set"
        ));
    };

    let config_home = std::env::var("XDG_CONFIG_HOME")
        .unwrap_or_else(|_| format!("{}/.config", std::env::var("HOME").expect("HOME not set")));

    let addr_file_path = format!("{}/ibus/bus/{}-{}", config_home, machine_id, display_part);

    let file = tokio::fs::File::open(&addr_file_path)
        .await
        .map_err(|e| anyhow::anyhow!("Failed to open {}: {}", addr_file_path, e))?;

    let reader = tokio::io::BufReader::new(file);
    let mut lines = reader.lines();
    while let Some(line) = lines.next_line().await? {
        if let Some(addr) = line.trim().strip_prefix("IBUS_ADDRESS=") {
            return Ok(addr.to_string());
        }
    }

    Err(anyhow::anyhow!(
        "IBUS_ADDRESS not found in {}",
        addr_file_path
    ))
}
```

PR #1706, opened 2026-02-09
🐛 zv: Fix GVariant bool encoding

kawadakk:
The boolean type is encoded as a single byte in GVariant, whereas it is encoded as 4 bytes with 4-byte alignment in D-Bus.

```console
$ python
>>> GLib.Variant.new_boolean(False).get_data_as_bytes().get_data()
b'\x00'
>>> GLib.Variant.new_boolean(True).get_data_as_bytes().get_data()
b'\x01'
```

### DIFF
--- a/zvariant/tests/number/bool_value.rs
+++ b/zvariant/tests/number/bool_value.rs
@@ -1,0 +1,13 @@
+use zvariant::LE;
+
+#[test]
+fn bool_value() {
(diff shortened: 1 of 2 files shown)


zeenix (approved):
Such a great first contribution! I don't have any room for any nitpicks at all. 😆 Thanks.

PR #1721, opened 2026-02-19
🐛 Do not use SendFlags::NOSIGNAL on Redox

jackpot51:
For now, this is not available on Redox. If it is implemented and exposed in rustix in the future, I will make another PR to revert this change.


jackpot51:
Thanks @zeenix! It would be helpful to have a new crates.io release as the current one does not build on Redox

zeenix:
> Thanks @zeenix! 

You're welcome.

> It would be helpful to have a new crates.io release as the current one does not build on Redox

Done.

PR #1743, opened 2026-03-12
🩹 zbus: Fix error name for fallible setters

### LINKED ISSUE #992: zb: `interface` property setters should return `fdo::Error`
The book explains how to write fallible methods and property getters in interfaces at https://dbus2.github.io/zbus/service.html#method-errors.

However, it seems that this is not actually the full picture: Fallible property setters only compile when returning `zbus::Error`, whereas `zbus::fdo::Error` produces very opaque error messages like the following:

```rust
#[zbus::interface(name = "org.mpris.MediaPlayer2.Player")]
impl MprisPlayer {
    // ...

    #[zbus(property(emits_changed_signal = "true"))]
    async fn volume(&self) -> Volume {
        self.volume
    }

    #[zbus(property)]
    async fn set_volume(&self, _value: Volume) -> zbus::fdo::Result<()> {
        Err(zbus::fdo::Error::NotSupported("...".to_owned()))
    }
}

```

```
error[E0271]: expected `{async block@src/mpris_event_handler.rs:309:1: 309:59}` to be a future that resolves to `Result<(), Error>`, but it resolves to `Result<(), Error>`
   --> src/mpris_event_handler.rs:309:1
    |
309 | #[zbus::interface(name = "org.mpris.MediaPlayer2.Player")]
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `zbus::Error`, found `zbus::fdo::Error`
    |
    = note: expected enum `std::result::Result<_, zbus::Error>`
               found enum `std::result::Result<_, zbus::fdo::Error>`
    = note: required for the cast from `Pin<Box<{async block@src/mpris_event_handler.rs:309:1: 309:59}>>` to `Pin<Box<dyn futures_util::Future<Output = std::result::Result<(), zbus::Error>> + std::marker::Send>>`
    = note: this error originates in the attribute macro `zbus::interface` (in Nightly builds, run with -Z macro-backtrace for more info)
```

which does not point to the code span that causes the error.

It would be very convenient (and, in my opinion, more consistent) if method setters could also return `zbus::fdo::Error`. In any case, it would be great if the documentation could be updated to mention the expected return type, and if the `interface` examples could be extended to contain a fallible property setter. Maybe it would also be possible to enhance the proc macro such that it yields errors that point to the offending code?
Thanks!

This is `zbus = 4.4.0`.

jbaublitz:
Currently zbus returns an error name referencing zbus when a setter fails to set a property. This commit allows collapsing nesting structures like:

zbus::fdo::Error::ZBus(zbus::Error::FDO(zbus::fdo::Error))

into a zbus::fdo::Error. This works around the problem where DispatchResult expects a zbus::Error for all futures that it wraps, resulting in this nesting when the zbus::Error is converted to an FDO error.

The change provides a flatten method to collapse these errors into a final error that is a non-nested FDO error with the appropriate name on the D-Bus. This method is used in all places where the DispatchResult::Async variant is unwound.

Closes #992

### DIFF
--- a/zbus/src/fdo/error.rs
+++ b/zbus/src/fdo/error.rs
@@ -169,6 +169,21 @@ pub enum Error {
     NotContainer(String),
 }
 
+impl Error {
+    /// Collapse redundant nesting of errors.
+    ///
+    /// Collapses a nested data structure like `zbus::fdo::Error::ZBus(zbus::Error::FDO(_))`
+    /// into a `zbus::fdo::Error`. `zbus::fdo::Error::ZBus(zbus::error::Failure(s))` will
+    /// also be converted to `zbus::fdo::Error::Failed(s)`.
+    pub fn flatten(self) -> Self {
+        match self {
+            Error::ZBus(zbus::Error::FDO(e)) => *e,
+            Error::ZBus(zbus::Error::Failure(s)) => Error::Failed(s),
+            e => e,
+        }
+    }
+}
+
 /// Alias for a `Result` with the error type [`zbus::fdo::Error`].
 ///
 /// [`zbus::fdo::Error`]: enum.Error.html
--- a/zbus/src/fdo/properties.rs
+++ b/zbus/src/fdo/properties.rs
@@ -4,6 +4,7 @@
 //! be useful across various D-Bus applications. This module provides their proxy.
 
 use std::{borrow::Cow, collections::HashMap};
+use zbus::fdo;
 use zbus_names::InterfaceName;
 use zvariant::{OwnedValue, Value};
 
@@ -64,7 +65,7 @@ impl Properties {
         #[zbus(connection)] connection: &Connection,
         #[zbus(header)] header: Header<'_>,
         #[zbus(signal_emitter)] emitter: SignalEmitter<'_>,
-    ) -> Result<()> {
+    ) -> fdo::Result<()> {
         let path = header.path().ok_or(crate::Error::MissingField)?;
         let root = server.root().read().await;
         let iface = root
@@ -89,7 +90,7 @@ impl Properties {
                 )));
             }
             zbus::object_server::DispatchResult::Async(f) => {
-                return f.await.map_err(Into::into);
+                return f.await.map_err(|e| fdo::Error::ZBus(e).flatten());
             }
         }
         let res = iface
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -313,10 +313,7 @@ impl ObjectServer {
                 )));
             }
             DispatchResult::Async(f) => {
-                return f.await.map_err(|e| match e {
-                    Error::FDO(e) => *e,
-                    e => fdo::Error::Failed(format!("{e}")),
-                });
+                return f.await.map_err(|e| fdo::Error::ZBus(e).flatten());
             }
             DispatchResult::RequiresMut => {}
         }
@@ -328,10 +325,7 @@ impl ObjectServer {
             DispatchResult::NotFound => {}
             DispatchResult::RequiresMut => {}
             DispatchResult::Async(f) => {
-                return f.await.map_err(|e| match e {
-                    Error::FDO(e) => *e,
-                    e => fdo::Error::Failed(format!("{e}")),
-                });
+                return f.await.map_err(|e| fdo::Error::ZBus(e).flatten());
             }
         }
         drop(write_lock);
--- a/zbus/tests/iface_and_proxy/iface.rs
+++ b/zbus/tests/iface_and_proxy/iface.rs
@@ -1,4 +1,7 @@
-use std::collections::HashMap;
+use std::{
+    collections::HashMap,
+    sync::atomic::{AtomicU32, Ordering},
+};
 use tokio::sync::mpsc::Sender;
 use tracing::{debug, instrument};
 use zbus::{
@@ -21,6 +24,7 @@ pub struct MyIface {
     emits_changed_const: u32,
     emits_changed_false: u32,
     r#let: u32,
+    fallible_setter_prop: AtomicU32,
 }
 
 impl MyIface {
@@ -34,6 +38,7 @@ impl MyIface {
             emits_changed_const: 0,
             emits_changed_false: 0,
             r#let: 0,
+            fallible_setter_prop: AtomicU32::new(0),
         }
     }
 }
@@ -482,6 +487,26 @@ impl MyIface {
     fn set_let(&mut self, val: u32) -> zbus::fdo::Result<()> {
         debug!("`Let` setter called.");
         self.r#let = val;
+        Ok(())
+    }
+
+    #[instrument]
+    #[zbus(property)]
+    fn fallible_setter_prop(&self) -> u32 {
+        debug!("`fallible_setter_prop` getter called.");
+        self.fallible_setter_prop.load(Ordering::SeqCst)
+    }
+
+    #[instrument]
+    #[zbus(property)]
+    async fn set_fallible_setter_prop(&self, val: u32) -> zbus::Result<()> {
+        debug!("`fallible_setter_prop` setter called.");
+        if val > 60 {
+            return Err(zbus::Error::Failure(format!(
+                "Provided value is {val}; values above 60 not accepted"
+            )));
+        }
+        self.fallible_setter_prop.store(val, Ordering::SeqCst);
         Ok(())
     }
 
--- a/zbus_macros/src/error.rs
+++ b/zbus_macros/src/error.rs
@@ -146,13 +146,28 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
             quote! {
                 impl ::std::convert::From<#zbus::Error> for #name {
                     fn from(value: #zbus::Error) -> #name {
-                        if let #zbus::Error::MethodError(name, desc, _) = &value {
-                            match name.as_str() {
-                                #error_converts
-                                _ => Self::#ident(value),
+                        match &value {
+                            #zbus::Error::MethodError(name, desc, _) => {
+                                match name.as_str() {
+                                    #error_converts
+                                    _ => Self::#ident(value),
+                                }
                             }
-                        } else {
-                            Self::#ident(value)
+                            #zbus::Error::FDO(e) => {
+                                // Handle FDO errors by extracting name and description
+                                // and converting similarly to MethodError
+                                let e_ref = ::std::convert::AsRef::as_ref(e);
+                                let name = #zbus::DBusError::name(e_ref);
+                                let desc = &::std::option::Option::map(
+                                    #zbus::DBusError::description(e_ref),
+                                    ::std::string::ToString::to_string,
+                                );
+                                match name.as_str() {
+                                    #error_converts
+                                    _ => Self::#ident(value),
+                                }
+                            }
+                            _ => Self::#ident(value),
                         }
                     }
                 }
--- a/zbus/tests/iface_and_proxy/client.rs
+++ b/zbus/tests/iface_and_proxy/client.rs
@@ -3,7 +3,7 @@ use futures_util::{StreamExt, TryStreamExt};
 use std::convert::TryInto;
 use tracing::{debug, instrument};
 use zbus::{
-    Connection, Error, Message, MessageStream,
+    Connection, DBusError, Error, Message, MessageStream,
     fdo::{ObjectManagerProxy, PropertiesProxy},
     message,
     proxy::CacheProperties,
@@ -108,6 +108,14 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
             .await
             .is_err()
     );
+
+    proxy.set_fallible_setter_prop(60).await?;
+    if let Err(e) = proxy.set_fallible_setter_prop(61).await {
+        assert_eq!(
+            zbus::fdo::Error::from(e).name(),
+            "org.freedesktop.DBus.Error.Failed"
+        )
+    }
 
     let props_proxy = PropertiesProxy::builder(&conn)
         .destination("org.freedesktop.MyService")?


zbus/src/object_server/mod.rs
```diff
@@ -328,10 +325,7 @@ impl ObjectServer {
             DispatchResult::NotFound => {}
             DispatchResult::RequiresMut => {}
             DispatchResult::Async(f) => {
-                return f.await.map_err(|e| match e {
-                    Error::FDO(e) => *e,
-                    e => fdo::Error::Failed(format!("{e}")),
-                });
+                return f.await.map_err(|e| fdo::Error::ZBus(e).flatten());
```
zeenix, line 328: Why do we need to wrap into fdo::Error to flatten immediately after? 🤔 The 2 operations seem like the complete opposite.

zbus/src/fdo/error.rs
```diff
@@ -169,6 +169,18 @@ pub enum Error {
     NotContainer(String),
 }
 
+impl Error {
+    /// Collapses a nested data structure like `zbus::fdo::Error::ZBus(zbus::Error::FDO(_))` into
+    /// a `zbus::fdo::Error`.
+    pub fn flatten(self) -> Self {
+        match self {
+            Error::ZBus(zbus::Error::FDO(e)) => *e,
+            Error::ZBus(zbus::Error::Failure(s)) => Error::Failed(s),
```
zeenix, line 181: Should be mentioned in the docs.

zbus/src/fdo/error.rs
```diff
@@ -169,6 +169,18 @@ pub enum Error {
     NotContainer(String),
 }
 
+impl Error {
+    /// Collapses a nested data structure like `zbus::fdo::Error::ZBus(zbus::Error::FDO(_))` into
+    /// a `zbus::fdo::Error`.
```
zeenix: I like to keep docs in the same format as commits (which incidentally is also the Rust docs convention): short summary line and explanation/more details in the paragraph(s) below it.

zbus/src/fdo/error.rs
```diff
@@ -169,6 +169,18 @@ pub enum Error {
...
+    pub fn flatten(self) -> Self {
```
zeenix, line 178: I think the addition of this method is very much a separate logical change since it can be useful in other places you're using here. Please split into a separate commit that comes before the main one. 🙏

zbus/src/fdo/properties.rs
```diff
@@ -64,7 +65,7 @@ impl Properties {
         #[zbus(connection)] connection: &Connection,
         #[zbus(header)] header: Header<'_>,
         #[zbus(signal_emitter)] emitter: SignalEmitter<'_>,
-    ) -> Result<()> {
+    ) -> ::std::result::Result<(), fdo::Error> {
```
zeenix: Hmm... Looked like an obvious API break at first but then I realised it's a dbus method. 😅

zeenix (changes requested):
Thanks so much. LGTM in general and I appreciate how you did a lot of work to figure out a very simple solution for the problem. Apart from the inline comments, please try your best to pick a more specific emoji. 🙏 Otherwise, we end up with mostly 1-2 emojis on most commits and it beats the whole point.

I'm very surprised that no test needed to be updated for this change. We'll want a test for this so we don't regress in this regard in the future. Could you please add one? Perhaps best to add a new property to the e2e test in `zbus/tests/e2e.rs` and test it on the client-side of it.

zbus/src/object_server/mod.rs
jbaublitz: We need to wrap it because the awaited error is of type `zbus::Error`. Unless we define flatten on `zbus::Error` and have that return a `zbus::fdo::Error`, it needs to be structured this way.

zbus/src/fdo/properties.rs
jbaublitz: Yeah this should be okay based on what I've seen.

zbus/src/object_server/mod.rs
zeenix: I don't mind having `flatten` in both error types but:

* `zbus::Error::flatten` returning an fdo::Error would be very strange API.
* I wonder why we can't construct exactly what we need here directly. Wrapping a value only to unwrap it immediately, feels very wrong. We don't have any obligation to use `flatten` here.

jbaublitz: So do you mean you would like to have the match statement in each of these `.map_err` implementations? Or something else? Can you clarify exactly what you're looking for here?

zeenix: I didn't have a exact suggestion in mind. It can be a match or anything appropriate. You've to admit that wrapping to then just unwrap immediately after, seems very strange, doesn't it?

jbaublitz: I'm willing to do what you would like me to do but I don't necessarily agree that this is strange. Because the input type for flatten is an fdo::Error and the output type is an fdo::Error, I wouldn't describe this as unwrapping it. I wrote it this way because it avoids duplicating the logic in multiple places which I believe we'll have to do if we avoid using flatten in these areas. I'm happy to do whatever you'd prefer even if I don't necessarily agree so just let me know what you'd prefer.

zbus/src/fdo/error.rs
jbaublitz: Okay this should be resolved.

jbaublitz:
@zeenix I just have one question about the commit pattern. Should the new test be included in the second commit or should it be a separate commit?

zbus/src/object_server/mod.rs
zeenix: > I don't necessarily agree that this is strange. Because the input type for flatten is an fdo::Error and the output type is an fdo::Error,

The flatten method itself is not the strange thing here. The strange part is wrapping to immediatelly unwrap (it happening through `flatten` method is not relevant). It's a bit like like `Option::Some(x).unwrap()`. Whether strange or not, it is a bit confusing at least.

> I'm happy to do whatever you'd prefer even if I don't necessarily agree so just let me know what you'd prefer.

I thought I already did. :) It's not really that big a deal though. I just hope that people don't get confused with this.

zeenix:
@jbaublitz could you please address the remaing comments? :pray:

zeenix:
> @zeenix I just have one question about the commit pattern. Should the new test be included in the second commit or should it be a separate commit?

Sorry I had missed this. I prefer atomic commits so separate commit is better but not necessary.

jbaublitz:
@zeenix I apologize for the delay. I have added a test and believe I have resolved all of your concerns.

zbus/tests/iface_and_proxy/iface.rs
```diff
@@ -485,6 +490,26 @@ impl MyIface {
         Ok(())
     }
 
+    #[instrument]
+    #[zbus(property)]
+    fn atomic(&self) -> u32 {
```
zeenix: The "atomic" part is an internal detail so not the best name for the method. I'd name it `fallible_setter_prop`.

zbus_macros/src/error.rs
```diff
@@ -146,13 +146,24 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
...
+                                    _ => Self::#ident(value),
+                                }
                             }
-                        } else {
-                            Self::#ident(value)
+                            #zbus::Error::FDO(e) => {
+                                // Handle FDO errors by extracting name and description
+                                // and converting similarly to MethodError
+                                let name = #zbus::DBusError::name(e.as_ref());
+                                let desc = &#zbus::DBusError::description(e.as_ref()).map(|s| s.to_string());
```
zeenix: why allocate/clone (`to_string`) if we only need a ref?

zbus/src/fdo/properties.rs
zeenix: there is `fdo::Result` we can use here.

zeenix (changes requested):

zbus_macros/src/error.rs
jbaublitz: It doesn't look like we only need a ref from what I can see. This is the error I get repeated for a large number of variants when I remove the allocation:

```
   --> zbus/src/fdo/error.rs:4:24                                                                                                            
    |                                                                                                                                        
  4 | #[derive(Clone, Debug, DBusError, PartialEq)]                                                                                          
    |                        ^^^^^^^^^                                                                                                       
    |                        |                                                                                                               
    |                        expected `String`, found `&str`                                                                                 
    |                        arguments to this enum variant are incorrect                                                                    
    |                                                                                                                                        
note: tuple variant defined here                                                                                                             
   --> zbus/src/fdo/error.rs:169:5                                                                                                           
    |                                                                                                                                        
169 |     NotContainer(String),                                                                                                              
    |     ^^^^^^^^^^^^                                                                                                                       
    = note: this error originates in the derive macro `DBusError` (in Nightly builds, run with -Z macro-backtrace for more info)             
                                                                                   
```

jbaublitz:
@zeenix Okay i believe I've resolved all of the issues!

zbus_macros/src/error.rs
zeenix: Right, the `&` at the left-most part is for the `Option`. That is very strange that something in the code is expecting `&Option<String>`. I'm guessing you won't have the time to investigate so I'll put it on my TODO. :)

This reminds me though that there is a good reason you'd find that in most of the macros' generated code we (and other projects too) prefer to use the methods with their paths instead of through the `.` sugar (e.g `::std::string::ToString::to_string`)  Can we do the same here so there is no confusion? Btw, you can use the method as the closure so this can be:


```suggestion
                                let desc = &::std::option::Option::map(
                                    #zbus::DBusError::description(e.as_ref()),
                                    ::std::string::ToString::to_string,
                                );
```

Also do the same for `e.as_ref()` and save it into a variable as it's being done for both lines.

zeenix (changes requested):
Nearly there.

jbaublitz:
@zeenix Okay I believe this is ready.

zeenix:
@jbaublitz In case you're not keeping an eye on your matrix notifications, I'm copy&pasting my messages to you there:

> John Baublitz: Hi there. Sorry I've been not able to give the proper time for your PR for weeks now, one of which I was on vacation.
> John Baublitz: tbh with you, I am also not super convinced about this whole flatten approach. It seems to work, yes but it all feels a bit hacky to me. I've been wondering about the alternative approach of solving this through DispatchResult I mentioned before, taking advantage of Interface as being declared unstable.
> I will try to cook up a PR and we can discuss there but I'm hoping that your main concern is to solve the issue at hand and won't hate me if we end up not merging your PR in the end after all your work. 🙈
> of course I do very much appreciate your efforts on this, in any case

PR #1771 is the alternative. Please let me know if you disagree and I'm happy to discuss which one would be the best approach.

jbaublitz:
@zeenix My only goal here is addressing the behavior change at the D-Bus level so that we can appropriately match on errors on the client side. If the linked PR addresses that, I have no preference on which goes in. I'll let you know what the results are on our side tomorrow.

jbaublitz:
@zeenix I pointed our codebase at your PR and I'm still getting the zbus-named error with your PR from dbus-monitor on fallible asynchronous setters with an immutable receiver:
```
error time=1776706850.552211 sender=:1.1493 -> destination=:1.1502 error_name=org.freedesktop.zbus.Error reply_serial=7
   string "Requested size or size_limit of filesystem testfs is 1 sectors which is less than minimum required: 1048576 sectors"
```

This is how I've set up the zbus dependency in the Cargo.toml file:
```
[dependencies.zbus]
version = "5.12.0"
optional = true
features = ["tokio"]
git = "https://github.com/zeenix/zbus"
branch = "property-setter-fdo-error"
```

Am I missing any set up? If not, I believe that your PR does not resolve the problem.

zeenix:
> @zeenix I pointed our codebase at your PR and I'm still getting the zbus-named error with your PR from dbus-monitor on fallible asynchronous setters with an immutable receiver:

Thanks for checking.

I added a test case (in a separate commit at the end). Could you please see if that is not testing the right thing?

> asynchronous setters with an immutable receiver:

I'm curios, does declaring the receiver mutable and/or the method non-async help here? :thinking: Perhaps it is me who missed something here. :)

jbaublitz:
> > @zeenix I pointed our codebase at your PR and I'm still getting the zbus-named error with your PR from dbus-monitor on fallible asynchronous setters with an immutable receiver:
> 
> Thanks for checking.
> 
> I added a test case (in a separate commit at the end). Could you please see if that is not testing the right thing?

It is not testing exactly what we're looking to test here. Specifically we are looking to test the name of the error as reported by dbus-monitor. This specifically affects how languages like Python match the error and originally caused a stack trace in our code when the unexpected error name popped up. Testing error type does not necessarily mean that the error name will be correct. This is why I took the flatten approach. Based on how name is determined in the macros, this solved the problem by removing the nested `zbus::Error` which generates the incorrect name.

> 
> > asynchronous setters with an immutable receiver:
> 
> I'm curios, does declaring the receiver mutable and/or the method non-async help here? 🤔 Perhaps it is me who missed something here. :)

I'm not sure I would say it helps as we specifically want an immutable &self receiver and an async setter that returns an unexpected error name. However I believe this problem doesn't appear anywhere else except in this very specific case as I believe it is the only place that returns the `DispatchResult::Async` variant. I believe other combinations will return other `DispatchResult` variants based on my testing and looking at the code.

zeenix:
> It is not testing exactly what we're looking to test here. Specifically we are looking to test the name of the error as reported by dbus-monitor. This specifically affects how languages like Python match the error and originally caused a stack trace in our code when the unexpected error name popped up. Testing error type does not necessarily mean that the error name will be correct. This is why I took the flatten approach. Based on how name is determined in the macros, this solved the problem by removing the nested `zbus::Error` which generates the incorrect name.

Wait, but why is there any nesting involved? The main motivation of #992 was always to allow returning an `fdo::Result` **directly** from your interface methods. Does it work if you do that?

zeenix:
> Wait, but why is there any nesting involved? The main motivation of #992 was always to allow returning an `fdo::Result` **directly** from your interface methods. Does it work if you do that?

Right, [your test case](https://github.com/z-galaxy/zbus/pull/1743/changes#diff-a381a1529da8ed0780048f2d739414d763b8569201846b6a8ca2d3b0e035d402R502) is indeed testing with `zbus::Result` wrapping. Try returning `fdo::Result` directly and see if that helps. I think it should.

jbaublitz:
Closing in favor of #1771

PR #1760, opened 2026-04-09
✨ zb: Add Builder::build_message_stream

zeenix:
Add `connection::Builder::build_message_stream` (and the blocking
counterpart `build_message_iterator`) for callers that must not lose
messages arriving immediately after the connection is built — most
notably bus implementations handling `Hello` from just-authenticated
clients that pipeline traffic.

The existing `build()` + `MessageStream::from(&conn)` path has a race
window: between `build()` returning and a receiver being activated on
the unfiltered broadcast channel, the socket reader task may run,
broadcast incoming messages, and find no active receivers — silently
dropping them. This has been observed in practice under tokio's LIFO
steal scheduling.

The fix is to activate a receiver on the inner broadcast channel
before the socket reader task is spawned. `build_()` now takes an
`activate_msg_stream` flag and returns the activated receiver alongside
the `Connection`; `build_message_stream` wraps it via the existing
`MessageStream::for_subscription_channel` helper. The `build()` fast
path is unchanged.

A regression test covers both the async and blocking APIs using
`Channel::pair()`, verifying that a message sent before the stream is
polled is still delivered.

### DIFF
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -264,4 +264,23 @@ impl<'a> Builder<'a> {
     pub fn build(self) -> Result<Connection> {
         block_on(self.0.build()).map(Into::into)
     }
+
+    /// Build the connection and return a [`MessageIterator`] to receive messages from it.
+    ///
+    /// This is the blocking counterpart of [`crate::connection::Builder::build_message_stream`].
+    /// The iterator is set up **before** the socket-reader task is started, so no messages can
+    /// be lost in the window between the connection being built and the iterator being created.
+    /// Use this when the peer may pipeline traffic right after authentication — e.g. a bus
+    /// implementation reading a `Hello` method call from a just-connected client.
+    ///
+    /// To get the [`Connection`] out of the returned iterator, use `Connection::from(&iter)`.
+    ///
+    /// This method is only available when the `bus-impl` feature is enabled.
+    ///
+    /// [`MessageIterator`]: crate::blocking::MessageIterator
+    #[cfg(feature = "bus-impl")]
+    pub fn build_message_iterator(self) -> Result<crate::blocking::MessageIterator> {
+        block_on(self.0.build_message_stream())
+            .map(|azync| crate::blocking::MessageIterator { azync: Some(azync) })
+    }
 }
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -1,3 +1,4 @@
+use async_broadcast::Receiver as ActiveReceiver;
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
 use enumflags2::BitFlags;
@@ -23,10 +24,13 @@ use vsock::VsockStream;
 
 use zvariant::ObjectPath;
 
+#[cfg(feature = "bus-impl")]
+use crate::MessageStream;
 use crate::{
     Connection, Error, Executor, Guid, OwnedGuid, Result,
     address::{self, Address},
     fdo::RequestNameFlags,
+    message::Message,
     names::{InterfaceName, WellKnownName},
     object_server::{ArcInterface, Interface},
 };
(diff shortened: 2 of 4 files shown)


zbus/src/blocking/connection/builder.rs
```diff
@@ -264,4 +264,21 @@ impl<'a> Builder<'a> {
     pub fn build(self) -> Result<Connection> {
         block_on(self.0.build()).map(Into::into)
     }
+
+    /// Build the connection and return a [`MessageIterator`] to receive messages from it.
+    ///
+    /// This is the blocking counterpart of [`crate::connection::Builder::build_message_stream`].
+    /// The underlying broadcast receiver is activated **before** the socket-reader task is
+    /// started, so no messages can be dropped in the race window between the connection being
```
zeenix: That's too much internal details. The "activation" part isn't even important here. From the type, it's obvious and implicit that it's an active receiver (which is the default), we're getting. We shouldn't make it sound like somthing is being "activated" in these docs and in the commit log/PR description.

zbus/src/blocking/connection/builder.rs
```diff
@@ -264,4 +264,21 @@ impl<'a> Builder<'a> {
...
+    /// built and the iterator being created. Use this when the peer may pipeline traffic right
+    /// after authentication — e.g. a bus implementation reading `Hello` from a just-connected
```
zeenix: "reading a `Hello` method".

zbus/src/connection/builder.rs
```diff
@@ -436,15 +438,76 @@ impl<'a> Builder<'a> {
...
+    ///
+    /// To get the [`Connection`] out of the returned stream, use `Connection::from(&stream)` —
+    /// this is cheap (an `Arc` clone).
+    ///
+    /// As with any [`MessageStream`], the caller **must poll it continuously** or the socket
+    /// reader will eventually block once the broadcast queue fills.
+    ///
+    /// # Example
+    ///
+    /// ```no_run
```
zeenix: would be great if this was actually run as a regression test.

zbus/src/connection/builder.rs
```diff
@@ -436,15 +438,76 @@ impl<'a> Builder<'a> {
...
+        let executor = Executor::new();
+        #[cfg(not(feature = "tokio"))]
+        let internal_executor = self.internal_executor;
+        // Box the future as it's large and can cause stack overflow.
+        let (conn, msg_receiver) =
+            Box::pin(executor.run(self.build_(executor.clone(), true))).await?;
+
+        #[cfg(not(feature = "tokio"))]
+        start_internal_executor(&executor, internal_executor)?;
+
```
zeenix, line 523: Why are we duplicating the `build` logic? We should try to make `build_` do most of the work (or create another internal helper method that both public build methods use.

zbus/src/connection/builder.rs
```diff
@@ -436,15 +438,76 @@ impl<'a> Builder<'a> {
...
+    /// let _conn: Connection = (&stream).into();
+    ///
+    /// // Poll the stream for incoming messages..
+    /// # drop(stream);
+    /// # Ok::<(), zbus::Error>(())
+    /// # }).unwrap();
+    /// #
+    /// # Ok::<_, Box<dyn Error + Send + Sync>>(())
+    /// ```
+    pub async fn build_message_stream(self) -> Result<MessageStream> {
```
zeenix, line 499: since this is mainly needed by bus impls, probably best to gate it behind `bus-impl` feature and document this fact.

zbus/src/connection/builder.rs
```diff
@@ -432,19 +436,89 @@ impl<'a> Builder<'a> {
...
+    ///         .build()
+    ///         .await
+    ///         .unwrap();
+    ///
+    ///     let stream = Builder::authenticated_socket(c2, guid)
+    ///         .unwrap()
+    ///         .p2p()
+    ///         .build_message_stream()
+    ///         .await
+    ///         .unwrap();
```
zeenix: This is not a good example. The stream is not being used at all afterwards so doesn't really demonstrate the use case.

zbus/src/connection/builder.rs
```diff
@@ -432,19 +436,89 @@ impl<'a> Builder<'a> {
...
+    /// that the stream is set up **before** the socket-reader task is started. No messages can
+    /// therefore be lost in the window between `build()` returning and `MessageStream::from`
+    /// being called. Use this when the peer may pipeline traffic right after authentication —
+    /// e.g. a bus implementation reading a `Hello` method call from a just-connected client.
+    ///
+    /// To get the [`Connection`] out of the returned stream, use `Connection::from(&stream)` —
+    /// this is cheap (an `Arc` clone).
+    ///
+    /// As with any [`MessageStream`], the caller **must poll it continuously** or the socket
+    /// reader will eventually block once the broadcast queue fills.
```
zeenix: no need for this paragraph.

zbus/src/connection/builder.rs
```diff
@@ -432,19 +436,89 @@ impl<'a> Builder<'a> {
...
+    ///
+    ///     // The `Connection` is still accessible from the stream.
+    ///     let _conn: Connection = (&stream).into();
+    /// });
+    /// ```
+    #[cfg(feature = "bus-impl")]
+    pub async fn build_message_stream(self) -> Result<MessageStream> {
+        let (conn, msg_receiver) = self.build_inner(true).await?;
+        let msg_receiver = msg_receiver.expect("build_inner(true) always returns Some");
+        Ok(MessageStream::for_subscription_channel(
```
zeenix, line 503: empty line above this for readability.

zbus/tests/builder_message_stream.rs
```diff
@@ -0,0 +1,116 @@
...
+        // The message must still be there — with the fix, the receiver was activated inside
+        // `build_message_stream` before the reader task was spawned, so the broadcast goes
+        // into an active channel and is queued for us.
+        let received = futures_lite::future::or(async { stream_b.next().await }, async {
+            futures_lite::future::yield_now().await;
+            // Arbitrarily long timeout through a yield loop so this works under both
+            // async-io and tokio without pulling in runtime-specific timers.
+            for _ in 0..10_000 {
+                futures_lite::future::yield_now().await;
+            }
```
zeenix: tokio is dev-dep and this test (or API bieng tested) is not runtime-specific so we should make use of that instead of relying on busy looping and other flaky mechanisms.

zbus/src/connection/builder.rs
```diff
@@ -432,19 +436,89 @@ impl<'a> Builder<'a> {
...
+    async fn build_inner(
+        self,
+        activate_msg_stream: bool,
+    ) -> Result<(Connection, Option<ActiveReceiver<Result<Message>>>)> {
         let executor = Executor::new();
         #[cfg(not(feature = "tokio"))]
         let internal_executor = self.internal_executor;
         // Box the future as it's large and can cause stack overflow.
-        let conn = Box::pin(executor.run(self.build_(executor.clone()))).await?;
+        let result =
```
zeenix: why rename `conn`?

zbus/src/connection/builder.rs
```diff
@@ -432,19 +436,99 @@ impl<'a> Builder<'a> {
...
+    ///
+    /// ```
+    /// use futures_util::StreamExt;
+    /// use zbus::{
+    ///     Connection, Guid, block_on,
+    ///     connection::{Builder, socket::Channel},
+    ///     message::Message,
+    /// };
+    ///
+    /// block_on(async {
```
zeenix: not sure why we moved to block_on that's visible to the doc reader. The unimportant/boilerplate parts should be hidden by default in the generated docs (this includes the imports).

zbus/tests/builder_message_stream.rs
```diff
@@ -0,0 +1,101 @@
...
+        // Run both concurrently so the SASL handshake completes cooperatively.
+        futures_util::join!(server, client);
+    });
+}
+
+/// Smoke-test for the blocking wrapper using the same busd scenario.
+#[cfg(feature = "blocking-api")]
+#[test]
+#[timeout(15000)]
+fn blocking_build_message_iterator_does_not_drop_pipelined_hello() {
```
zeenix: We don't need a blocking version of the test. Underneath it all works the same way.

zbus/tests/builder_message_stream.rs
```diff
@@ -0,0 +1,101 @@
...
+            let conn = Connection::from(&stream);
+            let reply = Message::method_return(&hello.header())
+                .unwrap()
+                .build(&(":1.1",))
+                .unwrap();
+            conn.send(&reply).await.unwrap();
+        };
+
+        // Client: a bus client whose build() pipelines Hello during SASL auth.
+        let client = async {
```
zeenix: why wrap this in async? The build method gives us a future already.

zbus/tests/builder_message_stream.rs
```diff
@@ -0,0 +1,101 @@
...
+                .unwrap();
+            assert_eq!(hello.header().member().unwrap().as_str(), "Hello");
+
+            // Reply so the client's build() can complete.
+            let conn = Connection::from(&stream);
+            let reply = Message::method_return(&hello.header())
+                .unwrap()
+                .build(&(":1.1",))
+                .unwrap();
+            conn.send(&reply).await.unwrap();
```
zeenix: * why so low-level? why not use `Connection::reply`?
* `stream` is no longer used after `conn` is created so `from` should consume it.

PR #1783, opened 2026-05-04
Limit "vsock" feature to Linux-only

elmarco:
Currently vsock feature is only available on Linux.

### DIFF
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -129,7 +129,24 @@ tempfile.workspace = true
 criterion.workspace = true
 
 [package.metadata.docs.rs]
-all-features = true
+# we can't use:
+# all-features = true
+# because some features (such as vsock) are target-specific
+features = [
+    "async-io",
+    "blocking-api",
+    "uuid",
+    "url",
+    "time",
+    "chrono",
+    "heapless",
+    "option-as-array",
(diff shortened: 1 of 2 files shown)


zbus/Cargo.toml
```diff
@@ -129,7 +131,21 @@ tempfile.workspace = true
 criterion.workspace = true
 
 [package.metadata.docs.rs]
-all-features = true
+features = [
```
zeenix, line 137: This is still needed even though with vsock now linux-only? If so, we should have a comment here, that says why we're not simply using `all-features = true`.

elmarco: yes, or you get:
error: The "vsock" and "tokio-vsock" features are only supported on Linux.

when `cargo doc --all-features`.

ok, I added a comment

zeenix (approved):
Thanks. BTW, I have add this to the general section my `~/.claude/CLAUDE.md`:

```
* DO NOT add yourself as author in commit messages. Add `Assisted-by: <model details>` lines instead.
```

As per the the Linux kernel and gstreamer coding practices. Just for the future.

zeenix:
@elmarco #1784 fixes the cargo-fuzz failure in the CI. Please rebase once that's merged.

PR #1789, opened 2026-05-08
🔊 zb: log GetAll error during property cache init

### LINKED ISSUE #1325: subscribe_to_property_change doesn't propagate error from GetAll properties
take the following example client
` 
let connection = zbus::Connection::session().await?;
let proxy = zbus::Proxy::new( // <- miss behaving daemon
    &connection,
    "com.example.rs",
    "/example_rs/reference",
    "com.example.referenceA2",
).await?;

let mut stream: PropertyStream<'_, zvariant::Value> = proxy.receive_property_changed("SantasAddress").await;
while let Some(signal) = stream.next().await {
    println!("Received signal: {:?}", signal.get().await);
}
`

I was testing out some test dbus interfaces and discovered that when you have a daemon that exposes a few properties 4 in my case but 1 of these causes a dbus error when calling Get on that property. The above code will not show an error or do really anything at all. I was able to track down the issue using `dbus-monitor` where I noticed zbus will do a GetAll call which silently fails and causes the above code to stall forever without any error or notice.

mvanhorn:
Closes #1325.

`PropertiesCache::init()` stashes GetAll failures in `caching_result` but never logs them, so a misbehaving daemon causes property streams to silently produce nothing (#1325).

Per @zeenix's suggestion in the issue thread, emit a `tracing::error!` in the spawned caching task before stashing the error so the failure shows up under `RUST_LOG`.

### DIFF
--- a/zbus/src/proxy/mod.rs
+++ b/zbus/src/proxy/mod.rs
@@ -13,7 +13,7 @@ use std::{
     sync::{Arc, OnceLock, RwLock, RwLockReadGuard},
     task::{Context, Poll},
 };
-use tracing::{Instrument, debug, info_span, instrument, trace};
+use tracing::{Instrument, debug, info_span, instrument, trace, warn};
 
 use zbus_names::{BusName, InterfaceName, MemberName, UniqueName};
 use zvariant::{ObjectPath, OwnedValue, Str, Value};
@@ -298,6 +298,10 @@ impl PropertiesCache {
                         (prop_changes, interface, uncached_properties)
                     }
                     Err(e) => {
+                        warn!(
+                            "Failed to populate properties cache via GetAll: {e}. \
+                             Property change streams will not produce values."
+                        );
                         ready.notify(usize::MAX);
                         *caching_result = CachingResult::Cached { result: Err(e) };
 


zeenix:
@mvanhorn Dude, I'm fine with the use AI and use it myself but please read the PR description/commit message before sumitting so you can remove unnecessary noise like "I don't have access to a session". Also you really should test changes on your end before submitting the PR.

zeenix (approved):

zeenix (changes requested):
Also please choose the most appropriate emoji. AI is pretty bad at this. Please have a look at the logs and changes and test before submitting. 🙏

mvanhorn:
> @mvanhorn Dude, I'm fine with the use AI and use it myself but please read the PR description/commit message before sumitting so you can remove unnecessary noise like "I don't have access to a session". Also you really should test changes on your end before submitting the PR.

hey I flagged it ;)

mvanhorn:
Sorry for the noise, @zeenix. Updated the title to 🥅 (matching the catch-errors precedent in this repo; 🐛 was wrong since the existing behavior isn't a bug, just unobservable) and dropped the 'no session bus available' paragraph from the body. I won't push another change to this PR without running the integration test locally first.

zeenix:
> Sorry for the noise, @zeenix. Updated the title to 🥅 (matching the catch-errors precedent in this repo; 🐛 was wrong since the existing behavior isn't a bug, just unobservable) and dropped the 'no session bus available' paragraph from the body. I won't push another change to this PR without running the integration test locally first.

Thanks so much. Still a few thing: 

* The error **was** being caught so that's not the most appropriate emoji either. There's a very specific one for log addition, please use that. 

* The description here is still way too verbose and obviously an LLM summary for you. There's no need to have the actual code in the description, especially for a super trivial change like this. Please ensure you read and commit message guidelines on this. 🙏

zeenix:
Btw you didn't update the commit message. Please change that too, not just the PR title/description.

mvanhorn:
🔊 (precedent: `4da53255 🔊 zb: add some logging to handshake code`), description trimmed, and amended the commit message to match. d6feb382 — diff is identical to the version you approved, only metadata changed.

zbus/src/proxy/mod.rs
```diff
@@ -298,6 +298,10 @@ impl PropertiesCache {
                         (prop_changes, interface, uncached_properties)
                     }
                     Err(e) => {
+                        error!(
```
zeenix: Oh sorry, I didn't notice this before but this should be a `warn!`. Error is for (usually) non-recoverable critical errors. Please update the commit message accordingly as well.

zeenix (changes requested):
Much better, thanks! One little adjustment and then we can merge.

mvanhorn:
163ea1b: `warn!` applied at zbus/src/proxy/mod.rs:301, and commit subject updated to match. Verified locally before pushing: cargo check + clippy clean; cargo test -p zbus --lib shows zero pass/fail diff vs unchanged HEAD on macOS (the failures are pre-existing session-bus tests on this host). CI linux_test is the authoritative surface for this path.

zeenix:
@mvanhorn Pretty sure the last comment itself also came from an AI but whatever. :)

zeenix (approved):

PR #1790, opened 2026-05-08
🦺 zv: Add `debug_assert` and docs for `Fd` deserialization safety

cachebag:
`BorrowedFd::borrow_raw(-1)` is undefined behavior per std's contract. Both `Fd::deserialize` and `ValueSeed::visit_i32` (for `Signature::Fd`) were passing the raw wire integer directly to `borrow_raw` without checking it, meaning any deserializer other than zvariant's own D-Bus one (which resolves the wire index through the message's FD list before calling `visit_i32`) could trigger UB in safe code by supplying `-1`.

This PR adds a `debug_assert!(raw >= 0)` at both call sites, and documents on `Fd` that the `Deserialize` impl is designed for the D-Bus path and using it with other formats is unsupported.

### DIFF
--- a/zvariant/src/fd.rs
+++ b/zvariant/src/fd.rs
@@ -9,6 +9,18 @@ use crate::{Basic, Type};
 /// do not implement  [`Serialize`] and [`Deserialize`]. So we provide a
 /// wrapper for both that implements these traits.
 ///
+/// # D-Bus FD Encoding
+///
+/// In D-Bus, file descriptors are **not** encoded as their raw numeric values on the wire.
+/// Instead, the message body contains a `u32` **index** into an out-of-band array of file
+/// descriptors that is transferred alongside the message via `SCM_RIGHTS` on Unix sockets.
+///
+/// The [`Deserialize`] implementation on this type is designed for zvariant's D-Bus
+/// deserializer, which resolves that index through the message's FD list before calling the
+/// serde visitor. **Using this type with other deserializers (JSON, CBOR, etc.) is not
+/// supported** — you would receive a `BorrowedFd` pointing to whatever FD number happened to
+/// be in the payload, which is almost certainly not what you want.
+///
 /// [`Serialize`]: https://docs.serde.rs/serde/trait.Serialize.html
 /// [`Deserialize`]: https://docs.serde.rs/serde/de/trait.Deserialize.html
 #[derive(Debug)]
--- a/zvariant/src/fd.rs
+++ b/zvariant/src/fd.rs
@@ -136,6 +136,7 @@ impl<'de> Deserialize<'de> for Fd<'de> {
         D: Deserializer<'de>,
     {
         let raw = i32::deserialize(deserializer)?;
+        debug_assert!(raw >= 0);
         // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
         let fd = unsafe { BorrowedFd::borrow_raw(raw) };
 
--- a/zvariant/src/value.rs
+++ b/zvariant/src/value.rs
@@ -830,6 +830,7 @@ where
         let v = match &self.signature {
             #[cfg(unix)]
             Signature::Fd => {
+                debug_assert!(value >= 0);
                 // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
                 let fd = unsafe { std::os::fd::BorrowedFd::borrow_raw(value) };
                 Fd::Borrowed(fd).into()


zvariant/src/fd.rs
```diff
@@ -124,7 +124,19 @@ impl<'de> Deserialize<'de> for Fd<'de> {
         D: Deserializer<'de>,
     {
         let raw = i32::deserialize(deserializer)?;
-        // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
+        // `BorrowedFd::borrow_raw` is undefined behavior if the raw value is `-1`, so we must
+        // refuse it here. We can't validate that the FD is otherwise actually open, but the
```
zeenix: The FD is encoded in a special way in D-Bus format: the actual FD is not present in the encoding but rather the index to it in vector that's transfered in a out-of-bound way on the socket. So you can't encode/decode FDs in other formats in a meaningful way anyway. Varlink (which uses JSON as its encoding) does the same and you'll not have -1 as the value in there either.

So I really appreciate your effort here but I'm not entirely sure this is needed.

zeenix: @cachebag woa! How did you receive, read and act on my comment so fast? :open_mouth:

cachebag: I get that- I think the concern is most about `Fd: Deserialize` as a generic serde impl. Any deserializer other than zvariant's own D-Bus one (i.e. JSON, CBOR, a custom transport,  Varlink/JSON) calls straight into `Fd::deserialize` with the raw wire integer. There's no interception of `deserialize_i32` to look up an FD list in that case, so -1 goes directly to `BorrowedFd::borrow_raw`, which is UB by std's contract.

> woa! How did you receive, read and act on my comment so fast?

Lol, there wasn't much to act on- I understand if you don't see the change as desired so I closed. 

Reading it though, IMO it is a purely defensive check that I think is cheap. Since `Fd: Deserialize` is public and callable by any serde deserializer, it should not be possible to produce UB through it regardless of the callers format.

zeenix: > > woa! How did you receive, read and act on my comment so fast?
> 
> Lol, there wasn't much to act on- I understand if you don't see the change as desired so I closed.

If it was a simple shut-and-close case, I would have closed the PR myself. :laughing: I could always change my mind, you know.

> Since `Fd: Deserialize` is public and callable by any serde deserializer, it should not be possible to produce UB through it regardless of the callers format.

It's alsmost certain that if anyone uses this with another format, they're asking for trouble. It's not only -1 that's bad, any number that is not a valid and open FD, would lead to UB. So if you use it with any other deserializer, you'll end up with UB in any case unless you take some steps to ensure that the FD is valid, and if you do all that, you could do the check yourself in that code.

Having said that, what we could do:

* Add `debug_assert!(fd >= 0);` in the right places. That way it's super cheap in dev and inexistant in release mode.
* Warn about the special encoding of FDs in the `Fd` docs.

How about that?

cachebag: I like that. Makes sense to me!

zvariant/src/fd.rs
```diff
@@ -124,7 +136,19 @@ impl<'de> Deserialize<'de> for Fd<'de> {
         D: Deserializer<'de>,
     {
         let raw = i32::deserialize(deserializer)?;
-        // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
+        // This impl is intended for zvariant's D-Bus deserializer, which resolves the wire
+        // index through the message's FD list before calling the visitor—so `raw` should
+        // always be a valid, open FD. Using other deserializers is unsupported (see type docs).
+        //
+        // We add a debug assertion for `-1` specifically because `BorrowedFd::borrow_raw(-1)`
```
zeenix: It's not only -1 that's problematic, any negative number is AFAIK. Besides we're checking for positive integers. I actually don't think we need to have this comment. This assertion is a sanity check and will only trigger for anyone that trying to use this zvariant-specific type with another deserializers and just lucky enogh to have a negative number and get this assertion failure to stop them from shooting themselves in the foot.

zvariant/src/fd.rs
```diff
@@ -124,7 +136,19 @@ impl<'de> Deserialize<'de> for Fd<'de> {
...
+        // is instant UB per std's contract. Other invalid values (closed FDs, etc.) are also
+        // wrong but don't have the same "guaranteed UB" status.
+        debug_assert!(
+            raw >= 0,
+            "Fd::deserialize received -1, which is UB for BorrowedFd"
```
zeenix: Also not needed IMO. Just a one liner is enough.

zeenix (changes requested):
atomic commits please and then you can also use the appropriat emoji. :) Also, don't forget to update the PR description.

zvariant/src/fd.rs
```diff
@@ -136,7 +136,8 @@ impl<'de> Deserialize<'de> for Fd<'de> {
         D: Deserializer<'de>,
     {
         let raw = i32::deserialize(deserializer)?;
-        // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
```
zeenix, line 127: The safety comment was conveying why this is safe. Nothing to do with encoding I think.

zvariant/src/fd.rs
```diff
@@ -243,3 +244,14 @@ impl std::fmt::Display for OwnedFd {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn deserialize_valid_raw_value_is_accepted() {
+        let fd = serde_json::from_str::<Fd<'_>>("0").expect("0 is a permitted raw fd value");
+        assert_eq!(fd.as_raw_fd(), 0);
+    }
+}
```
zeenix: We don't need a test case here. We're just adding a `debug_assert` to catch the user violating the assumption they really shouldn't in dev. We don't **have to** do it.

zvariant/src/value.rs
```diff
@@ -830,7 +830,8 @@ where
         let v = match &self.signature {
             #[cfg(unix)]
             Signature::Fd => {
-                // SAFETY: The `'de` lifetimes will ensure the borrow won't outlive the raw FD.
```
zeenix, line 833: same comment about this comment.

zeenix (changes requested):
Almost there. :) Also the last commit is not fixing a bug, so 🦺seems more appropriate here. While you do that, please also keep the commit short log at 50 chars since it's possible here:

```
🦺 zv: Add 2 debug asserts

Add debug_assert for non-negative raw FD in deserialize paths.
```

cachebag:
Sorry for the careless edits. Please let me know if you'd like anything further adjusted. @zeenix

zeenix (approved):

PR #1792, opened 2026-05-11
✨ zb: Support multiple auth mechanisms on server side

directhex:
The D-Bus specification allows servers to support multiple SASL authentication mechanisms on a single socket, listing them all in the `REJECTED` response so the client can retry with an alternative. Until now, zbus only supported configuring a single mechanism per server.

This adds `Builder::auth_mechanisms()` to accept a list of mechanisms, while the existing `auth_mechanism()` continues to work unchanged.

### DIFF
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -126,6 +126,13 @@ impl<'a> Builder<'a> {
         Self(self.0.auth_mechanism(auth_mechanism))
     }
 
+    /// Specify multiple mechanisms to accept during authentication.
+    ///
+    /// See [`crate::connection::Builder::auth_mechanisms`] for details.
+    pub fn auth_mechanisms(self, auth_mechanisms: impl Into<Vec<AuthMechanism>>) -> Self {
+        Self(self.0.auth_mechanisms(auth_mechanisms))
+    }
+
     /// Specify the user id during authentication.
     ///
     /// This can be useful when using [`AuthMechanism::External`] with `socat`
--- a/zbus/src/connection/builder.rs
+++ b/zbus/src/connection/builder.rs
@@ -83,7 +83,7 @@ pub struct Builder<'a> {
     internal_executor: bool,
     interfaces: Interfaces<'a>,
     names: HashSet<WellKnownName<'a>>,
-    auth_mechanism: Option<AuthMechanism>,
+    auth_mechanisms: Option<Vec<AuthMechanism>>,
     #[cfg(feature = "bus-impl")]
     unique_name: Option<crate::names::UniqueName<'a>>,
     request_name_flags: BitFlags<RequestNameFlags>,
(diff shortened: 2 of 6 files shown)


zeenix:
Thanks for this contribution but only supporting one mechanism is very much on purpose. We only support 2, `EXTERNAL` and `ANONYMOUS`. If you use the latter, the former becomes completely useless because the latter then already allows anyone to connect w/o any auth. So it only makes sense to use one of the mechanisms and hence why this was removed on purpose.

directhex:
An ANONYMOUS-only server will only talk to ANONYMOUS-only clients, meaning you need to either offer two sockets, or coordinate a breaking change where every client needs to be patched at the same time. Offering both mechanisms on the same socket is the only graceful way to handle back-compat where clients and servers might not be built from the exact same commit and deployed at the exact same time.

zeenix:
> An ANONYMOUS-only server will only talk to ANONYMOUS-only clients

Are you certain? IIRC the client on receiving a rejection, can try another method in the same handshake process. I can check this later but if you beat me to it, even better. :)

directhex:
> > An ANONYMOUS-only server will only talk to ANONYMOUS-only clients
> 
> Are you certain? IIRC the client on receiving a rejection, can try another method in the same handshake process. I can check this later but if you beat me to it, even better. :)

Let's say you have a server, and enable `ANONYMOUS` only. Your client hasn't been updated, and still uses `EXTERNAL`.

1. Client sends `AUTH EXTERNAL <uid>`
2. Server doesn't support `EXTERNAL` → replies `REJECTED ANONYMOUS`
3. Client could retry with `AUTH ANONYMOUS` — but only if it's programmed to fall back, which it can't be because zbus doesn't support multiple client mechanisms.

Enabling the server to offer both mechanisms means both `ANONYMOUS`-only and `EXTERNAL`-only clients can connect (let's say in my case, where a design originally used `EXTERNAL` but it turns out that's an enormous pain in the ass for containers so for the container case only, I want to use `ANONYMOUS` instead).

This is all standard practice in the spec on FDO.

zeenix:
> Let's say you have a server, and enable `ANONYMOUS` only. Your client hasn't been updated, and still uses `EXTERNAL`.
> 
> 1. Client sends `AUTH EXTERNAL <uid>`
> 2. Server doesn't support `EXTERNAL` → replies `REJECTED ANONYMOUS`
> 3. Client could retry with `AUTH ANONYMOUS` — but only if it's programmed to fall back, which it can't be because zbus doesn't support multiple client mechanisms.

Right but if your'e building both sides with zbus, you can develop them so that auth is configurable on both sides and you switch it for both. With other clients, the above should work just fine.

zeenix:
> Right but if your'e building both sides with zbus, you can develop them so that auth is configurable on both sides and you switch it for both. With other clients, the above should work just fine.

Having said that, the change is minimal so I'll review the changes soon and merge when ready.

zeenix:
> Having said that, the change is minimal

Looking at the code, I realized that the changes are mimimal because while you're adding API that claims to work for both client and server, you're only really implementing it for the server side.

Given that anonymous should not be used on unix socket and on tcp, you can only use anonymous, I really don't think I want this complexity in zbus for a corner case.

PR #1798, opened 2026-05-22
🧵 zb: Use async versions of UnixStream when possible

### LINKED ISSUE #1796: Zbus still spawns a second thread even with internal executor disabled
Not sure if this is a bug but there seems to be another thread being spawned besides de internal executor. Enabling the tokio feature or setting internal executor to false does not get rid of it. I have not found anywhere in the documentation any mention of zbus spawning another thread besides de internal executor. 

```Rust
#[tokio::main(flavor = "current_thread")]
async fn main() -> Result<(), Box<dyn Error>> {
    let conn = zbus::Connection::session().await?;
    let mut sigint = signal(SignalKind::interrupt())?;
    sigint.recv().await;
    Ok(())
}
```

Tokio disabled:

Image

Tokio enabled:

Image

So where is that second thread coming from? Is there a way to run just a single thread? It's not great having to use another thread for my usecase since it's fairly simple and I don't want to use extra resources I don't need.

germaniuss:
When connecting to a `UnixSocket` a blocking/threaded API is being used. This commit changes this to use `tokio::net::UnixStream` and `Async<std::os::unix::net::UnixStream>` when possible, preventing the use of extra threads. The windows version continues using the blocking/threaded API.

The connect_addr API is dropped since neither tokio nor async-io supports it. Instead, we use the connect functions in both environments, since they allow for abstract names when the path is prepended with '\0'.

Move the connection code to unix.rs for better future maintainability.

Closes #1796

### DIFF
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -12,13 +12,13 @@ use async_io::Async;
 use std::collections::HashMap;
 #[cfg(not(feature = "tokio"))]
 use std::net::TcpStream;
-#[cfg(unix)]
-use std::os::unix::net::{SocketAddr, UnixStream};
+#[cfg(all(unix, not(feature = "tokio")))]
+use std::os::unix::net::UnixStream;
 #[cfg(feature = "tokio")]
 use tokio::net::TcpStream;
 #[cfg(feature = "tokio-vsock")]
 use tokio_vsock::VsockStream;
-#[cfg(windows)]
+#[cfg(all(windows, not(feature = "tokio")))]
 use uds_windows::UnixStream;
 #[cfg(all(feature = "vsock", not(feature = "tokio")))]
 use vsock::VsockStream;
@@ -55,8 +55,6 @@ pub use ibus::Ibus;
 #[path = "vsock.rs"]
 // Gotta rename to avoid name conflict with the `vsock` crate.
 mod vsock_transport;
-#[cfg(target_os = "linux")]
-use std::os::linux::net::SocketAddrExt;
 #[cfg(any(
     all(feature = "vsock", not(feature = "tokio")),
     feature = "tokio-vsock"
@@ -102,59 +100,11 @@ impl Transport {
     #[cfg_attr(any(unix, windows), async_recursion::async_recursion)]
     pub(super) async fn connect(self) -> Result<Stream> {
         match self {
-            Transport::Unix(unix) => {
-                // This is a `path` in case of Windows until uds_windows provides the needed API:
-                // https://github.com/haraldh/rust_uds_windows/issues/14
-                let addr = match unix.take_path() {
-                    #[cfg(unix)]
-                    UnixSocket::File(path) => SocketAddr::from_pathname(path)?,
-                    #[cfg(windows)]
-                    UnixSocket::File(path) => path,
-                    #[cfg(target_os = "linux")]
-                    UnixSocket::Abstract(name) => {
-                        SocketAddr::from_abstract_name(name.as_encoded_bytes())?
-                    }
-                    UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
-                        // You can't connect to a unix:dir.
-                        return Err(Error::Unsupported);
-                    }
-                };
-                let stream = crate::Task::spawn_blocking(
-                    move || -> Result<_> {
-                        #[cfg(unix)]
-                        let stream = UnixStream::connect_addr(&addr)?;
-                        #[cfg(windows)]
-                        let stream = UnixStream::connect(addr)?;
-                        stream.set_nonblocking(true)?;
+            #[cfg(all(not(unix), feature = "tokio"))]
+            Transport::Unix(_) => Err(Error::Unsupported),
+            #[cfg(any(unix, not(feature = "tokio")))]
+            Transport::Unix(unix) => unix.connect().await.map(Stream::Unix),
 
-                        Ok(stream)
-                    },
-                    "unix stream connection",
-                )
-                .await??;
-                #[cfg(not(feature = "tokio"))]
-                {
-                    Async::new(stream)
-                        .map(Stream::Unix)
-                        .map_err(|e| Error::InputOutput(e.into()))
-                }
-
-                #[cfg(feature = "tokio")]
-                {
-                    #[cfg(unix)]
-                    {
-                        tokio::net::UnixStream::from_std(stream)
-                            .map(Stream::Unix)
-                            .map_err(|e| Error::InputOutput(e.into()))
-                    }
-
-                    #[cfg(not(unix))]
-                    {
-                        let _ = stream;
-                        Err(Error::Unsupported)
-                    }
-                }
-            }
             #[cfg(unix)]
             Transport::Unixexec(unixexec) => unixexec.connect().await.map(Stream::Unixexec),
             #[cfg(all(feature = "vsock", not(feature = "tokio")))]
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -115,7 +115,7 @@ impl Transport {
                         SocketAddr::from_abstract_name(name.as_encoded_bytes())?
                     }
                     UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
-                        // you can't connect to a unix:dir
+                        // You can't connect to a unix:dir.
                         return Err(Error::Unsupported);
                     }
                 };
--- a/zbus/src/address/transport/unix.rs
+++ b/zbus/src/address/transport/unix.rs
@@ -15,10 +15,8 @@ use super::encode_percents;
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
 
-#[cfg(target_os = "linux")]
-use std::os::linux::net::SocketAddrExt;
-#[cfg(unix)]
-use std::os::unix::net::{SocketAddr, UnixStream};
+#[cfg(all(unix, not(feature = "tokio")))]
+use std::os::unix::net::UnixStream;
 #[cfg(all(windows, not(feature = "tokio")))]
 use uds_windows::UnixStream;
 
@@ -71,28 +69,29 @@ impl Unix {
 
     #[cfg(not(feature = "tokio"))]
     pub(super) async fn connect(self) -> Result<Async<UnixStream>> {
-        let stream = self.get_stream().await?;
-        Async::new(stream).map_err(|e| Error::InputOutput(e.into()))
+        let addr = self.take_addr()?;
+
+        #[cfg(unix)]
+        let stream = Async::<UnixStream>::connect(addr).await;
+
+        #[cfg(not(unix))]
+        let stream = Async::new(self.get_stream(addr).await?);
+
+        stream.map_err(|e| Error::InputOutput(e.into()))
     }
 
     #[cfg(all(unix, feature = "tokio"))]
     pub(super) async fn connect(self) -> Result<tokio::net::UnixStream> {
-        let stream = self.get_stream().await?;
-        tokio::net::UnixStream::from_std(stream).map_err(|e| Error::InputOutput(e.into()))
+        let addr = self.take_addr()?;
+        tokio::net::UnixStream::connect(addr)
+            .await
+            .map_err(|e| Error::InputOutput(e.into()))
     }
 
-    #[cfg(any(unix, not(feature = "tokio")))]
-    async fn get_stream(self) -> Result<UnixStream> {
-        #[cfg(unix)]
-        let addr = self.take_socket_addr()?;
-        #[cfg(windows)]
-        let addr = self.take_path_addr()?;
-
+    #[cfg(not(unix))]
+    async fn get_stream(self, addr: PathBuf) -> Result<UnixStream> {
         crate::Task::spawn_blocking(
             move || -> Result<_> {
-                #[cfg(unix)]
-                let stream = UnixStream::connect_addr(&addr)?;
-                #[cfg(windows)]
                 let stream = UnixStream::connect(addr)?;
                 stream.set_nonblocking(true)?;
 
@@ -103,26 +102,20 @@ impl Unix {
         .await?
     }
 
-    #[cfg(all(windows, not(feature = "tokio")))]
-    fn take_path_addr(self) -> Result<PathBuf> {
-        // This is a `path` in case of Windows until uds_windows provides the needed API:
-        // https://github.com/haraldh/rust_uds_windows/issues/14
+    #[cfg(any(unix, not(feature = "tokio")))]
+    fn take_addr(self) -> Result<PathBuf> {
+        // This is a `path` because neither uds_windows, tokio, nor async_io provide
+        // the SocketAddrExt functions.
         match self.take_path() {
             UnixSocket::File(path) => Ok(path),
-            UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
-                // You can't connect to a unix:dir.
-                Err(Error::Unsupported)
-            }
-        }
-    }
-
-    #[cfg(unix)]
-    fn take_socket_addr(self) -> Result<SocketAddr> {
-        match self.take_path() {
-            UnixSocket::File(path) => Ok(SocketAddr::from_pathname(path)?),
             #[cfg(target_os = "linux")]
             UnixSocket::Abstract(name) => {
-                Ok(SocketAddr::from_abstract_name(name.as_encoded_bytes())?)
+                use std::os::unix::ffi::OsStringExt;
+
+                let mut v = name.into_vec();
+                v.insert(0, 0);
+
+                Ok(PathBuf::from(OsString::from_vec(v)))
             }
             UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
                 // You can't connect to a unix:dir.
--- a/zbus/src/address/transport/unix.rs
+++ b/zbus/src/address/transport/unix.rs
@@ -1,3 +1,6 @@
+#[cfg(any(unix, not(feature = "tokio")))]
+use crate::{Error, Result};
+
 #[cfg(target_os = "linux")]
 use std::ffi::OsString;
 use std::{
@@ -8,6 +11,16 @@ use std::{
 
 #[cfg(unix)]
 use super::encode_percents;
+
+#[cfg(not(feature = "tokio"))]
+use async_io::Async;
+
+#[cfg(target_os = "linux")]
+use std::os::linux::net::SocketAddrExt;
+#[cfg(unix)]
+use std::os::unix::net::{SocketAddr, UnixStream};
+#[cfg(all(windows, not(feature = "tokio")))]
+use uds_windows::UnixStream;
 
 /// A Unix domain socket transport in a D-Bus address.
 #[derive(Clone, Debug, PartialEq, Eq)]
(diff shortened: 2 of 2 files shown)


germaniuss:
@zeenix Can't check if windows functions correctly since I don't have it installed atm but I think everything should work fine. If you can give it a look and tell me if anything is not right. I've also moved the connect code to unix.rs since that is what is being done with Tcp and Unixexec and seemed easier to maintain than a lot of code in mod.rs.

germaniuss:
@zeenix So, can this be merged? I've been using the fork on my PC for a personal project and all seems to work fine. There was an issue with windows but it should be solved now.

zeenix:
> @zeenix So, can this be merged?

I was on a work trip this week so didn't get to review but in any case, please be patient.

> I've been using the fork on my PC for a personal project and all seems to work fine.

Great to hear.

> There was an issue with windows but it should be solved now.

Cool. I'll check. @elmarco could you please look at this from Windows' perspective?

zeenix:
> > @zeenix So, can this be merged?

Multiple CI failures here, so this will need more work for sure. :)

germaniuss:
> I was on a work trip this week so didn't get to review but in any case, please be patient.

Sure thing!

> Multiple CI failures here, so this will need more work for sure. :)

Yeah, I'll check again to see what's wrong here.

zbus/src/address/transport/unix.rs
```diff
@@ -31,6 +43,25 @@ impl Unix {
         self.path
     }
 
+    fn take_addr(self) -> Result<PathBuf> {
+        match self.take_path() {
+            // This is a `path` because neither uds_windows, tokio, nor async_io provide
+            // SocketAddrExt api
```
zeenix: missing `.`.

zbus/src/address/transport/unix.rs
```diff
@@ -55,6 +86,45 @@ impl Unix {
...
+    pub(super) async fn connect(self) -> Result<UnixStream> {
+        #[cfg(unix)]
+        {
+            let addr = self.take_addr()?;
+            UnixStream::connect(addr)
+                .await
+                .map_err(|e| Error::InputOutput(e.into()))
+        }
+
+        #[cfg(windows)]
```
zeenix:
```rust
#[cfg(not(unix))]
```

Like the current code is doing?

zbus/src/address/transport/unix.rs
```diff
@@ -55,6 +86,45 @@ impl Unix {
...
     }
+
+    #[cfg(not(feature = "tokio"))]
+    pub(super) async fn connect(self) -> Result<Async<UnixStream>> {
+        let addr = self.take_addr()?;
+
+        #[cfg(unix)]
+        let stream = Async::<UnixStream>::connect(addr).await;
+
+        #[cfg(windows)]
```
zeenix: `#[cfg(not(unix))]` here too (and other places where it makes sense?).

zbus/src/address/transport/unix.rs
```diff
@@ -31,6 +43,25 @@ impl Unix {
...
+            UnixSocket::File(path) => Ok(path),
+            #[cfg(target_os = "linux")]
+            UnixSocket::Abstract(name) => {
+                use std::{ffi::OsString, os::unix::ffi::OsStringExt};
+                let mut v = name.into_vec();
+                v.insert(0, 0);
+                Ok(PathBuf::from(OsString::from_vec(v)))
+            }
+            UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
+                // you can't connect to a unix:dir
```
zeenix: Please follow the English grammer in comments: capitalize the first letter and end the sentence or phrase with a `.` at the end.

zbus/src/address/transport/mod.rs
```diff
@@ -102,59 +96,7 @@ impl Transport {
     #[cfg_attr(any(unix, windows), async_recursion::async_recursion)]
     pub(super) async fn connect(self) -> Result<Stream> {
```
zeenix, line 101: The commit message says "The connect_addr API is dropped" but I don't see any `connect` API being dropped. 🤔 If it's dropped, it should be done in a separate commit after its last use is removed. This way, we can easily bring it back if it's needed again, with a simple `git revert`.

zbus/src/address/transport/mod.rs
```diff
@@ -102,59 +96,7 @@ impl Transport {
...
-                    }
-
-                    #[cfg(not(unix))]
-                    {
-                        let _ = stream;
-                        Err(Error::Unsupported)
-                    }
-                }
-            }
+            Transport::Unix(unix) => unix.connect().await.map(Stream::Unix),
```
zeenix, line 106: Why are these changes needed? If you're doing a refactor, it should be going into a separate commit.

germaniuss:
@zeenix The CI issues should be fixed now.

zeenix (changes requested):
Apart from these, please also choose a more appropriate emoji. ♻️ is for refactoring and that's not a good catagorization of the change.

zbus/src/address/transport/unix.rs
zeenix: oh, this is an existing comment you moved. I should have been more careful with reviewing the PR that introduced it. :-)

zbus/src/address/transport/mod.rs
germaniuss: I meant specifically that the PR no longer uses the `UnixSocket::connect_addr` from `SocketAddrExt`, instead I just use the plain `connect` functions that async-io and tokio provide since those support abstract names when the path is prepended by a null byte. Not sure how I could've worded that better. English is not my first language sorry 😅

zbus/src/address/transport/mod.rs
germaniuss: They aren't strictly needed, I can move all code back to the function in `mod.rs`. I felt this was easier to maintain in the future and is what is being done for Tcp and Unixexec, they both have a connect function that contains code for either tokio or async-io.

zbus/src/address/transport/unix.rs
germaniuss: No problem, I'll correct it if you want.

zbus/src/address/transport/unix.rs
germaniuss: Not sure about that, most places in the codebase use `cfg(windows)`. I can change it for sure, but should I also change it for the other places then? Maybe I'll just revert this particular instance and the others can be looked at in another PR?

zbus/src/address/transport/mod.rs
zeenix: > Not sure how I could've worded that better. English is not my first language sorry 😅

The wording you just used right here, perhaps? :wink: It's quite clear to me, even though I think that's too much of an implementation detail that I'd rather skip but no strong feelings about this. I'd rather have too many details in the commit log over too little. :)

zbus/src/address/transport/mod.rs
zeenix: I thought so. Refactors and improvements to code readability are always welcome! Just that it should be ideally done separately. Perhpas you can split the commit and do a refactor before the actual changes?

zbus/src/address/transport/unix.rs
zeenix: That would be greatly appreciated but please first correct it in the current location in a separate commit (✏️ being the most appropiate emoji).

zbus/src/address/transport/unix.rs
zeenix: If the code is actually windows-specific, then `cfg(windows)` is most appropriate but it's possible that in many places we're using it where we should be using `cfg(not(unix))`. E.g line#51 in mod.rs that you removed, we were doing `cfg(not(unix))`. We should also keep in mind that by using `cfg(unix)` and `cfg(windows)`, we ensure that the code will fail to build on any platform that is neither of these.

zbus/src/address/transport/mod.rs
germaniuss: I've split the refactor to another commit. I had to put it after the actual changes since moving the code before would have been a harder and there would have been a lot of duplicated code. All platforms/features used the code using spawn_blocking and it would have had to be copied over to both connect functions for tokio and async-io.

zbus/src/address/transport/mod.rs
germaniuss: I changed the commit message but kept, the mention of `connect_addr` since there was an explicit mention of it in a comment in the code (on why it was not being used for the windows version). I assume the original intention was to use it for windows as well when it was supported.

zbus/src/address/transport/unix.rs
germaniuss: Fixed the comment and split to another commit.

zbus/src/address/transport/unix.rs
germaniuss: I changed both instances of `cfg(windows)` for `cfg(not(unix))`. The rest of the instances do seem to be windows specific. In any case I believe the windows specific code could be removed entirely once the experimental UDS api for windows is stabilized in `std`

zbus/src/address/transport/unix.rs
germaniuss: Done

zbus/src/address/transport/mod.rs
zeenix: > I've split the refactor to another commit.

Thanks.

> I had to put it after the actual changes since moving the code before would have been a harder

Harder, yes but that's much better. If (for any reason) we realize later that there are problems with the actual changes, we'll have to revert both changes and we'll loose the refactoring as well.

> and there would have been a lot of duplicated code. All platforms/features used the code using spawn_blocking and it would have had to be copied over to both connect functions for tokio and async-io.

Not sure I follow. Why would you need to care about use of spawn_blocking before the actual change that ports over? Remember that each commit needs to be as independent as possible.

germaniuss:
Addressed all of the comments. Not really sure what emoji I should use for the main commit though 🤔. Maybe ⚡️? Though I'm not sure since there is not a clear performance improvement.

zeenix:
> Addressed all of the comments.

Thanks. I unresolved one though. I'm sorry to ask you to do more but I did describe the reason for the changes I'm asking for.

> Not really sure what emoji I should use for the main commit though 🤔. Maybe ⚡️? Though I'm not sure since there is not a clear performance improvement.

I think it's more appropriate than :recycle: still and I think there **is** a tiny performance improvement from not launching threads. Having said that 🧵is probably the best here.

zeenix:
> > Addressed all of the comments.
> 
> Thanks. I unresolved one though.

Oh and the windows CI continues to be failing.

zbus/src/address/transport/unix.rs
zeenix: > I changed both instances of `cfg(windows)` for `cfg(not(unix))`. The rest of the instances do seem to be windows specific.

:+1: 

> In any case I believe the windows specific code could be removed entirely once the experimental UDS api for windows is stabilized in `std`

For sure but I'm not holding my breath. :)

germaniuss:
Switched out the order of the commits. Going to find a windows PC to test out why its failing.

germaniuss:
@zeenix Not really sure if I like how that second commit turned out, but everything should be done now. Windows turned out to be more of a pain than I imagined.

zeenix:
> @zeenix Not really sure if I like how that second commit turned out, but everything should be done now. Windows turned out to be more of a pain than I imagined.

Ouch, sorry! I actually was considering to drop the Windows support as I also don't have a machine to test it on but some people do use zbus on Windows. :shrug:

zbus/src/address/transport/unix.rs
```diff
@@ -1,3 +1,6 @@
+#[allow(unused_imports)]
```
zeenix: It would be good to avoid this because if the import actually goes unused for all cfgs/platforms, we'll end up with 2 rendundant lines here forever. 

Please use the appropriate `cfg` here (just like we're doing for other imports here).

zbus/src/address/transport/unix.rs
```diff
@@ -9,6 +12,18 @@ use std::{
...
 
+#[cfg(not(feature = "tokio"))]
+use async_io::Async;
+
+#[cfg(target_os = "linux")]
+use std::os::linux::net::SocketAddrExt;
+#[cfg(unix)]
+use std::os::unix::net::SocketAddr;
+#[cfg(unix)]
+pub use std::os::unix::net::UnixStream;
```
zeenix: these can and should be combined. fmt doesn't complain for this? 🤔

zbus/src/address/transport/mod.rs
```diff
@@ -102,58 +96,12 @@ impl Transport {
     #[cfg_attr(any(unix, windows), async_recursion::async_recursion)]
     pub(super) async fn connect(self) -> Result<Stream> {
         match self {
+            #[allow(unused_variables)]
```
zeenix: same here: please use the appropriate `cfg` on the variables that are not always used, instead.

zbus/src/address/transport/unix.rs
```diff
@@ -31,6 +46,34 @@ impl Unix {
...
+            UnixSocket::Abstract(name) => {
+                Ok(SocketAddr::from_abstract_name(name.as_encoded_bytes())?)
+            }
+            UnixSocket::Dir(_) | UnixSocket::TmpDir(_) => {
+                // You can't connect to a unix:dir.
+                Err(Error::Unsupported)
+            }
+        }
+    }
+
```
zeenix: * Given that the 2 variants of this functions return completely different types for different platform families, probably best to not name them the same.
* nitpick: we try to follow top-down code layout in this project: pub before pub(crate) and pub(crate) before private API, and usage before definition. So these should be moved below `connect`.

zbus/src/address/transport/unix.rs
```diff
@@ -55,6 +98,48 @@ impl Unix {
...
+        let stream = crate::Task::spawn_blocking(
+            move || -> Result<_> {
+                #[cfg(unix)]
+                let stream = UnixStream::connect_addr(&addr)?;
+                #[cfg(windows)]
+                let stream = UnixStream::connect(addr)?;
+                stream.set_nonblocking(true)?;
+
+                Ok(stream)
```
zeenix: We're ending up with a lot more code duplication with this refactor and beats the point of this commit. The next commit fixes this (mostly) but the goal of splitting it into a separate commit was to make this change independent and good on its own.

zbus/src/address/transport/unix.rs
```diff
@@ -103,42 +93,33 @@ impl Unix {
...
+            .await??;
+            Async::new(stream)
+        };
+
+        stream.map_err(|e| Error::InputOutput(e.into()))
     }
 
     #[cfg(all(unix, feature = "tokio"))]
-    pub(super) async fn connect(self) -> Result<tokio::net::UnixStream> {
+    pub(super) async fn connect(self) -> Result<UnixStream> {
```
zeenix: This change of switching from full path of `UnixStream` to import, is not required by the containing commit. I think you may have added the change to the wrong commit. 😉

zeenix (changes requested):

zbus/src/address/transport/unix.rs
germaniuss: Yeah, that's why I wanted to do the code move after the actual changes were applied since I wanted to avoid code duplication, I don't really know what would be a good way to approach this if I were to move the code prior to the other changes.

zbus/src/address/transport/unix.rs
germaniuss: One of them has `pub` since its imported in `mod.rs`

zbus/src/address/transport/unix.rs
zeenix: > Yeah, that's why I wanted to do the code move after the actual changes

As I explained before with an example, there are practical reasons for keeping separate changes indendent.

> I don't really know what would be a good way to approach this if I were to move the code prior to the other changes.

By moving the common parts into a separate function(s)?

zbus/src/address/transport/unix.rs
germaniuss: Yeah sorry I did a bunch of rewrites, since I had to switch to a windows PC and back again haha, this slipped in at some point but it's not present in the current version.

zbus/src/address/transport/unix.rs
zeenix: > One of them has `pub` since its imported in `mod.rs`

Oops, sorry I missed that detail.

zbus/src/address/transport/unix.rs
zeenix: > but it's not present in the current version.

I was commenting on the last commit of the current revision and I can to still see it. Can you please check the last commit again? If you also see it in Github, see if you also see it locally or not. If not, it's some GH bug then.

zbus/src/address/transport/unix.rs
germaniuss: Ideally I would like them to be the same function, I didn't really know what's the best approach to these functions was here. I can rename them to take_sock_addr and take_path_addr or something similar, but I'm still not convinced this approach is the best one so I'd like to hear your opinion on how the second commit could be handled better.

zbus/src/address/transport/unix.rs
germaniuss: Oh you meant the last commit, at some point it was in the second one haha. I'll change it to use the full path.

zbus/src/address/transport/unix.rs
germaniuss: Done!

zbus/src/address/transport/unix.rs
germaniuss: I changed that in this revision, since tcp does not reexport TcpStream either and I thought changing that only for UnixStream could be confusing.

zbus/src/address/transport/mod.rs
germaniuss: Done

zbus/src/address/transport/unix.rs
germaniuss: I've renamed the functions to `take_socket_addr` and `take_path_addr` and moved them down.

zbus/src/address/transport/unix.rs
germaniuss: I've made a function called `get_stream` that abstracts the common code.

zbus/src/address/transport/unix.rs
germaniuss: Changed all commits to use full path for `tokio::net::UnixStream` as it was before

zbus/src/address/transport/unix.rs
```diff
@@ -71,58 +69,49 @@ impl Unix {
...
-        let stream = self.get_stream().await?;
-        tokio::net::UnixStream::from_std(stream).map_err(|e| Error::InputOutput(e.into()))
+        let addr = self.take_addr()?;
+        tokio::net::UnixStream::connect(addr)
+            .await
+            .map_err(|e| Error::InputOutput(e.into()))
     }
 
     #[cfg(any(unix, not(feature = "tokio")))]
-    async fn get_stream(self) -> Result<UnixStream> {
```
zeenix: I don't think we need this change to be so intrusive and remove functions just added in the previous commit. Why not just simplify this method instead?

zbus/src/address/transport/unix.rs
```diff
@@ -71,58 +69,49 @@ impl Unix {
...
-    }
-
-    #[cfg(unix)]
-    fn take_socket_addr(self) -> Result<SocketAddr> {
-        match self.take_path() {
-            UnixSocket::File(path) => Ok(SocketAddr::from_pathname(path)?),
             #[cfg(target_os = "linux")]
             UnixSocket::Abstract(name) => {
-                Ok(SocketAddr::from_abstract_name(name.as_encoded_bytes())?)
+                use std::{ffi::OsString, os::unix::ffi::OsStringExt, path::PathBuf};
```
zeenix: nitpick: empty line after imports please.

zbus/src/address/transport/unix.rs
```diff
@@ -71,58 +69,49 @@ impl Unix {
...
+                let mut v = name.into_vec();
+                v.insert(0, 0);
+                Ok(PathBuf::from(OsString::from_vec(v)))
```
zeenix, line 118: nitpick: I tend to keep an empty line before the last expression/return statements for readibility. It would be great if you also add it here. 🙏

zeenix (commented):
Great work! I appreciate your patience and perseverance. 👍 I'm pretty sure the next iteration can be merged.

zeenix (changes requested):

zbus/src/address/transport/unix.rs
germaniuss: Done, I've even removed two of the imports since they were already being imported at the top ;)

zbus/src/address/transport/unix.rs
germaniuss: Done

zbus/src/address/transport/unix.rs
germaniuss: I've readded the function and made it windows specific since from a maintainability point of view it seems the best for the eventual windows specific code removal.

germaniuss:
> Great work! I appreciate your patience and perseverance. 👍 I'm pretty sure the next iteration can be merged.

Great to hear! I rather appreciate the strict guidelines and expectation of code quality. That what makes open source have some of the best software out there!

zbus/src/address/transport/unix.rs
germaniuss: > seems the best for the eventual windows specific code removal.

And to be fair, the UDS on windows support for Rust does seem to be going along quite well so maybe it's not so far off after all :D

zbus/src/address/transport/unix.rs
```diff
@@ -103,26 +102,20 @@ impl Unix {
         .await?
     }
 
-    #[cfg(all(windows, not(feature = "tokio")))]
-    fn take_path_addr(self) -> Result<PathBuf> {
-        // This is a `path` in case of Windows until uds_windows provides the needed API:
-        // https://github.com/haraldh/rust_uds_windows/issues/14
+    #[cfg(any(unix, not(feature = "tokio")))]
+    fn take_addr(self) -> Result<PathBuf> {
```
zeenix, line 106: you're renaming a method in the last commit that you created in the previous one and the commit still seems more intrusive than it should be IMO. This commit should mostly just be replacing spawn_blocking use for `unix` in `get_stream` (which should exist for both configs) so the caller is simply abstracted from creation of stream and this commit becomes simpler and very much to the point.

zeenix:
> > Great work! I appreciate your patience and perseverance. 👍 I'm pretty sure the next iteration can be merged.
> 
> Great to hear! I rather appreciate the strict guidelines and expectation of code quality. That what makes open source have some of the best software out there!

Well?

zeenix:
Ping? You were so close.

zeenix:
@germaniuss Please get this to the finish line soon. If we do not see any activity on this PR in a week's time, this will be treated as spam and closed. In addition your account may be permanently blocked from corresponding with the project in the future (depending on how much time the maintainers spent on reviewing this PR).

Issue #495, opened 2023-10-20
xmlgen: "incorrect type" for arg with two or more mandatory u64 parameters


svalouch:
`zbus-xmlgen` fails to parse an argument with two (or more) unsigned integers.

Minimal reproducer (as of 1856c0f348eff0d25fac0921c963cb05b3811c59):
```xml
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN" "http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<node>
    <interface name="org.test.testinterface">
        <method name="testmethod">
            <arg name="testarg" direction="out" type="tt"/>
        </method>
    </interface>
</node>
```
Result: `Error: QuickXml(Custom("incorrect type"))`

It works fine if the `type` is either a single `t` or any number of optional `t`s, such as `(t)` or `(ttt)` and so on.

<details>
<summary>Real-world example from nfs-ganesha</summary>
I came across this when trying to parse the introspection output of [nfs-ganesha](https://github.com/nfs-ganesha/nfs-ganesha)

Command used: `gdbus introspect --system --dest org.ganesha.nfsd --object-path /org/ganesha/nfsd/ClientMgr --xml`

The offending line is right at the very bottom:
```xml
<arg name="clnt_cmp_ops_stats" direction="out" type="ttt"/>
```

Complete output:
```xml
<!DOCTYPE node PUBLIC "-//freedesktop//DTD D-BUS Object Introspection 1.0//EN"
"http://www.freedesktop.org/standards/dbus/1.0/introspect.dtd">
<node>
  <interface name="org.freedesktop.DBus.Introspectable">
    <method name="Introspect">
      <arg name="data" direction="out" type="s"/>
    </method>
  </interface>
  <interface name="org.ganesha.nfsd.clientmgr">
    <method name="AddClient">
      <arg name="ipaddr" direction="in" type="s"/>
      <arg name="status" direction="out" type="b"/>
      <arg name="error" direction="out" type="s"/>
    </method>
    <method name="RemoveClient">
      <arg name="ipaddr" direction="in" type="s"/>
      <arg name="status" direction="out" type="b"/>
      <arg name="error" direction="out" type="s"/>
    </method>
    <method name="ShowClients">
      <arg name="time" direction="out" type="(tt)"/>
      <arg name="clients" direction="out" type="a(sbbbbbbbb(tt))"/>
    </method>
  </interface>
[...]

PR #1802, opened 2026-05-26
✅ zx: Add regression test for bare struct arg signatures

cachebag:
This PR adds regression coverage for introspection args with multi-complete signatures like `type="tt"`.

Related: #495

### DIFF
--- a/zbus_xml/tests/tests.rs
+++ b/zbus_xml/tests/tests.rs
@@ -2,6 +2,7 @@ use quick_xml::de::DeError;
 use std::error::Error;
 
 use zbus_xml::{ArgDirection, Node};
+use zvariant::Signature;
 
 #[test]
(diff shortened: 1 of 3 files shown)


zeenix (changes requested):
LGTM, thanks but please add description to commit message too (best to write everything in the commit message and then let `gh pr create` populate the PR description from the commit message).

zeenix:
> Closes #495

Wait, this doesn't fix it, right? If the issue is already fixed, we should close it and not have this claim here.

cachebag:
Am I missing something here? @zeenix

zeenix (commented):
Sorry, it slipped my radar. The only thing left is to break the line in the commit description to make it fit the 74 chars limit, as per our (and git's own) conventions. :)

zeenix (approved):

PR #1803, opened 2026-05-26
⚡ zb: Route method replies directly by serial

cachebag:
This PR routes pending method replies directly by serial instead of consuming them through an internal MessageStream.

Previously, each method call expecting a reply created a filtered broadcast receiver for MethodReturn/Error messages and then polled that stream until it found the matching reply_serial. This meant unrelated replies still passed through per-call stream machinery.

This change adds a pending method-call map keyed by outgoing serial. `call_method_raw` registers the pending call before sending the message, and SocketReader dispatches incoming MethodReturn/Error messages directly to the matching pending caller. Socket read failures now drain the pending map and wake callers with the read error.

This removes the dedicated method-return broadcast channel and avoids extra stream polling/allocation work on the method-call reply path while keeping normal message broadcast behavior unchanged.


zeenix:
Thanks, this is a big one and I'll need time to review properly. In the meantime, I'll have LLMs take a look at it and see if they can find something real. :)

zbus/src/connection/mod.rs
```diff
@@ -249,42 +324,31 @@ impl OrderedFuture for PendingMethodCall {
...
-                    Poll::Ready(PollResult::Terminated) => {
-                        return Poll::Ready(None);
-                    }
-                    Poll::Pending => return Poll::Pending,
+            // `before` is only provided after another stream has produced that sequence. Since the
+            // socket reader dispatches replies synchronously as it reads messages, any earlier
+            // matching reply would already be stored above.
+            if before.is_some() {
+                return Poll::Ready(None);
+            }
```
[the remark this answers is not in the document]
cachebag: I think that this is fair.

zeenix:
@claude can you review this too, keeping in mind the existing review from copilot here?

zbus/src/connection/mod.rs
```diff
@@ -249,42 +324,29 @@ impl OrderedFuture for PendingMethodCall {
...
+                    this.listener = None;
                 }
+                // `before` is only provided after another stream has produced that sequence. Since
+                // the socket reader dispatches replies synchronously as it reads messages, any
+                // earlier matching reply would already be stored above. Return `None` only after
+                // polling the listener so the current task is still woken when the reply arrives.
+                Poll::Pending if before.is_some() => return Poll::Ready(None),
+                Poll::Pending => return Poll::Pending,
             }
         }
```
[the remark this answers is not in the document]
cachebag: Fixed

zbus/src/connection/socket_reader.rs
```diff
@@ -97,6 +109,47 @@ impl SocketReader {
...
+        let reply_serial = match msg.header().reply_serial() {
+            Some(serial) => serial,
+            None => return,
+        };
+
+        let pending_call = self
+            .pending_method_calls
+            .lock()
+            .expect("lock poisoned")
+            .remove(&reply_serial);
```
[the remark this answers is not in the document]
cachebag: Fixed

zbus/src/connection/mod.rs
```diff
@@ -94,6 +92,7 @@ impl Drop for ConnectionInner {
 type Subscriptions = HashMap<OwnedMatchRule, (u64, InactiveReceiver<Result<Message>>)>;
 
 pub(crate) type MsgBroadcaster = Broadcaster<Result<Message>>;
+pub(crate) type PendingMethodCallMap = StdMutex<HashMap<NonZeroU32, Arc<PendingMethodReply>>>;
```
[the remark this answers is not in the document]
cachebag: I do not agree with this. This was done intentionally. The lock is only held for short in-memory HashMap/Option updates and is never held across an `.await`, so using an async mutex would add overhead on the hot path without improving scheduling behavior here.

I did address the poisoning part separately by recovering with `into_inner()` instead of panicking.

zbus/src/connection/socket_reader.rs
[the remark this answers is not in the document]
zeenix: Actually, in zbus we don't handle lock poisoning anywhere. Your code was good. Probably also an `unwrap` would be fine.

zbus/src/connection/mod.rs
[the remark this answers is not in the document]
zeenix: yeah, this is something LLMs seem to always get wrong.

zbus/src/connection/mod.rs
```diff
@@ -2,13 +2,13 @@
...
 use event_listener::{Event, EventListener};
-use ordered_stream::{OrderedFuture, OrderedStream, PollResult};
+use ordered_stream::OrderedFuture;
 use std::{
     collections::HashMap,
     io::{self, ErrorKind},
     num::NonZeroU32,
     pin::Pin,
-    sync::{Arc, OnceLock, Weak},
+    sync::{Arc, Mutex as StdMutex, OnceLock, Weak},
```
zeenix: The pattern we've followed so far for this is naming aliases `SyncMutex` or `AsyncMutex`, depending on which import we alias.

zbus/src/connection/mod.rs
```diff
@@ -221,8 +221,103 @@ pub struct Connection {
...
+
+    pub(crate) fn set_result(&self, ordering: zbus::message::Sequence, reply: Result<Message>) {
+        let notify = {
+            let mut pending_reply = self
+                .reply
+                .lock()
+                .unwrap_or_else(|poisoned| poisoned.into_inner());
+            if pending_reply.is_none() {
+                *pending_reply = Some((ordering, reply));
+                true
```
zeenix: nitpick: empty line before the last expression.

zbus/src/connection/mod.rs
```diff
@@ -221,8 +221,103 @@ pub struct Connection {
...
     serial: NonZeroU32,
+    reply: Arc<PendingMethodReply>,
+    pending_method_calls: Arc<PendingMethodCallMap>,
+    listener: Option<EventListener>,
+}
+
+#[derive(Debug)]
+pub(crate) struct PendingMethodReply {
+    reply: StdMutex<Option<(zbus::message::Sequence, Result<Message>)>>,
+    event: Event,
```
zeenix: Instead of 2 separate states, why not using a channel instead? I know we don't need a broadcast channel here but since async-broadcast is a dep, it's best to use that.

zeenix (changes requested):
Sounds good to me in general.

* I've been meaning to split `mod.rs` here for a while since it's nearly 2k lines already so I'd feel a bit bad about adding more lines and types to it. Would it be possible to split the PendingMethodCall to a separate module in a separate commit **before** this one? I know that would be a bit of work but I believe that's the correct order since there are more chances of a regression from an actual change in behaviour than a refactor (especially in Rust) so then it's easier to only revert the problematic commit.
* Could you please ensure the commit message contains the same details as the PR description?

zbus/src/connection/mod.rs
cachebag: IMHO it is more general than what this path needs. This is a one-shot handoff from the socket reader to exactly one pending caller, so the Option + Event keeps the reply path smaller and avoids async-broadcast queue/receiver bookkeeping. 

If you prefer that for merely the consistency then I can switch it out if you'd like.

zeenix: > IMHO it is more general than what this path needs.

Yes but using a general API for a more specific (subset) is not wrong. :) We will only have 1 sender and receiver, but that's ok. There were event designs to merge async-channel and async-broadcast together and addition of a oneshot API in it. It just never happened.

> so the Option + Event keeps the reply path smaller and avoids async-broadcast queue/receiver bookkeeping.

I don't think there will be more bookkeeping than this approach. It'll likely be less code/fields. I also believe it'll be more readable.

cachebag: Done

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,14 +1,23 @@
...
 use futures_core::Future;
-use ordered_stream::{OrderedFuture, OrderedStream, PollResult};
+use futures_lite::Stream;
+use ordered_stream::OrderedFuture;
 
-use crate::{Message, MessageStream, Result, message::Type};
+use crate::{Error, Message, Result, message::Sequence};
+
+pub(crate) type PendingMethodReply = (Sequence, Result<Message>);
+
```
zeenix: redundant empty line.

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,14 +1,23 @@
...
+pub(crate) type PendingMethodReplySender = Arc<Sender<PendingMethodReply>>;
```
zeenix: Why `Arc`-wrap? `Sender` already implements `Clone` and it's cheap.

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,14 +1,23 @@
...
+pub(crate) type PendingMethodCallMap = SyncMutex<HashMap<NonZeroU32, PendingMethodReplySender>>;
```
zeenix: This seems like it really deserves to be it's own type (struct with a named field). It should also hide the `Arc` and impl `Clone`.

zeenix: Instead of using `pub(crate)` on every item here, it'd be best to use `pub` but make sure that symbols from this module are not `pub use`ed.

zbus/src/connection/pending_method_call.rs
```diff
@@ -17,13 +26,53 @@ use crate::{Message, MessageStream, Result, message::Type};
...
+        {
+            let mut pending_method_calls = pending_method_calls.lock().unwrap();
+            match pending_method_calls.entry(serial) {
+                Entry::Vacant(entry) => {
+                    entry.insert(reply_sender.clone());
+                }
+                Entry::Occupied(_) => {
+                    return Err(Error::Failure(format!(
+                        "Serial number `{serial}` reused while a method call is still pending"
+                    )));
```
zeenix: This is not a runtime recoverable error IMO. This can only happen if we make a very catastrophic mistake somewhere. This this should be an `unreachable!` IMO.

zbus/src/connection/pending_method_call.rs
```diff
@@ -17,13 +26,53 @@ use crate::{Message, MessageStream, Result, message::Type};
...
+        })
+    }
+}
+
+impl Drop for PendingMethodCall {
+    fn drop(&mut self) {
+        let mut pending_method_calls = self.pending_method_calls.lock().unwrap();
+        if pending_method_calls
+            .get(&self.serial)
+            .is_some_and(|reply_sender| Arc::ptr_eq(reply_sender, &self.reply_sender))
```
zeenix: Why do we need to check the sender? Serial already makes each call unique.

zbus/src/connection/pending_method_calls.rs
```diff
@@ -43,49 +92,24 @@ impl Future for PendingMethodCall {
...
-            }
+        match Pin::new(&mut this.reply_receiver).poll_next(cx) {
+            Poll::Ready(Some(reply)) => Poll::Ready(Some(reply)),
+            Poll::Ready(None) => Poll::Ready(None),
+            // `before` is only provided after another stream has produced that sequence. Since the
+            // socket reader dispatches replies synchronously as it reads messages, any earlier
+            // matching reply would already be queued above. Return `None` only after polling the
+            // receiver so the current task is still woken when the reply arrives.
+            Poll::Pending if before.is_some() => Poll::Ready(None),
+            Poll::Pending => Poll::Pending,
```
zeenix, line 152: OrderedStream always confuses me so I'm not too confident about this code. I wish @danieldg was still around to review this part. One thing I do notice is that there is no handling or return of `PollResult::NoneBefore`. 🤔

zbus/src/connection/socket_reader.rs
```diff
@@ -50,8 +56,14 @@ impl SocketReader {
             trace!("Waiting for message on the socket..");
             let msg = self.read_socket().await;
             match &msg {
-                Ok(msg) => trace!("Message received on the socket: {:?}", msg),
-                Err(e) => trace!("Error reading from the socket: {:?}", e),
+                Ok(msg) => {
+                    trace!("Message received on the socket: {:?}", msg);
+                    self.dispatch_method_reply(msg);
```
zeenix: Why is it assumed that it's a method call?

zbus/src/connection/socket_reader.rs
```diff
@@ -97,6 +109,47 @@ impl SocketReader {
         }
     }
 
+    fn dispatch_method_reply(&self, msg: &Message) {
+        if !matches!(msg.message_type(), Type::MethodReturn | Type::Error) {
+            return;
+        }
```
zeenix: what about method calls? This isn't client-side only. The CI is passing so I must be missing something here. 🤔

zbus/src/connection/pending_method_call.rs
```diff
@@ -17,13 +26,53 @@ use crate::{Message, MessageStream, Result, message::Type};
...
+        {
+            pending_method_calls.remove(&self.serial);
```
zeenix: It feels a bit odd that the containee is managing itself in its container. Perhaps it'd be best if the new Map type, would handle a call in an async method? Also I actually wonder if there is even a need to keep a map of pending calls around in `Connection`. The PendingCall is kept alive by the caller and it's a future they need to .await so we don't need to manage the lifecycle of the call anyway.

zeenix (changes requested):
Apart from the inline comments, I'm also wonder about the actual benefits from this. Any chance you could add a benchmark before the last commit that demonstrates the  benefits? Typically client applications don't have more than 20 calls at a time but I can imagine in some cases, them going to 100 or even 1000 at a time so the benchmark should test for 1000 (near) simultaneous calls.

cachebag:
@zeenix CodSpeed CPU instrumentation warns about system calls; looks like walltime mode would be the more representative mode for this benchmark.

~~Also, how are we going to compare before/after? I am not sure of the best way to get a comparison since we have no baseline to run against~~

I measured them locally. 


```log
zbus on  perf/direct-method-reply-routing [$] 
➜ git checkout 5086686d
cargo bench -p zbus --bench benchmarks --features p2p 1000-concurrent-p2p 2>&1 | tee /tmp/zbus-method-call-bench-before.txt

git checkout perf/direct-method-reply-routing
cargo bench -p zbus --bench concurrent_method_calls --features p2p 1000-concurrent-p2p 2>&1 | tee /tmp/zbus-method-call-bench-after.txt

grep -E "1000-concurrent-p2p|time:" /tmp/zbus-method-call-bench-before.txt /tmp/zbus-method-call-bench-after.txt

Note: switching to '5086686d'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 5086686d 📊 zb: Add concurrent method-call benchmark
   Compiling zbus v5.15.0 (/home/cachebag/personal/zbus/zbus)
    Finished `bench` profile [optimized + debuginfo] target(s) in 11.38s
     Running benches/benchmarks.rs (target/release/deps/benchmarks-7d2f411d55f43d31)
Gnuplot not found, using plotters backend
Benchmarking method-call/1000-concurrent-p2p
Benchmarking method-call/1000-concurrent-p2p: Warming up for 3.0000 s
Benchmarking method-call/1000-concurrent-p2p: Collecting 10 samples in estimated 5.5798 s (165 iterations)
Benchmarking method-call/1000-concurrent-p2p: Analyzing
method-call/1000-concurrent-p2p
                        time:   [34.497 ms 34.544 ms 34.591 ms]
                        thrpt:  [28.909 Kelem/s 28.948 Kelem/s 28.988 Kelem/s]
                 change:
                        time:   [+445.02% +451.04% +455.87%] (p = 0.00 < 0.05)
                        thrpt:  [-82.010% -81.853% -81.652%]
                        Performance has regressed.

Warning: you are leaving 1 commit behind, not connected to
any of your branches:

  5086686d 📊 zb: Add concurrent method-call benchmark

If you want to keep it by creating a new branch, this may be a good time
to do so with:

 git branch <new-branch-name> 5086686d

Switched to branch 'perf/direct-method-reply-routing'
Your branch is up to date with 'origin/perf/direct-method-reply-routing'.
   Compiling zbus v5.15.0 (/home/cachebag/personal/zbus/zbus)
    Finished `bench` profile [optimized + debuginfo] target(s) in 11.31s
     Running benches/concurrent_method_calls.rs (target/release/deps/concurrent_method_calls-769cb29fe7257ae0)
Gnuplot not found, using plotters backend
Benchmarking method-call/1000-concurrent-p2p
Benchmarking method-call/1000-concurrent-p2p: Warming up for 3.0000 s
Benchmarking method-call/1000-concurrent-p2p: Collecting 10 samples in estimated 5.1253 s (825 iterations)
Benchmarking method-call/1000-concurrent-p2p: Analyzing
method-call/1000-concurrent-p2p
                        time:   [6.1778 ms 6.1852 ms 6.1938 ms]
                        thrpt:  [161.45 Kelem/s 161.68 Kelem/s 161.87 Kelem/s]
                 change:
                        time:   [-82.117% -82.086% -82.056%] (p = 0.00 < 0.05)
                        thrpt:  [+457.29% +458.22% +459.18%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) low mild

/tmp/zbus-method-call-bench-before.txt:Benchmarking method-call/1000-concurrent-p2p
/tmp/zbus-method-call-bench-before.txt:Benchmarking method-call/1000-concurrent-p2p: Warming up for 3.0000 s
/tmp/zbus-method-call-bench-before.txt:Benchmarking method-call/1000-concurrent-p2p: Collecting 10 samples in estimated 5.5798 s (165 iterations)
/tmp/zbus-method-call-bench-before.txt:Benchmarking method-call/1000-concurrent-p2p: Analyzing
/tmp/zbus-method-call-bench-before.txt:method-call/1000-concurrent-p2p
/tmp/zbus-method-call-bench-before.txt:                        time:   [34.497 ms 34.544 ms 34.591 ms]
/tmp/zbus-method-call-bench-before.txt:                        time:   [+445.02% +451.04% +455.87%] (p = 0.00 < 0.05)
/tmp/zbus-method-call-bench-after.txt:Benchmarking method-call/1000-concurrent-p2p
/tmp/zbus-method-call-bench-after.txt:Benchmarking method-call/1000-concurrent-p2p: Warming up for 3.0000 s
/tmp/zbus-method-call-bench-after.txt:Benchmarking method-call/1000-concurrent-p2p: Collecting 10 samples in estimated 5.1253 s (825 iterations)
/tmp/zbus-method-call-bench-after.txt:Benchmarking method-call/1000-concurrent-p2p: Analyzing
/tmp/zbus-method-call-bench-after.txt:method-call/1000-concurrent-p2p
/tmp/zbus-method-call-bench-after.txt:                        time:   [6.1778 ms 6.1852 ms 6.1938 ms]
/tmp/zbus-method-call-bench-after.txt:                        time:   [-82.117% -82.086% -82.056%] (p = 0.00 < 0.05)
```

[zbus-method-call-bench-before.txt](https://github.com/user-attachments/files/28437991/zbus-method-call-bench-before.txt)
[zbus-method-call-bench-after.txt](https://github.com/user-attachments/files/28437993/zbus-method-call-bench-after.txt)

zbus/benches/concurrent_method_calls.rs
```diff
@@ -0,0 +1,73 @@
...
+
+struct BenchmarkInterface;
+
+#[zbus::interface(name = "org.zbus.Benchmark")]
+impl BenchmarkInterface {
+    async fn ping(&self, value: u32) -> u32 {
+        value
+    }
+}
+
```
zeenix: reminder: topdown code layout, so usage **before** definition for the same visibility items.

zbus/benches/concurrent_method_calls.rs
```diff
@@ -0,0 +1,73 @@
+use std::hint::black_box;
```
zeenix, line 1: Probably a better emoji for this commit: 🧐

.github/workflows/bench.yml
```diff
@@ -22,10 +22,30 @@ jobs:
           cache-target: release
           bins: cargo-codspeed
       - name: Build the benchmark target(s)
-        run: cargo codspeed build
+        run: cargo codspeed build --bench benchmarks
```
zeenix: why is this needed now? 🤔

.github/workflows/bench.yml
```diff
@@ -22,10 +22,30 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - name: Setup rust toolchain, cache and cargo-codspeed binary
+        uses: moonrepo/setup-rust@v1
+        with:
+          channel: stable
+          cache-target: release
+          bins: cargo-codspeed
+      - name: Build the benchmark target(s)
+        run: cargo codspeed build --bench concurrent_method_calls --features p2p --measurement-mode walltime
+      - name: Run the benchmarks
```
zeenix: Is this all for using walltime? I'm not sure you actually need that (despite of warning from Codspeed) since you're using an in-memory transport that doesn't do any syscalls. In fact, Codspeed says "there are no systemcalls in this bechmark" in a tooltip if you hover over the "Include system calls" checkbox (which is disabled) in the [console](https://codspeed.io/z-galaxy/zbus/branches/cachebag%3Aperf%2Fdirect-method-reply-routing?uri=zbus%2Fbenches%2Fconcurrent_method_calls.rs%3A%3Abenches%3A%3Aconcurrent_method_calls%3A%3Amethod-call%3A%3A1000-concurrent-p2p&runnerMode=WallTime).

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,29 +1,94 @@
...
+use async_broadcast::{Receiver, Sender, broadcast};
 use futures_core::Future;
-use ordered_stream::{OrderedFuture, OrderedStream, PollResult};
+use futures_lite::Stream;
+use ordered_stream::OrderedFuture;
 
-use crate::{Message, MessageStream, Result, message::Type};
+use crate::{Message, Result, message::Sequence};
+type PendingMethodReply = (Sequence, Result<Message>);
+
```
zeenix: * empty line should be above the previous line, not after. 
* topdown layout: Please move the private type alias below pub API (way below 😄).

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,29 +1,94 @@
...
+    sender: Sender<PendingMethodReply>,
+}
+
+impl Clone for PendingMethodReplySender {
+    fn clone(&self) -> Self {
+        Self {
+            sender: self.sender.clone(),
+        }
+    }
+}
```
zeenix: why an explicit impl? It can simply be derived, no?

zbus/src/connection/pending_method_call.rs
```diff
@@ -1,29 +1,94 @@
...
+use ordered_stream::OrderedFuture;
 
-use crate::{Message, MessageStream, Result, message::Type};
+use crate::{Message, Result, message::Sequence};
+type PendingMethodReply = (Sequence, Result<Message>);
+
+pub type PendingMethodCallMap = SyncMutex<HashMap<NonZeroU32, PendingMethodReplySender>>;
+
+#[derive(Debug)]
+pub struct PendingMethodReplySender {
```
zeenix: I think there was a misunderstanding. I asked for the Map type to be a struct. The sender is quite simple so type alias is good.

zeenix (changes requested):
Nice results for the benchmarks! Thanks for going the extra mile so I feel fine about a bit of a major change like this. :) However, I think some of my prevous comments are not yet addressed and I've some new reservation/comments too. :)

cachebag:
Cool. Work looks to be a bit slammed this week but I will get to this as soon as I can. Don't want you to think I abandoned

.github/workflows/bench.yml
cachebag: Interesting..okay sure I went ahead and switched it back to instrumentation-mode job...I guess that warning I saw was unnecessary then.

zbus/src/connection/pending_method_call.rs
cachebag: Thanks ChatGPT

.github/workflows/bench.yml
zeenix: Hmm.. I had actually assumed that you added the in-memory transport **after** that warning from Codspeed. It's back again so I think there **are** some syscalls involved and wall clock would indeed be better here. Having said that, it'd be hard to rely on Codspeed reports for this with wallclock time. I think it'd be best to just skip this benchmark in Codspeed workflow.

cachebag: Ok done. @zeenix

zbus/src/connection/socket_reader.rs
zeenix: You didn't answer my question here. If you return from this method, the message is not handled at all then (line 61 above).

zbus/src/connection/pending_method_call.rs
```diff
@@ -0,0 +1,146 @@
...
+
+use async_broadcast::{Receiver, Sender, broadcast};
+use futures_core::Future;
+use futures_lite::Stream;
+use ordered_stream::OrderedFuture;
+
+use crate::{Error, Message, Result, message::Sequence};
+
+#[derive(Clone, Debug)]
+pub struct PendingMethodCallMap {
```
zeenix: * `PendingMethodCalls` is good enough IMO. It being a map is an internal detail, no? 🤔 
* The module name should be named after this type IMO.

zbus/src/connection/pending_method_call.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+/// A method call whose completion can be awaited or joined with other streams.
+///
+/// This is useful for cache population method calls, where joining the call with an update signal
+/// stream can be used to ensure that cache updates are not overwritten by a cache population whose
+/// task is scheduled later.
+#[derive(Debug)]
+pub struct PendingMethodCall {
```
zeenix: Perhaps we can keep this type private and have `register_call` above, return an opaque `impl Future...`?

zbus/src/connection/pending_method_calls.rs
zeenix: @cachebag you didn't reply to this either. :)

zbus/src/connection/socket_reader.rs
cachebag: The `return` here only returns from `dispatch_pending_reply`, not from `receive_msg`. After `self.dispatch_pending_reply(msg)` returns, `receive_msg` continues into the existing broadcast loop below, so `MethodCall` and `Signal` messages are still delivered/handled exactly as before.

This helper only fast-paths `MethodReturn`/`Error` messages to pending outgoing calls; it does not consume or skip the message from the normal broadcast path. Even dispatched replies are still broadcast below.

@zeenix

zbus/src/connection/pending_method_calls.rs
cachebag: There is no such thing as a direct `PollResult::NoneBefore` return value. The equivalent is `Poll::Ready(None)`. The `ordered_stream::FromFuture` adapter maps `Poll::Ready(None)` from `OrderedFuture::poll_before` to `PollResult::NoneBefore`.

That’s what this branch is doing:
```rust
Poll::Pending if before.is_some() => Poll::Ready(None),
```
We poll `reply_receiver` first so the current task’s waker is registered. If no reply is queued and `before` is `Some(seq)`, then for the current uses with message streams, any matching reply with a receive position before `seq` would already have been read and queued by `SocketReader`, because replies are dispatched synchronously before the message is broadcast to streams. 

If the reply arrives later, the receiver wake-up is already registered.

@zeenix

zbus/src/connection/socket_reader.rs
zeenix: Ah right. I'm stupid. Thanks for explaining. I think it would be more efficient to instead:

* Only call this method for MethodCall msg types.
* Make this a `debug_assert`.

zbus/src/connection/pending_method_calls.rs
zeenix: Right, thanks for explaining. As I wrote, ordered stream stuff always confuses me. Since the CI is all green, I'll just trust you, my understanding and our tests that this is all correct. :)

cachebag: No problem! Thanks for thoroughly reviewing this addition. I know it is quite ambitious of me, so I appreciate the consideration.

zeenix (approved):

PR #1810, opened 2026-05-31
✨ zb: Add remove_named method to ObjectServer

hasali19:
I have a use case where I need to be able to track the interfaces an object implements and select interfaces to remove at runtime. This is difficult to do with the current API as it requires the interface type to be known statically.

This PR adds a method to remove an interface using its name instead of the type.

### DIFF
--- a/zbus/src/object_server/mod.rs
+++ b/zbus/src/object_server/mod.rs
@@ -190,16 +190,33 @@ impl ObjectServer {
         P: TryInto<ObjectPath<'p>>,
         P::Error: Into<Error>,
     {
+        self.remove_named(path, I::name()).await
+    }
+
+    /// Unregister a D-Bus [`Interface`] at a given path, using its name.
+    ///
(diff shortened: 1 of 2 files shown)


zeenix (changes requested):
Thanks, LGTM but please add description to the commit message too for offline viewing in the history.

zbus/src/object_server/mod.rs
```diff
@@ -189,17 +189,30 @@ impl ObjectServer {
         I: Interface,
         P: TryInto<ObjectPath<'p>>,
         P::Error: Into<Error>,
+    {
+        self.remove_named(path, I::name()).await
+    }
+
+    pub async fn remove_named<'p, P>(
```
zeenix, line 200: Please add docs to the new public method.

hasali19: Done

zeenix (approved):

PR #1811, opened 2026-06-02
🐛 zu: Fix GVariant bool alignment

kawadakk:
In GVariant, the boolean type has an alignment of 1, not 4.

<https://developer.gnome.org/documentation/specifications/gvariant-specification-1.0.html#booleans>

> A boolean has a fixed size of 1 and an alignment of 1.

This means that the elements of `amb` should be packed tightly like this:

```console
$ python
>>> GLib.Variant.parse(None, '[just true, true, true]').get_data_as_bytes().get_data()
b'\x01\x01\x01\x01\x02\x03'
```

### DIFF
--- a/zvariant/tests/number/bool_value.rs
+++ b/zvariant/tests/number/bool_value.rs
@@ -1,4 +1,6 @@
 use zvariant::LE;
+#[cfg(feature = "gvariant")]
+use zvariant::{serialized::Context, to_bytes};
 
 #[test]
 fn bool_value() {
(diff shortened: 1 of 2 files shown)


zeenix:
Thanks but I'm not sure this is correct. According to [the GVariant docs](https://developer.gnome.org/documentation/specifications/gvariant-specification-1.0.html#booleans):

> A boolean has a fixed size of 1 and an alignment of 1.

zeenix:
> Thanks but I'm not sure this is correct. According to [the GVariant docs](https://developer.gnome.org/documentation/specifications/gvariant-specification-1.0.html#booleans):
> 
> > A boolean has a fixed size of 1 and an alignment of 1.

Actually, that's what you're doing. :facepalm: I needed more coffee, I guess. :laughing:

zeenix (changes requested):
LGTM but one tiny nitpick: Could you please add a link [to the relevant section](https://developer.gnome.org/documentation/specifications/gvariant-specification-1.0.html#booleans) of the reference in the commit message and PR description?

zeenix (approved):

PR #1812, opened 2026-06-03
🐛 zv: Add tail padding to fixed-size struct in GVariant encoding

kawadakk:
<https://developer.gnome.org/documentation/specifications/gvariant-specification-1.0.html#structures>

> If all of the items contained in a structure are fixed-size then the structure itself is fixed- size. Considerations have to be made to satisfy the constraints that are placed on the value of this fixed size.

> [..] the fixed sized must be a multiple of the alignment of the structure. This is accomplished by adding zero-filled padding bytes to the end of any fixed-width structure until this property becomes true.

```console
$ python
>>> GLib.Variant.parse(None, '{uint32 0xffffffff: byte 0xa0}').get_data_as_bytes().get_data()
b'\xff\xff\xff\xff\xa0\x00\x00\x00'
>>> GLib.Variant.parse(None, '(uint32 0xffffffff, byte 0xa0)').get_data_as_bytes().get_data()
b'\xff\xff\xff\xff\xa0\x00\x00\x00'
```

### DIFF
--- a/zvariant/src/gvariant/de.rs
+++ b/zvariant/src/gvariant/de.rs
@@ -769,6 +769,12 @@ impl<'d, 'de, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> SeqAccess<'de
             // All fields have been deserialized.
             self.de.0.container_depths = self.de.0.container_depths.dec_structure();
 
+            if self.de.0.signature.is_fixed_sized() {
+                debug_assert_eq!(self.offsets_len, 0);
+                let alignment = self.de.0.signature.alignment(Format::GVariant);
+                self.de.0.parse_padding(alignment)?;
+            }
+
             // Skip over the framing offsets (if any)
             self.de.0.pos += self.offsets_len;
         }
--- a/zvariant/src/gvariant/de.rs
+++ b/zvariant/src/gvariant/de.rs
@@ -638,6 +638,11 @@ impl<'d, 'de, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> MapAccess<'de
         self.de.0.pos += de.0.pos;
         // No need for retaking the container depths as the value can't be incomplete.
 
+        // A fixed-sized dictionary entry should be padded to its alignment.
+        if self.offsets.is_none() {
+            self.de.0.parse_padding(self.element_alignment)?;
+        }
+
         if let Some(key_offset_size) = self.key_offset_size {
             self.de.0.pos += key_offset_size as usize;
         }
--- a/zvariant/src/gvariant/ser.rs
+++ b/zvariant/src/gvariant/ser.rs
@@ -728,6 +728,11 @@ where
         value.serialize(&mut *self.seq.ser)?;
         self.seq.ser.0.signature = self.key_signature;
 
+        // A fixed-sized dictionary entry should be padded to its alignment.
+        if self.seq.offsets.is_none() {
+            self.seq.ser.0.add_padding(self.seq.element_alignment)?;
+        }
+
         if let Some(key_offset) = key_offset {
             let entry_size = self.seq.ser.0.bytes_written - self.key_start.unwrap_or(0);
             let offset_size = FramingOffsetSize::for_encoded_container(entry_size);
--- a/zvariant/src/gvariant/ser.rs
+++ b/zvariant/src/gvariant/ser.rs
@@ -603,6 +603,13 @@ where
             Some(offsets) => offsets,
             None => return Ok(()),
         };
+
+        if self.ser.0.signature.is_fixed_sized() {
+            debug_assert!(offsets.is_empty());
+            let alignment = self.ser.0.signature.alignment(Format::GVariant);
+            self.ser.0.add_padding(alignment)?;
+        }
+
         let struct_len = self.ser.0.bytes_written - self.start;
         if struct_len == 0 {
             // Empty sequence
--- a/zvariant/tests/struct_value.rs
+++ b/zvariant/tests/struct_value.rs
@@ -35,6 +35,25 @@ fn struct_value() {
     let decoded: as_value::Deserialize<'_, Foo> = encoded.deserialize().unwrap().0;
     assert_eq!(decoded.0, foo);
 
+    // A struct with underaligned tail.
+    let foo: (u32, u8) = (u32::MAX, 0x40);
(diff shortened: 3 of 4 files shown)


zvariant/src/gvariant/de.rs
```diff
@@ -769,6 +769,12 @@ impl<'d, 'de, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> SeqAccess<'de
             // All fields have been deserialized.
             self.de.0.container_depths = self.de.0.container_depths.dec_structure();
 
+            if self.de.0.signature.is_fixed_sized() {
+                assert_eq!(self.offsets_len, 0);
```
zeenix: perhaps a `debug_assert_eq!` is more appropriate here? 🤔

zvariant/src/gvariant/ser.rs
```diff
@@ -603,6 +603,13 @@ where
             Some(offsets) => offsets,
             None => return Ok(()),
         };
+
+        if self.ser.0.signature.is_fixed_sized() {
+            assert!(offsets.is_empty());
```
zeenix: same question for this assert.

zeenix (changes requested):
Thanks, LGTM other than one small comment.

zeenix:
Oh and it would be great to have a test case for this. Perhaps `zvariant/tests/struct_value.rs` can be modified to test this?

kawadakk:
> Oh and it would be great to have a test case for this. Perhaps `zvariant/tests/struct_value.rs` can be modified to test this?

The PR already includes a test case: https://github.com/z-galaxy/zbus/pull/1812/changes/1398aecba0a6486582e1cee1b18ef4421bbbfae9

zvariant/src/gvariant/de.rs
kawadakk: Replaced with `debug_assert_eq!`

zvariant/src/gvariant/ser.rs
kawadakk: Ditto.

zvariant/tests/struct_value.rs
```diff
@@ -35,6 +35,22 @@ fn struct_value() {
     let decoded: as_value::Deserialize<'_, Foo> = encoded.deserialize().unwrap().0;
     assert_eq!(decoded.0, foo);
 
+    // A struct with underaligned tail
+    let foo: (u32, u8) = (u32::MAX, 0x40);
+    let encoded = to_bytes(ctxt, &foo).unwrap();
+    assert_eq!(encoded.bytes(), b"\xff\xff\xff\xff\x40");
```
zeenix: Wouldn't an array/slice literal instead of a string be better?

zvariant/tests/struct_value.rs
```diff
@@ -35,6 +35,22 @@ fn struct_value() {
...
+    let decoded: (u32, u8) = encoded.deserialize().unwrap().0;
+    assert_eq!(decoded, foo);
+
+    #[cfg(feature = "gvariant")]
+    {
+        let ctxt = Context::new_gvariant(LE, 0);
+        let encoded = to_bytes(ctxt, &foo).unwrap();
+        assert_eq!(encoded.bytes(), b"\xff\xff\xff\xff\x40\0\0\0");
```
zeenix: same here.

zvariant/src/gvariant/de.rs
```diff
@@ -400,6 +400,7 @@ struct ArrayDeserializer<'d, 'de, 'sig, 'f, F> {
     // Element signature in case of normal array, key signature in case of dict.
     child_signature: &'sig Signature,
     value_signature: Option<&'sig Signature>,
+    should_pad_value_tail: bool,
```
zeenix: I don't think canclulation of this is heavy at all so I'm not sure if it's worth caching this. 🤔

zeenix (changes requested):
LGTM otherwise. Thanks!

zeenix:
> The PR already includes a test case: [1398aec](https://github.com/z-galaxy/zbus/commit/1398aecba0a6486582e1cee1b18ef4421bbbfae9)

Oops, sorry I missed it somehow.

zvariant/src/gvariant/de.rs
kawadakk: Actually it was sufficient to check `self.offsets.is_none()`, which depends on `fixed_sized_child`.

zvariant/tests/struct_value.rs
kawadakk: Updated to use array literals, matching the existing code's style.

zvariant/tests/struct_value.rs
```diff
@@ -35,6 +35,25 @@ fn struct_value() {
     let decoded: as_value::Deserialize<'_, Foo> = encoded.deserialize().unwrap().0;
     assert_eq!(decoded.0, foo);
 
+    // A struct with underaligned tail
```
zeenix: nitpick: missing `.` at the end.

zvariant/src/gvariant/de.rs
```diff
@@ -638,6 +638,11 @@ impl<'d, 'de, 'sig, 'f, #[cfg(unix)] F: AsFd, #[cfg(not(unix))] F> MapAccess<'de
         self.de.0.pos += de.0.pos;
         // No need for retaking the container depths as the value can't be incomplete.
 
+        // A fixed-sized dictionary entry should be padded to its alignment
```
zeenix: missing `.`.

zvariant/src/gvariant/ser.rs
```diff
@@ -324,10 +324,15 @@ where
 
         let seq = self.serialize_seq(len)?;
 
+        // A fixed-sized dictionary entry should be padded to its alignment
```
zeenix: missing `.`.

zvariant/src/gvariant/ser.rs
```diff
@@ -691,6 +696,7 @@ pub struct MapSerializer<'ser, 'b, W> {
     seq: SeqSerializer<'ser, 'b, W>,
     key_signature: &'ser Signature,
     value_signature: &'ser Signature,
+    should_pad_value_tail: bool,
```
zeenix: here, we're still caching this information. Can't this one also be determined where it's needed?

zeenix (changes requested):
LGTM otherwise. Almost there. :)

zeenix (approved):
Thanks!

PR #1824, opened 2026-06-19
✨ zb: Make the tokio and async-io features additive

ogoffart:
Cargo unifies features across the whole build, so the `tokio` feature can be turned on by another, independent crate in the workspace even when the application itself uses zbus with async-io. That used to compile the async-io backend out and force every zbus user in the process into a tokio runtime, so the async-io application would panic with "there is no reactor running".

Gate the async-io backend on `feature = "async-io"` rather than `not(feature = "tokio")` so both can be compiled together. When both are enabled, pick the backend at run time: use tokio when a tokio runtime is active on the current thread, and async-io otherwise.

### DIFF
--- a/zbus/README.md
+++ b/zbus/README.md
@@ -128,6 +128,16 @@ zbus = { version = "5", default-features = false, features = ["tokio"] }
 That's it! No threads launched behind your back by zbus (directly or indirectly) now and no need to
 tick any executors etc. 😼
 
+The `tokio` and `async-io` features are additive: when both are enabled (for example, because
+another crate in your dependency tree enables `tokio`), zbus selects the runtime at run time — using
+tokio when a tokio runtime is driving the current thread, and `async-io` otherwise. With only the
+`tokio` feature enabled (i.e. `async-io` disabled), zbus must be used from a thread running a tokio
+runtime, as there is no `async-io` fallback to drive its I/O.
+
+This run-time selection applies to the async API. The blocking API (`zbus::blocking`) drives its
+connections through its own `block_on`, which uses tokio whenever the `tokio` feature is enabled, so
+those connections always run on tokio when that feature is on.
+
 **Note**: On Windows, the `async-io` feature is currently required for UNIX domain socket support,
 see [the corresponding tokio issue on GitHub][tctiog].
 
--- a/zbus/src/abstractions/executor.rs
+++ b/zbus/src/abstractions/executor.rs
@@ -80,7 +80,7 @@ impl Executor<'_> {
 
     /// Runs a single task.
     ///
-    /// With `tokio` feature enabled, its a noop and never returns.
+    /// With `tokio` feature enabled, it's a noop and never returns.
     pub async fn tick(&self) {
         #[cfg(not(feature = "tokio"))]
         {
--- a/zbus/src/abstractions/executor.rs
+++ b/zbus/src/abstractions/executor.rs
@@ -1,8 +1,12 @@
-#[cfg(not(feature = "tokio"))]
+#[cfg(feature = "async-io")]
 use async_executor::Executor as AsyncExecutor;
-#[cfg(not(feature = "tokio"))]
+#[cfg(feature = "async-io")]
 use async_task::Task as AsyncTask;
-#[cfg(not(feature = "tokio"))]
+#[cfg(feature = "tokio")]
+use std::io::Error;
+#[cfg(not(feature = "async-io"))]
+use std::marker::PhantomData;
+#[cfg(feature = "async-io")]
 use std::sync::Arc;
 use std::{
     future::Future,
@@ -11,8 +15,6 @@ use std::{
     task::{Context, Poll},
 };
 #[cfg(feature = "tokio")]
-use std::{future::pending, io::Error, marker::PhantomData};
-#[cfg(feature = "tokio")]
 use tokio::task::JoinHandle;
 
 /// A wrapper around the underlying runtime/executor.
@@ -20,16 +22,12 @@ use tokio::task::JoinHandle;
 /// This is used to run asynchronous tasks internally and allows integration with various runtimes.
 /// See [`crate::Connection::executor`] for an example of integration with external runtimes.
 ///
-/// **Note:** You can (and should) completely ignore this type when building with `tokio` feature
-/// enabled.
-#[cfg(not(feature = "tokio"))]
+/// **Note:** You can (and should) completely ignore this type when the `tokio` backend is in use.
 #[derive(Debug, Clone)]
 pub struct Executor<'a> {
-    executor: Arc<AsyncExecutor<'a>>,
-}
-#[cfg(feature = "tokio")]
-#[derive(Debug, Clone)]
-pub struct Executor<'a> {
+    #[cfg(feature = "async-io")]
+    async_io: Option<Arc<AsyncExecutor<'a>>>,
+    #[cfg(not(feature = "async-io"))]
     phantom: PhantomData<&'a ()>,
 }
 
@@ -41,87 +39,91 @@ impl Executor<'_> {
         future: impl Future<Output = T> + Send + 'static,
         #[allow(unused)] name: &str,
     ) -> Task<T> {
-        #[cfg(not(feature = "tokio"))]
-        {
-            Task(Some(self.executor.spawn(future)))
+        #[cfg(feature = "async-io")]
+        if let Some(executor) = &self.async_io {
+            return Task::from_async_io(executor.spawn(future));
         }
 
         #[cfg(feature = "tokio")]
-        {
-            #[cfg(tokio_unstable)]
-            {
-                Task(Some(
-                    tokio::task::Builder::new()
-                        .name(name)
-                        .spawn(future)
-                        // SAFETY: Looking at the code, this call always returns an `Ok`.
-                        .unwrap(),
-                ))
-            }
-            #[cfg(not(tokio_unstable))]
-            {
-                Task(Some(tokio::task::spawn(future)))
-            }
-        }
+        return Task::from_tokio(tokio_spawn(future, name));
+
+        #[cfg(all(feature = "async-io", not(feature = "tokio")))]
+        unreachable!("async-io executor is always `Some` when tokio is disabled")
     }
 
     /// Return `true` if there are no unfinished tasks.
     ///
-    /// With `tokio` feature enabled, this always returns `true`.
+    /// With the `tokio` backend in use, this always returns `true`.
     pub fn is_empty(&self) -> bool {
-        #[cfg(not(feature = "tokio"))]
-        {
-            self.executor.is_empty()
+        #[cfg(feature = "async-io")]
+        if let Some(executor) = &self.async_io {
+            return executor.is_empty();
         }
 
-        #[cfg(feature = "tokio")]
         true
     }
 
     /// Runs a single task.
     ///
-    /// With `tokio` feature enabled, it's a noop and never returns.
+    /// With the `tokio` backend in use, it's a noop and never returns.
     pub async fn tick(&self) {
-        #[cfg(not(feature = "tokio"))]
-        {
-            self.executor.tick().await
+        #[cfg(feature = "async-io")]
+        if let Some(executor) = &self.async_io {
+            executor.tick().await;
+            // Skip the `tokio` branch below (only present when both backends are compiled in).
+            #[cfg(feature = "tokio")]
+            return;
         }
 
         #[cfg(feature = "tokio")]
-        {
-            pending().await
-        }
+        std::future::pending::<()>().await;
     }
 
     /// Create a new `Executor`.
     pub(crate) fn new() -> Self {
-        #[cfg(not(feature = "tokio"))]
-        {
-            Self {
-                executor: Arc::new(AsyncExecutor::new()),
-            }
+        Self {
+            #[cfg(feature = "async-io")]
+            async_io: (!super::use_tokio()).then(|| Arc::new(AsyncExecutor::new())),
+            #[cfg(not(feature = "async-io"))]
+            phantom: PhantomData,
         }
+    }
 
-        #[cfg(feature = "tokio")]
-        {
-            Self {
-                phantom: PhantomData,
-            }
-        }
+    /// Whether this executor needs an external driver thread (only the `async-io` backend does).
+    #[cfg(feature = "async-io")]
+    pub(crate) fn needs_internal_driver(&self) -> bool {
+        self.async_io.is_some()
     }
 
     /// Runs the executor until the given future completes.
     ///
-    /// With `tokio` feature enabled, it just awaits on the `future`.
+    /// With the `tokio` backend in use, it just awaits on the `future`.
     pub(crate) async fn run<T>(&self, future: impl Future<Output = T>) -> T {
-        #[cfg(not(feature = "tokio"))]
-        {
-            self.executor.run(future).await
+        #[cfg(feature = "async-io")]
+        if let Some(executor) = &self.async_io {
+            return executor.run(future).await;
         }
-        #[cfg(feature = "tokio")]
-        {
-            future.await
-        }
+
+        future.await
+    }
+}
+
+#[cfg(feature = "tokio")]
+fn tokio_spawn<T: Send + 'static>(
+    future: impl Future<Output = T> + Send + 'static,
+    #[allow(unused)] name: &str,
+) -> JoinHandle<T> {
+    #[cfg(tokio_unstable)]
+    {
+        tokio::task::Builder::new()
+            .name(name)
+            .spawn(future)
+            // SAFETY: Looking at the code, this call always returns an `Ok`.
+            .unwrap()
+    }
+    #[cfg(not(tokio_unstable))]
+    {
+        tokio::task::spawn(future)
     }
 }
 
@@ -132,28 +134,46 @@ impl Executor<'_> {
 /// * it will be cancelled, rather than detached. For detaching, use the `detach` method.
 /// * errors from the task cancellation will will be ignored. If you need to know about task errors,
 ///   convert the task to a `FallibleTask` using the `fallible` method.
-#[cfg(not(feature = "tokio"))]
 #[doc(hidden)]
 #[derive(Debug)]
-pub struct Task<T>(Option<AsyncTask<T>>);
-#[cfg(feature = "tokio")]
-#[doc(hidden)]
-#[derive(Debug)]
-pub struct Task<T>(Option<JoinHandle<T>>);
+pub struct Task<T> {
+    #[cfg(feature = "async-io")]
+    async_io: Option<AsyncTask<T>>,
+    #[cfg(feature = "tokio")]
+    tokio: Option<JoinHandle<T>>,
+}
 
 impl<T> Task<T> {
+    #[cfg(feature = "async-io")]
+    fn from_async_io(task: AsyncTask<T>) -> Self {
+        Self {
+            async_io: Some(task),
+            #[cfg(feature = "tokio")]
+            tokio: None,
+        }
+    }
+
+    #[cfg(feature = "tokio")]
+    fn from_tokio(handle: JoinHandle<T>) -> Self {
+        Self {
+            #[cfg(feature = "async-io")]
+            async_io: None,
+            tokio: Some(handle),
+        }
+    }
+
     /// Detaches the task to let it keep running in the background.
     #[allow(unused_mut)]
-    #[allow(unused)]
     pub fn detach(mut self) {
-        #[cfg(not(feature = "tokio"))]
-        {
-            self.0.take().expect("async_task::Task is none").detach()
+        #[cfg(feature = "async-io")]
+        if let Some(task) = self.async_io.take() {
+            task.detach();
         }
 
         #[cfg(feature = "tokio")]
-        {
-            self.0.take().expect("tokio::task::JoinHandle is none");
+        if let Some(handle) = self.tokio.take() {
+            // Dropping a tokio `JoinHandle` detaches it.
+            drop(handle);
         }
     }
 }
@@ -163,43 +183,46 @@ where
     T: Send + 'static,
 {
     /// Launch the given blocking function in a task.
+    ///
+    /// `blocking::unblock` needs no runtime, so async-io's pool is used unless a tokio runtime is
+    /// active.
     #[allow(unused)]
     pub(crate) fn spawn_blocking<F>(f: F, #[allow(unused)] name: &str) -> Self
     where
         F: FnOnce() -> T + Send + 'static,
     {
-        #[cfg(not(feature = "tokio"))]
-        {
-            Self(Some(blocking::unblock(f)))
+        super::select_runtime! {
+            tokio: Self::from_tokio(tokio_spawn_blocking(f, name)),
+            async_io: Self::from_async_io(blocking::unblock(f)),
         }
+    }
+}
 
-        #[cfg(feature = "tokio")]
-        {
-            #[cfg(tokio_unstable)]
-            {
-                Self(Some(
-                    tokio::task::Builder::new()
-                        .name(name)
-                        .spawn_blocking(f)
-                        // SAFETY: Looking at the code, this call always returns an `Ok`.
-                        .unwrap(),
-                ))
-            }
-            #[cfg(not(tokio_unstable))]
-            {
-                Self(Some(tokio::task::spawn_blocking(f)))
-            }
-        }
+#[cfg(feature = "tokio")]
+fn tokio_spawn_blocking<F, T>(f: F, #[allow(unused)] name: &str) -> JoinHandle<T>
+where
+    F: FnOnce() -> T + Send + 'static,
+    T: Send + 'static,
+{
+    #[cfg(tokio_unstable)]
+    {
+        tokio::task::Builder::new()
+            .name(name)
+            .spawn_blocking(f)
+            // SAFETY: Looking at the code, this call always returns an `Ok`.
+            .unwrap()
+    }
+    #[cfg(not(tokio_unstable))]
+    {
+        tokio::task::spawn_blocking(f)
     }
 }
 
 impl<T> Drop for Task<T> {
     fn drop(&mut self) {
         #[cfg(feature = "tokio")]
-        {
-            if let Some(join_handle) = self.0.take() {
-                join_handle.abort();
-            }
+        if let Some(join_handle) = self.tokio.take() {
+            join_handle.abort();
         }
     }
 }
@@ -208,24 +231,16 @@ impl<T> Future for Task<T> {
     type Output = Result<T>;
 
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        #[cfg(not(feature = "tokio"))]
-        {
-            Pin::new(&mut self.get_mut().0.as_mut().expect("async_task::Task is none"))
-                .poll(cx)
-                .map(|r| Ok(r))
+        let this = self.get_mut();
+
+        #[cfg(feature = "async-io")]
+        if let Some(task) = &mut this.async_io {
+            return Pin::new(task).poll(cx).map(Ok);
         }
 
         #[cfg(feature = "tokio")]
-        {
-            Pin::new(
-                &mut self
-                    .get_mut()
-                    .0
-                    .as_mut()
-                    .expect("tokio::task::JoinHandle is none"),
-            )
-            .poll(cx)
-            .map(|r| match r {
+        if let Some(handle) = &mut this.tokio {
+            return Pin::new(handle).poll(cx).map(|r| match r {
                 Ok(v) => Ok(v),
                 Err(e) => {
                     if e.is_cancelled() {
@@ -234,7 +249,9 @@ impl<T> Future for Task<T> {
                         panic!("tokio::task::JoinHandle error: {e}")
                     }
                 }
-            })
+            });
         }
+
+        unreachable!("Task always has exactly one backend")
     }
 }
--- a/zbus/src/abstractions/mod.rs
+++ b/zbus/src/abstractions/mod.rs
@@ -64,3 +64,16 @@ pub(crate) mod timeout;
 // Not unix-specific itself but only used on unix.
 #[cfg(target_family = "unix")]
 pub(crate) mod process;
+
+#[cfg(all(test, feature = "tokio", feature = "async-io"))]
+mod tests {
+    #[test]
+    fn use_tokio_reflects_active_runtime() {
+        assert!(!super::use_tokio(), "no runtime is active here");
+        let runtime = tokio::runtime::Runtime::new().unwrap();
+        assert!(
+            runtime.block_on(async { super::use_tokio() }),
+            "a tokio runtime is active",
+        );
+    }
+}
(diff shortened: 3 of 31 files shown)


zbus/src/connection/mod.rs
```diff
@@ -1698,7 +1720,7 @@ mod p2p_tests {
     }
 
     #[cfg(any(
-        all(feature = "vsock", not(feature = "tokio")),
+        all(feature = "vsock", not(feature = "tokio-vsock")),
         feature = "tokio-vsock"
```
zeenix: are you sure about this?

zeenix (changes requested):
Thanks @ogoffart . Could you please split this a bit into multiple commits. E.g some docs and test additons, and CI addition are certainly obvious candidates for going into their own commits. Perhpas there are other changes that can split too?

zbus/src/abstractions/executor.rs
```diff
@@ -41,87 +39,91 @@ impl Executor<'_> {
...
         }
 
-        #[cfg(feature = "tokio")]
         true
     }
 
     /// Runs a single task.
     ///
-    /// With `tokio` feature enabled, its a noop and never returns.
+    /// With the tokio backend in use, it's a noop and never returns.
```
zeenix: certainly an unrelated change and hence a separate commit please.

zbus/src/abstractions/mod.rs
```diff
@@ -1,7 +1,53 @@
...
+//! enabled.
+
+/// Whether zbus should use tokio (rather than `async-io`) for its I/O.
+///
+/// Only consulted when `async-io` is compiled in, since that's the only time there's a choice. With
+/// both backends we use tokio when a tokio runtime is active on the current thread; with only
+/// `async-io` the answer is always `false`. Keeping the features additive this way means enabling
+/// `tokio` elsewhere in the dependency graph no longer forces every zbus user into a tokio runtime.
+#[cfg(feature = "async-io")]
+pub(crate) fn use_tokio() -> bool {
```
zeenix: bit nitpick: We try to follow top-down code layout, [similar to rustls](https://github.com/rustls/rustls/blob/main/CONTRIBUTING.md#top-down-ordering-within-modules) so for pub before private and usage before definition. Here the macro should be defined first.

zbus/src/connection/mod.rs
ogoffart: Indeed, the `not(tokio-vsock)` is pointless

zbus/src/abstractions/executor.rs
ogoffart: It is related as now it is not a noop if the feature is enabled but tokio isn't used at runtime.

But i moved it to the documentation commit

zeenix: > It is related as now it is not a noop if the feature is enabled but tokio isn't used at runtime.

Ah, you removed the backticks from tokio and added `the` and missing `'` to `its` so I thought you're only fixing typos here. Best to only stick to the actual change in the main commit and split fixes to typos into a separate commit (which should come first so we can easily revert the main/big change if needed, w/o also having to remove/rework the fixes).

zeenix:
## Review by Claude (Anthropic), requested by @zeenix

_The following review was generated by Claude. @zeenix is relaying it; the
analysis and any mistakes in it are the model's, so please sanity-check before
acting on it._

Overall this is solid and the approach is right: gate on `feature = "async-io"`
instead of `not(feature = "tokio")`, type-erase the chosen backend into
`BoxedSplit` at connect time, and only spawn the internal driver thread when an
async-io executor actually exists (`needs_internal_driver()`). The executor
rewrite is clean and the `Drop`/`detach`/`poll` invariants hold (exactly one
backend field is `Some`, `unreachable!()` arms are genuinely unreachable).

### Main concern: runtime selection is applied to unix sockets but not TCP/vsock

The decision point is inconsistent across transports:

- `address/transport/mod.rs` — `unix_stream_to_split()` branches on
  `use_tokio()` at connect time. Good.
- But `Tcp::connect()` in `transport/tcp.rs` is gated `#[cfg(feature =
  "async-io")]`, with the tokio variant `#[cfg(all(feature = "tokio",
  not(feature = "async-io")))]`. So with **both** features enabled, TCP
  *always* produces an `Async<TcpStream>` regardless of `use_tokio()`.
- Same in `connection/builder.rs`: `Target::UnixStream` now routes through
  `unix_stream_to_split()`, but `Target::TcpStream` is still an unconditional
  `Async::new(stream)?.into()` under `#[cfg(feature = "async-io")]`.

Consequence: a tokio app that pulls in `async-io` transitively, connecting over
TCP, gets async-io sockets driven by async-io's global reactor thread — even
though a tokio runtime is active and `Executor::new()` correctly set
`async_io: None` (so no zbus driver thread). It doesn't panic (async-io's
reactor starts lazily and `Async<T>` is runtime-agnostic), but it silently runs
the "wrong" backend plus an extra reactor thread, which is the kind of surprise
this PR is otherwise removing.

Since the motivating bug is the async-io-primary direction, this may be an
acceptable scope cut — but if so it should be documented, because the README/FAQ
additions read as if selection is symmetric. Ideally TCP (and the
`Builder::tcp_stream` path) mirror the unix path; the `unix_stream_to_split`
shape extends naturally. vsock is a separate axis (`vsock` vs `tokio-vsock` are
distinct impls, not a runtime choice), so I'd leave it but note it.

### `select_runtime!` is evaluated per-call, not per-connection

`timeout()` and `spawn_blocking()` re-evaluate `use_tokio()` on every call, so
the timer/blocking-pool backend can diverge from the backend the connection was
built with if the calling thread's runtime context differs. It's benign today
(timers and `blocking::unblock` are independent of the socket's reactor), but
it's a latent foot-gun: if any future `select_runtime!` site ever needs to match
the socket's backend, this thread-local check won't guarantee that. Worth a
comment stating the contract: the load-bearing decision is made once at build
time on the building thread, and these sites are only safe because their choice
is independent of the socket.

### Commit hygiene

`unix_stream_to_tokio()` adds `stream.set_nonblocking(true)?` before
`from_std` — a real latent fix (the old tokio path omitted it), not part of the
additive-features change. By the same rule you applied to the doc typo, this is
a separate fix that ideally lands in its own commit ordered before the main one.

### Smaller notes

- `select_runtime!` is defined before `use_tokio()` now — top-down ordering
  addressed. 👍
- `tick()` control flow is correct across all three cfg combinations; the
  `#[cfg(feature = "tokio")] return;` after `executor.tick().await` is load-
  bearing and easy to miss — a one-line comment would help (you already added
  one, good).
- The CI job exercises `--features tokio,p2p`; the new `unix_p2p_*` tests cover
  both `use_tokio()` branches via `async_io::block_on` vs `#[tokio::test]`,
  which is exactly the right way to hit both arms. Nice.

ogoffart:
The review was useful.  I hope the windows build is fixed now (i didn't test windows myself)
And I also fixed the vsock thing  (but i'm not very familiar with that so i hope it is right)

ogoffart:
Updated.  The previous one had actually a bug that it changed the signature of `Builder::unix_stream` by changing the type it receive, but it is now fixed.
(This is a pre-existing problem that the type it receive changes between feature)

zbus/src/address/transport/mod.rs
```diff
@@ -263,27 +251,56 @@ impl Transport {
...
+    }
+
+    Ok(stream.into())
+}
+
+#[cfg(feature = "vsock")]
+fn vsock_connect_async_io(addr: &Vsock) -> Result<Stream> {
+    let stream = vsock::VsockStream::connect_with_cid_port(addr.cid(), addr.port())?;
+    Async::new(stream)
+        .map(|s| Stream::Vsock(s.into()))
```
zeenix: **Compile break: `vsock` without the `async-io` feature no longer builds.**

`vsock_connect_async_io` is gated on `feature = "vsock"` and uses bare `Async::new`, but the `use async_io::Async;` import (line 9) is gated on `feature = "async-io"`. Since `Cargo.toml` defines `vsock = ["dep:vsock", "dep:async-io"]` — which enables the async_io *crate* but **not** the `async-io` *feature* — a build with `vsock` but without the feature strips the import while compiling this function.

Reproduced with `cargo check -p zbus --no-default-features --features tokio,vsock` (Linux):

```
error[E0433]: failed to resolve: use of undeclared type `Async`
   --> zbus/src/address/transport/mod.rs:293
error[E0277]: the trait bound `Async<VsockStream>: Into<BoxedSplit>` is not satisfied
   --> zbus/src/address/transport/mod.rs:294
```

(The second error is the missing `impl Socket for Async<T>`, also gated on `feature = "async-io"` — see the builder.rs note.)

Interestingly `builder.rs` already widened its own `Async` import to `any(not(tokio), all(vsock, not(tokio-vsock)))`, so this looks like a spot that was missed. Fix options: make `vsock` (and `tokio-vsock`? no) enable the `async-io` feature rather than just `dep:async-io`, or widen the `Async` import here and the `Socket for Async<T>` impl gate to `any(feature = "async-io", feature = "vsock")`. Might be worth a `tokio,vsock` CI matrix entry, since the added CI line only covers `tokio,p2p`.

---

zbus/src/connection/builder.rs
```diff
@@ -691,23 +688,20 @@ impl<'a> Builder<'a> {
             Target::TcpStream(stream) => Async::new(stream)?.into(),
             #[cfg(feature = "tokio")]
             Target::TcpStream(stream) => stream.into(),
-            #[cfg(all(feature = "vsock", not(feature = "tokio")))]
+            #[cfg(all(feature = "vsock", not(feature = "tokio-vsock")))]
             Target::VsockStream(stream) => Async::new(stream)?.into(),
```
zeenix, line 753: Second symptom of the same root cause (see the `transport/mod.rs` note): with `--no-default-features --features tokio,vsock`, `Async::new(stream)?.into()` fails because the only `impl<T> Socket for Async<T>` (`connection/socket/mod.rs:415`) is gated on `feature = "async-io"`, which `vsock` alone doesn't enable:

```
error[E0277]: the trait bound `Async<VsockStream>: Into<BoxedSplit>` is not satisfied
   --> zbus/src/connection/builder.rs:692
```

Here the `Async` import is correctly gated, so only the `.into()` fails — but the same feature fix resolves both.

---

zbus/src/abstractions/executor.rs
```diff
@@ -41,87 +39,91 @@ impl Executor<'_> {
...
     /// Create a new `Executor`.
     pub(crate) fn new() -> Self {
-        #[cfg(not(feature = "tokio"))]
-        {
-            Self {
-                executor: Arc::new(AsyncExecutor::new()),
-            }
+        Self {
+            #[cfg(feature = "async-io")]
+            async_io: (!super::use_tokio()).then(|| Arc::new(AsyncExecutor::new())),
```
zeenix, line 86: **Backend choice is latched here but re-derived per-call elsewhere.**

`Executor::new()` decides the backend exactly once, based on `use_tokio()` on the thread/moment the `Connection` is built, and `spawn` / `is_empty` / `run` / `tick` all consult that latched `async_io` field. But `timeout()` (`abstractions/timeout.rs:40`) and `Task::spawn_blocking` re-evaluate `use_tokio()` at *every* call via `select_runtime!`. In single-runtime apps these always agree, but now that both backends can coexist they can diverge under mixed usage.

Concrete footgun: build a connection outside any tokio runtime (e.g. `zbus::blocking::Connection` at startup → async-io backend, own driver thread), then await its async futures from within a tokio current-thread runtime created *without* the time driver. The method-call `timeout()` sees `Handle::try_current().is_ok() == true`, routes to `tokio::time::timeout`, and panics (`A Tokio 1.x context was found, but timers are disabled`) even though the connection runs entirely on async-io. `spawn_blocking` similarly binds work to a foreign tokio runtime whose shutdown then fails the op.

The trigger is unusual (single-runtime apps are fine), but it's a new fragility. Since `timeout`/`spawn_blocking` are free functions with no executor handle, consider threading the latched decision through rather than re-sniffing the ambient runtime.

---

zeenix (commented):
Nice cleanup — making the backends additive and picking at run time is the right fix for the "there is no reactor running" panic. I applied the diff locally and compiled the affected feature configs: `tokio` only, `async-io,vsock`, default+`tokio`, `tokio,tokio-vsock`, and `tokio,async-io,vsock` all build clean, and `cargo clippy` on the both-backends config (`tokio,p2p`) is warning-free.

Two things stood out (inline):

1. **`--features tokio,vsock` (without the `async-io` feature) fails to compile** — confirmed by `cargo check`. `vsock` pulls in the async_io crate but not the `async-io` feature, while the `Async` import and the `Socket for Async<T>` impl the vsock path needs are gated on the feature. `builder.rs` was partially adjusted for this but `transport/mod.rs` / `socket/mod.rs` weren't. This is the one I'd consider a blocker.
2. **Backend selection latched at `Executor::new()` vs re-derived per-call in `timeout`/`spawn_blocking`** — can diverge under mixed-runtime usage (async-io-built connection driven from a tokio runtime), with a possible "timers are disabled" panic. Narrow trigger, but a new fragility worth a look.

Everything else checked out: `Task::detach` vs `Drop` correctly detaches tokio handles without aborting, the `unreachable!()` arms are genuinely unreachable per config, `tick()`'s both-backends flow returns correctly instead of falling into `pending()`, and the TCP nonce refactor is behavior-preserving.

---

zbus/src/abstractions/executor.rs
zeenix: Amendment / correction: this largely restates a point already made in the earlier "Review by Claude" comment (_"`select_runtime!` is evaluated per-call, not per-connection"_), which correctly judged it **benign** in normal use — timers and `blocking::unblock` are independent of the socket's reactor, so an async-io connection using a tokio timer (or vice versa) is fine. I agree with that assessment; please treat the above as a footnote to it rather than a fresh finding.

The one non-benign edge worth keeping in mind: if the ambient tokio runtime is a current-thread runtime built *without* the time driver, `timeout()` routing to `tokio::time::timeout` will panic. That's a narrow, arguably-user-error case, not a blocker. The substantive follow-up from that earlier review — that runtime selection is applied to unix sockets but not TCP — is the more useful one to act on.

---

zbus/src/connection/builder.rs
```diff
@@ -1,5 +1,8 @@
 use async_broadcast::Receiver as ActiveReceiver;
-#[cfg(not(feature = "tokio"))]
+#[cfg(any(
```
cachebag: This still leaves the explicit stream builder APIs non-additive.

@zeenix Should this probably be explicitly documented as still feature-selected?

zbus/src/connection/builder.rs
```diff
@@ -19,7 +22,7 @@ use tokio::net::UnixStream;
 use tokio_vsock::VsockStream;
 #[cfg(all(windows, not(feature = "tokio")))]
 use uds_windows::UnixStream;
-#[cfg(all(feature = "vsock", not(feature = "tokio")))]
+#[cfg(all(feature = "vsock", not(feature = "tokio-vsock")))]
```
cachebag, line 22: Same thing here, callers who already have a non-tokio stream lose that API when the tokio-specific feature is enabled.

cachebag (changes requested):
A couple questions..

zbus/src/connection/builder.rs
ogoffart: Right, this is an API that changes depending on the feature :-(

My suggestion would be to split this into two function:  `tokio_unix_stream`  and `std_unix_stream` ,  and deprecate this function.

But this could be done in another change.

zbus/src/address/transport/mod.rs
ogoffart: Fixed by making vsock enable the async-io feature

cachebag (approved):

zbus/src/connection/builder.rs
zeenix: @cachebag Good catch! Thanks.

I think this could lead to some very "interesting" issues.

> My suggestion would be to split this into two function: `tokio_unix_stream` and `std_unix_stream` , and deprecate this function.

Unfortunately that exposes the difference on the API-level, which is something I worked very hard to avoid.

Also, and more importantly, deprecation won't solve the issue for existing users. :( Deprecated API should still work as advertised.

---

I am now thinking if doing this properly requires bumping the semver. 🤔 We do already have [a few things piled up](https://github.com/z-galaxy/zbus/milestone/7) that will need a semver bump.

ogoffart: > Unfortunately that exposes the difference on the API-level, which is something I worked very hard to avoid.

There is no way to avoid it.  The types themselves are different.
Maybe a trait would work if we really want to keep the same name. But i'm thinking in this case having two different method is the right way to do.

> Also, and more importantly, deprecation won't solve the issue for existing users. :(

True.
But depreciation notice may guide existing user to use the version that won't break if features get enabled. Making it better for the ecosystem IMHO.

(Note that this PR doesn't change anything for existing users)

ogoffart:
(Rebased to fix CI after rust update)

zbus/src/connection/builder.rs
zeenix: > (Note that this PR doesn't change anything for existing users)

Right, it does but only for cases where both backends get enabled but then things will be broken anyway.

zeenix: > > Also, and more importantly, deprecation won't solve the issue for existing users. :(
> 
> True. But depreciation notice may guide existing user to use the version that won't break if features get enabled. Making it better for the ecosystem IMHO.

Ok, let's go this way then (at least for now). I think this should be part of this PR as w/o these changes, the PR doesn't achieve the goal in its title. Could you please add these changes?

ogoffart: > I think this should be part of this PR as w/o these changes, the PR doesn't achieve the goal in its title. 

I think it makes the scope a bit broader But Ok, i've done the change:
I added `async_io_unix_stream` and `async_io_tcp_stream`. (I didn't name them `std_` because i ralized on windows it is a different type.
I also kept the normal one for tokio as this can't break when using it with tokio.
See the two last commits of this PR.

Let me know if you prefer different names.

zbus/src/connection/builder.rs
zeenix: was this addressed?

zbus/src/connection/builder.rs
zeenix: Verified fixed: `vsock = ["dep:vsock", "async-io"]` now enables the `async-io` *feature*, and `cargo check -p zbus --no-default-features --features tokio,vsock` (the original repro) compiles cleanly. Resolving.

---

zbus/src/abstractions/executor.rs
zeenix: Closing this thread: both follow-ups from the original review have landed. `select_runtime!` now carries the contract comment (per-call re-evaluation is only safe for call sites independent of the socket's reactor — `zbus/src/abstractions/mod.rs`), and the substantive point — runtime selection missing for TCP — is fixed via `tcp_async_to_split()` in `zbus/src/address/transport/mod.rs`. The remaining timers-disabled-runtime edge was already agreed to be narrow user-error territory, not a blocker.

---

zbus/src/connection/builder.rs
zeenix: Yes, this was addressed by the last two commits (1de6f10, 284d81a): `async_io_unix_stream` and `async_io_tcp_stream` were added to both the async and blocking builders. They take feature-independent types (`std::os::unix::net::UnixStream`/`uds_windows::UnixStream`, `std::net::TcpStream`), so a caller holding a non-tokio stream keeps a working API regardless of whether `tokio` gets enabled elsewhere in the dependency graph. The old `unix_stream`/`tcp_stream` remain for tokio streams and are deprecated in the `not(tokio)` configuration, pointing users at the additive variants. Resolving.

---

zeenix:
## Review by Claude (Anthropic), round 2 — requested by @zeenix

_As before: @zeenix is relaying this; the analysis and any mistakes are the model's, so please sanity-check before acting on it._

First, status of the previous rounds: **all prior findings are addressed** and their threads have been resolved — the `vsock`-without-`async-io` compile break (fixed via `vsock = ["dep:vsock", "async-io"]`, repro re-verified), TCP runtime selection (`tcp_async_to_split()`), the `select_runtime!` contract comment, and the non-additive stream builders (`async_io_unix_stream`/`async_io_tcp_stream` + deprecations). This round only surfaces things not raised before.

### 1. `utils::block_on` is the one remaining compile-time backend switch (`zbus/src/utils.rs:31`)

Every other `not(feature = "tokio")` gate was converted to `async-io` / `all(tokio, not(async-io))`, but `block_on` — and with it the entire `zbus::blocking` API, whose `Builder::build()` is literally `block_on(self.0.build())` — still selects on `#[cfg(feature = "tokio")]`. With **both** features enabled it takes the tokio path: it lazily creates the global multi-thread runtime and, since `Runtime::block_on` enters the runtime context, `use_tokio()` is `true` for everything inside — so **blocking connections always latch the tokio backend**.

For the motivating scenario (async-io app + transitively-enabled `tokio`), the async API stays on async-io as intended, but the blocking API silently migrates to tokio and zbus spawns a hidden `num_cpus`-worker runtime. It works — this is just the pre-existing tokio-only blocking behavior — but it's the same "another crate flipped my backend" surprise this PR sets out to remove, and the new README/FAQ text ("uses tokio when a tokio runtime is driving the current thread") is misleading here, because the runtime driving the thread is one zbus itself created.

I don't think a mechanical flip to async-io-preferred (`#[cfg(feature = "async-io")]` → `async_io::block_on`) is obviously right either: it would change behavior for the many existing default-features+`tokio` users in the *other* direction, and the in-tree tests currently rely on `utils::block_on` establishing a tokio context (the `#[cfg(feature = "tokio")]` branches of `tcp_p2p`/`unix_p2p_pipe` create tokio sockets inside it). So this is a genuine design decision: either pick a selection rule for `block_on` under dual features and adjust tests, or keep the current behavior and document the blocking-API asymmetry in the README/FAQ. Right now it's silent.

### 2. The async-io arm has no end-to-end coverage when both backends are compiled in

A direct consequence of the above: in the new CI entry (`--features tokio,p2p,vsock`), every integration test runs inside a tokio context via `utils::block_on`, so `use_tokio()` is `true` throughout and the tests exercise the *tokio* arm of the runtime selection. The tests' async-io branches (e.g. `async_io_unix_stream` in `unix_p2p_pipe`, `connection/mod.rs:1694`) are gated `#[cfg(not(feature = "tokio"))]`, so they compile out of that run entirely. The flagship scenario — both features compiled, **no** tokio runtime active — is only covered by the `use_tokio_reflects_active_runtime()` unit test; no connection is ever built on the async-io fallback in that configuration.

Suggestion: one p2p test gated `all(feature = "tokio", feature = "async-io")` that builds via `async_io::block_on` (bypassing `utils::block_on`) and ideally asserts the executor latched async-io (`needs_internal_driver()`). That would also exercise `start_internal_executor` in the dual-feature config.

### 3. Nit: `deprecated(since = "5.17.0")`

`main` is already at 5.18.0 (zbus-5.18.0 is tagged/released), so the four new deprecations (async + blocking × `unix_stream`/`tcp_stream`) will first ship in a later release. `since` should be whatever version release-plz cuts next, otherwise the attribute claims the deprecation existed in versions that shipped without it.

### Verified locally (Linux + cross-checks)

- `cargo check -p zbus --no-default-features --features tokio,vsock` ✅ (previously-broken combo from round 1)
- `cargo check -p zbus --features tokio` and `--features tokio,vsock,tokio-vsock` ✅
- Windows `x86_64-pc-windows-gnu`: default, `--features tokio`, `--no-default-features --features tokio` ✅
- clippy clean on default / dual-backend / tokio-only configs
- Full test suite with `--features tokio,p2p,vsock` under `dbus-run-session`: all binaries green. (The `vsock_*` and `unixexec` tests fail in my container, but identically without this PR — missing `vsock_loopback` module and a `systemd-stdio-bridge` handshake EOF respectively, i.e. environmental.)

Aside: the CodSpeed regression flagged on this PR is in `zbus_names::parse_name`, which this PR doesn't touch — the report itself warns the comparison crossed runner environments, so it looks like noise that can be acknowledged.

---

zbus/README.md
```diff
@@ -128,6 +128,12 @@ zbus = { version = "5", default-features = false, features = ["tokio"] }
 That's it! No threads launched behind your back by zbus (directly or indirectly) now and no need to
 tick any executors etc. 😼
 
+The `tokio` and `async-io` features are additive: when both are enabled (for example, because
```
zeenix, line 131: This paragraph is accurate for the async API but not for `zbus::blocking`. `utils::block_on` is the one gate left switching on `#[cfg(feature = "tokio")]` instead of on `async-io`, and because `Runtime::block_on` enters the runtime context, everything built through the blocking API sees `use_tokio() == true`. Confirmed in a `tokio,async-io` build:

```rust
assert!(tokio::runtime::Handle::try_current().is_err());  // passes: no ambient runtime
assert!(zbus::block_on(async { tokio::runtime::Handle::try_current().is_ok() }));  // also passes
```

So for an async-io app whose dependency tree turns on `tokio`, the async connections correctly stay on async-io — the fix works — but every `zbus::blocking::Connection` silently moves to tokio, with zbus spinning up a global multi-threaded runtime behind its back. That's the same "another crate flipped my backend" surprise this paragraph says is gone, and the reason it reads wrong here is that the runtime "driving the current thread" is one zbus itself created.

Not a regression — it's what the `tokio` feature has always done — so I don't think it blocks. But either `block_on` gets a selection rule under dual features, or this paragraph and the FAQ one should carve out the blocking API explicitly. Right now it's silent.

---
_Generated by Claude Opus 5._

zbus/src/connection/builder.rs
```diff
@@ -189,37 +200,84 @@ impl<'a> Builder<'a> {
...
     /// Create a builder for a connection that will use the given VSOCK stream.
     ///
     /// This method is only available when either `vsock` or `tokio-vsock` feature is enabled. The
     /// type of `stream` is `vsock::VsockStream` with `vsock` feature and `tokio_vsock::VsockStream`
     /// with `tokio-vsock` feature.
-    #[cfg(any(
-        all(feature = "vsock", not(feature = "tokio")),
-        feature = "tokio-vsock"
-    ))]
+    #[cfg(any(feature = "vsock", feature = "tokio-vsock"))]
```
zeenix, line 280: With `unix_stream`/`tcp_stream` sorted, `vsock_stream` is the last builder whose parameter type flips under feature unification: a `vsock` user passes `vsock::VsockStream`, and the moment anything in the tree enables `tokio-vsock` it becomes `tokio_vsock::VsockStream` and their call stops compiling. Exactly the failure mode the rest of this PR fixes, and `Transport::Vsock` already does the run-time pick between the two, so an `async_io_vsock_stream` would fall out the same shape as the other two.

The doc comment above does say the type varies, and round 1 scoped vsock out as a separate axis, so leaving it is defensible — but now that the PR title promises additivity, this is worth a follow-up issue rather than quietly staying the exception.

---
_Generated by Claude Opus 5._

zbus/src/abstractions/process.rs
```diff
@@ -1,17 +1,17 @@
...
+#[cfg(all(feature = "tokio", not(feature = "async-io")))]
 use tokio::process::Child;
 
 use crate::address::transport::Unixexec;
 
 /// A wrapper around the command API of the underlying async runtime.
 pub struct Command(
-    #[cfg(not(feature = "tokio"))] async_process::Command,
-    #[cfg(feature = "tokio")] tokio::process::Command,
+    #[cfg(feature = "async-io")] async_process::Command,
```
zeenix, line 17: Smallest leftover: `unixexec` is the one transport that doesn't get the run-time treatment. With both features compiled in, `Command` is always `async_process`, never tokio's, regardless of `use_tokio()` — unlike unix/TCP/vsock, which all select now.

It works (async-process brings its own reactor and reaper, and the pipes are `Async<_>`, which is runtime-agnostic), so this is consistency rather than correctness. But it does mean a tokio app connecting over a `unixexec:` address quietly gains an async-io reactor thread. Fine to leave if that's deliberate — a one-line comment saying so would save the next reader the same trip.

---
_Generated by Claude Opus 5._

zeenix (commented):
## Review by Claude (Anthropic), round 3 — requested by @zeenix

_As before: @zeenix is relaying this; the analysis and any mistakes in it are the model's, so please sanity-check before acting on it._

**Nothing blocking.** Round 2's items 2 and 3 landed: the async-io arm now has real end-to-end coverage in the dual-feature build (`unix_p2p_async_io_backend`, which also asserts the executor actually latched async-io — nice), and the deprecations moved to `since = "5.19.0"`. Item 1 (`utils::block_on`) is the one that didn't get a response; details inline on the README, and I don't think it blocks — a sentence of docs closes it.

### Verified locally

- `cargo check -p zbus --no-default-features` for `tokio`, `async-io`, `tokio,async-io`, `tokio,vsock`, `tokio,tokio-vsock`, `tokio,async-io,vsock,tokio-vsock`, `tokio,p2p,vsock` — all clean.
- `cargo clippy --all-targets` on `p2p,vsock,bus-impl`, `tokio,p2p,vsock`, `tokio,p2p`, `async-io,p2p` — clean.
- `cargo +nightly fmt --all --check` — clean.
- `x86_64-pc-windows-gnu`: default, `tokio`, `tokio,async-io`, `async-io` — clean. The Windows `Transport::Unix` arm Copilot flagged in round 1 is right now: async-io drives the socket there, `Err(Error::Unsupported)` when it's off.
- `cargo test --release -p zbus --lib --no-default-features --features tokio,p2p,vsock` under `dbus-run-session` — 38/38, including `unix_p2p_async_io_backend`, `unix_p2p`, `tcp_p2p`, `vsock_connect`, `vsock_p2p`.

Also confirmed the TCP hand-off is sound: `Async::into_inner` deregisters from the reactor without restoring blocking mode, so the fd handed to `TcpStream::from_std` in `tcp_async_to_split` is still non-blocking, as tokio requires.

### Commit contents vs. titles

Much better split than round 1. Three mismatches:

1. **`4cc37d47 📝 zb: Document the tokio and async-io runtime selection` — the message describes two changes that aren't in it.** The commit touches only `zbus/README.md` and `book/src/faq.md`, but the body also claims "select the stream type the same way in the peer-to-peer example" (that's `book/src/connection.md`, which actually lands in `a279ccc1`) and "Describe the `Executor` methods by the active backend" (those rewrites — `/// With the `tokio` backend in use, …` — are in `f86481da`). The book hunk can't move earlier, since it uses `async_io_unix_stream` which doesn't exist until `dc339ed4`, so the fix is just to trim the message down to what's there.

2. **`dc339ed4 ✨ zb: Add Builder::async_io_unix_stream and async_io_tcp_stream` also adds a new end-to-end test** — `unix_p2p_async_io_backend` + `async_io_unix_p2p_pipe`, +35 lines in `connection/mod.rs` — which the message doesn't mention. It genuinely can't go in the earlier ✅ commit because it needs the new API, so either a line in the body or split it into its own ✅ commit after this one.

3. **`f86481da` changes the feature graph silently.** `vsock = ["dep:vsock", "async-io"]` makes `vsock` imply the `async-io` *feature*, not just the crate. That's the fix for the round-1 compile break and it changes what a `tokio,vsock` user builds, so it deserves a sentence in the body.

`db0f101f` (typo first, keeping the big change revertible), `6281cab5`, `a279ccc1` and `371fe4e1` all match their titles. The test-cfg churn inside `f86481da` (`connection/mod.rs`, `proxy/mod.rs`, `handshake/mod.rs`, `address/mod.rs`) is what makes that commit compile, so it belongs there.

---
_Generated by Claude Opus 5._

zbus/src/connection/builder.rs
zeenix: We don't need to fix this in this PR at all (or maybe not at all). I only know of one person who uses vsock feature and I'm sure they don't run into this combo.

Still, worth mentioning in the relevant commit message.

zeenix (approved):

PR #1839, opened 2026-07-03
♻️ zx: Drop quick-xml in favour of a bespoke XML parser

zeenix:
D-Bus introspection documents only use a tiny subset of XML — elements and attributes — so a full-blown XML library is overkill. This replaces quick-xml with a small purpose-built parser (plus a matching writer) in `zbus_xml`, built on **winnow** — which zvariant, zbus_names and zbus already depend on — so the low-level parsing machinery lives in a widely-used, well-audited crate rather than hand-rolled scanning code.

The series: tidy the error module docs, add the real-world test corpus and benchmarks, add the (winnow-based) parser, port to it, test against the real-world documents, deprecate the quick-xml error variants, bump to 5.2.0, drop the dependency, and bound the signature parser's recursion.

## What the parser supports

Exactly what introspection XML needs:

- elements, attributes, entity/character references (`&lt;`, `&#66;`, …) in attribute values, with XML attribute-value whitespace normalization (literal tab/newline/CR → space; character-referenced whitespace kept)
- character references are validated against the XML `Char` production (so e.g. `&#0;` cannot smuggle a NUL into names/values) and signs are rejected; duplicate attributes are rejected
- skipping over everything else a real-world document may carry: XML declaration, doctype (including internal subsets and quoted literals), comments, CDATA, processing instructions, text content, and foreign elements such as Telepathy&#39;s `<tp:docstring>`
- the parser recognises the introspection grammar directly — one small winnow combinator per element (`<node>`, `<interface>`, `<method>`, `<signal>`, `<property>`, `<arg>`, `<annotation>`) decides which attributes are required vs optional (via `opt`) and which children are expected, building the `Node` tree in a single pass with no generic DOM or intermediate representation (same spirit as the signature parser in `zvariant_utils`); the low-level scanning is memchr-backed
- the two axes on which a document can nest arbitrarily deep — nested `<node>`s and skipped-over foreign elements — are walked iteratively, so the call stack stays flat regardless of input; a depth cap (1024) bounds only the open-element bookkeeping. Parse errors carry a byte offset, and invalid names/signatures surface directly as `Error::Name`/`Error::Variant`

## Error API

New `Error` variants: `Xml(XmlError)`, `Name(zbus_names::Error)` and `Io(Arc<io::Error>)`. Signature errors now surface as the pre-existing `Variant` variant instead of being stringified through serde.

For backwards compatibility, the `QuickXml`/`QuickXmlSer` variants are kept but deprecated — they are never returned anymore. Since their payload types came from quick-xml, they now hold our own copies of `DeError`/`SeError` (also deprecated and `doc(hidden)`), so code matching on these variants keeps compiling. The payload-type swap is technically breaking, but the chances of anyone depending on the exact quick-xml error types are negligible — and quick-xml bumps its semver on every release anyway, so such breaks were already routine for users of these variants. The version is bumped to 5.2.0 since the deprecations require a minor release.

## Writer fixes (for free)

- `Node::to_writer` now emits the standard `<node>` root element instead of `<Node>` — an artifact of quick-xml&#39;s serializer using the struct name
- no more empty `direction=""` attributes for args without a direction
- whitespace in attribute values is escaped as character references, so it survives attribute-value normalization by conforming readers (quick-xml wrote it raw, silently corrupting values on interop)
- parsing still ignores the root element&#39;s name (as quick-xml did), so documents written by older versions round-trip fine

## Real-world test data

To make sure this doesn&#39;t break anything in the wild, the introspection XML advertised by a dozen live services was captured verbatim (by launching each service on a real bus and calling `Introspect()` on it — no hand-written approximations) and checked in as test fixtures:

- **systemd 255**: `systemd --user` manager, a unit object, the unit-list node, plus `systemd-hostnamed`, `-timedated`, `-localed`, `-logind`, `-networkd`
- **dbus-daemon 1.14** (`org.freedesktop.DBus`), **polkitd 124**, **PackageKit 1.2.8**, **dconf 0.40**

Together these cover the XML produced by all three common generators: sd-bus, libdbus and GDBus. The tests parse each document through both entry points, and check the dbus-daemon and systemd manager interfaces exhaustively — every method, signal and property, with each argument&#39;s name, signature and direction — through shared helpers, plus assert write→parse round-trip fidelity. Provenance is documented in `tests/data/real_world/README.md`. (The systemd manager and logind documents contain `h` signatures, which zvariant only supports on Unix; on other platforms the tests assert clean failure — a pre-existing zvariant limitation, not new to this PR.)

## Benchmarks

Parse/write benchmarks over the largest real-world document (systemd manager, ~36K) are added at the start of the series — before any change — so the switch from quick-xml to the bespoke parser can be measured, and they run on CI via CodSpeed. Parsing measures within run-to-run noise of the previous implementation.

## Review &amp; testing

- New tests for whitespace normalization, attribute escaping, invalid/signed character references, duplicate attributes, DOCTYPE quoted literals, foreign-element/CDATA/comment skipping, malformed documents, and write→parse round-trips
- `zbus_xmlgen` codegen tests pass, workspace compiles with `--all-features`, clippy clean, Windows target cross-checked
- `zbus` e2e test (parses live introspection output from the object server) passes under `dbus-run-session`
- Each commit in the series builds and passes the tests on its own

Known, documented behavior differences vs quick-xml: `Node::from_reader` consumes the reader to end-of-stream before parsing (doc note added), and non-whitespace text around the root element is ignored rather than rejected.

## Signature parser hardening (zvariant_utils)

Parsing feeds untrusted `type=` attribute values straight into `zvariant::Signature::try_from`, whose signature parser recursed once per level of container nesting with no bound — so a deeply nested type string (tens of thousands of `(`) overflowed the stack and aborted the process. This is pre-existing (the old quick-xml path called `Signature::try_from` too) but reachable from this crate's untrusted-input entry point, so the last commit fixes it at the source: the signature parser now caps struct and array nesting at 32 each — the same limits the (de)serializer already enforces (see `zvariant::container_depths`) — and rejects deeper nesting as `Error::InvalidSignature` as soon as the limit is passed, without recursing through the rest of the string. This keeps the parser's recursion (and hence its stack usage) bounded while staying no stricter than what the (de)serializer round-trips, so it doesn't reject any encodable type (verified against both fuzz targets).

### DIFF
--- a/zbus_xml/benches/benchmarks.rs
+++ b/zbus_xml/benches/benchmarks.rs
@@ -1,0 +1,27 @@
+use std::hint::black_box;
+
+use criterion::{Criterion, criterion_group, criterion_main};
+
+use zbus_xml::Node;
+
+fn benchmarks(c: &mut Criterion) {
+    // The largest of the real-world introspection documents in the test data.
+    let xml = include_str!("../tests/data/real_world/systemd1_manager.xml");
+
+    c.bench_function("parse_systemd_manager", |b| {
+        b.iter(|| Node::try_from(black_box(xml)).unwrap())
+    });
+
+    let node = Node::try_from(xml).unwrap();
+    c.bench_function("write_systemd_manager", |b| {
+        b.iter(|| {
+            let mut writer = Vec::with_capacity(xml.len());
+            node.to_writer(black_box(&mut writer)).unwrap();
+
+            writer
+        })
+    });
+}
+
+criterion_group!(benches, benchmarks);
+criterion_main!(benches);
--- a/zbus_xml/src/error.rs
+++ b/zbus_xml/src/error.rs
@@ -2,12 +2,13 @@ use quick_xml::{de::DeError, se::SeError};
 use std::{convert::Infallible, error, fmt};
 use zvariant::Error as VariantError;
 
-/// The error type for `zbus_names`.
+/// The error type for `zbus_xml`.
 ///
 /// The various errors that can be reported by this crate.
 #[derive(Clone, Debug)]
 #[non_exhaustive]
 pub enum Error {
+    /// A zvariant error, e.g. an invalid signature.
     Variant(VariantError),
     /// An XML error from quick_xml
     QuickXml(DeError),
--- a/zbus_xml/src/error.rs
+++ b/zbus_xml/src/error.rs
@@ -1,5 +1,6 @@
 use quick_xml::{de::DeError, se::SeError};
 use std::{convert::Infallible, error, fmt};
+use zbus_names::Error as NamesError;
 use zvariant::Error as VariantError;
 
 /// The error type for `zbus_xml`.
@@ -10,6 +11,10 @@ use zvariant::Error as VariantError;
 pub enum Error {
     /// A zvariant error, e.g. an invalid signature.
     Variant(VariantError),
+    /// A D-Bus name error.
+    Name(NamesError),
+    /// An XML parsing error.
+    Xml(XmlError),
     /// An XML error from quick_xml
     QuickXml(DeError),
     /// An XML serialization error from quick_xml
@@ -20,6 +25,8 @@ impl PartialEq for Error {
     fn eq(&self, other: &Self) -> bool {
         match (self, other) {
             (Self::Variant(s), Self::Variant(o)) => s == o,
+            (Self::Name(s), Self::Name(o)) => s == o,
+            (Self::Xml(s), Self::Xml(o)) => s == o,
             (Self::QuickXml(_), Self::QuickXml(_)) => false,
             (Self::QuickXmlSer(_), Self::QuickXmlSer(_)) => false,
             (_, _) => false,
@@ -31,6 +38,8 @@ impl error::Error for Error {
     fn source(&self) -> Option<&(dyn error::Error + 'static)> {
         match self {
             Error::Variant(e) => Some(e),
+            Error::Name(e) => Some(e),
+            Error::Xml(e) => Some(e),
             Error::QuickXml(e) => Some(e),
             Error::QuickXmlSer(e) => Some(e),
         }
@@ -41,6 +50,8 @@ impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Error::Variant(e) => write!(f, "{e}"),
+            Error::Name(e) => write!(f, "{e}"),
+            Error::Xml(e) => write!(f, "XML error: {e}"),
             Error::QuickXml(e) => write!(f, "XML error: {e}"),
             Error::QuickXmlSer(e) => write!(f, "XML serialization error: {e}"),
         }
@@ -50,6 +61,18 @@ impl fmt::Display for Error {
 impl From<VariantError> for Error {
     fn from(val: VariantError) -> Self {
         Error::Variant(val)
+    }
+}
+
+impl From<NamesError> for Error {
+    fn from(val: NamesError) -> Self {
+        Error::Name(val)
+    }
+}
+
+impl From<XmlError> for Error {
+    fn from(val: XmlError) -> Self {
+        Error::Xml(val)
     }
 }
 
@@ -70,6 +93,40 @@ impl From<Infallible> for Error {
         match i {}
     }
 }
+
+/// An error encountered while parsing an XML document.
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct XmlError {
+    message: String,
+    position: usize,
+}
+
+impl XmlError {
+    pub(crate) fn new(message: impl Into<String>, position: usize) -> Self {
+        Self {
+            message: message.into(),
+            position,
+        }
+    }
+
+    /// A message describing the error.
+    pub fn message(&self) -> &str {
+        &self.message
+    }
+
+    /// The byte offset in the document at which the error was encountered.
+    pub fn position(&self) -> usize {
+        self.position
+    }
+}
+
+impl fmt::Display for XmlError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} (at byte offset {})", self.message, self.position)
+    }
+}
+
+impl error::Error for XmlError {}
 
 /// Alias for a `Result` with the error type `zbus_xml::Error`.
 pub type Result<T> = std::result::Result<T, Error>;
--- a/zbus_xml/src/error.rs
+++ b/zbus_xml/src/error.rs
@@ -1,5 +1,4 @@
-use quick_xml::{de::DeError, se::SeError};
-use std::{convert::Infallible, error, fmt, io, sync::Arc};
+use std::{borrow::Cow, convert::Infallible, error, fmt, io, num::NonZeroUsize, sync::Arc};
 use zbus_names::Error as NamesError;
 use zvariant::Error as VariantError;
 
@@ -18,8 +17,20 @@ pub enum Error {
     /// An I/O error.
     Io(Arc<io::Error>),
     /// An XML error from quick_xml
+    #[deprecated(
+        since = "5.2.0",
+        note = "This variant is no longer returned from any of our API. \
+                Match on `Error::Xml` instead."
+    )]
+    #[allow(deprecated)]
     QuickXml(DeError),
     /// An XML serialization error from quick_xml
+    #[deprecated(
+        since = "5.2.0",
+        note = "This variant is no longer returned from any of our API. \
+                Match on `Error::Xml` or `Error::Io` instead."
+    )]
+    #[allow(deprecated)]
     QuickXmlSer(SeError),
 }
 
@@ -29,8 +40,6 @@ impl PartialEq for Error {
             (Self::Variant(s), Self::Variant(o)) => s == o,
             (Self::Name(s), Self::Name(o)) => s == o,
             (Self::Xml(s), Self::Xml(o)) => s == o,
-            (Self::QuickXml(_), Self::QuickXml(_)) => false,
-            (Self::QuickXmlSer(_), Self::QuickXmlSer(_)) => false,
             (_, _) => false,
         }
     }
@@ -43,7 +52,9 @@ impl error::Error for Error {
             Error::Name(e) => Some(e),
             Error::Xml(e) => Some(e),
             Error::Io(e) => Some(e),
+            #[allow(deprecated)]
             Error::QuickXml(e) => Some(e),
+            #[allow(deprecated)]
             Error::QuickXmlSer(e) => Some(e),
         }
     }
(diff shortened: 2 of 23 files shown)


TTWNO:
The entire reason this got started was due to a security vulnerability found in quick-xml.  While not ideal, at least it is a crate that is being audited by security people.

With respect, who is going to test your XML parser for vulnerabilities? Used in this one project... I question the theory that the LLM has produced "the" canonical XML parser for this project and that no further issues will be found.

If it is indeed good enough for zbus, then why not upstream? Who else could use a basic XML parser in this style?

Idk, this seems like vendoring a niche implementation for hardly any gain. The quick-xml fix is already released on crates.io

zeenix:
> The entire reason this got started was due to a security vulnerability found in quick-xml.

Not exactly. This was only the last straw. quick-xml bumps semver every release and by exposing their error types, we are (at least in theory) breaking semver each time we upgrade to the latest version. This seems like a persistent issue with quick-xml.

Just because it's LLM-generated, doesn't necessarily mean it's more error prone. I explicitly asked it how to thoroughly test the impl with various xml interfaces out there. I got it to review using multiple agents and address the concerns. I'll also review myself and also get copilot to review before merging this. Also keep in mind that I used the most powerful model available in the market.

We're talking about a very specific subset of XML here and not a generic parser impl so the attack surface is much lower than that of quick-xml. I'm also going to make the LLM rewrite it to make full use of winnow, shifting a lot of the attack surface to winnow, which has a lot not eyeballs on it.

TTWNO:
> Just because it's LLM-generated, doesn't necessarily mean it's more error prone.

I'm inclined to believe it would be less error prone, haha! Not about you, but just humans make silly mistakes all the time. The auditing is my concern.

> We're talking about a very specific subset of XML here and not a generic parser impl so the attack surface is much lower than that of quick-xml. I'm also going to make the LLM rewrite it to make full use of winnow, shifting a lot of the attack surface to winnow, which has a lot not eyeballs on it

Fair enough. Thank you for the reassurance.

zeenix:
> I'm inclined to believe it would be less error prone, haha! Not about you, but just humans make silly mistakes all the time. 

Oh, I make silly mistakes all the time. These machines aren't perfect either but I believe together we can do a a much job than just alone.

> The auditing is my concern

Yeah I get it. Thanks for raising it. One important thing to keep in mind is that zbus doesn't typically parse arbitrary XML at runtime. This is only done by developer tools so the attack surface is really very small here.

zbus_xml/src/xml.rs
```diff
@@ -0,0 +1,331 @@
...
+                } else {
+                    format!("expected quoted value for attribute `{attr_name}`")
+                })
+            })?;
+
+            if attrs.iter().any(|(name, _)| *name == attr_name) {
+                return Err(self.error(format!("duplicate attribute `{attr_name}`")));
+            }
+            let value = unescape(raw).map_err(|e| self.error(e))?;
+            attrs.push((attr_name, value));
```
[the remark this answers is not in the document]
zeenix: Both points verified and addressed:

- da389f5 anchors unescape errors at the offending reference itself (better than the pre-winnow behavior, which pointed at the start of the value), with a regression test asserting the exact offset.
- 88a9333 rejects attributes not separated by whitespace, per the XML grammar.

---

zeenix:
Pushed `e4d5895` (🐛 zx: Parse deeply nested documents iteratively) to this branch, as discussed in #1840: an adversarial review found that `MAX_DEPTH` bounded the *recursion* of `skip_element`/nested-`<node>` parsing but not the *stack usage* — with debug builds' large frames, ~1000 nesting levels (well within the cap) overflow a 2 MiB thread stack.

Rather than tuning the cap to whatever today's debug frames make safe, the parser now iterates on the two axes where documents can nest arbitrarily deep (`skip_element` tracks open elements in a `Vec`; `Node::from_xml` builds nested nodes on an explicit stack). All remaining recursion follows the fixed introspection hierarchy, so the call stack stays flat regardless of input, and `MAX_DEPTH` only bounds bookkeeping memory. The new test parses documents nested to the cap on a deliberately small (512 KiB) thread stack.

One description bullet is now slightly stale as a result: "so recursion stays bounded on adversarial input" — it's stronger than that now.

---

zbus_xml/src/error.rs
```diff
@@ -1,14 +1,22 @@
 use quick_xml::{de::DeError, se::SeError};
-use std::{convert::Infallible, error, fmt};
+use std::{convert::Infallible, error, fmt, io, sync::Arc};
+use zbus_names::Error as NamesError;
 use zvariant::Error as VariantError;
 
-/// The error type for `zbus_names`.
+/// The error type for `zbus_xml`.
```
zeenix, line 5: unrelated fix. Split into separate commit.

zbus_xml/src/error.rs
```diff
@@ -1,14 +1,22 @@
...
 ///
 /// The various errors that can be reported by this crate.
 #[derive(Clone, Debug)]
 #[non_exhaustive]
 pub enum Error {
+    /// A signature error.
```
zeenix: It can contain potentially any zvariant::Error, not just signature error.

zbus_xml/src/lib.rs
```diff
@@ -203,6 +410,61 @@ impl Property<'_> {
     pub fn annotations(&self) -> &[Annotation] {
         &self.annotations
     }
+
+    fn from_xml<'i>(reader: &mut Reader<'i>, mut tag: StartTag<'i>) -> Result<Self> {
```
zeenix: There is a lot of code being added in the second commit, that is doing most of the parsing while that was the task for the previous commit to handle it all. The second commit was only supposed to make use of that.

zbus_xml/src/error.rs
```diff
@@ -8,6 +7,7 @@ use zvariant::Error as VariantError;
 /// The various errors that can be reported by this crate.
 #[derive(Clone, Debug)]
 #[non_exhaustive]
+#[allow(deprecated)]
```
zeenix: is this really needed on the enum that's defining the deprecated API, since it's not using deprecated API?

zbus_xml/src/error.rs
```diff
@@ -29,14 +39,13 @@ impl PartialEq for Error {
...
             (Self::Xml(s), Self::Xml(o)) => s == o,
-            (Self::QuickXml(_), Self::QuickXml(_)) => false,
-            (Self::QuickXmlSer(_), Self::QuickXmlSer(_)) => false,
             (_, _) => false,
         }
     }
 }
 
 impl error::Error for Error {
+    #[allow(deprecated)]
```
zeenix: can this be on the match arms themselves so it's only applied where we want it to apply?

zbus_xml/src/error.rs
```diff
@@ -50,6 +59,7 @@ impl error::Error for Error {
 }
 
 impl fmt::Display for Error {
+    #[allow(deprecated)]
```
zeenix: can this be on the match arms themselves so it's only applied where we want it to apply?

zbus_xml/tests/tests.rs
```diff
@@ -162,6 +162,162 @@ fn ignores_unknown_elements_and_text() -> Result<(), Box<dyn Error>> {
...
+            .find(|i| i.name() == "org.freedesktop.systemd1.Manager")
+            .expect("Manager interface");
+
+        let start_unit = manager
+            .methods()
+            .iter()
+            .find(|m| m.name() == "StartUnit")
+            .expect("StartUnit method");
+        assert_eq!(start_unit.args().len(), 3);
+        assert_eq!(start_unit.args()[2].ty(), "o");
```
zeenix: all the checks in these new tests should be more thorough and check every value and signature etc. Try reusing code between the tests through utility function(s).

zbus_xml/benches/benchmarks.rs
zeenix, line 1: The benchmarks should come before any changes so every change can be benchmarked.

zbus_xml/src/xml.rs
```diff
@@ -7,6 +7,13 @@
 
 use std::borrow::Cow;
 
+use winnow::{
```
zeenix, line 14: The winnow change should be squashed with the bespoke parser commit as this is a fixup. We never wanted non-winnow parser. Benchmarks should then be used to measure any performance diff between the git main and winnow-based custom parsing.

zbus_xml/src/xml.rs
```diff
@@ -30,61 +37,66 @@ pub(crate) enum Event<'i> {
...
+                self.markup_declaration()?;
+            } else if self.input.starts_with("<?") {
+                self.parse(("<?", take_until(0.., "?>"), "?>"))
+                    .map_err(|_| self.error("unterminated processing instruction"))?;
+            } else if self.input.starts_with("</") {
+                let end_name = self
+                    .parse(preceded("</", name))
+                    .map_err(|_| self.error("expected a name in closing tag"))?;
+                self.parse((whitespace, '>'))
+                    .map_err(|_| self.error(format!("expected `>` to close `</{end_name}`")))?;
```
zeenix: this is way too if-else based low-level by hand parsing and not making use of winnow to write simple and elegant parsers that make full use of combinators. We should have small separate functions for each fragment that are used in combinations and they should be documented.

zeenix (commented):
The last 3 commits seem like they're fixups to earlier commits. If so, they should be squashed into the appropriate commits in the PR.

zbus_xml/src/error.rs
```diff
@@ -8,7 +9,14 @@ use zvariant::Error as VariantError;
 #[derive(Clone, Debug)]
 #[non_exhaustive]
 pub enum Error {
+    /// A zvariant error, e.g. an invalid signature.
```
zeenix, line 11: How is this related to addition of xml parser? Split. Ensure all commits are atomic and only contain hunks that are covered by the commit message.

zbus_xml/src/error.rs
```diff
@@ -8,7 +9,14 @@ use zvariant::Error as VariantError;
...
     Variant(VariantError),
+    /// A D-Bus name error.
+    Name(NamesError),
+    /// An XML parsing error.
+    Xml(XmlError),
+    /// An I/O error.
+    Io(Arc<io::Error>),
```
zeenix, line 18: Is the containing commit making use of the Name and Io variants? If not, why add them in this commit?

zbus_xml/src/lib.rs
```diff
@@ -116,6 +218,50 @@ impl Method<'_> {
...
```
zeenix: [This](https://github.com/z-galaxy/zbus/pull/1839#discussion_r3528976103) comment was never really addressed. The parsing code should directly result in this high-level tree/representation instead of provding very low-level API for these methods here to then do a lot of work. See the signature parser code in zvariant_utils for example. The parser code can then make decisions directly about each node as it knows about what's expected and what's optional etc (e.g by making use of `winnow::combinator::opt`). Remember that we don't need a generic xml parser here but only D-Bus specific XML.

zeenix:
This also resolves #1328 (`Node::from_reader` failing on large introspection files with `"Deserializer buffers 1024 events, limit exceeded"`).

That error came from quick-xml's serde deserializer buffering events while resolving sequences, capped via `event_buffer_size`. Since this PR drops quick-xml entirely and parses introspection XML with the bespoke `winnow` parser that builds the `Node` tree directly, there is no longer any event-buffer ceiling — the only remaining bound is `MAX_DEPTH` (element *nesting* depth, not document size), which a large-but-flat interface won't hit.

Verified on `main`: a ~536 KB document (one interface, 5000 methods — far beyond the old 4k-event limit) now parses cleanly via `Node::from_reader`, whereas the old path errored out. So #1328 can be closed as fixed by the parser rewrite (shipping in zbus_xml 5.2.0); no separate limit-raising change is needed.

---

PR #1840, opened 2026-07-04
✨ Handle the Telepathy introspection extensions in zbus_xml & zbus-xmlgen

### LINKED ISSUE #255: xmlgen should handle Telepathy D-Bus interface XML extensions
In GitLab by @Absolucy on Mar 29, 2022, 22:54

Hi, I've been trying to use zbus_xmlgen to generate interfaces for MPRIS, and I've ran into several things in their respective XMLs that have caused errors!

## Docstrings

```xml
    <property name="Fullscreen" type="b" tp:name-for-bindings="Fullscreen" access="readwrite">
      <tp:added version="2.2" />
      <annotation name="org.mpris.MediaPlayer2.property.optional" value="true"/>
      <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
        <p>Whether the media player is occupying the fullscreen.</p>
        <p>
          This is typically used for videos.  A value of <strong>true</strong>
          indicates that the media player is taking up the full screen.
        </p>
        <p>
          Media centre software may well have this value fixed to <strong>true</strong>
        </p>
        <p>
          If <tp:member-ref>CanSetFullscreen</tp:member-ref> is <strong>true</strong>,
          clients may set this property to <strong>true</strong> to tell the media player
          to enter fullscreen mode, or to <strong>false</strong> to return to windowed
          mode.
        </p>
        <p>
          If <tp:member-ref>CanSetFullscreen</tp:member-ref> is <strong>false</strong>,
          then attempting to set this property should have no effect, and may raise
          an error.  However, even if it is <strong>true</strong>, the media player
          may still be unable to fulfil the request, in which case attempting to set
          this property will have no effect (but should not raise an error).
        </p>
        <tp:rationale>
          <p>
            This allows remote control interfaces, such as LIRC or mobile devices like
            phones, to control whether a video is shown in fullscreen.
          </p>
        </tp:rationale>
      </tp:docstring>
    </property>
```

## Structs

```xml
    <tp:struct name="Playlist" array-name="Playlist_List">
      <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
        <p>A data structure describing a playlist.</p>
      </tp:docstring>
      <tp:member type="o" tp:type="Playlist_Id" name="Id">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>A unique identifier for the playlist.</p>
          <p>This should remain the same if the playlist is renamed.</p>
        </tp:docstring>
      </tp:member>
      <tp:member type="s" name="Name">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>The name of the playlist, typically given by the user.</p>
        </tp:docstring>
      </tp:member>
      <tp:member type="s" tp:type="Uri" name="Icon">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>The URI of an (optional) icon.</p>
        </tp:docstring>
      </tp:member>
    </tp:struct>
```

## Enums

```xml
    <tp:enum name="Playback_Status" tp:name-for-bindings="Playback_Status" type="s">
      <tp:enumvalue suffix="Playing" value="Playing">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>A track is currently playing.</p>
        </tp:docstring>
      </tp:enumvalue>
      <tp:enumvalue suffix="Paused" value="Paused">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>A track is currently paused.</p>
        </tp:docstring>
      </tp:enumvalue>
      <tp:enumvalue suffix="Stopped" value="Stopped">
        <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
          <p>There is no track currently playing.</p>
        </tp:docstring>
      </tp:enumvalue>
      <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
        <p>A playback state.</p>
      </tp:docstring>
    </tp:enum>
```

## Simple Types

```xml
    <tp:simple-type name="Playlist_Id" type="o" array-name="Playlist_Id_List">
      <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
        <p>Unique playlist identifier.</p>
        <tp:rationale>
          <p>
            Multiple playlists may have the same name.
          </p>
          <p>
            This is a D-Bus object id as that is the definitive way to have
            unique identifiers on D-Bus.  It also allows for future optional
            expansions to the specification where tracks are exported to D-Bus
            with an interface similar to org.gnome.UPnP.MediaItem2.
          </p>
        </tp:rationale>
      </tp:docstring>
    </tp:simple-type>

    <tp:simple-type name="Uri" type="s" array-name="Uri_List">
      <tp:docstring xmlns="http://www.w3.org/1999/xhtml">
        <p>A URI.</p>
      </tp:docstring>
    </tp:simple-type>
```

### LINKED ISSUE #256: xmlgen should ignore non-supported XML w/ a warning
Creating a separate issue for the simpler/bug part of #255. As [discussed there](https://gitlab.freedesktop.org/dbus/zbus/-/issues/255#note_1318643), we should ignore any unexpected XML with a warning on the standard error output, instead of just completely giving up.

zeenix:
Fixes #255. Fixes #256.

Built on top of #1839 (now merged); this branch is rebased onto `main`. Since #1839 reworked introspection-tree construction into a bespoke winnow parser, the zbus_xml side here is a reimplementation against it (one commit) rather than three; the six zbus_xmlgen commits, which only consume zbus_xml's public API, replayed unchanged. The debug-build deep-nesting stack-overflow this series had carried a fix for now lives in #1839's parser itself, so it's dropped here.

## What this does

The Telepathy D-Bus introspection extensions — used most prominently by MPRIS — embed documentation (`tp:docstring`) and named type definitions (`tp:simple-type`, `tp:enum`, `tp:struct`, `tp:mapping`, referenced via `tp:type` attributes) in introspection XML. Previously all of it was skipped silently (and before #1839, it broke parsing outright — the original complaint of #255).

### zbus_xml

- **Docstrings** are captured on nodes, interfaces, methods, signals, properties and args, exposed via `docstring()` accessors. Content is kept verbatim (it's typically HTML) with surrounding whitespace trimmed, matched under any namespace prefix. Parse-only: the writer doesn't emit them (documented on `Node::to_writer`, including the equality implication for round-trips).
- **Type definitions** are parsed into a new `telepathy` module from both `<node>` and `<interface>` level (`telepathy_types()` accessors), and `tp:type` references are exposed on `Arg`, `Property` and `telepathy::Member`.
- **Warnings** (#256): anything that still gets skipped — unsupported elements like `tp:flags`, or a malformed type definition (missing attribute, invalid/empty signature, non-basic mapping key) — is recorded as a `Warning` (element name, byte offset, message) returned by the new `Node::from_reader_with_warnings()`. Malformed definitions never fail the document. The existing entry points are unchanged.

Implementation-wise, the parser threads the document and a warning collector through a `winnow::Stateful` input and hooks docstring capture and type-def parsing into the points where it would otherwise skip a foreign element.

### zbus-xmlgen

- **New `CodeGenerator` API**: configuration lives behind private fields with a constructor, builder-style `with_*` setters (`with_node_types`, `with_service`, `with_path`, `with_format`) and getters; `interface_code()` generates one interface's code, `file_code()` a complete source file. `GenTrait` and `write_interfaces()` keep their released shapes and are deprecated shims delegating to it (with a test pinning `GenTrait`'s output to the new implementation), so nothing breaks and future knobs stay additive. Version bumped to 5.4.0 for the deprecations.
- **Docstrings become doc comments**: the interface docstring on the trait, member docstrings under the synthesized `/// Xyz method` lines, arg docstrings as a `# Arguments` list. rustdoc passes the HTML through.
- **Type definitions become Rust types**: aliases for `tp:simple-type`/`tp:mapping`; structs and fieldless enums with serde + zvariant `Type`/`Value`/`OwnedValue` derives (through the re-export, via `#[zvariant(crate = "zbus::zvariant")]`). Numeric enums get `#[repr]`, explicit discriminants and serde_repr derives; string enums get `#[zvariant(signature = "s")]` with renames where the wire value differs from the variant name. Generated signatures use the named types wherever a `tp:type` reference resolves to a generated definition and matches the actual signature:

  ```rust
  fn get_playlists(&self, index: u32, max_count: u32, order: PlaylistOrdering, reverse_order: bool)
      -> zbus::Result<Vec<Playlist>>;
  #[zbus(property)]
  fn active_playlist(&self) -> zbus::Result<MaybePlaylist>;
  ```

  The `Value`/`OwnedValue` derives make the types work for properties (`TryFrom<OwnedValue>`/`Into<Value>`), not just method bodies. Node-level definitions are generated only when the interface (transitively) references them; interface-level ones always; in single-file/stdout output, all interfaces form one document so shared definitions (and the header) appear once. The generated file's doc header calls out the serde/serde_repr dependency.
- **Anything that cannot become valid Rust falls back**: unresolvable or signature-mismatched `tp:type` references, definitions whose names collide (with each other or with the proxy macro's own items), enums with duplicate/out-of-range/unrepresentable values — all degrade to the structural types instead of emitting broken code.
- **Warnings go to stderr** (#256), deduplicated by message.

## Review & testing

Three independent review agents earlier went over this work (parser correctness + adversarial input, codegen validity, API/backwards-compat); all confirmed findings are folded into the commits here. Highlights:

- The adversarial parser pass (thousands of truncations, mutations and hostile splices; 10k-value enums; MAX_DEPTH probes; UTF-8 boundary checks) found no panics or hangs; it caught the empty-signature (`Unit`) acceptance and two warning-contract violations, all fixed.
- The codegen pass compiled generator output for adversarial definitions and caught 10 ways to emit non-compiling Rust (name collisions, discriminant issues, enum mapping keys, bare CR in docstrings, duplicated shared types in single-file output, and the pre-existing invalid `Signature<'_>`/`OwnedSignature` emission) — all fixed, pinned by the `telepathy_edge_cases` fixture, which is diff- and compile-tested through the real `#[proxy]` macro.
- The compat pass verified serde wire format unchanged, a 5.3.1-style consumer compiles with only deprecation warnings, docs clean under `-D warnings`, and MSRV 1.87.

Also, on this rebase onto `main`: 24 zbus_xml + 7 zbus_xmlgen tests pass (docstring capture, warning positions/messages, well-formed and nine kinds of malformed definitions, the MPRIS-modeled fixtures); clippy and tests clean under CI's `-D warnings`; rustfmt, docs, Windows cross-check and the zbus e2e test under `dbus-run-session` all pass; every commit builds and tests standalone.

## Deliberate scope edges

- `tp:flags` is *not* generated (stays a warning): the natural mapping is enumflags2, but zvariant's `BitFlags` support is feature-gated and zbus doesn't re-export enumflags2, so generated code couldn't rely on it without extra setup. Easy follow-up if wanted.

https://claude.ai/code/session_01BVM3TqQzUHJSB7sUqvVRDJ

### DIFF
--- a/zbus_xmlgen/tests/data/telepathy_docstrings.rs
+++ b/zbus_xmlgen/tests/data/telepathy_docstrings.rs
@@ -1,0 +1,72 @@
+/// Provides access to the media player's playlists.
+///
+/// Since D-Bus does not directly support enums, or a **maybe** type, they are described in this interface.
+#[proxy(interface = "com.example.Playlists", assume_defaults = true)]
+pub trait Playlists {
+    /// ActivatePlaylist method
+    ///
+    /// Starts playing the given playlist.
+    ///
+    /// It is up to the media player whether this completely replaces the current tracklist, or whether it is merely inserted into the tracklist and the first track starts.
+    ///
+    /// # Arguments
+    ///
+    /// * `playlist_id` - The id of the playlist to activate.
+    fn activate_playlist(&self, playlist_id: &zbus::zvariant::ObjectPath<'_>) -> zbus::Result<()>;
+
+    /// GetPlaylists method
+    ///
+    /// Gets a set of playlists.
+    ///
+    /// # Arguments
+    ///
+    /// * `index` - The index of the first playlist to be fetched (according to the ordering).
+    /// * `max_count` - The maximum number of playlists to fetch.
+    /// * `order` - The ordering that should be used.
+    /// * `reverse_order` - Whether the order should be reversed.
+    /// * `playlists` - A list of (at most *max_count*) playlists.
+    fn get_playlists(
+        &self,
+        index: u32,
+        max_count: u32,
+        order: &str,
+        reverse_order: bool,
+    ) -> zbus::Result<Vec<(zbus::zvariant::OwnedObjectPath, String, String)>>;
+
+    /// PlaylistChanged signal
+    ///
+    /// Indicates that either the Name or Icon attribute of a playlist has changed.
+    ///
+    /// Client implementations should be aware that this signal may not be implemented.
+    ///
+    /// Without this signal, media players have no way to notify clients of a change in the attributes of a playlist.
+    ///
+    /// # Arguments
+    ///
+    /// * `playlist` - The playlist which details have changed.
+    #[zbus(signal)]
+    fn playlist_changed(
+        &self,
+        playlist: (zbus::zvariant::ObjectPath<'_>, &str, &str),
+    ) -> zbus::Result<()>;
+
+    /// Orderings property
+    ///
+    /// The available orderings. At least one must be offered.
+    ///
+    /// Media players may not have access to all the data required for some orderings.
+    #[zbus(property)]
+    fn orderings(&self) -> zbus::Result<Vec<String>>;
+
+    /// PlaylistCount property
+    ///
+    /// The number of playlists available.
+    #[zbus(property)]
+    fn playlist_count(&self) -> zbus::Result<u32>;
+
+    /// Undocumented property
+    #[zbus(property)]
+    fn undocumented(&self) -> zbus::Result<String>;
+    #[zbus(property)]
+    fn set_undocumented(&self, value: &str) -> zbus::Result<()>;
+}
--- a/zbus_xmlgen/tests/data/telepathy_docstrings.rs
+++ b/zbus_xmlgen/tests/data/telepathy_docstrings.rs
@@ -1,3 +1,95 @@
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    serde::Serialize,
+    serde::Deserialize,
+    zbus::zvariant::Type,
+    zbus::zvariant::Value,
+    zbus::zvariant::OwnedValue,
+)]
+#[zvariant(signature = "s", crate = "zbus::zvariant")]
+pub enum PlaylistOrdering {
+    /// Alphabetical ordering by name, ascending.
+    Alphabetical,
+    /// A user-defined ordering.
+    #[serde(rename = "User")]
+    #[zvariant(rename = "User")]
+    UserDefined,
+}
+
+/// A repeat / loop status.
+#[derive(
+    Debug,
+    Clone,
+    Copy,
+    PartialEq,
+    Eq,
+    serde_repr::Deserialize_repr,
+    serde_repr::Serialize_repr,
+    zbus::zvariant::Type,
+    zbus::zvariant::Value,
+    zbus::zvariant::OwnedValue,
+)]
+#[zvariant(crate = "zbus::zvariant")]
+#[repr(u32)]
+pub enum LoopStatus {
+    /// The playback will stop when there are no more tracks to play.
+    None = 0,
+    /// The current track will start again from the beginning once it has finished playing.
+    Track = 1,
+    /// The playback loops through a list of tracks.
+    Playlist = 2,
+}
+
+/// A data structure describing a playlist.
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    serde::Serialize,
+    serde::Deserialize,
+    zbus::zvariant::Type,
+    zbus::zvariant::Value,
+    zbus::zvariant::OwnedValue,
+)]
+#[zvariant(crate = "zbus::zvariant")]
+pub struct Playlist {
+    /// A unique identifier for the playlist.
+    pub id: PlaylistId,
+    /// The name of the playlist, typically given by the user.
+    pub name: String,
+    /// The URI of an (optional) icon.
+    pub icon: String,
+}
+
+/// A data structure describing a playlist, or nothing.
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    serde::Serialize,
+    serde::Deserialize,
+    zbus::zvariant::Type,
+    zbus::zvariant::Value,
+    zbus::zvariant::OwnedValue,
+)]
+#[zvariant(crate = "zbus::zvariant")]
+pub struct MaybePlaylist {
+    /// Whether this structure refers to a valid playlist.
+    pub valid: bool,
+    /// The playlist, providing Valid is true.
+    pub playlist: Playlist,
+}
+
+/// A mapping from metadata attribute names to values.
+pub type MetadataMap = std::collections::HashMap<String, zbus::zvariant::OwnedValue>;
+
+/// Unique playlist identifier.
+pub type PlaylistId = zbus::zvariant::OwnedObjectPath;
+
 /// Provides access to the media player's playlists.
 ///
 /// Since D-Bus does not directly support enums, or a **maybe** type, they are described in this interface.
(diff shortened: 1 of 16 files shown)


zbus_xmlgen/tests/data/telepathy_docstrings.rs
```diff
@@ -1,3 +1,95 @@
...
+    zbus::zvariant::Value,
+    zbus::zvariant::OwnedValue,
+)]
+#[zvariant(signature = "s", crate = "zbus::zvariant")]
+pub enum PlaylistOrdering {
+    /// Alphabetical ordering by name, ascending.
+    Alphabetical,
+    /// A user-defined ordering.
+    #[serde(rename = "User")]
+    #[zvariant(rename = "User")]
```
bilelmoussaoui, line 19: why have both renames? for OwnedValue?

zeenix: The two attributes drive two independent (de)serialization paths, and a generated enum may be used through either:

- `#[serde(rename)]` controls the wire string when the value is (de)serialized through serde — i.e. in a message body: method args/returns, signal args, struct fields.
- `#[zvariant(rename)]` controls it in the `Value`/`OwnedValue` conversions (`From<enum> for Value` / `TryFrom<Value>`), which is the path used for **properties** (returned as an `OwnedValue`) and whenever the enum is nested inside a `Value`. The `Value`/`OwnedValue` derive matches variants against strings itself and consults its own `rename`, not serde's ([`enum_name_for_variant`](https://github.com/z-galaxy/zbus/blob/main/zvariant_derive/src/value.rs) uses `variant_attrs.rename`).

So it's not about `OwnedValue` specifically vs `Value` — it's serde-path vs zvariant-Value-path. `PlaylistOrdering` in this fixture is used both ways (`GetPlaylists`'s `order` arg → serde path; the `Orderings` property → `Vec<Value>` path), so both must map `UserDefined` → `"User"`. Dropping either emits the Rust variant name on that path instead:

```rust
// serde path, no serde(rename): serialized string is "UserDefined", not "User"
// Value path, no zvariant(rename): Value::from(UserDefined) is "UserDefined", not "User"
```

(verified both with a quick round-trip test). The generator can't know up front which path a given type will travel, so it emits both to keep the wire value consistent.

---

zbus_xmlgen/tests/data/telepathy_docstrings.rs
```diff
@@ -0,0 +1,197 @@
...
+    ) -> zbus::Result<Vec<Playlist>>;
+
+    /// PlaylistChanged signal
+    ///
+    /// <p>Indicates that either the Name or Icon attribute of a
+    /// playlist has changed.</p>
+    /// <p>Client implementations should be aware that this signal
+    /// may not be implemented.</p>
+    /// <tp:rationale>Without this signal, media players have no way to
+    /// notify clients of a change in the attributes of a playlist.</tp:rationale>
```
zeenix: seems like xml tags end up in the generated docs.

zeenix: Fixed in c26b0b2. Telepathy `<tp:docstring>` bodies are HTML fragments, and they were being written into rustdoc comments verbatim, so tags leaked through.

The generator now runs each docstring through a small HTML→Markdown pass before emitting it:

- `<strong>`/`<b>` → `**bold**`, `<em>`/`<i>` → `*italic*`, `<code>`/`<tt>` → `` `code` ``
- block-level elements (`<p>`, `<br>`, `<tp:rationale>`, lists, headings, …) → paragraph breaks
- comments, CDATA and processing instructions are stripped; entities (`&amp;`, `&lt;`, `&#39;`, …) are decoded
- the match is on the *local* name, so a namespace prefix like `tp:` doesn't matter

So e.g. the `<strong>maybe</strong>` in the `Playlists` interface doc now renders as `**maybe**`, and there are no raw tags left in the regenerated fixtures. Anything not recognized is dropped rather than passed through, so unknown tags can't leak either.

---

PR #1841, opened 2026-07-07
✨ zb: Add Connection::closed() and is_closed() API

bilelmoussaoui:
Expose a way to detect when the underlying socket has been closed, either by the remote peer or due to an I/O error. This lets D-Bus service daemons exit cleanly instead of hanging on pending futures.

### DIFF
--- a/zbus/src/blocking/connection/mod.rs
+++ b/zbus/src/blocking/connection/mod.rs
@@ -250,6 +250,24 @@ impl Connection {
         self.inner.monitor_activity()
     }
 
+    /// Returns `true` if the connection has been closed.
+    ///
+    /// A connection is considered closed either when the remote peer disconnects, an I/O error
+    /// occurs on the socket, or [`Connection::close`] is called.
+    pub fn is_closed(&self) -> bool {
+        self.inner.is_closed()
+    }
+
+    /// Block until the connection is closed.
+    ///
+    /// A connection is considered closed either when the remote peer disconnects, an I/O error
+    /// occurs on the socket, or [`Connection::close`] is called.
+    ///
+    /// Returns immediately if the connection is already closed.
+    pub fn closed(&self) {
+        block_on(self.inner.closed())
+    }
+
     /// Return the peer credentials.
     ///
     /// The fields are populated on the best effort basis. Some or all fields may not even make
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -6,7 +6,10 @@ use std::{
     collections::HashMap,
     future::Future,
     io,
-    sync::{Arc, OnceLock, Weak},
+    sync::{
+        Arc, OnceLock, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
     time::Duration,
 };
 use tracing::{Instrument, debug, info_span, instrument, trace, trace_span, warn};
@@ -32,7 +35,7 @@ pub mod socket;
 pub use socket::Socket;
 
 mod socket_reader;
-use socket_reader::SocketReader;
+use socket_reader::{SocketReader, SocketStatus};
 
 mod pending_method_calls;
 use pending_method_calls::PendingMethodCalls;
@@ -54,7 +57,7 @@ pub(crate) struct ConnectionInner {
     unique_name: OnceLock<OwnedUniqueName>,
     registered_names: Mutex<HashMap<WellKnownName<'static>, NameStatus>>,
 
-    activity_event: Arc<Event>,
+    socket_status: Arc<SocketStatus>,
     socket_write: Mutex<Box<dyn socket::WriteHalf>>,
 
     // Our executor
@@ -220,7 +223,7 @@ impl Connection {
             return Err(Error::Unsupported);
         }
 
-        self.inner.activity_event.notify(usize::MAX);
+        self.inner.socket_status.activity_event.notify(usize::MAX);
         let mut write = self.inner.socket_write.lock().await;
 
         write.send_message(msg).await
@@ -1100,7 +1103,11 @@ impl Connection {
 
         let connection = Self {
             inner: Arc::new(ConnectionInner {
-                activity_event: Arc::new(Event::new()),
+                socket_status: Arc::new(SocketStatus {
+                    activity_event: Event::new(),
+                    closed: AtomicBool::new(false),
+                    closed_event: Event::new(),
+                }),
                 socket_write: Mutex::new(auth.socket_write),
                 server_guid: auth.server_guid,
                 #[cfg(unix)]
@@ -1144,7 +1151,31 @@ impl Connection {
     ///
     /// This function is meant for the caller to implement idle or timeout on inactivity.
     pub fn monitor_activity(&self) -> EventListener {
-        self.inner.activity_event.listen()
+        self.inner.socket_status.activity_event.listen()
+    }
+
+    /// Returns `true` if the connection has been closed.
+    ///
+    /// A connection is considered closed either when the remote peer disconnects, an I/O error
+    /// occurs on the socket, or [`Connection::close`] is called.
+    pub fn is_closed(&self) -> bool {
+        self.inner.socket_status.closed.load(Ordering::Relaxed)
+    }
+
+    /// Waits until the connection is closed.
+    ///
+    /// A connection is considered closed either when the remote peer disconnects, an I/O error
+    /// occurs on the socket, or [`Connection::close`] is called.
+    ///
+    /// Returns immediately if the connection is already closed.
+    pub async fn closed(&self) {
+        loop {
+            let listener = self.inner.socket_status.closed_event.listen();
+            if self.inner.socket_status.closed.load(Ordering::Acquire) {
+                return;
+            }
+            listener.await;
+        }
     }
 
     /// Return the peer credentials.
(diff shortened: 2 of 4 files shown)


zbus/src/connection/mod.rs
```diff
@@ -1147,6 +1154,23 @@ impl Connection {
         self.inner.activity_event.listen()
     }
 
+    /// Returns `true` if the underlying socket has been closed.
+    pub fn is_closed(&self) -> bool {
+        self.inner.closed.load(Ordering::Acquire)
+    }
+
+    /// A future that resolves when the connection's underlying socket is closed,
```
zeenix: I wouldn't word it like this since we're not returning a Future but rather it's just an async method on the high-level. I'd change the method name to `await_closed` and adust the doc comment accordingly.

zbus/src/connection/mod.rs
```diff
@@ -55,6 +58,8 @@ pub(crate) struct ConnectionInner {
     registered_names: Mutex<HashMap<WellKnownName<'static>, NameStatus>>,
 
     activity_event: Arc<Event>,
+    closed: Arc<AtomicBool>,
+    closed_event: Arc<Event>,
```
zeenix: I don't think the `Arc` is needed here. ConnectionInner itself is in the Arc already. I don't quite remember why I used `Arc` on `activity_event` though. 🤔 The commit that did that was 9f461a2127acd047af38008d63f44914a10d42a4 and it might have been an oversight.

zbus/src/connection/mod.rs
```diff
@@ -1147,6 +1154,23 @@ impl Connection {
         self.inner.activity_event.listen()
     }
 
+    /// Returns `true` if the underlying socket has been closed.
+    pub fn is_closed(&self) -> bool {
+        self.inner.closed.load(Ordering::Acquire)
```
zeenix: `Ordering::Relaxed` won't work?

zbus/src/connection/mod.rs
```diff
@@ -1147,6 +1154,23 @@ impl Connection {
...
+    /// A future that resolves when the connection's underlying socket is closed,
+    /// either by the remote peer or due to an I/O error.
+    ///
+    /// Resolves immediately if the connection is already closed.
+    pub async fn closed(&self) {
+        let listener = self.inner.closed_event.listen();
+        if self.inner.closed.load(Ordering::Acquire) {
+            return;
+        }
+        listener.await;
```
zeenix: perhaps we should also check the `closed` value after this to avoid any spurious events (just to be safe) and use a loop here?

zeenix (changes requested):

zbus/src/connection/mod.rs
bilelmoussaoui: `await_closed().await` looks really weird, no better ideas tho

zeenix: Yeah, I thought of but I can't come up with a better naming. `closed` on its own sounds strange and will look strange too unless there is a `.await` directly after it (which will not always be the case, e.g `tokio::select` arm).

bilelmoussaoui: tbh i think naming it closed is better than await_closed

zeenix: > tbh i think naming it closed is better than await_closed

You already wrote that but ok, no biggie. :) Can you please update the doc wording?

bilelmoussaoui: Should be better.

zbus/src/connection/mod.rs
bilelmoussaoui: Dropped.

bilelmoussaoui: Actually we can't drop the Arc here. I thought of passing ConnectionInner directly, but that causes a reference cycle and would require more refactoring to properly clean things up. So I am skipping that for now.

bilelmoussaoui:
Sorry for taking too long, updated the patch.

zbus/src/connection/mod.rs
zeenix: Right, I get it now and that also explains the Arc on `activity_event` now.

However, since all 3 fields are being cloned only in one place to be passed to the socket reader task, probably best to split this into a separate struct and wrap that in an Arc. It sounds like a mico-optimization, I know but unless we measure it, we never know and it's good to have code in such a wide-used crate as optimized as possible.

bilelmoussaoui: Abstracted the fields in a SocketState struct.

zbus/src/connection/mod.rs
```diff
@@ -43,6 +46,14 @@ use handshake::Authenticated;
...
 const DEFAULT_MAX_QUEUED: usize = 64;
 
+/// Socket-related state shared between [`ConnectionInner`] and the socket reader task.
+#[derive(Debug)]
+pub(crate) struct SocketState {
+    pub(crate) activity_event: Event,
+    pub(crate) closed: AtomicBool,
+    pub(crate) closed_event: Event,
+}
+
```
zeenix: * Given that this file is already quite big, I'd rather we put this in the socket_reader module.
* The name should rather be `SocketStatus` IMO.
* Regardless of which module it's in, please follow the top-down code layout guideline, which I recently [made explicit](https://github.com/z-galaxy/zbus/blob/main/CONTRIBUTING.md#code-layout).

bilelmoussaoui: done

zeenix (commented):
LGTM but I'd let Claude have another look at this before merging.

zbus/src/connection/socket_reader.rs
```diff
@@ -153,3 +160,11 @@ impl SocketReader {
         Ok(msg)
     }
 }
+
+/// Socket-related state shared between [`super::ConnectionInner`] and the socket reader task.
+#[derive(Debug)]
+pub(crate) struct SocketStatus {
```
zeenix: Actually `pub(super)` would be enough here, I think?

zbus/src/connection/socket_reader.rs
```diff
@@ -153,3 +160,11 @@ impl SocketReader {
...
+    pub(crate) activity_event: Event,
+    pub(crate) closed: AtomicBool,
+    pub(crate) closed_event: Event,
```
zeenix: These can just `pub` because they can't be more exposed that the struct itself.

bilelmoussaoui:
Updated the patch again.

zbus/src/connection/mod.rs
```diff
@@ -1144,7 +1151,31 @@ impl Connection {
...
+    /// Returns immediately if the connection is already closed.
+    pub async fn closed(&self) {
+        loop {
+            let listener = self.inner.socket_status.closed_event.listen();
+            if self.inner.socket_status.closed.load(Ordering::Acquire) {
+                return;
+            }
+            listener.await;
+        }
     }
```
zeenix, line 1179: The loop is correct (the listener is registered before the flag load and `notify` always follows the store, so there is no missed-wakeup window) but it registers a listener *before* the first flag check. That means the already-closed fast path—which `closed_resolves_immediately_if_already_closed` exercises—always pays a listener registration for nothing. `event_listener`'s own documented pattern checks the flag first:

```rust
loop {
    if self.inner.socket_status.closed.load(Ordering::Acquire) {
        return;
    }
    let listener = self.inner.socket_status.closed_event.listen();
    if self.inner.socket_status.closed.load(Ordering::Acquire) {
        return;
    }
    listener.await;
}
```

FWIW the loop can only ever go around once: `closed` is a monotonic latch and `event_listener` doesn't wake listeners spuriously, so after `listener.await` the flag is necessarily `true`. It's dead beyond the re-check, but it costs nothing and guards against the flag ever becoming non-monotonic, so no objection to keeping it.

Generated by Claude Opus 4.8.

zbus/src/connection/socket_reader.rs
```diff
@@ -12,7 +18,6 @@ use crate::{
 
 use super::socket::ReadHalf;
 
-#[derive(Debug)]
 pub(crate) struct SocketReader {
```
zeenix, line 22: `#[derive(Debug)]` got dropped from `SocketReader` here. `SocketStatus` derives `Debug` so nothing forces this—it still compiles with the derive restored. Please put it back, as it's unrelated to this change.

Generated by Claude Opus 4.8.

zbus/src/blocking/connection/mod.rs
```diff
@@ -250,6 +250,18 @@ impl Connection {
...
 
+    /// Returns `true` if the underlying socket has been closed.
+    pub fn is_closed(&self) -> bool {
+        self.inner.is_closed()
+    }
+
+    /// Block until the connection's underlying socket is closed.
+    ///
+    /// Returns immediately if the connection is already closed.
+    pub fn closed(&self) {
```
zeenix: These docs still describe the *underlying socket*, while the async versions were reworded to talk about the connection. Please keep them in sync, including the "remote peer disconnects, an I/O error occurs on the socket, or `close` is called" sentence that the async docs now carry.

Generated by Claude Opus 4.8.

zbus/src/connection/mod.rs
zeenix: Ignore this one.

zbus/src/blocking/connection/mod.rs
bilelmoussaoui: Done.

bilelmoussaoui:
the benchmark regression is definitely unrelated lol

zeenix:
> the benchmark regression is definitely unrelated lol

You took 87 microseconds away from member name parsing, nooooo.

zeenix (approved):

zeenix:
> Claude Opus 4.6

Why not 4.8 btw? :thinking:

PR #1846, opened 2026-07-10
🥅 zb: Connection Error returning Address

### LINKED ISSUE #1478: Improve the connection error with Address

0xPxt:
Added Error::Connection, which shows the address when the the code panics.
Fixes https://github.com/z-galaxy/zbus/issues/1478

### DIFF
--- a/zbus/src/address/mod.rs
+++ b/zbus/src/address/mod.rs
@@ -56,7 +56,9 @@ impl Address {
 
     #[cfg_attr(any(target_os = "macos", windows), async_recursion::async_recursion)]
     pub(crate) async fn connect(self) -> Result<Stream> {
-        self.transport.connect().await
+        // FIXME: Avoid the unconditional clone of the whole `Address`.
+        let address = self.clone();
+        self.transport.connect(address).await
     }
 
     /// Get the address for the session socket respecting the `DBUS_SESSION_BUS_ADDRESS` environment
--- a/zbus/src/address/transport/mod.rs
+++ b/zbus/src/address/transport/mod.rs
@@ -6,14 +6,14 @@
 use crate::connection::socket::Command;
 #[cfg(windows)]
 use crate::win32::autolaunch_bus_address;
-use crate::{Error, Result};
+use crate::{Address, Error, Result};
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
-use std::collections::HashMap;
 #[cfg(not(feature = "tokio"))]
 use std::net::TcpStream;
 #[cfg(unix)]
 use std::os::unix::net::{SocketAddr, UnixStream};
+use std::{collections::HashMap, sync::Arc};
 #[cfg(feature = "tokio")]
 use tokio::net::TcpStream;
 #[cfg(feature = "tokio-vsock")]
@@ -100,7 +100,7 @@ pub enum Transport {
 
 impl Transport {
     #[cfg_attr(any(unix, windows), async_recursion::async_recursion)]
-    pub(super) async fn connect(self) -> Result<Stream> {
+    pub(super) async fn connect(self, address: Address) -> Result<Stream> {
         match self {
             Transport::Unix(unix) => {
                 // This is a `path` in case of Windows until uds_windows provides the needed API:
@@ -122,9 +122,11 @@ impl Transport {
                 let stream = crate::Task::spawn_blocking(
                     move || -> Result<_> {
                         #[cfg(unix)]
-                        let stream = UnixStream::connect_addr(&addr)?;
+                        let stream = UnixStream::connect_addr(&addr)
+                            .map_err(|e| Error::Connection(Arc::new(e), address))?;
                         #[cfg(windows)]
-                        let stream = UnixStream::connect(addr)?;
+                        let stream = UnixStream::connect(addr)
+                            .map_err(|e| Error::Connection(Arc::new(e), address))?;
                         stream.set_nonblocking(true)?;
 
                         Ok(stream)
@@ -156,10 +158,11 @@ impl Transport {
                 }
             }
             #[cfg(unix)]
-            Transport::Unixexec(unixexec) => unixexec.connect().await.map(Stream::Unixexec),
+            Transport::Unixexec(unixexec) => unixexec.connect(&address).await.map(Stream::Unixexec),
             #[cfg(all(feature = "vsock", not(feature = "tokio")))]
             Transport::Vsock(addr) => {
-                let stream = VsockStream::connect_with_cid_port(addr.cid(), addr.port())?;
+                let stream = VsockStream::connect_with_cid_port(addr.cid(), addr.port())
+                    .map_err(|e| Error::Connection(Arc::new(e), address))?;
                 Async::new(stream).map(Stream::Vsock).map_err(Into::into)
             }
 
@@ -168,13 +171,13 @@ impl Transport {
                 VsockStream::connect(tokio_vsock::VsockAddr::new(addr.cid(), addr.port()))
                     .await
                     .map(Stream::Vsock)
-                    .map_err(Into::into)
+                    .map_err(|e| Error::Connection(Arc::new(e), address))
             }
 
             Transport::Tcp(mut addr) => match addr.take_nonce_file() {
                 Some(nonce_file) => {
                     #[allow(unused_mut)]
-                    let mut stream = addr.connect().await?;
+                    let mut stream = addr.connect(&address).await?;
 
                     #[cfg(unix)]
                     let nonce_file = {
@@ -208,7 +211,7 @@ impl Transport {
 
                     Ok(Stream::Tcp(stream))
                 }
-                None => addr.connect().await.map(Stream::Tcp),
+                None => addr.connect(&address).await.map(Stream::Tcp),
             },
 
             #[cfg(windows)]
@@ -224,8 +227,8 @@ impl Transport {
 
             #[cfg(target_os = "macos")]
             Transport::Launchd(launchd) => {
-                let addr = launchd.bus_address().await?;
-                addr.connect().await
+                let transport = launchd.bus_address().await?;
+                transport.connect(address).await
             }
 
             #[cfg(unix)]
--- a/zbus/src/address/transport/tcp.rs
+++ b/zbus/src/address/transport/tcp.rs
@@ -1,5 +1,5 @@
 use super::encode_percents;
-use crate::{Error, Result};
+use crate::{Address, Error, Result};
 #[cfg(not(feature = "tokio"))]
 use async_io::Async;
 #[cfg(not(feature = "tokio"))]
@@ -8,6 +8,7 @@ use std::{
     collections::HashMap,
     fmt::{Display, Formatter},
     str::FromStr,
+    sync::Arc,
 };
 #[cfg(feature = "tokio")]
 use tokio::net::TcpStream;
@@ -128,20 +129,25 @@ impl Tcp {
     }
 
     #[cfg(not(feature = "tokio"))]
-    pub(super) async fn connect(self) -> Result<Async<TcpStream>> {
+    pub(super) async fn connect(self, address: &Address) -> Result<Async<TcpStream>> {
+        let address_clone = address.clone();
+        let family = self.family();
         let addrs = crate::Task::spawn_blocking(
             move || -> Result<Vec<SocketAddr>> {
-                let addrs = (self.host(), self.port()).to_socket_addrs()?.filter(|a| {
-                    if let Some(family) = self.family() {
-                        if family == TcpTransportFamily::Ipv4 {
-                            a.is_ipv4()
+                let addrs = (self.host(), self.port())
+                    .to_socket_addrs()
+                    .map_err(|e| Error::Connection(Arc::new(e), address_clone))?
+                    .filter(|a| {
+                        if let Some(family) = self.family() {
+                            if family == TcpTransportFamily::Ipv4 {
+                                a.is_ipv4()
+                            } else {
+                                a.is_ipv6()
+                            }
                         } else {
-                            a.is_ipv6()
+                            true
                         }
-                    } else {
-                        true
-                    }
-                });
+                    });
                 Ok(addrs.collect())
             },
             "connect tcp",
@@ -149,12 +155,16 @@ impl Tcp {
         .await
         .map_err(|e| Error::Address(format!("Failed to receive TCP addresses: {e}")))??;
 
+        let mut last_err = Error::Address(match family {
+            Some(family) => format!("no `{family}` addresses found for `{address}`"),
+            None => format!("no addresses found for `{address}`"),
+        });
+
         // we could attempt connections in parallel?
-        let mut last_err = Error::Address("Failed to connect".into());
         for addr in addrs {
             match Async::<TcpStream>::connect(addr).await {
                 Ok(stream) => return Ok(stream),
-                Err(e) => last_err = e.into(),
+                Err(e) => last_err = Error::Connection(Arc::new(e), address.clone()),
             }
         }
 
@@ -162,10 +172,10 @@ impl Tcp {
     }
 
     #[cfg(feature = "tokio")]
-    pub(super) async fn connect(self) -> Result<TcpStream> {
+    pub(super) async fn connect(self, address: &Address) -> Result<TcpStream> {
         TcpStream::connect((self.host(), self.port()))
             .await
-            .map_err(|e| Error::InputOutput(e.into()))
+            .map_err(|e| Error::Connection(Arc::new(e), address.clone()))
     }
 }
 
--- a/zbus/src/error.rs
+++ b/zbus/src/error.rs
@@ -3,7 +3,7 @@ use zbus_names::{Error as NamesError, InterfaceName, OwnedErrorName};
 use zvariant::{Error as VariantError, ObjectPath};
 
 use crate::{
-    fdo,
+    Address, fdo,
     message::{Message, Type},
 };
 
@@ -60,6 +60,8 @@ pub enum Error {
     InvalidSerial,
     /// The given interface already exists at the given path.
     InterfaceExists(InterfaceName<'static>, ObjectPath<'static>),
+    /// Failed to connect to the D-Bus server at the given address.
+    Connection(Arc<io::Error>, Address),
 }
 
 impl PartialEq for Error {
@@ -85,6 +87,7 @@ impl PartialEq for Error {
             (Error::InputOutput(_), Self::InputOutput(_)) => false,
             (Self::Failure(s1), Self::Failure(s2)) => s1 == s2,
             (Self::InterfaceExists(s1, s2), Self::InterfaceExists(o1, o2)) => s1 == o1 && s2 == o2,
+            (Self::Connection(_, a1), Self::Connection(_, a2)) => a1 == a2,
             (_, _) => false,
         }
     }
@@ -114,6 +117,7 @@ impl error::Error for Error {
             Error::MissingParameter(_) => None,
             Error::InvalidSerial => None,
             Error::InterfaceExists(_, _) => None,
+            Error::Connection(e, _) => Some(e),
         }
     }
 }
@@ -149,6 +153,7 @@ impl fmt::Display for Error {
             }
             Error::InvalidSerial => write!(f, "Serial number in the message header is 0"),
             Error::InterfaceExists(i, p) => write!(f, "Interface `{i}` already exists at `{p}`"),
+            Error::Connection(e, addr) => write!(f, "Failed to connect to address `{addr}`: {e}"),
         }
     }
 }
@@ -182,6 +187,7 @@ impl Error {
             Error::MissingParameter(_) => Some("A required parameter is missing"),
             Error::InvalidSerial => Some("serial number in the message header is 0"),
             Error::InterfaceExists(_, _) => Some("interface already exists"),
+            Error::Connection(_, _) => Some("could not connect to specified address"),
         }
     }
 }
@@ -212,6 +218,7 @@ impl Clone for Error {
             Error::MissingParameter(p) => Error::MissingParameter(p),
             Error::InvalidSerial => Error::InvalidSerial,
             Error::InterfaceExists(i, p) => Error::InterfaceExists(i.clone(), p.clone()),
+            Error::Connection(e, addr) => Error::Connection(e.clone(), addr.clone()),
         }
     }
 }
--- a/zbus/tests/issue/issue_1478.rs
+++ b/zbus/tests/issue/issue_1478.rs
@@ -1,0 +1,44 @@
+use ntest::timeout;
+use zbus::{Error, connection};
+
+const UNIX_ADDRESS: &str = "unix:path=/this/path/does/not/exist";
+const TCP_ADDRESS: &str = "tcp:host=localhost,port=4142,family=ipv4";
+#[cfg(unix)]
+const UNIXEXEC_ADDRESS: &str = "unixexec:path=/this/path/does/not/exist";
+#[cfg(any(
+    all(feature = "vsock", not(feature = "tokio")),
+    feature = "tokio-vsock"
+))]
+const VSOCK_ADDRESS: &str = "vsock:cid=2,port=0";
+
+#[test]
+#[timeout(15000)]
+fn connection_error() {
+    // Addresses issue [#1478](https://github.com/z-galaxy/zbus/issues/1478). The issue mentions
+    // that connection error troubleshooting could be simplified by surfacing the connection
+    // address to the user. This test ensures connection failures throw the error Error::Connection,
+    // and that such error shows the address involved in the attempted connection.
+    zbus::block_on(connection_error_async());
+}
+
+async fn connection_error_async() {
+    #[allow(unused_mut)]
+    let mut addresses = vec![UNIX_ADDRESS, TCP_ADDRESS];
+    #[cfg(unix)]
+    addresses.push(UNIXEXEC_ADDRESS);
+    #[cfg(any(
+        all(feature = "vsock", not(feature = "tokio")),
+        feature = "tokio-vsock"
+    ))]
+    addresses.push(VSOCK_ADDRESS);
+
+    for addr in addresses {
+        let res = connection::Builder::address(addr).unwrap().build().await;
+
+        let Err(Error::Connection(_, error_addr)) = res else {
+            panic!("expected a connection error, got {res:?}");
+        };
+
+        assert_eq!(error_addr.to_string(), addr);
+    }
+}
--- a/zbus/src/address/transport/unixexec.rs
+++ b/zbus/src/address/transport/unixexec.rs
@@ -1,9 +1,9 @@
 use std::{
     borrow::BorrowMut, ffi::OsString, fmt::Display, os::unix::ffi::OsStrExt, path::PathBuf,
-    process::Stdio,
+    process::Stdio, sync::Arc,
 };
 
-use crate::process::Command;
+use crate::{Address, process::Command};
 
 use super::encode_percents;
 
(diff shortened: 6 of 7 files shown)


zeenix:
@0xPxt thanks for this pr, I'll review myself too later. One thing, please rebase on main, not merge it so history is easier to follow and the PR only contains the relevant commits. 🙏

zbus/src/address/mod.rs
```diff
@@ -56,7 +56,11 @@ impl Address {
 
     #[cfg_attr(any(target_os = "macos", windows), async_recursion::async_recursion)]
     pub(crate) async fn connect(self) -> Result<Stream> {
-        self.transport.connect().await
+        let addr = self.to_string();
+        self.transport
+            .connect()
+            .await
+            .map_err(|e| Error::Connection(Box::new(e), addr))
```
zeenix: Instead of mapping blindly and ignoring the specific errors, we should change Transport::connect (and/or methods it calls) to return the new error variant where that's the case.

zbus/src/error.rs
```diff
@@ -60,6 +60,8 @@ pub enum Error {
     InvalidSerial,
     /// The given interface already exists at the given path.
     InterfaceExists(InterfaceName<'static>, ObjectPath<'static>),
+    /// Connection failed (includes the attempted address).
+    Connection(Box<Error>, String),
```
zeenix: 1. Isn't the underlying error always std::io::Error? If so, we should use that instead of boxing.
2. Instead of String, it'd be best to have Address type here.

zeenix (commented):
Looks good in general otherwise. I appreciate you following the contribution guide but could you please add details to the commit message(s) too? 

Btw there's a specific gimoji for catching errors. 😉

zeenix (changes requested):
Forgot to click "request charges".

zbus/src/error.rs
0xPxt: Changing from String to Address forced me to clone the Address. Is that ok? Or should I try to restructure a little bit in order to avoid such clone?

zeenix: > Changing from String to Address forced me to clone the Address.

That's fine, I think. For String you'll need to allocate, so the worst case scenario, Address also allocates.

zbus/src/address/mod.rs
zeenix: This isn't yet addressed at all.

0xPxt: > This isn't yet addressed at all.

I am sorry, I misunderstood what you meant... Now the error is mapped in the places that raise a connection error only. I think I did not leave any uncovered but I'll let you tell @zeenix since you're much more familiar with the code.

zeenix: > I am sorry, I misunderstood what you meant...

Not a problem. Next time, just ask. :)

> Now the error is mapped in the places that raise a connection error only.

Thanks. My concern wasn't about mapping happening **here**, but rather more with the **blind** mapping. Could you please ensure that all errors being mapped are `io::Error` coming from actually trying to connect, and no others?

0xPxt: > Could you please ensure that all errors being mapped are io::Error coming from actually trying to connect, and no others?

I just double-checked and all of them are connection errors.

The only thing that I am not sure about is that Tokio's TCP does :
``` Rust
 pub async fn connect<A: ToSocketAddrs>(addr: A) -> io::Result<TcpStream> {
    let addrs = to_socket_addrs(addr).await?;

    let mut last_err = None;

    for addr in addrs {
        match TcpStream::connect_addr(addr).await {
            Ok(stream) => return Ok(stream),
            Err(e) => last_err = Some(e),
        }
    }

    Err(last_err.unwrap_or_else(|| {
        io::Error::new(
            io::ErrorKind::InvalidInput,
            "could not resolve to any address",
        )
    }))
}

 ```
So the DNS lookup that happens in `to_socket_addrs` can result in an error... I don't know if that falls exactly under "Connection error" but I think it's a minor nuance given that this happens inside a function call `connect`.
What do you think @zeenix?

Comment (zeenix) [reply zbus/src/address/mod.rs]: > > Could you please ensure that all errors being mapped are io::Error coming from actually trying to connect, and no others?
> 
> I just double-checked and all of them are connection errors.

:+1: I assume they're all `io::Error` and you're not creating a new `io::Error` in mapping?

> The only thing that I am not sure about is that Tokio's TCP does :
> 
> ```rust
>  pub async fn connect<A: ToSocketAddrs>(addr: A) -> io::Result<TcpStream> {
>     let addrs = to_socket_addrs(addr).await?;
> 
>     let mut last_err = None;
> 
>     for addr in addrs {
>         match TcpStream::connect_addr(addr).await {
>             Ok(stream) => return Ok(stream),
>             Err(e) => last_err = Some(e),
>         }
>     }
> 
>     Err(last_err.unwrap_or_else(|| {
>         io::Error::new(
>             io::ErrorKind::InvalidInput,
>             "could not resolve to any address",
>         )
>     }))
> }
> ```
> 
> So the DNS lookup that happens in `to_socket_addrs` can result in an error... I don't know if that falls exactly under "Connection error"

You're constructing a new `io::Error`, which likely means that the underlying error is not already `io::Error` and hence we shouldn't do this here. Recall that I asked for you to check that all mapped errors are `io::Error`   in a previous comment to ensure that it's the right call for `Error::Connection` to contain `io::Error`.

> but I think it's a minor nuance given that this happens inside a function call `connect`. What do you think @zeenix?

The goal here is to return a more informed error in cases where we're not doing that already. We need to ensure that's the case.

Comment (zeenix) [reply zbus/src/address/mod.rs]: > You're constructing a new `io::Error`, which likely means that the underlying error is not already `io::Error` and hence we shouldn't do this here. Recall that I asked for you to check that all mapped errors are `io::Error` in a previous comment to ensure that it's the right call for `Error::Connection` to contain `io::Error`.

Oh wait, that code is from tokio, not yours. :) That is an `io::Error` and yeah, it looks like it should be mapped to `Error::Connection` indeed.

**Review by zeenix — COMMENTED** [review]
> [!NOTE]
> This review (including the inline comments) was drafted by Claude Opus 4.8 at my request. The failure table below comes from actually building the branch and running each address against both backends locally, not from a read-through. I have reviewed it before posting, but flagging the authorship so you know where it came from.

Thanks for sticking with this! The earlier points are now properly addressed: the payload is `Arc<io::Error>` instead of a boxed `Error`, it carries `Address` instead of `String`, and the mapping happens at the individual connect calls rather than blindly in `Address::connect`.

I went through all six mapping sites, and I can confirm every one of them wraps an `io::Error` that comes straight out of a real connect call — nothing fabricated, no non-connect errors swept in. It also builds clean for me with `--features vsock`, `--no-default-features --features tokio`, and `tokio,tokio-vsock` (that last combo is what proves the `tokio-vsock` error really is an `io::Error`).

So the "are all mapped errors connect errors?" half is done. The remaining question is the other half: **are all connect errors mapped?** Not yet. I drove each transport through `Builder::address(..).build()` on both backends:

| address | async-io | tokio |
|---|---|---|
| `unix:path=/definitely/not/a/socket` | `Connection` ✅ | `Connection` ✅ |
| `tcp:host=127.0.0.1,port=1` | `Connection` ✅ | `Connection` ✅ |
| `unixexec:path=/definitely/not/a/binary` | `InputOutput` ❌ | `InputOutput` ❌ |
| `tcp:host=no.such.host.invalid,port=1234` | `InputOutput` ❌ | `Connection` ✅ |
| `tcp:host=127.0.0.1,port=1,family=ipv6` | `Address("Failed to connect")` ❌ | `Connection` ✅ |

The three ❌ rows are inline comments below. Note how the same user-visible failure currently produces a different error variant depending on which runtime feature is enabled — that's the bit I'd most like fixed.

A few things that don't fit on a line:

**Tests.** The PR adds a public error variant and changes which variant users get on a very common failure, but nothing asserts any of it. `Address::connect` is `pub(crate)`, but `connection::Builder::address("unix:path=/nope")?.build()` drives it fine from an integration test — that's exactly how I produced the table above, so it should only be a few lines. Please add at least the unix and tcp cases.

**Please mention the behaviour change in the PR description.** `Error` is `#[non_exhaustive]`, so nothing downstream fails to compile, but code that used to match `Error::InputOutput` on a connect failure will now silently fall through to its catch-all. That's fine for a minor release, but it should be visible in the release notes.

Unrelated to this PR, but that last table row also revealed that the tokio path ignores `family=` entirely — it happily connected over IPv4 with `family=ipv6` requested. I'll file that separately; please don't try to fix it here.

File: zbus/src/address/transport/mod.rs [hunk 3]
```diff
@@ -159,7 +161,8 @@ impl Transport {
             Transport::Unixexec(unixexec) => unixexec.connect().await.map(Stream::Unixexec),
```
zeenix: `unixexec` isn't covered. `Unixexec::connect` does `Command::for_unixexec(self)…spawn()?`, and a spawn failure is an `io::Error` — for this transport, spawning the process **is** connecting, so it qualifies under the same rule as the others.

Today a typo in `path=` gives you a bare `I/O error: No such file or directory (os error 2)` with no mention of the address, which is precisely the complaint in #1478. Could you map this one too?

zbus/src/address/transport/tcp.rs
```diff
@@ -128,7 +129,7 @@ impl Tcp {
     }
 
     #[cfg(not(feature = "tokio"))]
-    pub(super) async fn connect(self) -> Result<Async<TcpStream>> {
+    pub(super) async fn connect(self, address: &Address) -> Result<Async<TcpStream>> {
         let addrs = crate::Task::spawn_blocking(
             move || -> Result<Vec<SocketAddr>> {
                 let addrs = (self.host(), self.port()).to_socket_addrs()?.filter(|a| {
```
zeenix: DNS failures are backend-dependent right now.

In the tokio path, resolution happens inside `TcpStream::connect`, so a lookup failure lands in `Error::Connection` — which, as I said in the other thread, is correct. But here in the async-io path we do the resolution ourselves, and the `?` on `to_socket_addrs()` turns it into `Error::InputOutput`:

```
tcp:host=no.such.host.invalid,port=1234
  async-io: I/O error: failed to lookup address information: Name or service not known
  tokio:    Failed to connect to address `tcp:host=no.such.host.invalid,port=1234`: failed to lookup address information: Name or service not known
```

Same failure, different variant, and only one of them names the address. It's still an `io::Error` from the connect attempt, so please map it to `Error::Connection` here as well so the two backends agree.

zbus/src/address/transport/tcp.rs
```diff
@@ -154,18 +155,18 @@ impl Tcp {
         for addr in addrs {
             match Async::<TcpStream>::connect(addr).await {
                 Ok(stream) => return Ok(stream),
-                Err(e) => last_err = e.into(),
+                Err(e) => last_err = Error::Connection(Arc::new(e), address.clone()),
```
zeenix, line 167: This assignment is fine, but the `last_err` it overwrites is the problem: the seed a couple of lines up is `Error::Address("Failed to connect".into())`, and that's what gets returned whenever the loop body never runs — i.e. when the family filter leaves the address list empty:

```
tcp:host=127.0.0.1,port=1,family=ipv6
  async-io: address error: Failed to connect
```

No address, no cause — the exact defect this PR exists to fix, surviving in the one place a seeded-sentinel pattern hides it.

I'd drop the sentinel entirely and early-return on an empty list with an error that actually names the address. I'd rather you *didn't* fabricate an `io::Error` just to force it into `Error::Connection` (tokio does that internally, but we shouldn't) — an `Error::Address` mentioning the address reads more honestly here, since a family filter matching nothing is a configuration problem, not an I/O failure.

zbus/src/error.rs
```diff
@@ -60,6 +60,8 @@ pub enum Error {
     InvalidSerial,
     /// The given interface already exists at the given path.
     InterfaceExists(InterfaceName<'static>, ObjectPath<'static>),
+    /// Connection failed (includes the attempted address).
```
zeenix: The parenthetical just restates what the type right below already says. Something like:

```rust
/// Failed to connect to the D-Bus server at the given address.
```

carries more for the reader.

zbus/src/error.rs
```diff
@@ -60,6 +60,8 @@ pub enum Error {
...
+    Connection(Arc<io::Error>, Address),
 }
 
 impl PartialEq for Error {
```
zeenix, line 67: `Connection` gets no arm here, so it falls through to the `(_, _) => false` catch-all. That happens to match how `InputOutput` behaves, but `InputOutput` **says so explicitly**. Please add an explicit arm so the behaviour is intentional rather than incidental — and do consider comparing the addresses, since unlike `io::Error`, `Address` is `PartialEq`.

zbus/src/address/transport/tcp.rs
0xPxt: Dropping the centinel meant having to make last_err an Option (because the compiler can not see that is_empty is checked and thus marked the variable as possibly uninitialized).
I didn't drop it because I think the code is simpler this way, but I can change it if you feel strongly about this.

0xPxt:
> Please add at least the unix and tcp cases.

@zeenix Is it fine to add inline tests in the corresponding `src/` relevant file? Or should I add them under `tests/issue/`?
Do not review the new changes yet, as I am yet to include such tests.
btw thank you for your amazing responsiveness :smile:

zbus/src/address/transport/tcp.rs
```diff
@@ -128,44 +129,53 @@ impl Tcp {
...
+        let family = self.family();
         let addrs = crate::Task::spawn_blocking(
             move || -> Result<Vec<SocketAddr>> {
-                let addrs = (self.host(), self.port()).to_socket_addrs()?.filter(|a| {
-                    if let Some(family) = self.family() {
-                        if family == TcpTransportFamily::Ipv4 {
-                            a.is_ipv4()
+                let addrs = (self.host(), self.port())
+                    .to_socket_addrs()
+                    .map_err(|e| Error::Connection(Arc::new(e), address_clone))?
```
zeenix, line 139: btw can't this be:

```suggestion
                    .map_err(|e| Error::Connection(Arc::new(e), address.clone()))?
```

so we don't need `address_clone` then?

zbus/src/address/transport/tcp.rs
zeenix: @0xPxt Firstly, this comment was from Claude, which has no feelings. 😆 It made a compelling point though. It's not a problem to make `last_err` an option.

zeenix:
> @zeenix Is it fine to add inline tests in the corresponding `src/` relevant file? Or should I add them under `tests/issue/`?

I generally prefere the latter because that helps keep things separate.

zbus/src/address/transport/tcp.rs
0xPxt: I tried, but I don't think so because the closure captures `address` and then :

```
error[E0521]: borrowed data escapes outside of method
   --> zbus/src/address/transport/tcp.rs:134:21
    |
132 |       pub(super) async fn connect(self, address: &Address) -> Result<Async<TcpStream>> {
    |                                         -------  - let's call the lifetime of this reference `'1`
    |                                         |
    |                                         `address` is a reference that is only valid in the method body
133 |           let family = self.family();
134 |           let addrs = crate::Task::spawn_blocking(
    |  _____________________^
135 | |             move || -> Result<Vec<SocketAddr>> {
136 | |                 let addrs = (self.host(), self.port())
137 | |                     .to_socket_addrs()
...   |
152 | |             "connect tcp",
153 | |         )
    | |         ^
    | |         |
    | |_________`address` escapes the method body here
    |           argument requires that `'1` must outlive `'static`
```
tbh I don't fully understand the error but it seems to be related to spawn_blocking requiring a 'static closure.

zeenix: > I tried, but I don't think so because the closure captures `address` and then :

Thanks for trying. I wasn't sure it's possible either. :)

0xPxt:
Hi @zeenix, I just pushed the tests.
I am ultimately leaving clicking on "resolve conversation" to you, once you think it's adequate. Every comment has been addressed now.

zeenix:
> Hi @zeenix, I just pushed the tests.
> I am ultimately leaving clicking on "resolve conversation" to you, once you think it's adequate. Every comment has been addressed now.

Cool. I resolved what I was certain about and I'll have Claude take another look. In the meantime, the last commit has the "wrong" (well, not the best fitting) emoji. Kindly always thing about this per commit. :pray:

zeenix:
Also "map_err in connection error locations" is very vague, I think. How about "throw Error::Connection on connection failure"? If you also name the error by the name and path in the first commit, then it's easier to follow from the titles of the commit, what is going on.

zbus/src/address/transport/mod.rs
```diff
@@ -224,8 +227,8 @@ impl Transport {
 
             #[cfg(target_os = "macos")]
             Transport::Launchd(launchd) => {
-                let addr = launchd.bus_address().await?;
-                addr.connect().await
+                let transport = launchd.bus_address().await?;
```
0xPxt, line 230: I renamed this variable in my last force-push, I think addr was misleading in this variant since the returned value is a Result<Transport>.
`pub(super) async fn bus_address(&self) -> Result<Transport>`

Hope that's ok.

zbus/src/abstractions/process.rs
```diff
@@ -87,7 +87,7 @@ impl Command {
     }
 
     /// Executes the command as a child process, returning a handle to it.
-    pub fn spawn(&mut self) -> Result<Child, Error> {
+    pub fn spawn(&mut self) -> std::io::Result<Child> {
```
zeenix: unrelated change. Please remove.

zbus/tests/issue/issue_1478.rs
```diff
@@ -0,0 +1,40 @@
...
+const UNIXEXEC_ADDRESS: &str = "unixexec:path=/this/path/does/not/exist";
+#[cfg(any(
+    all(feature = "vsock", not(feature = "tokio")),
+    feature = "tokio-vsock"
+))]
+const VSOCK_ADDRESS: &str = "vsock:cid=2,port=0";
+
+#[test]
+#[timeout(15000)]
+fn connection_error() {
```
zeenix, line 16: It would be good to add a comment describing the issue briefly and adding a link to it would be a cherry on top.

See other tests in this dir.

zeenix (changes requested):
LGTM otherwise. We're pretty close to a merge, I know it. :)

zbus/src/address/transport/tcp.rs
zeenix: was this addressed? 🤔

0xPxt: Would you rather do something similar to :

```rust
        if addrs.is_empty() {
            return Err(Error::Address(match family {
                Some(family) => format!("no `{family}` addresses found for `{address}`"),
                None => format!("no addresses found for `{address}`"),
            }));
        }

        let mut last_err: Option<Error> = None;

        // we could attempt connections in parallel?
        for addr in addrs {
            match Async::<TcpStream>::connect(addr).await {
                Ok(stream) => return Ok(stream),
                Err(e) => last_err = Some(Error::Connection(Arc::new(e), address.clone())),
            }
        }

        Err(last_err.expect("non-empty `addrs` always sets `last_err` or returns"))
```

instead of the current :

```rust
        let mut last_err = Error::Address(match family {
            Some(family) => format!("no `{family}` addresses found for `{address}`"),
            None => format!("no addresses found for `{address}`"),
        });

        // we could attempt connections in parallel?
        for addr in addrs {
            match Async::<TcpStream>::connect(addr).await {
                Ok(stream) => return Ok(stream),
                Err(e) => last_err = Error::Connection(Arc::new(e), address.clone()),
            }
        }

        Err(last_err)
```
?

zbus/tests/issue/issue_1478.rs
```diff
@@ -0,0 +1,45 @@
...
+    // Addresses issue [#1478](https://github.com/z-galaxy/zbus/issues/1478).
+    // The issue mentions that connection error troubleshooting could be
+    // simplified by surfacing the connection address to the user.
+    // This test ensures connection failures throw the error Error::Connection,
```
zeenix: Please don't break lines just to start a new sentence. I know this is nowadays quite common but it's grammatically incorrect and its benefits are subjective at best. I tolerate this (not that I have a choice 😆) in issue/PR comments but I don't want my project's code to perpetuate this. 🙏

zbus/src/address/transport/tcp.rs
```diff
@@ -128,44 +129,53 @@ impl Tcp {
     }
 
     #[cfg(not(feature = "tokio"))]
-    pub(super) async fn connect(self) -> Result<Async<TcpStream>> {
+    pub(super) async fn connect(self, address: &Address) -> Result<Async<TcpStream>> {
+        let address_clone = address.clone();
+        let family = self.family();
```
zeenix, line 134: don't think this variable is needed. `self.family()` isn't long to type/read and you're calling `self.family` again below anyway.

zbus/src/address/transport/tcp.rs
zeenix: I was more curious about whether or not the actual issue pointed out in the first comment (vague error in one case) in this thread was resolved or not. Looks to me that it was but just making sure.

zeenix (changes requested):
one tiny change and one clarification to go. :)

zbus/tests/issue/issue_1478.rs
0xPxt: Oh, I saw that in the rest of `issue_<num>.rs` test files and I did the same, would you rather I just put that whole text in a single line? Or break line after a period (`.`)?

zbus/src/address/transport/tcp.rs
0xPxt: I added it because self is moved to the closure right below it.
The compiler complains otherwise :
```
error[E0382]: borrow of moved value: `self`
   --> zbus/src/address/transport/tcp.rs:158:45
    |
132 |     pub(super) async fn connect(self, address: &Address) -> Result<Async<TcpStream>> {
    |                                 ---- move occurs because `self` has type `Tcp`, which does not implement the `Copy` trait
...
135 |             move || -> Result<Vec<SocketAddr>> {
    |             ---------------------------------- value moved into closure here
136 |                 let addrs = (self.host(), self.port())
    |                              ---- variable moved due to use in closure
...
158 |             return Err(Error::Address(match self.family() {
    |                                             ^^^^ value borrowed here after move
    |
help: consider cloning the value before moving it into the closure
    |
134 ~         let value = self.clone();
135 ~         let addrs = crate::Task::spawn_blocking(
136 |             move || -> Result<Vec<SocketAddr>> {
137 ~                 let addrs = (value.host(), self.port())
    |
```

zbus/src/address/transport/tcp.rs
0xPxt: oh, my bad! it was indeed addressed :)

zbus/tests/issue/issue_1478.rs
zeenix: > would you rather I just put that whole text in a single line? 

Of course not. :) The correct thing is to break the lines at the charachter limit (100 chars) at the word-boundary (i-e not breaking the line in the middle of the word).

> Or break line after a period (`.`)?

That's what I was asking not to do. :)

zbus/src/address/transport/tcp.rs
zeenix: ah, `self` itself is moved into the closure. NM.

zbus/tests/issue/issue_1478.rs
0xPxt: Done!

zbus/src/address/mod.rs
```diff
@@ -56,7 +56,8 @@ impl Address {
 
     #[cfg_attr(any(target_os = "macos", windows), async_recursion::async_recursion)]
     pub(crate) async fn connect(self) -> Result<Stream> {
-        self.transport.connect().await
+        let address = self.clone();
+        self.transport.connect(address).await
```
zeenix, line 61: I'm so sorry but I only caught this now: Instead of prematurely cloning the addres, why can't we pass a reference and let whoever needs the clone, clone it? 🤔

zbus/tests/issue/issue_1478.rs
```diff
@@ -0,0 +1,45 @@
...
+const VSOCK_ADDRESS: &str = "vsock:cid=2,port=0";
+
+#[test]
+#[timeout(15000)]
+fn connection_error() {
+    // Addresses issue [#1478](https://github.com/z-galaxy/zbus/issues/1478). The issue mentions
+    // that connection error troubleshooting could be simplified by surfacing the connection
+    // address to the user. This test ensures connection failures throw the error
+    // Error::Connection, and that such error shows the address involved in the attempted
+    // connection.
```
zeenix: Wasn't going to mention this but since you're likely going to make a change still anyway:


```suggestion
    // address to the user. This test ensures connection failures throw the error Error::Connection,
    // and that such error shows the address involved in the attempted connection.
```

zeenix:
@0xPxt Thanks so much for your patience and perseverance in the face of my perfectionism. :+1:

zbus/src/address/mod.rs
0xPxt: > I'm so sorry but I only caught this now

No worries!

> let whoever needs the clone, clone it? 

By "whoever" you mean the branches in the match statement of transport.connect? As in, let the transport variant that needs the clone do the clone, and for the rest the clone would not be needed.

zeenix: > By "whoever" you mean the branches in the match statement of transport.connect? As in, let the transport variant that needs the clone do the clone

Yes. Whoever needs the an owned version, just clones it right where it's needed. :wink: 

> and for the rest the clone would not be needed.

Good!

0xPxt: Passing a reference directly is not allowed because self.transport is moved into Transport::connect.
```
pub(super) async fn connect(self, address: Address) -> Result<Stream>
```
I don't think it's that worth it then to do `self.transport.clone().connect(&self).await`, since then we would cloning the transport anyway in the cases that have the least clones and we would be cloning the transport as well as the address in the variants that require `adress.clone()`.

What do you think?

0xPxt: I just found that even if we clone the transport it's not possible to clone the address then in connect's body due to `move` in the closure :
```
error[E0521]: borrowed data escapes outside of method
   --> zbus/src/address/transport/mod.rs:122:30
    |
102 |       #[cfg_attr(any(unix, windows), async_recursion::async_recursion)]
    |                                      -------------------------------- lifetime `'life0` defined here
103 |       pub(super) async fn connect(self, address: &Address) -> Result<Stream> {
    |                                         ------- `address` is a reference that is only valid in the method body
...
122 |                   let stream = crate::Task::spawn_blocking(
    |  ______________________________^
123 | |                     move || -> Result<_> {
124 | |                         #[cfg(unix)]
125 | |                         let stream = UnixStream::connect_addr(&addr)
...   |
134 | |                     "unix stream connection",
135 | |                 )
    | |                 ^
    | |                 |
    | |_________________`address` escapes the method body here
    |                   argument requires that `'life0` must outlive `'static`

For more information about this error, try `rustc --explain E0521`.
```

zeenix: > Passing a reference directly is not allowed because self.transport is moved into Transport::connect.

Ah right, the Address::connect is consuming `self` by value. IIRC the idea was that address just gets transformed into a transport and then transport get transormed into a connected stream. Let's just add a `FIXME: ` comment on top of these lines that says that it'd be really great to avoid this unconditional cloning of Address, which isn't exactly zero-cost.

> just found that even if we clone the transport it's not possible to clone the address then in connect's body due to `move` in the closure :

I made this change locally and it works:

```
                         #[cfg(unix)]
                         let stream = UnixStream::connect_addr(&addr)?;
                         #[cfg(windows)]
-                        let stream = UnixStream::connect(addr)?;
+                       let stream = UnixStream::connect(&addr)?;
```

I only tested on the main branch though.

zeenix (approved):

PR #1881, opened 2026-07-25
Remove unnecessary clones

FranciscoTGouveia:
If I am not overlooking anything, I believe these clones are redundant and can thus be removed.

For reference, these changes were found by an optimizer developed as a research project.

### DIFF
--- a/zbus/src/connection/mod.rs
+++ b/zbus/src/connection/mod.rs
@@ -611,7 +611,7 @@ impl Connection {
                     loop {
                         let signal = lost_stream.next().await;
                         let inner = match weak_conn.upgrade() {
-                            Some(conn) => conn.inner.clone(),
+                            Some(conn) => conn.inner,
                             None => break,
                         };
 
@@ -662,7 +662,7 @@ impl Connection {
                         loop {
                             let signal = acquired_stream.next().await;
                             let inner = match weak_conn.upgrade() {
-                                Some(conn) => conn.inner.clone(),
+                                Some(conn) => conn.inner,
                                 None => break,
                             };
                             match signal {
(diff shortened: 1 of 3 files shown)


zeenix (changes requested):
Thanks! These are all very cheap clones so it depends on wether or not they are on the hot paths to make a significant enough difference but 1. hopefully our benchmarks will tell us something and 2. these also simplify the code and harmless at worst so I'll take these changes still. As the CI will inform you, you'll need to update the commit message to comply with our practices. While you're at it, please also add details to the commit message as well (just copy&paste the PR description).

In general, I'd advice to look at and follow the contributing.md when a project has one, before sending out a contribution. It saves time for the reviewers. 🙏 Don't let this discourage you from contributing though. :)

zeenix (commented):
Thanks for adding the emoji prefix but these are not random emojis. You need to choose one from a curated list. Please read through the contributing.md and follow the guidelines there. 🙏

zeenix:
Don't mind the "regression" report for a benchmark here. There is something up with Codspeed for this specific benchmark and this benchmark is unrelated to these changes anyway. Although I'm a bit sad that there was no performance improvements from these changes but that's ok and I thought it'd wouldn't necessarily bring significant enough improvements in performance.

FranciscoTGouveia:
Thanks for the review!

>Thanks for adding the emoji prefix but these are not random emojis. You need to choose one from a curated list.

I have actually read through the contributing guidelines -- albeit only after seeing the CI fail :smile: -- and believed that the :coffin: would be the most suitable for this change.
However, taking another look, this is not in fact "dead code" _per se_, so I decided to change it to :recycle:.

I believe the commit message is now much more appropriate.
Apologies for this back-and-forth and thanks once again for the review!

zeenix:
> However, taking another look, this is not in fact "dead code" _per se_, so I decided to change it to ♻️.

⚡️ would be the best IMO but since we don't see big performance improvements in the benchmarks (it's still possible that this makes a big difference for certain applications), ♻️ is fine too.

zeenix (approved):

Issue #1144, opened 2024-11-13
Drop GVariant support


zeenix:
As I expressed in #1125:

* the whole gvariant support is getting too much to handle for me, especially mixed with zbus etc.
* I've long abandoned the plans of using gvariant as a dbus2 protocol (which was its original intention).
* [`gvariant` crate](https://github.com/ostreedev/gvariant-rs/tree/master) exists and not only it is used by ostree folks, it actually lives in the ostree space and seems to be the officially blessed implementation.

So I'm inclined to drop this support in the zvariant 6.0.

zeenix:
> * [`gvariant` crate](https://github.com/ostreedev/gvariant-rs/tree/master) exists and not only it is used by ostree folks, it actually lives in the ostree space and seems to be the officially blessed implementation.

A bit more context about this: I originally added gvariant support in hopes that ostree folks will use it, since gvariant provides a nice API but seems they went for the `gvariant` crate, even though (IMO) it has a non-intuitive strange API that is not serde-compatible.

---

Anyway, another option here would be to split the gvariant support into a separate crate. However, I would not have the time and energy to do that. So if someone wants to take that up, please go ahead and let me me here.

bilelmoussaoui:
> A bit more context about this: I originally added gvariant support in hopes that ostree folks will use it, since gvariant provides a nice API but seems they went for the gvariant crate, even though (IMO) it has a non-intuitive strange API that is not serde-compatible.

After testing the gvariant API, it seems really not an alternative to zvariant in it current form. 

I will have to look and and see how much effort would take to keep the gvariant format support in-tree by helping keeping it working, if that is a possibility.

zeenix:
> I will have to look and and see how much effort would take to keep the gvariant format support in-tree by helping keeping it working, if that is a possibility.

I feel bad about making you do this. 😔 Tell you what, if ostree would switch to zvariant, I promise to keep the gvariant support and maintain it too.

Is that something you'd want to look into?

bilelmoussaoui:
Sure

zeenix:
> Sure

Great. I suggest making sure they'll be willing to accept such a contribution before putting any effort into it.

bilelmoussaoui:
actually now that i checked the ostree repository, it doesn't seem to use the gvariant crate anywhere. Or you were talking about flat-manager which manipulate an ostree repo gvariant format manually in https://github.com/flatpak/flat-manager/blob/master/src/ostree.rs?

[...]

PR #1905, opened 2026-08-15
🗑️ zv: Deprecate GVariant support

### LINKED ISSUE #1910: zvariant: D-Bus deserializer panics on a wire signature containing a maybe type
### Summary

When the `gvariant` feature is enabled, a D-Bus signature that contains a maybe (`m`) type parses successfully, but computing its D-Bus alignment hits an `unreachable!()` and panics instead of rejecting the signature as invalid for the D-Bus format. Since signatures can come from untrusted wire data, this is a robustness/DoS concern rather than a mere internal invariant violation.

### Reproduction

With `zvariant`'s `gvariant` feature enabled (or, equivalently, when another crate in the graph enables `zvariant_utils/gvariant`):

```rust
use zvariant::serialized::{Context, Data};
use zvariant::{LE, Signature};

let bytes = [0u8, 0, 0, 0]; // empty D-Bus array
let ctxt = Context::new_dbus(LE, 0);
let data = Data::new(&bytes[..], ctxt);
let sig = Signature::try_from("ami").unwrap(); // array of maybe-of-i32
let _: (Vec<Option<i32>>, usize) = data.deserialize_for_signature(&sig).unwrap();
```

Panics with:

```
internal error: entered unreachable code: Maybe type is not supported in D-Bus
```

The top-level deserialize path already rejects a bare maybe with a clean error, but a maybe nested inside another container (e.g. an array element) reaches the alignment computation before that check.

### Root cause

`Signature::alignment` for the D-Bus format treats the maybe variant as `unreachable!()`. The maybe type is GVariant-only, so a D-Bus signature should never contain one — but the parser accepts it (maybe parsing is only gated on the feature, not on the format), so the invariant can be violated from outside.

### Suggested fix

Recursively reject maybe-bearing signatures at the D-Bus serialization/deserialization boundary, returning an error instead of relying on an `unreachable!()` deep in alignment. This should cover nested positions and signature-valued fields, under both the `gvariant` feature and the unified-graph combination.

### Notes

- Pre-existing: reproduces under `--features gvariant` on `main`; not introduced by the GVariant deprecation work (#1144).
- Independent of the new `zgvariant` crate (which is GVariant-only and never uses the D-Bus alignment path).

zeenix:
Deprecates zvariant's GVariant support in 5.15.0, pointing users at the new standalone
[zgvariant](https://github.com/z-galaxy/zgvariant) crate. Part of the #1144 split; the
implementation itself is removed in 6.0 (issue stays open to track it).

What this PR does:
- `#[deprecated(since = "5.15.0")]` on the gvariant-gated public items (`Maybe`,
  `Value::Maybe`, `Context::new_gvariant`, the `gvariant` module and its leaked serializer
  types, the maybe signature constants). `Value::Maybe` carries a real attribute since `Value`
  is zvariant-owned. The variants of *shared* types (`Format::GVariant`, `Signature::Maybe`)
  get doc-notes instead of attributes: they are defined in `zvariant_utils` and used by
  zgvariant itself, so an attribute there would warn inside zgvariant.
- README/book/AGENTS.md point GVariant users at zgvariant.
- Removes the heavy coverage that now lives in the zgvariant repo: the benchmarks, the 6.8 MB
  ostree test fixture with its test, and the gvariant fuzz target. **The regular gvariant test
  suite is retained** and still runs in CI (`cargo t -p zvariant --features gvariant`, Linux
  and Windows) — the deprecated feature ships throughout 5.x, so shared-code changes must not
  regress it unnoticed until the 6.0 removal.
- `ostree-tests` stays declared (removing a published feature is a break); it still enables
  `gvariant`, only its own tests moved to the zgvariant repo.
- **Fixes a feature-unification compile break**: with `zvariant_utils/gvariant` enabled
  externally — the dependency graph of every zbus + zgvariant co-user — zvariant's matches over
  the shared `Format`/`Signature` enums were non-exhaustive (5× E0004). Wildcard error arms
  (directing users to zgvariant) fix it, guarded by a CI check. Verified in all feature states,
  including error-arm reachability.
- Bumps zvariant to 5.15.0 (deprecation attrs) and zvariant_utils to 4.2.0 (it gained
  `Signature::contains_maybe`), and raises zvariant's `zvariant_utils` floor to 4.2.0 so a
  resolver can't pair zvariant 5.15 with an older zvariant_utils lacking that method.
  zvariant_derive is bumped to 5.15.0 in lockstep with zvariant (their `version_group`).

cargo-semver-checks against the published versions: only additive minors. The only behavior
change to an existing feature configuration is the #1910 fix below: under `zvariant/gvariant`,
malformed D-Bus input carrying a maybe type now errors instead of panicking.

This also fixes a panic (#1910) that the coexistence commit would otherwise newly expose. A
D-Bus signature containing a maybe (`m`) type parses when the maybe type is available, then
panics at an `unreachable!()` while computing D-Bus alignment. The underlying `unreachable!()`
predates this work (it also fires under `--features gvariant`), but `zbus + zgvariant` does not
compile on `main`, so making that graph usable newly exposes the panic — a DoS vector via a
malicious peer's signature — to applications that never opted into `zvariant/gvariant`. The
D-Bus ser/de boundary now rejects maybe-bearing signatures (recursively, including variant
signatures read off the wire) with an error, verified by a regression test in the mixed-feature
CI configuration.

Fixes #1910. Refs #1144. Refs #1365.

https://claude.ai/code/session_012ukJQhhRSiD9SqJa2BzwgM


zvariant/fuzz/Cargo.lock
```diff
@@ -205,22 +226,22 @@ dependencies = [
...
  "quote",
- "syn",
+ "syn 3.0.3",
  "zvariant_utils",
 ]
 
 [[package]]
 name = "zvariant_utils"
-version = "3.4.0"
+version = "4.0.0"
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This lock entry is still stale: the path crate is 4.1.0, and the green fuzz job actually logs `Updating zvariant_utils v4.0.0 (...) -> v4.1.0`. Please regenerate this nested lock at HEAD. Since `cargo --locked fuzz` does not propagate `--locked` to cargo-fuzz’s inner Cargo invocation, please also add a direct `cargo metadata --manifest-path zvariant/fuzz/Cargo.toml --locked --format-version 1` CI check.

.github/workflows/rust.yml
```diff
@@ -120,8 +122,6 @@ jobs:
               -- --skip fdpass_systemd
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
             cargo --locked test --release --verbose --doc --no-default-features connection::Connection::executor
-          # zvariant only with ostree tests (which implicitly enables `gvariant` feature too).
-          cargo --locked t -p zvariant --features ostree-tests
```
zeenix, line 124: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Deprecation does not stop this implementation from shipping throughout 5.x. Removing every runtime GVariant test leaves future changes to shared zvariant code, `zvariant_utils`, derives, or dependencies able to regress the legacy feature while docs still compile. Tests in `zgvariant` exercise its fork, not this implementation. Please retain at least the core `cargo test -p zvariant --features gvariant` coverage until the implementation is actually removed in 6.0; the large fixture, benchmarks, and fuzzing can still move.

.github/workflows/rust.yml
```diff
@@ -189,7 +187,6 @@ jobs:
         run: |
           cargo install cargo-fuzz
           cargo --locked fuzz run --fuzz-dir zvariant/fuzz dbus -- -max_total_time=30 -max_len=100M
-          cargo --locked fuzz run --fuzz-dir zvariant/fuzz --features gvariant gvariant -- -max_total_time=30 -max_len=100M
```
zeenix, line 192: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This deletion needs to be folded into `921d3edc`, which removes the feature and fuzz target. At that commit the workflow still invokes the now-missing target, so the commit cannot pass CI independently. Likewise, fold the test-run deletions into `0ab8ac97`, and move the AGENTS fuzz-command deletion into `921d3edc`, so every commit follows the repository’s atomic-commit rule.

zvariant/src/value.rs
```diff
@@ -96,7 +97,10 @@ pub enum Value<'a> {
     Array(Array<'a>),
     Dict(Dict<'a, 'a>),
     Structure(Structure<'a>),
+    /// Deprecated: GVariant support has moved to the `zgvariant` crate; this deprecated copy
+    /// will be removed in zvariant 6.0.
     #[cfg(feature = "gvariant")]
+    #[allow(deprecated)]
```
zeenix, line 106: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Could `Value::Maybe` carry an actual `#[deprecated(since = "5.15.0", note = ...)]` attribute rather than only prose? `Value` is owned by zvariant—zgvariant defines its own enum—so the shared-`Signature` constraint does not apply here. As written, downstream code can pattern-match `Value::Maybe` without receiving any migration warning.

zvariant_utils/src/serialized.rs
```diff
@@ -6,6 +6,9 @@ pub enum Format {
     #[default]
     DBus,
     /// [GVariant](https://developer.gnome.org/glib/stable/glib-GVariant.html) format.
+    ///
+    /// Deprecated: GVariant support has moved to the `zgvariant` crate; this deprecated copy
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This is the shared `Format` definition actively used by `zgvariant` itself, not a deprecated copy, so it cannot be documented here as being removed when zvariant reaches 6.0. Please scope the note to zvariant’s old integration, e.g. “GVariant support in `zvariant` is deprecated …; use `zgvariant` for GVariant serialization.” Since `2451ae7c` changes zvariant_utils too, its package prefix should also include `zu` per CONTRIBUTING.md.

zvariant/src/ser.rs
```diff
@@ -79,6 +81,21 @@ where
...
+        // variant to this build regardless. `zvariant`'s own `#[cfg(feature = ...)]` can't
+        // detect that (Cargo features don't propagate that way), so the variant can't be
+        // named explicitly here without breaking the common case where it doesn't exist at
+        // all. Fall back to a wildcard instead: it's unreachable whenever `gvariant` is
+        // enabled (the two arms above are then exhaustive) or the variant doesn't exist.
+        #[cfg(not(feature = "gvariant"))]
+        #[allow(unreachable_patterns)]
+        _ => {
+            return Err(Error::Message(
+                "GVariant format requires the `gvariant` feature of `zvariant`".to_owned(),
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This new diagnostic directs users to enable the very feature this release deprecates. Since the arm exists specifically for the zvariant+zgvariant coexistence graph, please direct users to serialize with `zgvariant` (optionally mentioning `zvariant/gvariant` only as temporary legacy compatibility), and keep the equivalent serializer/deserializer messages in sync.

zvariant/Cargo.toml
```diff
@@ -14,8 +14,9 @@ readme = "README.md"
 
 [features]
 default = []
-# FIXME: Also allow disabling D-Bus support
+# Deprecated: GVariant support has moved to the `zgvariant` crate and will be removed in 6.0.
 gvariant = ["zvariant_derive/gvariant", "zvariant_utils/gvariant"]
+# Deprecated: inert compatibility alias (its tests moved to the zgvariant repo); removed in 6.0.
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This alias is not inert: enabling it still activates `gvariant` and materially changes the feature graph. Please describe it as a deprecated compatibility alias with no tests remaining. This comment change also belongs in `0ab8ac97`, where those tests are removed; in its current `9e4ecb91` commit, the tests still exist.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of zeenix: The feature-unification fix and current CI are sound, but the inline findings below should be addressed before merge.

zvariant/fuzz/Cargo.lock
zeenix: Regenerated. The nested lockfile is now refreshed in the fuzz-removal commit (picking up zvariant_utils 4.1.0) and again in the bump commit (zvariant 5.15.0), and a new CI commit adds the explicit `cargo metadata --manifest-path zvariant/fuzz/Cargo.toml --locked` check to the fuzz job.

.github/workflows/rust.yml
zeenix: Restored. The regular gvariant test suite stays until the 6.0 removal, and CI now runs `cargo --locked t -p zvariant --features gvariant` on Linux and Windows. Only the benchmarks, the 6.8M flatpak-summary fixture with its ostree test, and fuzzing remain removed (all live on in the zgvariant repo). The retained tests carry `#![cfg_attr(feature = "gvariant", allow(deprecated))]` since they exercise the now-deprecated API.

.github/workflows/rust.yml
zeenix: Restructured. The standalone CI commit is gone: the fuzz-run and AGENTS fuzz-command deletions moved into the fuzz-removal commit, and the workflow's test-run change (now a switch to `--features gvariant` rather than a deletion, per the other thread) lives in the bench/test-data-removal commit. Verified `cargo check --all-targets` passes at every commit of the stack.

zvariant/src/value.rs
zeenix: Done — `Value::Maybe` now carries `#[deprecated(since = "5.15.0", ...)]` and the internal match arms take targeted `#[allow(deprecated)]`.

zvariant_utils/src/serialized.rs
zeenix: Reworded to scope the note to zvariant's integration — the variant itself stays, as zgvariant uses it via zvariant_utils. The commit is now prefixed `zv,zu`.

zvariant/src/ser.rs
zeenix: Reworded all five sites (serializer, deserializer `Maybe` path, and both `Data` deserialize methods): they now direct users to the `zgvariant` crate and mention `zvariant`'s deprecated `gvariant` feature only as legacy compatibility.

zvariant/Cargo.toml
zeenix: Reworded — the comment now says the alias still enables `gvariant` and that only its own (ostree) tests moved to the zgvariant repo. The comment change also moved to the removal commit, where those tests actually disappear.

.github/workflows/rust.yml
```diff
@@ -99,6 +99,8 @@ jobs:
           toolchain: stable
           components: rustfmt
       - uses: Swatinem/rust-cache@v2
+      - name: Check zvariant with zvariant_utils/gvariant enabled without zvariant/gvariant
+        run: cargo --locked check -p zvariant --features zvariant_utils/gvariant
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This compile-only guard misses a runtime failure in this exact unified graph. `Signature::from_str("ami")` now succeeds because `zvariant_utils/gvariant` exposes `Maybe`; deserializing that D-Bus signature as `Array` reaches `Signature::Maybe::alignment_dbus()` and panics at its `unreachable!()` instead of rejecting malformed wire data. Please add a runtime regression test under this feature combination and recursively reject GVariant-only types at the D-Bus serialization/deserialization boundary, including signature-valued/body-signature fields.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of zeenix: The prior source and CI fixes are verified, and current-head Actions are green. One blocking runtime finding remains below.

Review metadata also needs syncing:
- Update the PR description: it still says all GVariant tests/CI were removed, `ostree-tests` is inert, `Value::Maybe` is documentation-only, and the PR is a draft; all are now stale.
- Add a descriptive body to `e5a64bf9` as required by CONTRIBUTING, and rename `164295f5` to drop “and corpus” because no tracked corpus is removed.
- Fold the standalone review-fix commit `64b4bcb4` into `164295f5`; the existing atomicity thread remains open for that item.

.github/workflows/rust.yml
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: The target/workflow ordering is fixed. One history issue remains: `64b4bcb4` is a new standalone review-fix commit, while CONTRIBUTING explicitly requires review feedback to be folded into the existing commits. Please absorb that fuzz-lock guard into `164295f5`, which refreshes the lockfile it validates.

.github/workflows/rust.yml
zeenix: Confirmed the panic reproduces (`Signature::try_from("ami")` parses, then D-Bus deserialize hits the `unreachable!()` in `alignment_dbus`). But it is **pre-existing**, not introduced here: the same panic fires under a plain `--features gvariant` build (a supported config on `main`), and this branch never touches that `unreachable!()` — it was authored in 2024 and is byte-identical on `origin/main`. This PR only makes it reachable in one additional build combination; it neither introduces nor worsens it relative to the existing `gvariant` feature.

A correct fix (recursively reject maybe-bearing signatures at the D-Bus boundary, returning an error) has to cover the `--features gvariant` path too, so it belongs in its own change rather than gated behind this deprecation PR. Filed as #1910 with the reproduction.

.github/workflows/rust.yml
zeenix: Done — the standalone commit is gone. The fuzz-lock CI guard is folded into the fuzz-target-removal commit (which refreshes the lockfile it validates), and while there I also dropped "and corpus" from that commit's subject since the corpus was never tracked (`fuzz/.gitignore` excludes it).

zeenix: Follow-up: an independent review caught that commits 1-3 kept the gvariant fuzz target (which calls the now-deprecated `new_gvariant`) while the fuzz CI job runs `-D warnings`, so those commits would have failed their own fuzz job before the target's removal. Fixed by adding `#![allow(deprecated)]` to that fuzz target in the deprecation commit (the file is deleted two commits later regardless), mirroring the retained test files. Verified the fuzz target now compiles clean under `-D warnings` at each of those commits; the final tree is unchanged.

.github/workflows/rust.yml
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: The underlying code is pre-existing, but its reachability in this graph is not: `zbus + zgvariant` does not compile on main, and `1673a035` newly makes that recommended coexistence graph usable. That exposes the peer-signature panic to applications which never opted into legacy `zvariant/gvariant`; the bare-`Maybe` runtime fallback added here also leaves nested cases incomplete. #1910 can remain a separate atomic fix, but it should land first and this branch be rebased, or the recursive validation plus a mixed-feature runtime test should be added here. Please also revise the PR body’s “not introduced or worsened” claim.

zeenix: You're right, and I've folded the fix in rather than deferring it. My earlier "not introduced or worsened" was too narrow: the `unreachable!()` is pre-existing under `--features gvariant`, but since `zbus + zgvariant` does not compile on main, the coexistence commit newly exposes the panic to applications that never opted into `zvariant/gvariant` — a DoS vector via a malicious peer's signature. That's a real worsening for the graph this split exists to enable.

The D-Bus ser/de boundary now rejects maybe-bearing signatures via `Signature::contains_maybe` (recursive, and applied to variant signatures read off the wire, not just the static top-level one), returning an error instead of reaching the `unreachable!()`. The mixed-feature CI guard is upgraded from `check` to `test` so the new regression test runs in exactly that configuration; the fix also covers the `--features gvariant` path. Folded into the coexistence commit and the PR body's claim is corrected. Fixes #1910.

zvariant/benches/benchmarks.rs
```diff
@@ -31,31 +31,6 @@ macro_rules! benchmark {
             })
         });
         group.finish();
-
-        // Now GVariant.
-        #[cfg(feature = "gvariant")]
-        #[allow(deprecated)]
```
zeenix: we wouldn't have had to add these lines in the deprecation commit if it came after removal of the code that uses deprecated api. Same goes for all such changes in this PR.

zvariant/src/dbus/ser.rs
```diff
@@ -123,7 +124,9 @@ where
```
zeenix, line 124: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This still leaves ordinary D-Bus `SIGNATURE` (`g`) values unvalidated. `to_bytes(Context::new_dbus(...), &Signature::try_from("mi")?)` enters `serialize_str` with `self.0.signature == Signature::Signature`, skips this `Variant`-only block, and emits invalid D-Bus data; `deserialize_str` likewise hands `"mi"` to the shared parser. That includes message body-signature fields. Please validate parsed strings for both `Signature::Signature` and `Signature::Variant` on serialization and deserialization, and test (a) a plain `Signature` value and (b) an actual `Variant` carrying an `m` signature. The new tests only reject `m` when it is the codec’s layout signature, so neither dynamic branch is exercised.

zeenix (commented):
One remaining runtime-validation blocker.

zvariant/benches/benchmarks.rs
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Agreed. The clean order is: (1) remove the benches/OSTree test and its CI/metadata, (2) remove the fuzz target with its workflow/docs/lock changes, (3) deprecate the surviving API, then (4) update migration docs/feature metadata. That lets the temporary deprecation edits in `benchmarks.rs`, the fuzz target, and `ostree_de.rs` disappear instead of being added and deleted; retained test/internal allowances stay in the deprecation commit. Please call the feature “legacy,” not “deprecated,” in commits that precede the deprecation.

zvariant/src/dbus/ser.rs
zeenix: Fixed. Both `serialize_str` and `deserialize_str` now reject a maybe for `Signature::Signature` and `Signature::Variant` layouts, so a `g` value (including a message body-signature field) or a `v` carrying an `m` signature is caught, not just the codec's layout signature. Added two regression tests: (a) a plain `Signature` value carrying `mi` (serialize and deserialize), and (b) an actual `v` on the wire whose embedded signature is `mi` — both now error with the maybe-rejection message rather than emitting invalid data or panicking. Verified they hit the dynamic branch (the error is the maybe rejection, not a byte-shortage).

zvariant/benches/benchmarks.rs
zeenix: Reordered as suggested: (1) remove benches/ostree test + data, (2) remove the fuzz target, (3) deprecate the surviving API, (4) point users to zgvariant, then coexistence/bump. Because the benches, fuzz target and ostree test are gone before the deprecation commit, that commit no longer adds throwaway `#[allow(deprecated)]` lines to them — only the retained tests and internal lib references keep their allowances. Pre-deprecation commits now call the feature "legacy" (the `ostree-tests` comment and the CI comment). Verified the final tree is byte-identical to the previous one apart from those legacy-wording lines, every commit compiles under `-D warnings` (the fuzz target builds clean at its commit with no allow), and all suites pass.

zvariant_utils/src/signature/mod.rs
```diff
@@ -266,6 +268,22 @@ impl Signature {
         }
     }
 
+    /// Whether a maybe (`m`) type appears anywhere within this signature.
+    ///
+    /// The maybe type is GVariant-specific and has no representation in the D-Bus wire format, so
+    /// a signature bound for D-Bus must not contain one. This walks the whole signature tree,
+    /// since a maybe can be nested inside any container.
+    pub fn contains_maybe(&self) -> bool {
```
zeenix, line 276: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: `zvariant` now calls this new API, but its manifest still declares `zvariant_utils = "4.0.0"`; neither published 4.0 nor 4.1 contains this method. Since that caret requirement remains satisfied, an existing lockfile can upgrade to zvariant 5.15 while retaining zvariant_utils 4.0/4.1 and then fail with E0599. Please publish/bump zvariant_utils to the version containing this API and raise zvariant’s minimum dependency to that version, with the corresponding root and fuzz lock updates.

zvariant/tests/dbus_maybe_rejection.rs
```diff
@@ -0,0 +1,102 @@
...
+        return;
+    }
+    let v = Signature::try_from("v").unwrap();
+    let ctxt = Context::new_dbus(LE, 0);
+    // A variant on the wire: signature length byte, "mi", trailing nul, then value bytes.
+    let bytes = [2u8, b'm', b'i', 0, 0, 0, 0, 0];
+    let data = Data::new(&bytes[..], ctxt);
+    let res: zvariant::Result<(Value<'_>, usize)> = data.deserialize_for_signature(&v);
+    assert!(
+        res.is_err(),
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Please assert that this is specifically the maybe-rejection error (or use a nested `ami` payload that reaches the old alignment panic). With bare `mi`, this `is_err()` can still pass through the direct-Maybe fallback even if the dynamic Variant validation regresses, so it does not currently protect the branch the test is meant to cover.

zeenix (commented):
One packaging blocker remains; the runtime implementation and rewritten commit order otherwise look correct.

zvariant/src/dbus/mod.rs
```diff
@@ -2,3 +2,40 @@ mod de;
 pub(crate) use de::*;
 mod ser;
 pub use ser::*;
+
+use crate::{Error, Result, Signature};
+
+fn maybe_not_in_dbus() -> Error {
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Please move this private helper below `reject_maybe_in_signature_str`. CONTRIBUTING’s explicit top-down layout rule requires callers to appear before the helpers they depend on.

zeenix (commented):
The byte-scan optimization is sound, but two small repository-convention fixes accompany the remaining packaging blocker.

zvariant/src/dbus/mod.rs
```diff
@@ -2,3 +2,40 @@ mod de;
...
+
+    Ok(())
+}
+
+/// Reject a maybe type in a signature carried as a `g` or `v` value's string form.
+///
+/// This is the hot path (every variant on the wire), so it avoids re-parsing the string into a
+/// `Signature`: `m` is the maybe type constructor and appears nowhere else in the signature
+/// grammar, so its presence is an exact test for a maybe type.
+pub(crate) fn reject_maybe_in_signature_str(bytes: &[u8]) -> Result<()> {
```
zeenix, line 27: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Please also update `c5906e25`’s body: it currently says the boundary checks layout and dynamic `g`/`v` signatures “via `Signature::contains_maybe`,” but only parsed layout signatures use that method now; carried signature strings use this exact `m` byte scan.

zvariant_utils/src/signature/mod.rs
zeenix: Fixed. zvariant_utils is bumped to 4.2.0 (it gains `contains_maybe`), and zvariant's dependency floor is raised from `4.0.0` to `4.2.0` so a resolver can't pair zvariant 5.15 with an older zvariant_utils lacking the method. Root and fuzz lockfiles are regenerated to 4.2.0. Both live in the version-bump commit (now `🔖 zv,zu: Bump versions`).

zvariant/tests/dbus_maybe_rejection.rs
zeenix: Strengthened. The tests now assert the error is specifically the boundary maybe-rejection (its message "not valid in the D-Bus format" is distinct from the direct-`Maybe` deserialize fallback), so a regression in the dynamic branch can't pass silently. The variant test also now uses a nested `ami` payload, which would reach the alignment `unreachable!()` if the dynamic check regressed rather than being absorbed by the top-level fallback.

zvariant/src/dbus/mod.rs
zeenix: Moved `maybe_not_in_dbus` below its callers (`reject_maybe` and `reject_maybe_in_signature_str`), per the top-down layout rule.

zvariant/src/dbus/mod.rs
zeenix: Reworded the commit body: it now says the layout signature is checked via `Signature::contains_maybe`, while a signature carried as a `g`/`v` value is checked by scanning its string form for `m` (exact, no re-parse on the hot path).

zvariant_utils/src/serialized.rs
```diff
@@ -6,6 +6,9 @@ pub enum Format {
     #[default]
     DBus,
     /// [GVariant](https://developer.gnome.org/glib/stable/glib-GVariant.html) format.
+    ///
+    /// GVariant support in `zvariant` is deprecated and will be removed in zvariant 6.0; use
```
zeenix, line 10: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Since `05527fec` changes packaged zvariant_utils files, release-plz will include its subject in the 4.2 changelog after stripping the package prefix. That currently yields “Deprecate GVariant support,” incorrectly implying that the shared support used by zgvariant is deprecated. Please scope the subject too, e.g. `🗑️ zv,zu: Deprecate GVariant support in zvariant`.

zeenix (commented):
The previous four requests are fixed. Three final cleanup items remain.

zvariant/tests/dbus_maybe_rejection.rs
```diff
@@ -0,0 +1,106 @@
...
+use zvariant::{
+    LE, Signature, Value,
+    serialized::{Context, Data},
+    to_bytes_for_signature,
+};
+
+// The maybe type is unavailable in a plain zvariant build (no own `gvariant` feature and no
+// `zvariant_utils/gvariant` in the graph): `m` does not parse, so the hazard cannot arise and
+// there is nothing to assert.
+fn maybe_supported() -> bool {
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Please put the test entry points before these private helpers and `MAYBE_SIGNATURES`, moving the helpers/constant below their callers. The top-down item-order rule in CONTRIBUTING applies here too.

zvariant/tests/dbus_maybe_rejection.rs
```diff
@@ -0,0 +1,106 @@
...
+// caught by the top-level direct-`Maybe` fallback.
+#[test]
+fn dbus_deserialize_rejects_variant_carrying_maybe() {
+    if !maybe_supported() {
+        return;
+    }
+    let v = Signature::try_from("v").unwrap();
+    let ctxt = Context::new_dbus(LE, 0);
+    // A variant on the wire: signature length byte, "ami", trailing nul, then value bytes.
+    let bytes = [3u8, b'a', b'm', b'i', 0, 0, 0, 0, 0];
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: This needs three more zero bytes to reach the claimed `alignment_dbus()` panic without the guard. The 5-byte variant header is followed by 3 bytes of array padding, leaving only 1 byte here; `ArrayDeserializer` therefore fails while reading the 4-byte array length first. A 12-byte buffer supplies that length and then reaches the child-alignment calculation. The exact-error assertion already prevents a false pass, but the current reproduction/comment is inaccurate.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of zeenix: One PR-description cleanup too: the version bullet should mention the zvariant_utils 4.2.0 bump and zvariant’s new 4.2 floor. Also remove or qualify “No behavior change for any existing feature configuration”; the folded #1910 fix intentionally changes malformed D-Bus input under the existing `zvariant/gvariant` configuration from a panic to an error.

zvariant_utils/src/serialized.rs
zeenix: Good catch — that commit touches packaged zvariant_utils files, so it lands in the 4.2 changelog. Scoped its subject to `🗑️ zv,zu: Deprecate GVariant support in zvariant` so the stripped changelog line no longer implies the shared zvariant_utils support is deprecated.

zvariant/tests/dbus_maybe_rejection.rs
zeenix: Reordered: the four `#[test]` entry points now come first, with `maybe_supported`, `assert_maybe_rejected` and `MAYBE_SIGNATURES` below them.

zvariant/tests/dbus_maybe_rejection.rs
zeenix: Corrected to a 12-byte buffer, and verified empirically: with the guards removed, the old 9-byte buffer fails early at the array-length read (`invalid length 4, expected >= 7`), while 12 bytes reaches the child-alignment `unreachable!()`. The comment now describes the layout accurately (5-byte variant header, 3 bytes array padding, 4-byte array length, then the maybe child's alignment).

zvariant/src/dbus/mod.rs
zeenix: Also updated the PR description: the version bullet now notes the zvariant_utils 4.2.0 bump and zvariant's new 4.2 floor, and the "no behavior change" line is qualified — under `zvariant/gvariant`, malformed D-Bus input carrying a maybe now errors instead of panicking (the #1910 fix).

PR #1912, opened 2026-08-18
Validate names in `Value`/`OwnedValue` conversions

robin-simular:
### Bug

The name types (`UniqueName`, `InterfaceName`, etc.) validate their input everywhere except in the `Value`/`OwnedValue` conversions. Those were derive-generated and just wrapped the string without checking it. So an invalid name coming off the wire became a typed name, and failed much later in confusing ways (e.g. a method call with an empty destination).

### Fix

- **zbus_names**: Replace the derives with hand-written impls that convert through `TryFrom<Str>`, which validates. Signatures and error type stay the same, so no API changes.
- **zvariant**: Fix a latent bug this exposed in `Optional<T>`. Its `TryFrom<Value>` impl converted to `T` first and checked for the null sentinel (e.g. `""` for "no name") after. That only worked because invalid names used to slip through. Now it checks the sentinel first, like the serde path already does. So `Optional<UniqueName>` from `""` still gives `Ok(None)`, and invalid non-empty names give an error.

### ⚠️ Breaking change (zvariant)

The bounds on the `Optional<T>` conversion impls change from `T: PartialEq<NoneType>` to `NoneType: Into<Value>`. All common none-types (`Self`, `&str`, `String`, numbers) satisfy the new bound.

### DIFF
--- a/zvariant/src/from_value.rs
+++ b/zvariant/src/from_value.rs
@@ -199,19 +199,22 @@ where
 
 impl<'a, T> TryFrom<Value<'a>> for Optional<T>
 where
-    T: TryFrom<Value<'a>> + NoneValue + PartialEq<<T as NoneValue>::NoneType>,
+    T: TryFrom<Value<'a>> + NoneValue,
+    <T as NoneValue>::NoneType: Into<Value<'a>>,
     T::Error: Into<crate::Error>,
 {
     type Error = crate::Error;
 
     fn try_from(value: Value<'a>) -> Result<Self, Self::Error> {
-        T::try_from(value).map_err(Into::into).map(|value| {
-            if value == T::null_value() {
-                Optional::from(None)
-            } else {
-                Optional::from(Some(value))
-            }
-        })
+        // Check for the null sentinel before converting, since `T`'s conversion may
+        // legitimately reject it (e.g. name types validating the empty string).
+        if value == T::null_value().into() {
+            Ok(Optional::from(None))
+        } else {
+            T::try_from(value)
+                .map(|value| Optional::from(Some(value)))
+                .map_err(Into::into)
+        }
     }
 }
 
--- a/zvariant/src/into_value.rs
+++ b/zvariant/src/into_value.rs
@@ -89,6 +89,17 @@ into_value!(Fd<'a>, Fd);
 #[cfg(unix)]
 try_into_value_from_ref!(Fd<'a>, Fd);
 
+#[cfg(feature = "enumflags2")]
+impl<'a, F> From<enumflags2::BitFlags<F>> for Value<'a>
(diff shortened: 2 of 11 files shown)


zvariant/src/from_value.rs
```diff
@@ -199,19 +199,22 @@ where
 
 impl<'a, T> TryFrom<Value<'a>> for Optional<T>
 where
-    T: TryFrom<Value<'a>> + NoneValue + PartialEq<<T as NoneValue>::NoneType>,
+    T: TryFrom<Value<'a>> + NoneValue,
+    <T as NoneValue>::NoneType: Into<Value<'a>>,
```
zeenix, line 203: 🤖 ChatGPT 5.6 on behalf of Zeenix: This drops an existing supported case under the `enumflags2` feature. `BitFlags<F>` implements `Default` and `PartialEq`, and zvariant provides `TryFrom<Value>` for it, so `Optional<BitFlags<F>>::try_from(value)` satisfies the old bounds. There is no `Into<Value>` impl for `BitFlags<F>`, so it no longer satisfies this one. Please either preserve that conversion (with a regression test for both `Value` and `OwnedValue`) or avoid requiring the sentinel to implement `Into<Value>`.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of Zeenix: Please also make the history atomic. `5ef4b364` makes `Optional<UniqueName>::try_from(Value::from(""))` regress from `Ok(None)` to an error, and only `17dcdb23` restores it. Either put the `zvariant` sentinel-ordering change first and move the name-specific regression tests into the subsequent `zbus_names` commit, or squash the two commits.

zvariant/src/from_value.rs
robin-simular: I've implemented the Trait for it

zvariant/src/into_value.rs
```diff
@@ -89,6 +89,17 @@ into_value!(Fd<'a>, Fd);
...
+impl<'a, F> From<enumflags2::BitFlags<F>> for Value<'a>
+where
+    F: enumflags2::BitFlag,
+    F::Numeric: Into<Value<'a>>,
+{
+    fn from(v: enumflags2::BitFlags<F>) -> Self {
+        v.bits().into()
+    }
+}
+
```
zeenix, line 102: please split into a separate commit. This is useful regardless of the other changes here.

zeenix (changes requested):
LGTM otherwise.

zvariant/src/into_value.rs
robin-simular: Done

zeenix (approved):

Issue #1819, opened 2026-06-18
DeserializeDict with nested a{sv} value inside outer a{sv}


dkspindola:
I’m trying to model a D-Bus method that returns an `a{sv}` where one field is itself a variant containing another `a{sv}`.

The D-Bus shape is:

```text
info: a{sv}
info.manifest-hash: variant s
info.update: variant a{sv}
```

I expected this to work:

```rust
use zbus::zvariant::{DeserializeDict, OwnedValue, SerializeDict, Type};

#[derive(SerializeDict, DeserializeDict, Type, PartialEq, Debug, Default)]
#[zvariant(signature = "a{sv}", rename_all = "kebab-case")]
pub struct InspectBundleInfo {
    pub manifest_hash: Option<String>,
    pub update: Option<InspectBundleUpdateInfo>,
}

#[derive(SerializeDict, DeserializeDict, Type, PartialEq, Debug, Default)]
#[zvariant(signature = "a{sv}", rename_all = "kebab-case")]
pub struct InspectBundleUpdateInfo {
    pub compatible: Option<String>,
    pub version: Option<String>,
    pub description: Option<String>,
    pub build: Option<String>,
}
```

But deserialization fails with:

```text
Error: Variant(Message("invalid value: string \"v\", expected the value signature"))
```

If I change the nested field to `OwnedValue`, it works:

```rust
pub update: Option<OwnedValue>
```

The decoded value is a dictionary with signature `a{sv}`, as expected.

Is `DeserializeDict` expected to support this nested shape?

```text
outer a{sv}
  -> field value: variant a{sv}
  -> nested DeserializeDict struct
```

zeenix:
> Is `DeserializeDict` expected to support this nested shape?

Hmm.. this might be a regression from when we added [nested dictionary support](https://z-galaxy.github.io/zbus/faq.html#nested-dictionaries), so you can do `a{sa{sv}}`.

zeenix:
Are you sure that you didn't get the cases wrong? The code you said doesn't work, actually works but the code you said works, doesn't. The error is the same though.

See #1821.

nomo-nvm:
I think OP tried to connect to rauc and failed. I'm failing too. So here's my issue. I'm trying to parse rauc response to [inspectbundle](https://rauc.readthedocs.io/en/latest/reference.html#inspectbundle-method) method call. I use zbus 5.19 and zvariant 5.15.

[...]

PR #1914, opened 2026-08-19
💬 zv: Name the expected signature in as_value errors

zeenix:
When a peer sends a variant whose contained signature doesn't match the field type, the error only said "expected the value signature" — never saying *which* signature we were after, leaving users to guess.

Both reports in #1819 ran into this. RAUC wraps its nested `a{sv}` dicts in a variant twice ([`manifest.c#L1162`](https://github.com/rauc/rauc/blob/3991031a43f244669889e81f5db18ce56e890b68/src/manifest.c#L1162) uses `g_variant_dict_insert(..., "v", ...)`, so the `{sv}` entry wraps an already-wrapped value), and all the resulting error said was:

```
invalid value: string "v", expected the value signature
```

Now:

```
invalid value: string "v", expected a value of signature `a{sv}`
```

which points straight at the mismatch.

No behavioural change beyond the message; the derives already handled the documented shape correctly (verified while researching #1819 — a singly-wrapped `v` → `a{sv}` decodes into a typed `DeserializeDict` field just fine).

Regression test added as `zvariant/tests/issue/issue_1819.rs`, feeding a doubly-wrapped `v` → `v` → `a{sv}` payload and asserting the message names both the signature found and the one expected.

https://claude.ai/code/session_01S3GPtmz7fTcdueTquhZp7B

---
_Generated by [Claude Code](https://claude.ai/code/session_01S3GPtmz7fTcdueTquhZp7B)_

### DIFF
--- a/zvariant/src/as_value/deserialize.rs
+++ b/zvariant/src/as_value/deserialize.rs
@@ -66,9 +66,10 @@ impl<'de, T: Type + serde::Deserialize<'de>> Visitor<'de> for DeserializeValueVi
             .next_element()?
             .ok_or_else(|| serde::de::Error::invalid_length(0, &self))?;
         if T::SIGNATURE != &sig {
+            let expected = format!("a value of signature `{}`", T::SIGNATURE);
             return Err(serde::de::Error::invalid_value(
                 serde::de::Unexpected::Str(&sig.to_string()),
-                &"the value signature",
+                &expected.as_str(),
             ));
         }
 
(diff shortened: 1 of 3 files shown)


zeenix (commented):
🤖 ChatGPT 5.6 on behalf of Zeenix: Reviewed the error construction and the #1819 regression case. The test correctly produces `v → v → a{sv}`, the new diagnostic reports both the actual and expected signatures without changing deserialization behavior, the commit is atomic, and all CI checks are green. No issues found.

PR #1917, opened 2026-08-20
🐛 zb: Prune empty ancestor nodes on interface removal

### LINKED ISSUE #1916: `ObjectServer::remove` only unlinks the leaf node and leaves empty ancestors behind
`ObjectServer::remove`/`remove_named` unlinks the node at the given path but never checks whether the now-childless ancestors should go too. Those stay in the tree for the lifetime of the connection: they show up in `Introspect` and are returned by `GetManagedObjects` with empty interface dicts.

failure:
```log
running 6 tests
test agent::builder::tests::agent_manager_errors_map_to_public_lifecycle_failures ... ok
test agent::builder::tests::builder_methods_replace_every_registration_option ... ok
test agent::builder::tests::defaults_match_networkmanager_secret_agent_contract ... ok
test agent::builder::tests::unregister_leaves_no_objects_exported_on_the_agent_connection ... FAILED
test agent::builder::tests::unregister_releases_a_custom_object_path ... ok
test agent::builder::tests::unregister_releases_the_agent_object_path ... ok

failures:

---- agent::builder::tests::unregister_leaves_no_objects_exported_on_the_agent_connection stdout ----

thread 'agent::builder::tests::unregister_leaves_no_objects_exported_on_the_agent_connection' (329571) panicked at nmrs/src/agent/builder.rs:544:9:
assertion `left == right` failed
  left: ["<node name=\"org\">", "<node name=\"freedesktop\">", "<node name=\"NetworkManager\">"]
 right: []
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    agent::builder::tests::unregister_leaves_no_objects_exported_on_the_agent_connection

test result: FAILED. 5 passed; 1 failed; 0 ignored; 0 measured; 633 filtered out; finished in 0.01s

error: test failed, to rerun pass `-p nmrs --lib`

```
Bumping zbus didn't help me here.

**v5.16**
https://github.com/freedesktop-rs/nmrs/actions/runs/32380987548/job/96463862431?pr=530

**after i bumped to 5.19**
https://github.com/freedesktop-rs/nmrs/actions/runs/32381480661/job/96465492887?pr=530

I presume this is a problem child:

https://github.com/z-galaxy/zbus/blob/e8bd73293307c1d7bdceb566e518572427d21d5c/zbus/src/object_server/mod.rs#L234-L248

zeenix:
Fixes #1916.

`ObjectServer::at` auto-creates ancestor nodes as needed, but `ObjectServer::remove` only unlinked
the leaf node, so the now-empty ancestors stayed in the tree forever, showing up in `Introspect`
and `GetManagedObjects`. The first commit prunes ancestors that are left childless with only the
default interfaces. The follow-up commits fix adjacent problems in the same code that the reviews
of the first one surfaced:

* `remove` used to destroy a node unconditionally once its last user interface was gone, taking
  down live descendant objects and any `ObjectManager` registered at the same path. Both now keep
  the object alive (and `remove` returns `Ok(false)` for them — a behavior change worth noting;
  the old bool also changed from `true` after destroying the whole subtree).
* A failed `at()` (e.g. name collision with a default interface) used to leave its freshly
  auto-created nodes behind with no way to remove them.
* Removing the last interface at the root path `/` panicked (empty parent-path computation); the
  root object is now simply never destroyed.

Notes:

* Everything about the removal is deliberately non-recursive — the tree walk and the disposal of
  the detached node chain alike — since per-component recursion overflows the stack on very deep
  (yet valid) object paths. A regression test runs a deep removal on a deliberately small stack.
* Pruning happens before the fallible `InterfacesRemoved` emission so an emission failure can't
  leak empty nodes; the flip side is that such a failure returns `Err` for a removal that did
  take effect (emission failure implies the connection is dying, so this seemed the lesser evil).
* Known pre-existing gap, left for a follow-up: the error paths of `at()` after the interface
  insert (signal emission / property-read failures) still leave the interface registered and the
  fresh nodes in place.

Generated by Claude Fable 5.

https://claude.ai/code/session_01MuEgaUsWgw2usk5PxNCzxS

### DIFF
--- a/zbus/src/object_server/node.rs
+++ b/zbus/src/object_server/node.rs
@@ -106,8 +106,62 @@ impl Node {
         })
     }
 
-    pub(super) fn remove_node(&mut self, node: &str) -> bool {
-        self.children.remove(node).is_some()
+    /// Remove the descendant node at `path`.
+    ///
+    /// Ancestors that are thereby left childless with only the default interfaces are removed as
+    /// well. Returns whether the node at `path` was removed. The root node itself is never
+    /// removed.
+    pub(super) fn remove_node(&mut self, path: &ObjectPath<'_>) -> bool {
+        let parts = path
+            .split('/')
+            .filter(|p| !p.is_empty())
+            .collect::<Vec<_>>();
+        if parts.is_empty() {
+            return false;
+        }
+
+        // First pass: check that the whole path exists and find the deepest ancestor that has to
+        // stay: the last one along the path with other children or non-default interfaces of its
+        // own. Everything below it only exists to lead to the target node.
+        let mut node = &*self;
+        let mut keep_depth = 0;
+        for (depth, part) in parts.iter().enumerate() {
+            let Some(child) = node.children.get(*part) else {
+                return false;
+            };
+            if depth + 1 < parts.len()
+                && (child.children.len() > 1 || !child.has_default_interfaces_only())
+            {
+                keep_depth = depth + 1;
+            }
+            node = child;
+        }
+
+        // Second pass: unlink the target node and the now-useless part of its ancestor chain in
+        // one go, by cutting the tree right below the deepest surviving ancestor.
+        let mut node = &mut *self;
+        for part in &parts[..keep_depth] {
+            // The first pass established that the whole path exists.
+            node = node.children.get_mut(*part).unwrap();
+        }
+        let mut disposal = Vec::from_iter(node.children.remove(parts[keep_depth]));
+        let removed = !disposal.is_empty();
+        // Dispose of the detached subtree iteratively: dropping it in one go would recurse per
+        // level, as each node owns its children.
+        while let Some(mut node) = disposal.pop() {
+            disposal.extend(node.children.drain().map(|(_, child)| child));
+        }
+        removed
+    }
+
+    /// Whether the node only has the default interfaces that every node gets on creation.
+    ///
+    /// Note that unlike [`Node::is_empty`], this considers `ObjectManager` a non-default
+    /// interface: a node explicitly serving one must survive the pruning of its children.
+    fn has_default_interfaces_only(&self) -> bool {
+        self.interfaces
+            .keys()
+            .all(|k| *k == Peer::name() || *k == Introspectable::name() || *k == Properties::name())
     }
 
     pub(super) fn add_arc_interface(
--- a/zbus/src/object_server/node.rs
+++ b/zbus/src/object_server/node.rs
@@ -97,20 +97,11 @@ impl Node {
         self.interfaces.remove(interface_name).is_some()
     }
 
-    pub(super) fn is_empty(&self) -> bool {
-        !self.interfaces.keys().any(|k| {
-            *k != Peer::name()
-                && *k != Introspectable::name()
-                && *k != Properties::name()
-                && *k != ObjectManager::name()
-        })
-    }
-
-    /// Remove the descendant node at `path`.
+    /// Remove the node at `path` if it's no longer needed, along with any ancestors that are
+    /// thereby no longer needed either.
     ///
-    /// Ancestors that are thereby left childless with only the default interfaces are removed as
-    /// well. Returns whether the node at `path` was removed. The root node itself is never
-    /// removed.
+    /// A node is still needed if it has children or non-default interfaces (or is the root
+    /// node). Returns whether the node at `path` was removed.
     pub(super) fn remove_node(&mut self, path: &ObjectPath<'_>) -> bool {
         let parts = path
             .split('/')
@@ -136,6 +127,10 @@ impl Node {
             }
             node = child;
         }
+        if !node.children.is_empty() || !node.has_default_interfaces_only() {
+            // The target node is still needed.
+            return false;
+        }
 
         // Second pass: unlink the target node and the now-useless part of its ancestor chain in
         // one go, by cutting the tree right below the deepest surviving ancestor.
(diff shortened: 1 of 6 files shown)


zbus/src/object_server/node.rs
```diff
@@ -97,17 +98,57 @@ impl Node {
...
+        }
+
+        // Second pass: unlink the target node and the now-useless part of its ancestor chain in
+        // one go, by cutting the tree right below the deepest surviving ancestor.
+        let mut node = &mut *self;
+        for part in &parts[..keep_depth] {
+            // The first pass established that the whole path exists.
+            node = node.children.get_mut(*part).unwrap();
+        }
+        node.children.remove(parts[keep_depth]).is_some()
```
zeenix: 🤖 ChatGPT 5.6 on behalf of Zeeshan: Although the two traversals are iterative, this removal still *drops* the detached ancestor chain recursively: each removed `Node` owns the next one through its `HashMap<String, Node>`. The ~10k-component case mentioned in the PR can therefore still overflow the stack during destruction. Please dismantle the detached chain iteratively (draining/taking its children before each `Node` is dropped), and ideally cover this with a deep-path test running on a deliberately small stack.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of Zeeshan: The Clippy fix is correct and CI is fully green, but these two issues should be addressed before merging.

zbus/src/object_server/node.rs
```diff
@@ -220,19 +261,20 @@ impl Node {
...
+                // The default interfaces and `ObjectManager` itself are not managed.
+                .filter(|n| !is_default_interface(n) && **n != ObjectManager::name())
+            {
                 let props = node
                     .get_properties(object_server, connection, iface_name.clone())
                     .await?;
                 interfaces.insert(iface_name.clone().into(), props);
             }
-            managed_objects.insert(node.path.clone(), interfaces);
+            if !interfaces.is_empty() {
```
zeenix: 🤖 ChatGPT 5.6 on behalf of Zeeshan: This now omits live intermediate nodes from `GetManagedObjects`, even though they remain exported objects: `Node::new` installs `Introspectable` and `Properties`, and method dispatch accepts calls at those paths. The [ObjectManager specification](https://dbus.freedesktop.org/doc/dbus-specification.html#standard-interfaces-objectmanager) says applications SHOULD NOT export child objects that are absent from `GetManagedObjects`. Previously zbus filtered the standard interface *names* but still included every live path; pruning alone fixes #1916's stale post-removal entries. Please either retain the unconditional insertion here or redesign intermediate routing nodes so they do not export interfaces; otherwise this is both a compatibility change and a spec regression.

zbus/src/object_server/node.rs
zeenix: Confirmed — reproduced the overflow with a 2000-component path on a 128 KiB stack: the detached chain's drop glue recursed per node. The chain is now dismantled iteratively in `remove_node` before it's dropped, and the new `issue_1916_deep_path` test runs registration + removal of a 2000-component path on a 128 KiB thread (it aborted with a stack overflow before the fix). Squashed into the pruning commit.

Note that dropping a legitimately deep tree wholesale (e.g. the object server itself at connection teardown) still uses the default recursive drop glue — that's pre-existing and seems best left for a follow-up if deemed worth handling.

Generated by Claude Fable 5.

zbus/src/object_server/node.rs
zeenix: Agreed on the spec point: the intermediate nodes do remain exported (they answer `Introspectable`/`Properties`/`Peer`), so omitting them from `GetManagedObjects` while still exporting them violates the SHOULD NOT. The filtering commit is dropped from the branch entirely; `GetManagedObjects` reports every live path as before, and pruning alone covers the stale post-removal entries from #1916.

Generated by Claude Fable 5.

zbus/src/object_server/node.rs
zeenix: Follow-up: CI caught that the integration form of the deep-path test can't run under the tokio feature (`zbus::block_on` on a bare thread panics with "there is no reactor running"). The test is now a `Node` unit test (`deep_path_removal_needs_no_deep_stack`) driving `get_child_mut`/`remove_node` directly on a 128 KiB thread — no runtime involved, so it runs under every feature configuration, and it still aborts with a stack overflow when the iterative disposal is reverted.

Generated by Claude Fable 5.

zbus/tests/issue/issue_1916.rs
```diff
@@ -0,0 +1,216 @@
...
+    // leaving the auto-created (and now empty & childless) ancestor nodes behind in the tree.
+    zbus::block_on(issue_1916_async());
+}
+
+#[test]
+#[timeout(15000)]
+fn issue_1916_deep_path() {
+    // Neither the removal walk nor the disposal of the removed nodes may recurse per path
+    // component: with this many components on a deliberately small stack, a recursive
+    // implementation aborts with a stack overflow.
```
zeenix, line 33: 🤖 ChatGPT 5.6 on behalf of Zeeshan:

Update: this was already fixed by the concurrently pushed `6c98f058`. The deep-path coverage now lives in `node.rs` and exercises `get_child_mut`/`remove_node` directly on the 128 KiB-stack thread, with no async runtime or session-bus dependency. No action remains here.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of Zeeshan:

Follow-up review: both earlier findings are fixed. One CI-blocking issue remains in the new deep-path regression test.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of Zeeshan:

Rechecked the latest head. Both earlier findings are fixed, the deep-path regression test now directly exercises `Node` on a small stack, and I found no remaining code issues. Windows and all other completed checks are green; `linux_test` is still running.

cachebag (approved):
This all looks about right to me

PR #1918, opened 2026-08-21
📝 zm: Document the signals trait generated by `#[interface]`

olkva:
`#[interface]` turns every `#[zbus(signal)]` declaration into a method of the generated
`<Type>Signals` trait, but only the signature is taken over — the doc comments of the
declaration are dropped. Crates that opt into `#![warn(missing_docs)]` therefore still get

    warning: missing documentation for a method
      --> src/lib.rs:7:1
       |
     7 | #[zbus::interface(name = "org.example.Iface")]
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

for every interface that declares a signal, even when the signal itself is documented.
#1787 fixed this for the `impl Interface` block, so the signals trait is the remaining case of https://github.com/z-galaxy/zbus/issues/1542.

The doc comments are now carried over to the trait method, and signals without any get a
generated one-liner (`Emit the “ResourceChanged” signal.`), so the lint stays quiet either
way.

### DIFF
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -439,6 +439,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
             reply,
             member_name,
             cfg_attrs,
+            doc_attrs,
             ..
         } = method_info;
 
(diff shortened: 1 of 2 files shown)


zeenix (approved):
LGTM, thanks. I'll also have the machine take a look.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of zeenix: I found no additional issues in the implementation or regression test. The generated signals trait now receives either the original documentation or a fallback, covering both documented and undocumented signals. This looks ready once the branch is rebased and CI passes.

zeenix (approved):

PR #1920, opened 2026-08-22
🔧 Show changelog emojis on section headings only

zeenix:
Every changelog entry repeated the emoji from its commit message, directly under a section heading that already said the same thing in words:

```
### Added
- ✨ Add Connection::closed() and is_closed() API.
- ✨ Make the tokio and async-io features additive.
```

This moves the emoji up to the heading, where it says something once:

```
### ✨ Added
- Add Connection::closed() and is_closed() API.
- Make the tokio and async-io features additive.
```

## What changed

Only `release-plz.toml`:

- **Group names carry the emoji** — `group = "Added"` became `group = "<!-- 00 -->✨ Added"`, rendered through `striptags | trim | upper_first`. The `<!-- NN -->` prefix pins section order, reproducing the order that previously fell out of the alphabetical sort of the bare group names.
- **Entries drop their emoji prefix** — one `replace_regex` both recognises and removes it, so the two cannot disagree about where the prefix ends. It strips **exactly one** gimoji plus the whitespace after it, matching the commit contract: a subject carries one prefix emoji, and everything after it — another emoji included — is title text. The alternation lists exactly the emoji this file routes to a group; any other emoji can only reach the catch-all, where nothing is stripped.
- **The prefix must be followed by whitespace.** Subjects without a separator stay visibly malformed rather than being mangled — the 21 here (`♻️l zb: …`, `♻️zm: …`) render unchanged rather than becoming `l zb: …` and `zm: …`.
- **The catch-all group keeps its emoji.** `⚙️ Other` stands for no category, so an entry emoji there is information rather than repetition — 201 of this repo's Other entries open with a distinct marker (`💬`, `💣`, `👽️`, `⏪️`, `🏷️`, `💸`, `⚰️`, `✏️` and ~30 more). Those entries render byte-identical to the pre-change output.
- **`upper_first` is guarded** — it applies only when nothing was stripped, and still sees the subject untrimmed. The emoji used to shield the rest of the subject from that filter; unguarded it rewrites `zbus_macros shouldn't set features on zbus` to `Zbus_macros shouldn't …` and `signature::Fields::len() now const` to `Signature::Fields::len() …`.
- **The dependency heading is 📦, not ⬆️.** That group also collects `➖`/`⬇️` — 25 entries here — which an upward arrow contradicts. Dependencies is the only group mixing opposite directions.
- **👷 now routes to CI instead of Added.** `^👷` was listed in both the Features and the CI parser block; first match wins, so CI work was filed as a feature while the CI section only ever collected 💚. 44 entries change section as a result.

Commit preprocessors and the version-bump regexes are untouched (👷 appears in neither `custom_minor_increment_regex` nor `custom_major_increment_regex`). The emoji still sits at the start of the commit message, so parsers and bump rules match exactly as before — the strip happens at render time only.

## Verification

`git-cliff` 2.13.1 was run over the complete history of all eight repos sharing this template, with the pre-change and post-change configs, and every rendered entry compared as a multiset. In every repo: each entry in an emoji-headed group is its pre-change text minus exactly one prefix emoji, `Other` is byte-identical to pre-change, and the entry count is unchanged.

| repo | commits | entries |
|---|---|---|
| rust-analyzer-mcp | 132 | 111 |
| pixel8 | 502 | 408 |
| zbus | 6364 | 4987 |
| gimoji | 562 | 206 |
| busd | 637 | 296 |
| zlink | 1104 | 910 |
| zgvariant | 37 | 26 |
| zcheapstr | 21 | 10 |

Across all ~9,000 commits, exactly one subject carries two emoji: `1108af61`, `🩹 🚸 zv: \`Signature\` equality regardless of outer parentheses`. Per the contract its `🚸` is title text, so it renders as `- 🚸 zv: \`Signature\` equality …`.

## Note

release-plz only prepends to `CHANGELOG.md`, so existing entries keep the old per-entry emoji. The new style starts with the next release.

https://claude.ai/code/session_016iaTmaTUYJK5yuq86tGSuJ

### DIFF
--- a/release-plz.toml
+++ b/release-plz.toml
@@ -25,16 +25,40 @@ pr_name = "🔖 Releases"
 [changelog]
 
 # Changelog body template.
+# Each section heading carries the emoji of its category, so the leading emoji is stripped off
+# the entries rather than repeated on every line. One regex both recognises and removes it, so
+# the two can't disagree. It strips exactly one gimoji plus the whitespace after it, matching the
+# commit contract: a subject carries one prefix emoji, and everything after it — another emoji
+# included — is title text. So "✨ 🎉 Add foo" keeps its 🎉, and subjects with no separator after
+# the prefix ("⚰️a Drop …", "✨🏳️‍🌈 Rainbow") stay visibly malformed rather than being mangled.
+# The alternation is exactly the set of emoji this file routes to a group below; any other emoji
+# lands in the catch-all, where nothing is stripped. Keep the two in step when adding a parser.
+# The catch-all group is the exception: its heading emoji stands for no particular category, so
+# the entry emoji there is information rather than repetition and is kept. Those entries render
+# exactly as they did before this template existed.
+# `upper_first` applies only when nothing was stripped, and sees the subject untrimmed, exactly
+# as it did before. An emoji used to shield the rest of the subject from it; without that guard
+# it would rewrite "zbus_macros shouldn't …" into "Zbus_macros shouldn't …" and
+# "signature::Fields::len()" into "Signature::Fields::len()".
+# Group names are prefixed with an "<!-- NN -->" ordering comment, which `striptags` drops from
+# the rendered heading.
 # Issue refs are separated by "||" delimiter so we can insert period before them.
 body = """
 
 ## {{ version | trim_start_matches(pat="v") }} - {{ timestamp | date(format="%Y-%m-%d") }}
 {% for group, commits in commits | group_by(attribute="group") %}
-### {{ group | upper_first }}
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set subject = parts | first %}\
+{% if group is ending_with("Other") %}\
+{% set title = subject | upper_first %}\
+{% else %}\
+{% set title = subject | replace_regex(from="^\\s*(?:✨|🎉|🚸|🐛|🚑|🩹|🥅|⚡|📝|💡|♻|🎨|🚚|🔧|🏗|💄|💥|🔥|🗑|⬆|⬇|📌|➕|➖|🔒|✅|🧪|👷|💚)\\u{FE0F}?\\s+", to="") %}\
+{% if title == subject %}{% set title = title | upper_first %}{% endif %}\
+{% endif %}\
+- {{ title | trim | trim_end_matches(pat=".") }}\
 {% if commit.breaking %} (**BREAKING**){% endif %}.\
 {% if parts | length > 1 %} {{ parts | last }}{% endif %}
 {% endfor %}\
@@ -60,6 +84,8 @@ commit_preprocessors = [
 
 # Categorize commits based on gimoji emojis.
 # See: https://zeenix.github.io/gimoji/
+# The "<!-- NN -->" prefix on each group name fixes the order of the changelog sections
+# (the template strips it); the emoji right after it becomes the section heading emoji.
 commit_parsers = [
     # Skip commits carrying a "Changelog: skip" trailer (see the preprocessor above).
     { message = "^changelog-skip$", skip = true },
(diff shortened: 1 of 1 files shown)


release-plz.toml
```diff
@@ -25,16 +25,23 @@ pr_name = "🔖 Releases"
...
 ## {{ version | trim_start_matches(pat="v") }} - {{ timestamp | date(format="%Y-%m-%d") }}
 {% for group, commits in commits | group_by(attribute="group") %}
-### {{ group | upper_first }}
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set title = parts | first | trim %}\
+{% if title is matching("^[^\\p{L}\\p{N}\\s]+\\s") %}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: This is not an emoji-specific test, and its separator disagrees with the next line: `\s` accepts a tab while `split(pat=" ")` does not, which can drop title words or produce an empty entry. It also strips punctuation-only tokens and misses accepted no-space history: `bc08ed16` (`♻️l zb: Refactor…`) still renders `♻️l…` under the `♻️ Changed` heading. Please use one rendering-time, emoji-specific `replace_regex` and enforce or consistently support the separator.

release-plz.toml
```diff
@@ -25,16 +25,23 @@ pr_name = "🔖 Releases"
...
+{% set title = title | split(pat=" ") | slice(start=1) | join(sep=" ") | trim %}\
+{% endif %}\
+- {{ title | upper_first | trim_end_matches(pat=".") }}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: Applying `upper_first` after stripping corrupts real Rust/crate identifiers. `28485173` turns `zbus_macros shouldn't…` into `Zbus_macros shouldn't…`, and `f53a2113` turns `signature::Fields::len()` into `Signature::Fields::len()`. Please preserve title casing after the emoji is removed, or reserve `upper_first` for originally unprefixed subjects.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: I independently rendered the full history (4,987 entries before and after). Exactly 44 `👷` entries moved Added → CI and no other section moved, so the reroute is correct. The inline title transformations are nevertheless blocking. The catch-all also loses useful emoji: `💬 zv: Name the expected signature…` becomes plain `Name…` beneath `### ⚙️ Other`. Please preserve emoji for Other, use a neutral dependency heading, and consider optional-VS16 parser forms for historical subjects such as `🗑 Deprecate…`.

release-plz.toml
zeenix: Agreed and fixed — the test/split split was a real bug, not just a smell. Recognition and removal are now one `replace_regex` (which git-cliff does provide), so they cannot disagree:

```
from="^\s*(?:[\p{Emoji_Presentation}\p{Extended_Pictographic}](?:\u{FE0F}|\u{200D})?\s*)*[\p{Emoji_Presentation}\p{Extended_Pictographic}](?:\u{FE0F}|\u{200D})?\s+"
```

I confirmed the tab case was worse than "can drop words": `✨<TAB>Tabbed` matched `\s` in the test, then `split(pat=" ")` yielded a single element, `slice(start=1)` emptied it, and the entry rendered as `- .`. Whitespace is now consistently supported rather than half-supported. Punctuation-only openers are no longer touched, and a spaced multi-emoji prefix (`✨ 🎉 …`) is now consumed whole instead of one token.

On the no-space history I'm deliberately not following the suggestion, and `bc08ed16` is the reason. Dropping the separator requirement turns `♻️l zb: Refactor server side handshake even more` into `l zb: Refactor …` and `⚰️a Drop …` into `a Drop …` — a stray letter is worse than a visible stray emoji. There are 21 such subjects in this repo (`♻️l`, `♻️zb:`, `♻️zm:`), all historical, all already published in `CHANGELOG.md`; release-plz only prepends, so they will never be regenerated. New subjects are commitlint's job, not the renderer's. Leaving them visibly odd also renders exactly what the old template rendered, which keeps this change provably output-preserving.

Verified over all 6364 commits: 4987 entries, byte-identical to the pre-change output apart from the removed emoji — no dropped words, no empty entries.

---

release-plz.toml
zeenix: Correct, and both commits you cite reproduce. This was a regression I introduced: the emoji had been shielding the rest of the subject from `upper_first`, so moving the filter past the stripped prefix exposed every identifier.

Fixed by guarding it, which is your second suggestion:

```
{% set title = subject | replace_regex(from=…, to="") %}
{% if title == subject %}{% set title = title | upper_first %}{% endif %}
```

So `upper_first` runs only on subjects that never had a prefix — exactly the set it used to affect. `28485173` keeps `zbus_macros shouldn't set features on zbus` and `f53a2113` keeps `signature::Fields::len() now const`.

One further detail your comment led me to: I had also moved `trim` ahead of `upper_first`. That looked harmless but changed one more entry — ` zb-derive: support for #[dbus_interface] with generics` has a leading space, which used to defeat `upper_first` and now no longer would, producing `Zb-derive:` and corrupting the crate name in precisely the way you describe. The template now keeps the original order and lets the regex absorb leading whitespace instead.

Verified across all 4987 entries in this repo: zero casing differences from the pre-change output.

---

release-plz.toml
```diff
@@ -25,16 +25,37 @@ pr_name = "🔖 Releases"
...
+### {{ group | striptags | trim | upper_first }}
 {% for commit in commits %}\
 {% set first_line = commit.message | split(pat="\n") | first %}\
 {% set parts = first_line | split(pat="||") %}\
-- {{ parts | first | upper_first | trim | trim_end_matches(pat=".") }}\
+{% set subject = parts | first %}\
+{% if group is ending_with("Other") %}\
+{% set title = subject | upper_first %}\
+{% else %}\
+{% set title = subject | replace_regex(from="^\\s*(?:[\\p{Emoji_Presentation}\\p{Extended_Pictographic}](?:\\u{FE0F}|\\u{200D})?\\s*)*[\\p{Emoji_Presentation}\\p{Extended_Pictographic}](?:\\u{FE0F}|\\u{200D})?\\s+", to="") %}\
```
zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: The previous split/whitespace issue is fixed, but this replacement is still lossy. `Extended_Pictographic` includes text-default, load-bearing symbols: exact git-cliff 2.13.1 renders `✨ © 2026 licenses` as `- 2026 licenses.`, `✨ ™ Product` as `- Product.`, and `✨ ↩ Return behavior` as `- Return behavior.`. The atom also allows FE0F *or* ZWJ, not the common FE0F+ZWJ sequence: `✨ ❤️‍🔥 Heart` leaves `❤️‍🔥`, while attached `✨🏳️‍🌈 Rainbow` strips nothing; keycaps/tag flags have analogous gaps. Please use a repeated explicit alternation of the configured gimoji sequences (safest). Alternatively, model a complete emoji-presentation sequence and require VS16 before treating text-default pictographs as emoji.

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: Rechecked the amended/rebased head with the exact renderer and full history. Casing, catch-all preservation, neutral Dependencies, ordering, and the intended `👷` reroute now behave correctly. One new blocking edge in the replacement regex remains below.

release-plz.toml
zeenix: Confirmed, on the same git-cliff 2.13.1. All three reproduce, and two more that you didn't list are worse:

| subject | before | after |
|---|---|---|
| `✨ © 2026 licenses` | `- 2026 licenses.` | `- © 2026 licenses.` |
| `✨ ™ Product` | `- Product.` | `- ™ Product.` |
| `✨ ↩ Return behavior` | `- Return behavior.` | `- ↩ Return behavior.` |
| `✨ ❤️‍🔥 Heart` | `- ❤️‍🔥 Heart.` | `- Heart.` |
| `✨🏳️‍🌈 Rainbow` | `- ✨🏳️‍🌈 Rainbow.` | `- Rainbow.` |
| `🐛 ™ Trademark mid-title` | `- Trademark mid-title.` | `- ™ Trademark mid-title.` |

The attached-rainbow case left even the leading `✨` in place, and `🐛 ™ …` destroyed a trademark sign in an ordinary Fixed entry — not a decorative prefix at all. In a repo with 4,987 changelog entries that is the one I'd least want to ship.

I took your second option, modelling a complete emoji-presentation sequence:

```
^\s*(?:(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F})(?:\u{200D}(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F}))*\s*)*(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F})(?:\u{200D}(?:\p{Emoji_Presentation}\u{FE0F}?|\p{Extended_Pictographic}\u{FE0F}))*\s+
```

A text-default pictograph only counts as emoji when it carries U+FE0F, which is what rules out `©`, `™` and `↩`; ZWJ now joins whole sequences rather than being an alternative to FE0F; and the optional FE0F after a default-emoji codepoint keeps `⚡️ Route method replies directly by serial` working, where the parser matches bare `^⚡`.

Three reasons I went with this over the explicit alternation you called safest — happy to switch if you disagree:

1. **It fixes a case the alternation would not.** `✨🏳️‍🌈 Rainbow` strips cleanly here; against a list of the configured gimoji, `🏳️‍🌈` isn't in the list, so the run would break at `✨` and nothing would be stripped — exactly the wart you flagged.
2. **No list to keep in sync.** The eight repos share this template but have slightly different parser sets, so an explicit alternation means eight lists that must track eight `commit_parsers` blocks.
3. **The failure mode is conservative.** Anything the pattern doesn't recognise is left in place. Keycaps and tag sequences still aren't stripped — `✨ 1️⃣ Keycap` renders as `1️⃣ Keycap` — so the worst case is a visible unstripped emoji, never destroyed text. That asymmetry is what went wrong before.

Verified over all 6,364 commits here, and over all 6,954 entries across the eight repos: output identical to the previously reviewed render. The tightening only changes the synthetic edge cases; no real entry moved.

One pre-existing note: the same over-broad atom lives in this repo's scope-stripping `commit_preprocessor`, which I copied it from. It's lower risk there (it also requires a `zb: `-style scope to follow), but I'm happy to tighten it in a follow-up if you want them consistent.

---

zeenix: 🤖 ChatGPT 5.6 on behalf of zeenix: Clarification from the maintainer: the project contract is **exactly one prefix emoji, drawn from the curated gimoji set**. Anything after that prefix—including another emoji—is ordinary title text. My earlier request to model arbitrary compound emoji runs was therefore overbroad.

The current amended regex is still wrong under that contract because its repeated group consumes more than one emoji sequence. For example:

- `✨ 🎉 Add foo` must render as `- 🎉 Add foo.`, not `- Add foo.`
- `✨ ❤️‍🔥 Heart behavior` must render as `- ❤️‍🔥 Heart behavior.`, not `- Heart behavior.`
- `✨🏳️‍🌈 Rainbow` has no separator after the one allowed prefix and should remain visibly malformed rather than being consumed as a combined run.

Please strip **exactly one** recognized gimoji plus its required following whitespace. An explicit alternation of the curated prefixes is appropriate here; there is no need to recognize or repeat over arbitrary Unicode emoji sequences.

zeenix: Understood, and done — the contract settles it. It also retires my main argument for the presentation model: I justified it partly because it stripped `✨🏳️‍🌈 Rainbow` where an alternation would not, but under "one prefix, rest is title text" that subject *should* stay unstripped. I was optimising for the wrong target.

Now exactly one gimoji plus its required whitespace:

```
^\s*(?:✨|🎉|🚸|🐛|🚑|🩹|🥅|⚡|📝|💡|♻|🎨|🚚|🔧|🏗|💄|💥|🔥|🗑|⬆|⬇|📌|➕|➖|🔒|✅|🧪|👷|💚)\u{FE0F}?\s+
```

Your three cases, rendered on git-cliff 2.13.1:

| subject | renders as |
|---|---|
| `✨ 🎉 Add foo` | `- 🎉 Add foo.` |
| `✨ ❤️‍🔥 Heart behavior` | `- ❤️‍🔥 Heart behavior.` |
| `✨🏳️‍🌈 Rainbow` | `- ✨🏳️‍🌈 Rainbow.` |

The alternation is the exact set of emoji this file routes to a group, with the variation selector optional so bare-parser forms like `^⚡` still strip `⚡️`. It doesn't need to be wider: an emoji that isn't in that list can only reach the catch-all, and nothing is stripped there. I've noted in the config comment to keep the two in step when adding a parser — and if they ever drift, the failure is a visibly unstripped emoji, not lost text.

**This does change one real entry in this repo**, and it's the only one across all ~9,000 commits in the eight repos: `1108af61` is `🩹 🚸 zv: \`Signature\` equality regardless of outer parentheses`. It previously rendered as `- zv: \`Signature\` equality …` with both emoji consumed; it now renders `- 🚸 zv: \`Signature\` equality …`, keeping the 🚸 as title text per the contract.

Verified across all eight repos: every entry in an emoji-headed group is now its pre-change text minus exactly one prefix emoji, Other is byte-identical to pre-change, and entry counts are unchanged (111 / 408 / 4987 / 206 / 296 / 910 / 26 / 10).

---

zeenix (commented):
🤖 ChatGPT 5.6 on behalf of zeenix: Rechecked the latest amended head against the strict one-curated-gimoji-prefix contract. Exact git-cliff rendering over the full history preserves all 4,987 entries; the only membership change is the intended 44 👷 entries moving Added → CI. The real `🩹 🚸 zv: …` case correctly retains 🚸 as title text. CI is green. No remaining issues—ready to merge.

PR #1922, opened 2026-08-22
✨ zv: impl conversion from/to BTreeMap for Value

z33ky:
`Dict` itself already supports `BTreeMap`, so this seems like it may have been an oversight of sorts.

### DIFF
--- a/zvariant/src/from_value.rs
+++ b/zvariant/src/from_value.rs
@@ -9,7 +9,10 @@ use crate::{
 #[cfg(unix)]
 use crate::Fd;
 
-use std::{collections::HashMap, hash::BuildHasher};
+use std::{
+    collections::{BTreeMap, HashMap},
+    hash::BuildHasher,
+};
 
 macro_rules! value_try_from {
     ($kind:ident, $to:ty) => {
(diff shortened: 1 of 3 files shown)


zvariant/src/from_value.rs
```diff
@@ -178,6 +178,24 @@ where
     }
 }
 
+impl<'a, K, V> TryFrom<Value<'a>> for BTreeMap<K, V>
```
zeenix, line 184: huge nitpick: why add the new BTreeMap impls above the HashMap ones, rather than below? 🤔

zeenix (changes requested):
Sounds good to me. The title is a bit misleading though, you're adding conversion to/from *Value here, not Dict. I took the liberty of fixing it and making it shorter on the PR title. Please also update the commit title.

zvariant/src/from_value.rs
z33ky: I was going my lexical ordering, but I guess dict.rs has them the other way around, too. I'll change it to match.

zeenix: > I was going my lexical ordering

Ah ok. That's probably better then.

zeenix (approved):

zeenix (approved):

PR #1925, opened 2026-08-26
🏗️ Merge zvariant and zbus_names into zbus for 6.0

### LINKED ISSUE #1144: Drop GVariant support

### LINKED ISSUE #1919: RFC: Merge zvariant into zbus directly
Now that gvariant support is split into a separate crate, it probably makes sense to completely merge zvariant into zbus,  If we do this, we'll want to feature-gate all the existing API so that zbus could be used as a drop-in replacement for anyone needing zvariant directly.

_Originally posted by @zeenix in https://github.com/z-galaxy/zbus/pull/1590#discussion_r3825238265_

zeenix:
## Summary

zbus 6.0 becomes one crate. `zvariant`, `zvariant_derive` and `zbus_names` are folded into
`zbus` and `zbus_macros`, the D-Bus API moves behind a new `comms` feature, and the GVariant
support deprecated in zvariant 5.15 is deleted. `zbus = { version = "6", default-features =
false }` is the replacement for a `zvariant` dependency: the same API under `zbus::wire`, the
same feature names, the same dependency set.

Existing zbus users change little more than the version number; the handful of real breaks
(matching on `Error::Variant`/`Error::Names`, zbus_xml's collapsed error variants) is in the
chapter. `zbus::zvariant` survives as a deprecated alias module until 7.0.

Design doc: `docs/superpowers/specs/2026-08-25-zvariant-merge-design.md`.

Fixes #1919
Fixes #1144

## Before merging

- [ ] Rename the required status check on `main` from `zvariant_fuzz` to `zbus_fuzz`. The fuzz
      job is renamed in this PR, so branch protection would otherwise wait forever for a check
      that no longer runs.
- [ ] After zbus 6.0 is published, cut the final zvariant 5.15.1, zvariant_derive 5.15.1 and
      zbus_names 4.3.5 releases, whose only change is a deprecation banner pointing here. They
      live on a separate `5.x` branch and are not part of this PR.
- [ ] The changelogs release-plz generates for 6.0 inherit a stale `- Bump zvariant to 5.15.`
      entry from pre-branch commits on `main`; drop or reword it in the release PR.

## What changed, by area

- **GVariant removal (#1144).** `src/gvariant/`, `maybe.rs`, `framing_offset{,s}.rs`,
  `Value::Maybe`, `Context::new_gvariant`, `Error::MissingFramingOffset`, the `gvariant` and
  `ostree-tests` features and their pass-throughs, and the gvariant-gated tests are gone. What
  serves zbus/zgvariant coexistence stays: `zvariant_utils/gvariant`, the `Format::GVariant` and
  `Maybe` rejection arms, and `dbus_maybe_rejection`.
- **Shared validators (zvariant_utils 4.3.0).** The bus-name and object-path parsers move to
  `zvariant_utils::{names, object_path}` so `zbus_macros` can validate literals at expansion
  time without depending on the merged zbus. Additive, so zgvariant's `^4.1` is unaffected.
- **One error type.** `zvariant::Error` and `zbus_names::Error` are flattened into `zbus::Error`
  (32 variants, three of them `comms`-only: `MethodError`, `FDO`, `Connection`), and
  `zbus::Error::{Variant, Names}` disappear. `IncompatibleFormat` goes with them: with one wire
  format left nothing ever constructed it. `Connection` boxes its `Address` now, which keeps the
  enum at 64 bytes on `comms` builds instead of 112 — it is what the wire (de)serializers return
  by value out of deep recursion. `zbus_xml`'s two wrapper variants, `Variant` and `Name`,
  collapse into one `Zbus`. Flattening is observable in three ways, all of them called out in the
  upgrade chapter: `Error::Failure(s)` and the old `Error::Variant(zvariant::Error::Message(s))`
  are one value now, so they compare equal where they used to differ; `Error::source()` is a link
  shorter, because the wrapper hop is gone; and `InputOutput` prints `I/O error: {e}` where
  `zvariant::Error` printed the bare `io::Error`.
- **The `comms` gate.** No feature ⇒ wire format and names only; every runtime/D-Bus feature
  implies `comms`. Wire-only users opt into `enumflags2` exactly as zvariant 5 users did, and
  `comms` turns it on exactly as zbus always did. `arrayvec` is new to zbus; `camino`,
  `heapless` and the rest keep the names they had as zvariant pass-throughs. `zbus_macros` keeps
  `comms` in its own defaults, so a crate that depends on the macro crate directly still gets
  `proxy`, `interface` and `DBusError` from a plain version bump; only a direct dependency that
  had turned the (previously empty) defaults off *and* uses those three macros has to add
  `comms` now. zbus's own dependency on the macro crate is such a defaults-off one.
- **The move.** `git mv zvariant/src zbus/src/wire` and `git mv zbus_names/src zbus/src/names`,
  so history follows the files (83 and 10 detected renames). The three crate directories are
  deleted.
- **The derives.** `Type`, `Value`, `OwnedValue`, `SerializeDict`, `DeserializeDict` and
  `signature!` move into `zbus_macros`, whose default codegen path becomes `<zbus>::wire`; both
  `#[zbus(...)]` and `#[zvariant(...)]` keep working. Two latent bugs go with them: `signature!`
  hardcoded `::zvariant`, and `iface.rs` emitted a literal `::zbus::zvariant::Value` that broke
  `crate = ...` for `Value` property setters. Property-setter codegen also converts the
  value-conversion error with `Into::<zbus::Error>::into` rather than `From`, so a user error
  type that implements only `Into<zbus::Error>` — which zbus 5 accepted — keeps compiling.
- **Compatibility.** `zbus::zvariant` is a `#[deprecated]` module of aliases, wrapper functions
  and re-exports. `use zbus::zvariant;` warns, and so does every alias named through it —
  `zbus::zvariant::Value`, `…::Optional`, `…::Result` and friends. What cannot warn, because
  Rust has no deprecation on a `pub use`, are the traits, the five derives, `signature!` and
  the plain re-exports; the module path segment on its own does not warn either. A `compile_fail`
  doctest and an integration test pin both behaviours down.
- **Tests, benches, fuzz.** zvariant's test suite becomes the `zbus/tests/wire/main.rs` target —
  24 top-level files plus the `issue/` and `number/` submodules — so all of it runs with
  `--no-default-features`; the benches become `zbus/benches/{wire,names}.rs` (CodSpeed series
  restart) and the fuzz target moves to `zbus/fuzz`.
- **Docs.** The two module docs are rewritten for their new home, the book drops every
  `zbus::zvariant`, and there is a new chapter, "Upgrading to zbus 6.0", registered as a doctest so
  its snippets are compile-checked — including the ones that claim something no longer compiles. The
  chapter also covers the 6.0 breaks that landed on `main` independently of the merge (#1189's
  `&Value` property setters, #1912's name validation, and the `Optional` null-sentinel fix).
- **CI and release.** Wire-only `check`/`clippy`/`test`/`doc` runs added (Linux, Windows and the
  five cross targets), the two `-p zvariant --features gvariant` runs removed, the coexistence
  check retargeted at `-p zbus --no-default-features --features zvariant_utils/gvariant`, the
  Windows tokio-only run narrowed to `-p zbus … --tests` so feature unification stops handing it
  the default features back, the fuzz job renamed `zbus_fuzz`, and `deploy.yml`'s tag filter
  moved to `zbus-6.*`.
- **Publish order.** `zbus_xml` depends on `zbus` (wire-only) now, and zbus's dev-dependency on
  `zbus_xml` is path-only, so cargo strips it at publish and no cycle reaches crates.io: release
  `zbus` first, then `zbus_xml`.

## Dependency weight

`cargo tree -e normal -p zbus --no-default-features` resolves **21** packages — the same set
`cargo tree -e normal -p zvariant` resolves on `main`, with `zbus` in place of `zvariant` and
`zbus_macros` in place of `zvariant_derive`. `cargo tree -e features` shows no `comms`. The
default build goes from **66** packages to **63**: the same set, minus `zvariant`,
`zvariant_derive` and `zbus_names`, which are no longer packages of their own.

That first number is the one the 16 external wire-format-only crates (zbus-lockstep,
accesskit_atspi_common, glycin-common, gufo-common, ssip, dbus-router, …) care about: switching
to zbus costs them nothing.

## Migration

| Was | Is |
| --- | --- |
| `zvariant::X`, `zbus::zvariant::X` | `zbus::wire::X` |
| `zbus_names::X` | `zbus::names::X` |
| `zvariant::Error`, `zbus_names::Error`, `zbus::Error::{Variant,Names}` | `zbus::Error` |
| `zvariant = "5"` | `zbus = { version = "6", default-features = false }` |
| the `gvariant` feature | the `zgvariant` crate |

The book chapter has the full story, including what warns, what is silent and what breaks:
<https://z-galaxy.github.io/zbus/upgrading-to-6.html>.

## Testing

Run locally, all with `--locked`: wire-only `check`/`clippy`/`test`/`doc`, with and without the
ten wire features; the default, `tokio`-only and `tokio,p2p,vsock` test runs plus the
`uuid,url,time,chrono,option-as-array,vsock,bus-impl` one; `--all-features` docs; cross-target
`check` for windows-gnu, darwin, freebsd, netbsd and android in both the default and the
wire-only configuration; every doctest, including the book and the new chapter, and the
wire-only doctest set as well (the crate-root README is gated on `comms`, with a short
wire-only front page under `not(comms)`); the zgvariant coexistence check; a 30 s fuzz run
from `zbus/fuzz`; a release-plz dry run in a throwaway worktree, which keeps the hand-set
6.0.0 and 4.3.0 versions and writes changelog entries for the five remaining packages only;
and `git rebase -x 'cargo check --workspace --all-targets'` over all 28 commits, so every
commit builds on its own.

busd and zbus_polkit were built against this branch through a temporary `[patch]` to find real
migration friction. Both compile, with no errors and no warnings other than
`use of deprecated type alias zbus::zvariant::…` — `Optional`, `ObjectPath` and `Result` in
busd, `Value` and `OwnedValue` in zbus_polkit — while the derives and traits imported on the
very same `use` lines stay silent, exactly as the chapter describes. Their follow-up PRs are
separate.

https://claude.ai/code/session_01C77ZXwTqmd3qSpSXgw2RfN


zbus_macros/src/iface.rs
```diff
@@ -635,7 +641,7 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                             }
                             ::std::result::Result::Err(e) => {
                                 ::std::result::Result::Err(
-                                    ::std::convert::Into::into(#zbus::Error::Variant(::std::convert::Into::into(e))),
+                                    ::std::convert::Into::into(#zbus::Error::from(e)),
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This strengthens the setter conversion bound from `E: Into<zbus::Error>` to `zbus::Error: From<E>`. I reproduced it on Rust 1.87 with a custom `TryFrom<Value>` error that implements only the legal direct `Into<zbus::Error>`: the `#[interface]` expansion fails with E0277, whereas the pre-merge expansion accepted the analogous `Into<zvariant::Error>`. Please preserve the `Into` obligation (for example, `Into::into(Into::<#zbus::Error>::into(e))`) and add a compile test for it.

zbus_macros/Cargo.toml
```diff
@@ -18,28 +18,27 @@ readme = "README.md"
 
 [features]
 default = []
+# The D-Bus macros: `proxy`, `interface` and `DBusError`. The wire-format derives are always
+# compiled.
+comms = ["syn/fold"]
```
zeenix, line 23: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: A plain 5→6 bump of the stable, direct `zbus_macros` crate now removes `proxy`, `interface`, and `DBusError`: v5 exported them unconditionally, while v6 keeps `default = []` and puts them behind this non-default feature. Could we either preserve the old default surface (for example, default `comms` here and disable macro defaults from zbus's wire-only dependency) or document `features = ["comms"]`/switching to zbus's re-exports in the upgrade guide?

.github/workflows/rust.yml
```diff
@@ -173,10 +195,13 @@ jobs:
           cargo --locked test
           # tokio feature
           cargo --locked test --no-default-features --features tokio
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Please add `-p zbus` here. `cargo tree -e features --no-default-features --features tokio -i zbus` shows that the workspace's `zbus_xmlgen` member re-enables zbus's default `async-io` and `blocking-api` features, so this green job exercises tokio + async-io rather than the intended Windows tokio-only graph. The new Linux doctest comment already identifies `-p zbus` as load-bearing for this exact reason.

.github/workflows/rust.yml
```diff
@@ -120,10 +135,17 @@ jobs:
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
             cargo --locked test --release --verbose -p zbus --features tokio,p2p,vsock \
               -- --skip fdpass_systemd
+          # The executor doctest drives the connection from tokio's scheduler, so it has to build
+          # without the default async-io runtime. `-p zbus` is load-bearing: a workspace-wide
+          # `--no-default-features` is undone by zbus_xmlgen's default-featured zbus dependency.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
-            cargo --locked test --release --verbose --doc --no-default-features connection::Connection::executor
-          # zvariant with its legacy `gvariant` feature (still shipped throughout 5.x).
-          cargo --locked t -p zvariant --features gvariant
```
zeenix, line 126: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Please fold this deletion and its Windows equivalent into `01cc29ea`. That commit removes zvariant's `gvariant` feature but leaves these checked-in invocations active, so its CI fails with “package zvariant does not contain this feature” until `0689a895` removes them. This breaks the project's atomic-commit requirement even though the Cargo-check sweep itself passes.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The custom-setter regression is blocking for 6.0; the inline comments also cover a direct-macro migration break, a false-green Windows feature check, commit atomicity, and compatibility wording.

Two release-doc fixes that cannot be anchored to changed lines:

- `zbus/README.md:142` and `book/src/faq.md:166` still tell readers to depend on zbus 5; these snippets should use version 6.
- The PR/design claim that equality semantics are unchanged needs qualification: in v5 `Error::Failure("x")` and `Error::Variant(zvariant::Error::Message("x"))` were unequal, but both map to `Error::Failure("x")` now. Flattening also shortens `Error::source()` chains by removing the `Variant`/`Names` wrapper hop. Please cover these observable changes in the upgrade guide.

I verified the exact head: all 11 checks are green and all 30 commits pass `cargo check --workspace --all-targets --locked` on Rust 1.87. Those checks do not exercise the reproduced custom-setter case or the invalid workflow command at the intermediate commit.

zbus/src/zvariant.rs
```diff
@@ -0,0 +1,357 @@
+//! Deprecated aliases for the [`wire`] module.
+//!
+//! Every item here is an alias for, or a re-export of, its [`crate::wire`] counterpart. The
+//! module exists so that code written against the `zvariant` crate, or against `zbus::zvariant`,
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This module cannot keep standalone `zvariant::…` paths compiling; it only preserves `zbus::zvariant::…`. The upgrade guide correctly says direct zvariant users must replace the dependency and rename their paths, so please remove the standalone-crate claim here.

zbus_macros/src/iface.rs
zeenix: Fixed in 555c8b28. The three setter conversion sites in `zbus_macros/src/iface.rs` now emit `::std::convert::Into::<zbus::Error>::into(e)` instead of `zbus::Error::from(e)`, so an error type that implements only `Into<zbus::Error>` compiles again (zbus 5 required no more than that), while `Infallible` and plain `From` impls keep working because the target type is pinned. `zbus_macros/tests/tests.rs` gains `interface_property_setter_with_into_only_error`, which fails with E0277 on the previous codegen. Both live in the "Unify the error types" commit.

zbus_macros/Cargo.toml
zeenix: Done: `zbus_macros` now has `default = ["comms"]`, so a crate depending on the macro crate directly keeps `proxy`, `interface` and `DBusError` on a plain 5→6 bump. zbus depends on it with `default-features = false` and re-enables `comms` through its own `comms` feature, so the wire-only build is unchanged (`cargo tree -e features -p zbus --no-default-features` shows no `zbus_macros feature "comms"`). The upgrade chapter mentions it in the Cargo.toml section.

.github/workflows/rust.yml
zeenix: Changed to `cargo --locked test -p zbus --no-default-features --features tokio --tests`, with a comment recording why. `--tests` is deliberate rather than an omission: five doc examples use `async_io` and fail under a tokio-only graph (pre-existing on `main`, never exercised because the old invocation unified `async-io` back in), so running doctests here would turn the false green into a real red for a reason unrelated to this PR.

.github/workflows/rust.yml
zeenix: Folded: the "Drop the zvariant GVariant test runs" commit is squashed into "Remove GVariant support", whose message gains a paragraph about the CI steps going with the feature. No commit on the branch now invokes `--features gvariant` against a crate that lacks it; the per-commit `cargo check --workspace --all-targets` sweep still passes over all 29 commits.

zbus/src/zvariant.rs
zeenix: Reworded: the module doc now says it keeps `zbus::zvariant::…` paths compiling, and that code written against the `zvariant` crate itself switches its dependency and renames `zvariant::` to `zbus::wire::`, pointing at the "Upgrading to zbus 6.0" chapter.

zbus_macros/Cargo.toml
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The plain-default case is fixed, but one supported v5 configuration still regresses. Because v5's defaults were already empty, a direct `default-features = false, features = ["blocking-api"]` dependency still exported these macros; current head fails with E0432 because `blocking-api = []` does not enable their `comms` gate. I reproduced this against v5 and current head, and `blocking-api = ["comms"]` fixes it. Please also qualify the upgrade-guide claim: direct users that explicitly disabled defaults must remove that setting or add `comms`.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Rechecked the rebased head. The setter regression, Windows tokio-only coverage, GVariant commit atomicity, stale version snippets, and equality/source migration notes are fixed. Two existing threads remain open for the direct-macro no-default edge and compatibility wording; two further documentation/history contradictions are inline.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
```diff
@@ -0,0 +1,527 @@
...
+  not carried over.
+
+Gating: exactly the variants that embed `comms`-only types get `#[cfg(feature = "comms")]` —
+`MethodError(OwnedErrorName, Option<String>, Message)`, `FDO(Box<fdo::Error>)`,
+`Connection(Arc<io::Error>, Address)` — together with `impl From<Message>` and
+`impl From<fdo::Error>`. `InterfaceExists(InterfaceName<'static>, ObjectPath<'static>)` only
+uses core types and stays ungated, as do the remaining D-Bus-flavoured unit variants (they are
+inert in a wire-only build). The five hand-written matches (`PartialEq`, `Error::source`,
+`Display`, `description`, `Clone`) gain three cfg arms each. Every variant keeps its current
+`Display` text and compares to itself as it does today. What flattening changes is comparison
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This `Display` claim still needs the `InputOutput` exception: `zvariant::Error::InputOutput` printed only the inner error, while the merged variant prefixes `I/O error:`, exactly as the upgrade guide now explains. Please qualify this sentence so the design and migration contract agree.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
```diff
@@ -0,0 +1,3058 @@
...
+/usr/bin/git add -A
+/usr/bin/git -c core.hooksPath=/dev/null commit --no-verify --no-gpg-sign -F - <<'MSG'
+💥 zb: Unify the error types
+
+Three enums in one crate is two too many. `zbus::Error::Variant` and
+`zbus::Error::Names` only ever wrapped errors that now come from the
+same crate, so matching on a wire-format failure meant peeling two
+layers and knowing which of three `Display` impls produced the text.
+
+Every variant keeps its message, its `source()` and its equality
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This remains stale here and verbatim in `e7c9c130`'s commit message. Flattening changes cross-wrapper equality and removes a `source()` hop, as the corrected design and upgrade guide now say. Please amend this plan text and the existing commit message; `merge-5-docs-and-release.md:2364` repeats the shorter stale Display/equality claim too.

zbus/src/zvariant.rs
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The standalone-crate wording is fixed. One qualification remains: saying all `zbus::zvariant::…` paths keep compiling is broader than the compatibility actually provided—the upgrade chapter itself lists tuple-struct constructors such as `DynamicTuple(…)` and the removed `Error` variants as exceptions. Could this say it preserves *most* such paths and point readers to those exceptions?

zbus_macros/Cargo.toml
zeenix: Fixed in 73b512d1: `blocking-api = ["comms"]` in zbus_macros, so a direct `default-features = false, features = ["blocking-api"]` dependency gets the macros again (`cargo tree -e features -p zbus_macros --no-default-features --features blocking-api` shows `comms` on). The upgrade chapter now says that a direct dependency which had turned the previously empty defaults off needs `features = ["comms"]` or drops `default-features = false`, and that `blocking-api` implies it.

zbus/src/zvariant.rs
zeenix: Qualified in 73b512d1: the module doc now says *most* `zbus::zvariant::…` paths keep compiling and names the exceptions — tuple-struct constructors called through an alias such as `DynamicTuple(…)`, and the removed `Error` variants — pointing at the chapter's "What warns, what is silent, what breaks" section.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
zeenix: Qualified in 73b512d1: the spec sentence now reads "Every variant keeps its current `Display` text — except `InputOutput`, whose zvariant text gains zbus's `I/O error: ` prefix — and compares to itself as it does today", matching the upgrade chapter.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
zeenix: Amended in 73b512d1: the plan-3 paragraph and the "Unify the error types" commit message now say every variant keeps its message and compares to itself as before, while comparison across the former wrappers (`Failure(s)` vs the old `Variant(zvariant::Error::Message(s))`) and the `source()` chain change; the plan-5 sentence at :2364 says the same in short form, with the `InputOutput` `Display` exception.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
```diff
@@ -0,0 +1,3061 @@
...
+- Modify: `zbus/tests/wire/recursion_limits.rs`
+- Modify: `zbus_macros/src/iface.rs` (three codegen sites)
+- Modify: `zbus_xml/src/{error,xml}.rs`, `zbus_xml/tests/tests.rs`
+- Modify: `zbus/src/wire/{dbus/mod,serialized/data,from_value,ser}.rs` (`Error::Message` sites)
+- Test: `cargo test --workspace`
+
+**Interfaces:**
+- Consumes (tasks 3–4): `zbus::wire`, `zbus::names`, `zbus_macros`' codegen.
+- Produces:
+  - `zbus::Error` — one `#[non_exhaustive]` enum with the 33 variants listed below, plus
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The final all-feature enum has 32 variants, not 33, and this plan still adds the deliberately dropped `IncompatibleFormat` below (including its `PartialEq`, `source`, `Display`, `description`, and `Clone` arms). Old zbus had 22 variants; removing `Variant`/`Names` leaves 20, then 10 zvariant and 2 names variants total 32. Please remove the stale variant/arms and update this count, plan 5’s `33`, and the design spec’s `23 → 21` to `22 → 20`.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Rechecked the rewritten head. The previously reported runtime, feature-graph, compatibility, Display, and history fixes are effective; the remaining findings are three factual inconsistencies in the checked-in plans.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
```diff
@@ -0,0 +1,3061 @@
...
+/usr/bin/git add -A
+/usr/bin/git -c core.hooksPath=/dev/null commit --no-verify --no-gpg-sign -F - <<'MSG'
+💥 zb: Unify the error types
+
+Three enums in one crate is two too many. `zbus::Error::Variant` and
+`zbus::Error::Names` only ever wrapped errors that now come from the
+same crate, so matching on a wire-format failure meant peeling two
+layers and knowing which of three `Display` impls produced the text.
+
+Every variant keeps its message and compares to itself as before;
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The amended `c7f672e2` commit now correctly records the one `InputOutput` text change, but this checked-in commit template still says every variant keeps its message without that exception. Please copy the same qualification here so the executable plan and rewritten history agree.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
```diff
@@ -0,0 +1,2600 @@
...
+  `-p zbus --no-default-features --features zvariant_utils/gvariant`, the fuzz job renamed
+  `zbus_fuzz` (**branch protection needs updating**), and `deploy.yml`'s tag filter moved to
+  `zbus-6.*`.
+
+## Dependency weight
+
+`cargo tree -e normal -p zbus --no-default-features` resolves **21** packages — the same set
+`cargo tree -e normal -p zvariant` resolves on `main`, with `zbus` in place of `zvariant` and
+`zbus_macros` in place of `zvariant_derive`. `cargo tree -e features` shows no `comms`. The
+default build is unchanged at **66** packages.
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The exact normalized command above returns 63 on this head, not 66; the three removed package nodes explain the drop, and the PR body correctly says 66 → 63. Please update this line, Step 6’s `Expected: 66`, and Step 13’s `21 and 66` so the checked-in plan matches the measured result.

zbus_macros/Cargo.toml
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The feature edge is fixed and I independently reproduced it. One qualifier I missed in my earlier wording remains in `book/src/upgrading-to-6.md:85-89`: only direct no-default users of `proxy`, `interface`, or `DBusError` need `comms`. A derive-only direct dependency still compiles with defaults off and no `comms`, so please scope the migration sentence accordingly and avoid saying only zbus turns that default off.

zbus/src/zvariant.rs
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The main overclaim is fixed. The named exception list is still not exhaustive, though: removed GVariant paths such as `zbus::zvariant::Value::Maybe` also no longer compile and are documented in a separate section. Please call these “notable exceptions” and point to the full upgrade chapter (or also the GVariant section).

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
zeenix: Fixed in c16083b6: the plan-3 listing no longer carries `IncompatibleFormat` or its five arms and says 32 variants; plan 5 says 32 as well; the spec now reads "From zbus (22 → 20)" (counted on the pre-merge `zbus/src/error.rs`: 22 variants, minus `Variant`/`Names`).

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
zeenix: Copied in c16083b6: the plan-3 commit template now carries the same `InputOutput` sentence as the rewritten commit message.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
zeenix: Updated in c16083b6: Step 6 expects 63 (down from 66 on `main`, the three removed packages), Step 13 says "21 and 63", and the PR-text template says the default build goes from 66 to 63.

zbus_macros/Cargo.toml
zeenix: Scoped in c16083b6: the chapter now says only a direct dependency that had turned the (previously empty) defaults off *and* uses `proxy`, `interface` or `DBusError` needs `comms` (or drops `default-features = false`); the derives never needed it. The sentence about zbus's own dependency now says it is one such defaults-off dependency that re-enables `comms` through its own feature, rather than "only zbus".

zbus/src/zvariant.rs
zeenix: Reworded in c16083b6: "notable exceptions", now including everything GVariant such as `Value::Maybe` alongside the alias tuple-struct constructors and the removed `Error` variants, and the pointer is to the whole "Upgrading to zbus 6.0" chapter rather than one section.

zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: “Everything GVariant” is now too broad. In the supported coexistence graph, feature unification intentionally retains `Format::GVariant` and `Signature::Maybe`; I compiled `zbus::zvariant::serialized::Format::GVariant` in an external no-default consumer. Please say “the removed GVariant APIs, such as `Value::Maybe` and `Context::new_gvariant`.” While touching the opening, include forwarding wrappers too: this module defines five of them, so not every item is an alias or re-export.

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
```diff
@@ -0,0 +1,2201 @@
...
+Expected: `zbus/Cargo.toml OK`. If an assertion fires it prints the symmetric difference — fix the
+manifest and re-run.
+
+- [ ] **Step 8: Add the `comms` feature to `zbus_macros`**
+
+In `zbus_macros/Cargo.toml`, the `[features]` table must read exactly:
+
+```toml
+[features]
+default = []
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Following this “exact” table would undo the direct-macro fixes: the owning commit now correctly has `default = ["comms"]`, `comms = ["syn/fold"]`, and `blocking-api = ["comms"]`. Please update this block and the stale `default = []` hand-off at :2183. The same commit message’s claim that “no existing manifest changes meaning” also needs the explicit-no-default direct-`zbus_macros` exception now documented in the book.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
```diff
@@ -0,0 +1,2600 @@
...
+cd /home/zeenix/checkout/z-galaxy/zbus/zvariant-merge/book/src
+/usr/bin/grep -rn 'zvariant\|zbus_names' *.md
+```
+
+Expected: **no output**. (The new chapter, added in the next step, is the one file that is
+allowed to mention them.)
+
+- [ ] **Step 7: Write the new chapter `book/src/upgrading-to-6.md`**
+
+Create the file with exactly this content. Every ` ```rust ` block in it is compiled by
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This “exact” chapter block is still broadly stale relative to the committed chapter: it says zbus users’ code keeps compiling (:1294), every zvariant feature kept its name (:1315), and zvariant/zbus have unrelated `Signature` types (:1495), claims zgvariant has the same API, and omits the direct-`zbus_macros` migration note. Plan 4’s exact `zvariant.rs` block likewise retains the standalone-zvariant overclaim, and plan 2’s embedded commit text still claims host `winnow`/`serde` disappear although the actual message was corrected. Please synchronize the exact source/message templates with their owning commits rather than only patching the final files.

book/src/upgrading-to-6.md
```diff
@@ -0,0 +1,393 @@
...
+
+* `zvariant::Error::Message(s)` — what serde's `Error::custom` produced — is now
+  `zbus::Error::Failure(s)`, which is where zbus already collected such errors.
+* `Error::InputOutput` prints as `I/O error: <the io::Error>`, where `zvariant::Error` printed
+  the inner error on its own. zbus has always prefixed it and its rendering is the one that
+  survived, so this is the one change a log line or a string comparison can see at runtime.
+* `Error::MissingFramingOffset` is gone with the rest of GVariant, and so is
+  `Error::IncompatibleFormat`: with a single wire format left, nothing can be incompatible with
+  it.
+* `Error::Connection` carries a `Box<Address>` where zbus 5 inlined the `Address`. `Display` is
```
zeenix, line 196: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This breaks semantic commit atomicity. The paragraph lands in the book commit `bb7a2c1a8`, where `Error::Connection` still contains `Address`; the `Box<Address>` change does not exist until final commit `c16083b6f`. Please move this migration note into the boxing commit (and keep the plan’s owning chapter template aligned), or fold the code change earlier. A per-commit Cargo check cannot catch this documentation/API mismatch.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
```diff
@@ -0,0 +1,3050 @@
...
+`zbus::Error::Names` only ever wrapped errors that now come from the
+same crate, so matching on a wire-format failure meant peeling two
+layers and knowing which of three `Display` impls produced the text.
+
+Every variant keeps its message and compares to itself as before; what
+changes is comparison across the former wrappers — `Failure(s)` and the
+old `Variant(zvariant::Error::Message(s))` are one value now — and the
+`source()` chain, which loses the wrapper hop. The two wrapper variants
+and the seven `Invalid*Name(String)` variants deprecated since
+zbus_names 4.1 (which nothing constructed) are the only things lost. The
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: These are not the only losses: the same error-unification commit deliberately removes `IncompatibleFormat` and all of its arms. Please fix this template and the actual `8b2eb7e36` commit message. Plan 1’s hand-off at :1464-1466 should agree too: 10 variants are carried in, `Message`/`InputOutput` are deduplicated, and `IncompatibleFormat` is dropped—not “exactly 11” plus the two deduplications.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Rechecked the full rewritten history. All live CI checks and all 29 Rust 1.87 per-commit workspace checks pass; the remaining findings are documentation/history consistency issues. Two commit subjects also miss CONTRIBUTING.md’s package prefix: `d5a5adafd` changes `zbus/Cargo.toml` and `zbus/src/wire/mod.rs` but says only `zm,zd`, while `bb7a2c1a8` changes `zbus/src/lib.rs` but has no `zb` prefix.

zbus_macros/Cargo.toml
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The feature behavior is fixed, but one wording error remains: “the derives never needed it” includes `DBusError`, which is itself a `comms`-gated derive; a direct no-default import still fails with E0432. Please say “the wire-format derives and `signature!` never needed it.” The PR description also still has the old false claim that only zbus’s wire-only dependency turns the default off.

zbus/src/zvariant.rs
zeenix: Both corrected in 53755764. The opening now reads "an alias for, a forwarding wrapper around, or a re-export of" (the module defines five wrappers: `serialized_size`, `to_bytes`, `to_bytes_for_signature`, `padding_for_n_bytes`, `signature::validate`). "Everything GVariant" is now "the removed GVariant APIs, such as `Value::Maybe` and `Context::new_gvariant`"; `serialized::Format::GVariant` and `Signature`'s `Maybe` come from zvariant_utils and stay for zgvariant coexistence.

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
zeenix: Done in 53755764. The block is copied verbatim from `zbus_macros/Cargo.toml` (`default = ["comms"]`, `comms = ["syn/fold"]`, `blocking-api = ["comms"]`, current comments) and the hand-off at the end of the plan says the same. The gate commit message now says "almost no existing manifest changes meaning" and names the exception: a direct `zbus_macros` dependency with `default-features = false` that uses `proxy`/`interface`/`DBusError` has to add `comms`; `blocking-api` implies it; the wire-format derives and `signature!` need nothing.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
zeenix: All synced in 53755764: plan 5's chapter fence is the final `book/src/upgrading-to-6.md`; plan 4's `zvariant.rs` fence is the final file (the module doc was one of five drifts — the `impl_type_with_repr`/`static_str_type` re-exports, the `enum@`/`struct@` link disambiguators and rustfmt's multi-line `#[deprecated]` were the others); plan 2's heredoc is the real commit message; plan 3's too, subject included; plan 1's hand-off says 10 variants carried, `Message`/`InputOutput` deduplicated, `IncompatibleFormat` dropped.

book/src/upgrading-to-6.md
zeenix: Moved in 53755764: the paragraph is removed from `📝 zb: Update the book for zbus 6.0` and added, identical, to `⚡️ zb: Box the address in Error::Connection` (whose stat now includes the chapter). The list it belongs to is introduced by a count, so the intermediate commits say "Four details" and the boxing commit restores "Five". Plan 5's chapter template is the final chapter.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
zeenix: Added in 53755764, in both the commit message and plan 3's template: "The two wrapper variants are the only things lost that anything ever constructed. The seven `Invalid*Name(String)` variants deprecated since zbus_names 4.1 go too, and so does `IncompatibleFormat`, which nothing has constructed since 2020." Plan 1's hand-off now counts 10 carried variants with the two deduplications and the one drop.

zbus_macros/Cargo.toml
zeenix: Reworded in 53755764: "the wire-format derives and `signature!` never needed it" (`DBusError` is `comms`-gated). The PR body now matches the chapter: any direct dependency that had turned the (previously empty) defaults off and uses those three macros adds `comms`; the "only zbus turns the default off" sentence is gone.

docs/superpowers/plans/2026-08-25-zvariant-merge-1-gvariant-removal.md
```diff
@@ -0,0 +1,1497 @@
...
+EOF
+/usr/bin/git show --stat HEAD | head -50
+```
+
+Expected: `git status --short` before the commit shows the six deleted source files, the two
+deleted test files and the modified files from this task's **Files** list, nothing else.
+
+---
+
+### Task 3: Drop the GVariant CI runs
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This task would recreate the atomicity failure fixed earlier. Task 2 removes the `gvariant` feature, then this separate commit leaves checked-in CI invoking that nonexistent feature in the intervening revision. The actual `03b0b8d62` correctly folds these workflow deletions into feature removal and has no standalone CI commit. Please fold this task into Task 2 and update Plan 1’s four-commit expectations/history.

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
```diff
@@ -0,0 +1,2220 @@
...
+
+Replace the `run:` block of the windows `Test` step with:
+
+```yaml
+        run: |
+          $PSNativeCommandUseErrorActionPreference = $true
+          Start-Process C:\vcpkg\installed\x64-windows\tools\dbus\dbus-daemon.exe '--config-file=CI/win32-session.conf --address=autolaunch:'
+          cargo --locked test
+          # tokio feature
+          cargo --locked test --no-default-features --features tokio
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This exact workflow template still recreates the false-green Tokio job we fixed: without `-p zbus`, `zbus_xmlgen` unifies zbus’s default `async-io` back in. The Linux executor command referenced at :1766 has the same workspace-wide problem. Please copy the final scoped commands here (`-p zbus`, and `--tests` on Windows), then update the verification command at :2016 and this planned commit’s body.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
```diff
@@ -0,0 +1,2743 @@
...
+`cargo test --doc -p zbus`: the five `noplayground` ones must compile and run, and the two
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The embedded chapter now has eight Rust doctests: five runnable and three `compile_fail` (removed error variants, duplicate `From`, and the nested `DynamicTuple` constructor). This still says 5+2; :1704 says the count rises by only five/3+2; and `fbd1e2296`’s body also says two. Please correct all three claims to 5+3 / eight.

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
```diff
@@ -0,0 +1,3051 @@
...
+types that now live in the same crate, so the `From` conversions collapse into identities and `?`
+keeps working everywhere without a single call-site change — except at the handful of places that
+*named* the wrapper variants, all of which are listed below.
+
+**Two resolutions the spec leaves implicit**, both decided here:
+1. `InputOutput(Arc<io::Error>)` existed in both enums with different `Display` text (`zvariant`:
+   `{e}`; `zbus`: `I/O error: {e}`). The spec says it "dedupes with zbus's identical variant", so
+   **zbus's text wins**: `I/O error: {e}`.
+2. `description()` is a zbus-only inherent method with no counterpart in the other two enums, so
+   the eleven new variants get new strings. They are written out in full below; they follow the
```
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This count is twelve, not eleven: ten carried wire-error variants plus `InvalidName` and `InvalidNameConversion`. The final accounting confirms it: 20 retained zbus variants + 12 added variants = 32.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Rechecked `537557644`. The implementation, rebase, current CI, and 29-commit Rust 1.87 build sweep are clean. The remaining findings are in the executable plans and amended commit history: following those plans would recreate already-fixed atomicity/coverage defects, and two factual counts are still wrong.

book/src/upgrading-to-6.md
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The actual history placement is fixed, but the executable plan still recreates the defect. Plan 5 Step 7’s “exact” chapter contains the `Box<Address>` bullet, and Step 12 commits that file before any boxing code; the plan has no later boxing task. Please make the Step 7 template the four-detail intermediate chapter and introduce the fifth detail only in a later boxing step/commit (or add that step explicitly).

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The final feature table/hand-off and actual `bf0d37337` message are corrected. The plan’s own commit template at :1053-1060 is still stale, though: its subject omits `zx` despite staging `zbus_xml/Cargo.toml`, and it retains the false “no existing manifest changes meaning” claim without the direct-`zbus_macros` exception. Also, `dac0d2f62` now contains 46+/28- of feature, compatibility, re-export and link corrections while its subject/body still claim exactly two unrelated fixes. Please align both pieces of history.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
zeenix: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The specifically named fences now match, but the broader executable-plan synchronization is still incomplete. Plan 3’s names/wire/derive heredocs and expected log differ from `aaa180dcb`, `ac3ce83e3` and `36fc98adf`; Plan 4 still schedules already-folded CI/release-plz/doc changes in later commits; and Plan 5 still expects the obsolete top-three history and an unprefixed book subject. Please mechanically reconcile every add list, commit heredoc and expected-history block with the actual series rather than fixing only the quoted fragments.

docs/superpowers/plans/2026-08-25-zvariant-merge-1-gvariant-removal.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. Plan 1's standalone CI task went with them; the branch keeps the workflow deletions inside the GVariant-removal commit.

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. The workflow itself already carries the scoped commands (`-p zbus`, and `--tests` on Windows).

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. One half of this was already right on the branch, though: the book commit's body says "including the three snippets that claim something no longer compiles", not two — it was the plan text that said 5+2, and the chapter does have five runnable and three compile-fail fences (the third is indented inside a list item at `book/src/upgrading-to-6.md:284`).

docs/superpowers/plans/2026-08-25-zvariant-merge-3-move.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. The count was indeed twelve — ten carried wire variants plus `InvalidName` and `InvalidNameConversion`, giving 20 + 12 = 32.

book/src/upgrading-to-6.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. The history placement stands: the `Box<Address>` bullet lives in `⚡️ zb: Box the address in Error::Connection` and nowhere earlier.

docs/superpowers/plans/2026-08-25-zvariant-merge-4-gate-and-compat.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. That includes `📝 Correct two stale statements in the gate plan`, which only ever touched plan 4 — it is gone from the series as well.

docs/superpowers/plans/2026-08-25-zvariant-merge-5-docs-and-release.md
zeenix: The implementation plans are no longer part of this PR: `docs/superpowers/plans/` is dropped from the branch (the design spec stays), so this finding is moot. They were pre-execution documents, and keeping them byte-consistent with a series that has been rewritten five times was buying churn rather than accuracy. That removes the whole class of finding rather than the quoted fragments; the design spec is the only `docs/superpowers/` file the PR still adds, and its one reference to "the plan" is gone.

zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: Rechecked at `033f01d6`. The retained implementation patches are unchanged and sound, and I independently confirmed 10/10 Actions jobs plus a clean Rust 1.87 check of all 27 commits. I resolved the seven threads whose implementation-plan targets were removed; no user-facing documentation or corrected behavior was lost.

One PR-description count is now stale: the Testing section still says the sweep covered 29 commits, while this head has 27 after removing the two plan-only commits. Please update it. The inline findings below are confined to the retained design record and commit atomicity.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
```diff
@@ -0,0 +1,528 @@
...
+- deprecated type aliases for every struct and enum listed above (`pub type Value<'a> =
+  crate::wire::Value<'a>;` and so on), plus nested deprecated `serialized` and `signature`
+  modules aliasing `Context`, `Data`, `Format`, `Size`, `Written`, `Signature`, `Child`,
+  `Fields`, `Error`;
+- deprecated `Error`, `Result` and `MaxDepthExceeded` aliases of the root types;
+- deprecated wrapper functions for `to_bytes`, `to_bytes_for_signature`, `serialized_size` and
+  `padding_for_n_bytes`, with the same signatures;
+- deprecated constants for the signature chars/strings and `LE`/`BE`/`NATIVE_ENDIAN`;
+- plain (silent) re-exports for the four traits, the five derives, `signature!`, and the
+  `as_value` and `dbus` submodules — nothing can make these warn.
```
zeenix, line 149: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This is not the warning/export behavior of the final compatibility module. `use zbus::zvariant::Value` does warn because `Value` is itself a deprecated alias; a silent example would be the plain `Type` re-export. The final module also has five wrappers (including `signature::validate`), seven silent traits, `NETWORK_ENDIAN`, and further silent exports such as `to_writer{,_for_signature}`, `export`, and the helper macros. Please sync this compatibility inventory with `zbus/src/zvariant.rs`, since this section is the design contract for what warns versus compiles silently.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
```diff
@@ -0,0 +1,528 @@
...
+  and `Error::Message` would be misleading next to `zbus::Message`. `MissingFramingOffset` is
+  GVariant-only and goes with the GVariant removal.
+- `IncompatibleFormat` is not carried: nothing has constructed it since 2020.
+- From zbus_names (2): `InvalidName(&'static str)`, `InvalidNameConversion { from, to }`. The
+  seven `Invalid*Name(String)` variants deprecated since 4.1 are constructed by nothing and are
+  not carried over.
+
+Gating: exactly the variants that embed `comms`-only types get `#[cfg(feature = "comms")]` —
+`MethodError(OwnedErrorName, Option<String>, Message)`, `FDO(Box<fdo::Error>)`,
+`Connection(Arc<io::Error>, Address)` — together with `impl From<Message>` and
```
zeenix, line 186: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: The final public variant is `Connection(Arc<io::Error>, Box<Address>)`, not the unboxed shape recorded here. The last commit deliberately makes that API change to keep `Error` at 64 bytes, and the upgrade chapter/PR body already document it. Please update the retained design record as well.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
```diff
@@ -0,0 +1,528 @@
...
+two `#[macro_export]` macros `impl_type_with_repr!` and `static_str_type!` necessarily land at
+the zbus root; they are `#[doc(hidden)]` there and `#[doc(inline)] pub use`d from `wire`, so
+`zbus::wire::impl_type_with_repr!` is the documented path. The 78 `pub(crate)` items in the
+moved code are left as they are; tightening them to `pub(super)` is an optional follow-up.
+
+### Tests, benches, fuzz
+
+zvariant's 41 test files become one integration-test target `zbus/tests/wire.rs` with modules
+under `zbus/tests/wire/` (`common.rs` becomes a real `#[macro_use] mod common`; feature-gated
+files become `#[cfg(feature = …)] mod`; zvariant's `issue/` tests stay under `wire/issue/`),
```
zeenix, line 374: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: These move details no longer match the final tree. `unusual_byte_groupings` is removed entirely (there is no such allow outside this sentence), and after GVariant removal there are 39 Rust test files; the two old aggregator files become one `zbus/tests/wire/main.rs`, leaving 38 files under that directory. The Cargo integration-test target is therefore `zbus/tests/wire/main.rs`, not `zbus/tests/wire.rs`. Please record the move that actually landed.

docs/superpowers/specs/2026-08-25-zvariant-merge-design.md
```diff
@@ -0,0 +1,528 @@
...
+4. Outside this repo: RustSec advisories, GitHub release notes, downstream path updates (busd,
+   zbus_polkit).
+
+## Testing strategy
+
+All with `--locked`; run locally and encoded in CI:
+
+- Wire-only: `check`, `clippy`, `test` with `-p zbus --no-default-features` and with all ten wire
+  features; `cargo doc --no-default-features -p zbus` with `-D warnings`.
+- Unchanged configurations: default; `--no-default-features --features tokio`;
```
zeenix, line 478: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This is still overbroad and repeats the exact Tokio scoping bug fixed in CI. Several items below are local-only (`release-plz`, dependency comparison, the per-commit sweep, and downstream dry runs), so the whole list is not “encoded in CI.” Also, the Tokio-only command needs the load-bearing `-p zbus`; without it, `zbus_xmlgen` unifies zbus's default `async-io` back in and the check is false-green. Please qualify the heading, copy the scoped command from the workflow (including Windows's `--tests` distinction), and fold `5a7e915b` into the original spec commit `2aeaa83c` as CONTRIBUTING requires for review fixes.

zbus/src/blocking/mod.rs
```diff
@@ -17,6 +17,8 @@
 //!
 //! [asf]: https://rust-lang.github.io/wg-async/vision/shiny_future/users_manual.html#caveat-beware-the-async-sandwich
 //! [`blocking` crate]: https://docs.rs/blocking/
+//! [`tokio::runtime::Runtime::block_on`]: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on
```
zeenix, line 20: 🤖 Chatgpt GPT-5.6 on behalf of Zeeshan: This fix lands too late for an atomic series. I reproduced the checked-in doc job under Rust 1.98 with `RUSTDOCFLAGS='-D warnings' cargo --locked doc --all-features -p zbus_xml`: `f265a905`, `6e2ffba5`, and `91e70a3c` all fail on this Tokio link and `Connection::set_unique_name`; only `d2666112` turns the job green. Please absorb or reorder this fix so commits 14–16 are not individually CI-red.

zeenix: Fixed in 8af49875, and the finding understated the range. `📝 zb: Make two intra-doc links feature-independent` is now the FIRST commit of the series, as prep, with its message reworded to say so.

Sweeping the checked-in command over every commit — `git rebase -x "RUSTDOCFLAGS='-D warnings' cargo --locked doc --all-features -p zbus_xml"` — showed the red actually starts one commit earlier than reported: at `🚚 zb,zn,zx: Move zbus_names into zbus as the names module`, which is where zbus_xml first depends on zbus and therefore first documents it. Moving the fix to just before the wire move (where the finding placed it) still left that commit red; only hoisting it to the front clears the whole series. The sweep now reports 27 successful doc builds and zero errors.

zeenix: 🤖 Chatgpt GPT-5.6 on behalf of zeenix: Verified at `8af49875`. The range-diff shows this rustdoc patch moved unchanged from commit 17 to commit 1; the other 26 PR patches remain identical, with only the upstream `uuid` lockfile bump added by the rebase. I reran `RUSTDOCFLAGS='-D warnings' cargo --locked doc --all-features -p zbus_xml` on commits 13–17 (`d3d61e25` through `eeff8dbf`), covering the complete formerly failing interval, and all five pass. Resolving.

PR #1927, opened 2026-08-26
📝 zb: Add usage example to ObjectManager docs

### LINKED ISSUE #1410: Improve ObjectManager docs
Maybe this is already possible, but it seems like you cannot implement an object manager if you use object server, which seems silly, given that 90% of the data is there already in object server? It has a complete view of its children, can do instrospection etc etc, you'd just want to make the object server an object manager at a certain path and enforce that all served objects are children of the root.

Would such a contribution be accepted? I guess if not I'm just going to do the ugly thing and copy a bunch of code :)

ChrisJr404:
Adds a small usage example to the `ObjectManager` docs showing how to register it at a path via the object server and add managed objects under it, since that was not obvious from the prose alone.

Closes #1410.

### DIFF
--- a/zbus/src/fdo/object_manager.rs
+++ b/zbus/src/fdo/object_manager.rs
@@ -21,8 +21,9 @@ pub type ManagedObjects =
 ///
 /// The recommended path to add this interface at is the path form of the well-known name of a D-Bus
 /// service, or below. For example, if a D-Bus service is available at the well-known name
-/// `net.example.ExampleService1`, this interface should typically be registered at
-/// `/net/example/ExampleService1`, or below (to allow for multiple object managers in a service).
+/// `org.freedesktop.zbus.ExampleService1`, this interface should typically be registered at
+/// `/org/freedesktop/zbus/ExampleService1`, or below (to allow for multiple object managers in a
+/// service).
 ///
 /// It is supported, but not recommended, to add this interface at the root path, `/`.
 ///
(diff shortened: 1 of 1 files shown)


zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
 /// `ObjectManager` at an ancestor of an interface from within that interface's own `&mut self`
 /// method will deadlock. See [`ObjectServer::at`](crate::ObjectServer::at) for details.
 ///
+/// # Example
+///
+/// You do not implement this interface yourself. Add the provided `ObjectManager` at a path via the
```
zeenix: I think the first part is redundant or at least this is a bit late to say that (if you think we need to explicitly say that, it should be in the previous section already.

```suggestion
/// Add the provided `ObjectManager` at a path via the
```

Adjust the formatting, of course.

zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
...
+/// [`ObjectServer`](crate::ObjectServer), and every object registered under that path is tracked
+/// for you, with `InterfacesAdded` and `InterfacesRemoved` signals emitted automatically.
+///
+/// ```no_run
```
zeenix: Why `no_run` if we have the `block_on` plumbing anyway? I prefer examples to be run as tests so we don't end up regressing on code we advertize.

zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
...
+/// # use std::error::Error;
+/// use zbus::{fdo::ObjectManager, interface, Connection};
+/// # use async_io::block_on;
```
zeenix: Please use `zbus::block_on` so this doesn't require `async-io`.

zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
...
+///     name: String,
+/// }
+///
+/// #[interface(name = "net.example.ExampleService1.Item")]
+/// impl Item {
+///     #[zbus(property)]
+///     async fn name(&self) -> String {
+///         self.name.clone()
+///     }
+/// }
```
zeenix: As per our coding conventions, please follow top-down layout, so usage befor definiton for items of the same visibility. Also, in this case the focus the ObjectManager so that should appear as early as possible. The specific interface is not very interesting.

zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
...
+///
+/// # block_on(async {
+/// let connection = Connection::session().await?;
+/// let object_server = connection.object_server();
+///
+/// // Register the object manager up front, then add the objects it should manage below its path.
+/// object_server
+///     .at("/net/example/ExampleService1", ObjectManager)
```
zeenix: Let's stick to the project conventions and use `org.freedesktop.zbus.` (translating to `/org/freedesktop/zbus/` for the object path here) as the reverse-domain. This is an **object** path/interface, so `ExampleObject1` would be more appropriate.

zbus/src/fdo/object_manager.rs
```diff
@@ -34,6 +34,48 @@ pub type ManagedObjects =
...
+///     .await?;
+/// object_server
+///     .at(
+///         "/net/example/ExampleService1/item1",
```
zeenix: same here.

zeenix (changes requested):
LGTM otherwise. Thanks!

zeenix (approved):

PR #1929, opened 2026-08-27
✨ xmlgen: Maintain definition order

z33ky:
I found it a bit odd that `zbus-xmlgen` reorders the fields and did something about it. By default it does keep the previous behavior, but it can now keep the file ordering with a `--file-order` flag to `zbus-xmlgen`, or if used by code then the `CodeGenerator::with_sorted()` method.

I don't believe the changes have a noticeable impact on performance. This is a breaking change though, considering the public interface of `zbus_xml::Interface` changed in a non-backwards compatible way.

I've also considered renaming the `Interface::methods()`-like methods to `iter_methods()` to make in clear from the name that an iterator is returned. Let me know if this is something you'd prefer - if you're interested in taking this change at all.

---
The --file-order flag, or for the CodeGenerator the sorted flag, toggles whether to sort elements by type and alphabetically, or whether to keep the order as defined by the XML specification.

### DIFF
--- a/zbus/tests/iface_and_proxy/client.rs
+++ b/zbus/tests/iface_and_proxy/client.rs
@@ -70,13 +70,12 @@ pub async fn my_iface_test(conn: Connection, event: Event) -> zbus::Result<u32> 
         .find(|i| i.name() == "org.freedesktop.MyIface")
         .unwrap();
     // Test if the r# prefix for the keyword was removed
-    assert!(my_iface.methods().iter().any(|m| m.name() == "Type"));
-    assert!(my_iface.properties().iter().any(|p| p.name() == "Let"));
-    assert!(my_iface.signals().iter().any(|s| s.name() == "Match"));
+    assert!(my_iface.methods().any(|m| m.name() == "Type"));
+    assert!(my_iface.properties().any(|p| p.name() == "Let"));
+    assert!(my_iface.signals().any(|s| s.name() == "Match"));
     assert_eq!(
         my_iface
             .methods()
-            .iter()
             .find(|m| m.name() == "RawIdentifierParameter")
             .unwrap()
             .args()
--- a/zbus_xml/src/lib.rs
+++ b/zbus_xml/src/lib.rs
@@ -436,18 +436,8 @@ pub struct Interface<'a> {
     #[serde(rename = "@name", borrow)]
     name: InterfaceName<'a>,
 
-    #[serde(rename = "method", default)]
-    methods: Vec<Method<'a>>,
-    #[serde(rename = "property", default)]
-    properties: Vec<Property<'a>>,
-    #[serde(rename = "signal", default)]
-    signals: Vec<Signal<'a>>,
-    #[serde(rename = "annotation", default)]
-    annotations: Vec<Annotation>,
-    #[serde(skip)]
-    docstring: Option<String>,
-    #[serde(skip)]
-    telepathy_types: Vec<telepathy::TypeDef>,
+    #[serde(default)]
+    children: Vec<Child<'a>>,
 }
 
 impl<'a> Interface<'a> {
@@ -458,22 +448,46 @@ impl<'a> Interface<'a> {
 
     /// Returns the interface methods.
     pub fn methods(&self) -> impl Iterator<Item = &Method<'a>> {
-        self.methods.iter()
+        self.children.iter().filter_map(|child| {
+            if let Child::Method(m) = child {
+                Some(m)
+            } else {
+                None
+            }
+        })
     }
 
     /// Returns the interface signals.
     pub fn signals(&self) -> impl Iterator<Item = &Signal<'a>> {
-        self.signals.iter()
+        self.children.iter().filter_map(|child| {
+            if let Child::Signal(s) = child {
+                Some(s)
+            } else {
+                None
+            }
+        })
     }
 
     /// Returns the interface properties.
     pub fn properties(&self) -> impl Iterator<Item = &Property<'a>> {
-        self.properties.iter()
+        self.children.iter().filter_map(|child| {
+            if let Child::Property(p) = child {
+                Some(p)
+            } else {
+                None
+            }
+        })
     }
 
     /// Return the associated annotations.
     pub fn annotations(&self) -> impl Iterator<Item = &Annotation> {
-        self.annotations.iter()
+        self.children.iter().filter_map(|child| {
+            if let Child::Annotation(a) = child {
+                Some(a)
+            } else {
+                None
+            }
+        })
     }
 
     /// Return the content of the Telepathy `tp:docstring` extension element, if any.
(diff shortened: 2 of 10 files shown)


zbus_xmlgen/src/lib.rs
```diff
@@ -303,22 +327,33 @@ impl<'i> CodeGenerator<'i> {
...
+            for signal in &signals {
+                Self::write_signal(&mut *w, signal, &types)?;
+            }
 
-        let mut props = iface.properties().collect::<Vec<_>>();
-        props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
-        for p in props {
-            Self::write_property(&mut *w, p, &types)?;
+            let mut props = iface.properties().collect::<Vec<_>>();
+            props.sort_by(|a, b| a.name().partial_cmp(&b.name()).unwrap());
```
zeenix: Seems like all we need to do is to remove the sorting here all together (i-e just the `else` block and remove the rest). I looked into the history and found out that sorting was just there from the beginning and nobody just thought of this.

zbus_xml/src/lib.rs
```diff
@@ -327,6 +327,102 @@ impl Signal<'_> {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub enum Child<'a> {
```
zeenix: As per our contributing guide, usage before definition please. So all of this moves down after `Interface` impls.

zeenix (changes requested):
Apart from that one nitpick, 2 more regarding the commit messages:

* The last commit is not exactly a breaking change. We never guaranteed any specific order for generated code and there is no reason people should expect it untill we do so 💥 isn't the best prefix for it.
* Please don't break lines just to start a new sentence. I know it's pretty trendy these days but it's grammatically incorrect and it's benefits are completely subjective.

zbus_xml/src/lib.rs
z33ky: FYI the other types in this file also seem to be defined before usage.

zbus/tests/iface_and_proxy/client.rs
zeenix, line 1: > ♻️ xml: Return Iterators to access Interface elements

This commit is changing and (even breaking) the public API, isn't it? So not at all a refactor and the commit emoji will make it get completely ignored by our release-plz filters.

Actually do we really need to change existing methods? Can't we offer additional variants that return iterators instead?

zbus_xml/src/lib.rs
```diff
@@ -448,6 +448,11 @@ impl<'a> Interface<'a> {
         self.name.as_ref()
     }
 
+    /// Returns the child elements.
+    pub fn children(&self) -> &[Child<'a>] {
```
zeenix: I think this ended up in the wrong commit.

zeenix (changes requested):

zbus_xml/src/lib.rs
zeenix: Yeah, the rule was adopted later and admittedly I'm not always able to catch this when reviewing PRs.

zbus/tests/iface_and_proxy/client.rs
z33ky: > This commit is changing and (even breaking) the public API, isn't it? So not at all a refactor and the commit emoji will make it get completely ignored by our release-plz filters.

True.

> Actually do we really need to change existing methods? Can't we offer additional variants that return iterators instead?

If you want to keep the previous methods, you need contiguous storage for each child type (i.e. the previously existing `Vec`s). `Interface` *could* store both, the old per-type `Vec`s plus the `Vec<Child>`, but that seems odd to me.

zeenix: > > Actually do we really need to change existing methods? Can't we offer additional variants that return iterators instead?
> 
> If you want to keep the previous methods, you need contiguous storage for each child type (i.e. the previously existing `Vec`s). `Interface` _could_ store both, the old per-type `Vec`s plus the `Vec<Child>`, but that seems odd to me.

Ah ok. Then it's good.

zbus_xml/src/lib.rs
z33ky: I could move its introduction to "Change Interface to store a single list of child elements", but it is first used in "Maintain definition order".

zeenix: It obviously doens't belong to either of them. I was thinking it belong to the first commit but this is additive and I only noticed now that this one is inconsistent with the change to iterators. How about splitting this into own commit just before the last one and have it return an iterator?

zbus_xml/src/lib.rs
```diff
@@ -484,40 +498,145 @@ impl<'a> Interface<'a> {
...
         }
-        for annotation in &self.annotations {
-            annotation.write_xml(w)?;
+        .map(move |()| Ok(()))
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum MissingWriteXmlError<'a> {
+    Docstring(&'a String),
```
zeenix: It's very strange to return a ref to an owned string. Why can't this be `&'a str`?

zbus_xml/src/lib.rs
```diff
@@ -484,40 +498,145 @@ impl<'a> Interface<'a> {
...
+            Self::TelepathyTypeDef(t) => return Ok(Err(t.into())),
         }
-        for annotation in &self.annotations {
-            annotation.write_xml(w)?;
+        .map(move |()| Ok(()))
+    }
+}
+
+#[derive(Debug, PartialEq)]
+enum MissingWriteXmlError<'a> {
```
zeenix: I typically prefer one unififed error enum per crate and our `Error` here is quite slimt anyway so please just add any new variants that might be needed there instead.

zbus_xml/src/lib.rs
```diff
@@ -448,6 +448,11 @@ impl<'a> Interface<'a> {
         self.name.as_ref()
     }
 
+    /// Returns the child elements.
```
zeenix, line 449: commit shortlog should have `xml` prefix, to be consistent with your own choice here. 😉

zeenix (changes requested):
We're getting close to a merge, I think.

zbus_xml/src/lib.rs
z33ky: Note this error is internal. Still want to have it in the public `zbus_xml::Error` enum?

zbus_xml/src/lib.rs
z33ky: Oh yeah. I can also do "zx". Does zbus_xmlgen also use "zx"?

zbus_xml/src/lib.rs
```diff
@@ -482,40 +498,145 @@ impl<'a> Interface<'a> {
...
         write!(w, ">")?;
-        for method in &self.methods {
-            method.write_xml(w)?;
-        }
-        for property in &self.properties {
-            property.write_xml(w)?;
+        for child in &self.children {
+            // Ignore MissingWriteXmlError, which is merely returned for element types
+            // that don't have a write_xml() method, which we just want to skip.
+            let _: std::result::Result<(), MissingWriteXmlError<'_>> = child.write_xml(w)?;
```
zeenix: I'm really confused by this:

1. Firstly, the error type is being completely ignored so it (along with all it's impls) become completely useless then.
2. There are 2 variants of the error, but the comment here only mentions 1? 
3. An element type not writing xml, sounds like something that's normal so it shouldn't result in an error.

zbus_xml/src/lib.rs
zeenix: Oh, I had missed that it's not pub.

zbus_xml/src/lib.rs
zeenix: Using `zx` is fine for both, just use it consistently for all commits. :)

zbus_xml/src/lib.rs
z33ky: 1&3: I wasn't sure how I can properly model `Child::write_xml()`, I would find it weird if it simply doesn't do anything on some variants. Perhaps `Child::maybe_write_xml()` would be more appropriate.

2: The comment makes no mention of a specific variant of the error.

zeenix: > I wasn't sure how I can properly model `Child::write_xml()`

I think that's ok and I don't think we need to reflect that in the name. Think of it as an opportunity for the type to write xml, if it wants to. If you like, we can have it return a boolean or even an enum (just not an error).

> 2: The comment makes no mention of a specific variant of the error.

Not by name but it's talking about only one possibility, which is the one variant.

zbus_xml/src/xml.rs
```diff
@@ -291,12 +288,7 @@ fn interface<'i>(
 
     Ok(Interface {
         name,
-        methods,
-        properties,
-        signals,
-        annotations,
-        docstring,
-        telepathy_types,
+        children: interface_children,
```
zeenix, line 291: nit: can be simpler if we just name `interface_children` to `children`.

zbus_xml/src/lib.rs
```diff
@@ -482,40 +496,110 @@ impl<'a> Interface<'a> {
...
+    #[serde(borrow)]
+    Method(Method<'a>),
+    #[serde(borrow)]
+    Property(Property<'a>),
+    #[serde(borrow)]
+    Signal(Signal<'a>),
+    Annotation(Annotation),
+    Docstring(String),
+    #[serde(skip)]
+    TelepathyTypeDef(telepathy::TypeDef),
```
zeenix, line 551: These 2 are not really children so a bit strange to have them here. Besides we don't write neither of them so what's the usecase of having them here? Having them aslo means that our check of "if children present" becomes more inefficient than it would otherwise be.

zeenix (changes requested):
sorry, found one potentially big issue.

zbus_xml/src/lib.rs
zeenix: 🤖 ChatGPT GPT-6 Astra Pro on behalf of zeenix: **[P2] This also introduces a Serde serialization failure.** `#[serde(skip)]` on an enum variant returns an error when that variant is serialized; it does not omit the item from the surrounding `Vec`. Consequently, parsing the existing `telepathy_type_definitions` test input and calling `serde_json::to_string(&node)` now fails at `Child::TelepathyTypeDef`. Previously, the skipped `Interface::telepathy_types` field was simply omitted.

Keeping these definitions outside the serializable children, as suggested above, avoids this regression; otherwise the serializer needs to filter them explicitly. Please cover this with a direct Serde test—`Node::to_writer` takes a separate path and will not catch it.

zbus_xml/src/lib.rs
```diff
@@ -436,18 +436,8 @@ pub struct Interface<'a> {
...
-    properties: Vec<Property<'a>>,
-    #[serde(rename = "signal", default)]
-    signals: Vec<Signal<'a>>,
-    #[serde(rename = "annotation", default)]
-    annotations: Vec<Annotation>,
-    #[serde(skip)]
-    docstring: Option<String>,
-    #[serde(skip)]
-    telepathy_types: Vec<telepathy::TypeDef>,
+    #[serde(default)]
```
zeenix, line 439: 🤖 ChatGPT GPT-6 Astra Pro on behalf of zeenix: **[P2] Preserve the Serde mapping for interface members.** The retained `Deserialize` implementation now recognizes a `children` field, not the existing `<method>`, `<signal>`, `<property>`, or `<annotation>` elements. Consequently, deserializing `<node><interface name="org.example.Test"><method name="Ping"/></interface></node>` with `quick_xml::de::from_str::<Node<'_>>` ignores `method` and defaults `children` to an empty vector instead of producing the `Ping` method.

Please map this collection to the XML child elements (e.g. `$value` with lowercase variant tag names, or an equivalent custom implementation) and add a direct Serde regression test. The existing test named `serde` only exercises `Node::from_reader`/`to_writer`, which now use the hand-written parser/writer and do not validate these derives.

zeenix (commented):
🤖 ChatGPT GPT-6 Astra Pro on behalf of zeenix: Reviewed all four commits at `c184911`. The code-generation ordering change looks consistent with the agreed behavior, but there are two P2 regressions in the retained Serde implementations: interface members are silently lost when deserializing standard introspection XML, and interface-level Telepathy definitions make serialization fail. The first is detailed inline; the second is added to the existing Telepathy-children thread.

Commit hygiene: `2dc5a55` still introduces the public `Child` API and changes `Node::to_writer` ordering under a ♻️ refactor title. Please make `Child` public in the accessor commit (`0ff6516`) and accurately describe/categorize the observable XML-writer behavior change. The final commit changes code generation, not parsing as its message currently says.

Current CI is green. This was a source review; no local tests were run because this environment has no Rust toolchain.

zbus_xml/src/xml.rs
zeenix: 🤖 ChatGPT GPT-6 Astra Pro on behalf of zeenix: Renaming the local also requires qualifying the `children(...)` parser-helper call immediately below as `self::children(...)`; otherwise the new local shadows the function.

PR #1930, opened 2026-08-27
🚚 zb,zm,zu: Rename zvariant_utils to zbus_utils

zeenix:
## Summary

`zvariant_utils` becomes `zbus_utils`, versioned 6.0.0 with the rest of the workspace. Nothing
called zvariant survives in this repo after #1925, so the shared plumbing crate should not carry
the name either.

There is no compatibility shim: 6.0 is unreleased, so no published zbus ever depended on this
crate under the old name, and the rename costs existing zbus users nothing.

## What this means for zgvariant

zgvariant reaches `Signature` through this crate, which is what makes it the *same* type as
zbus's `zbus::wire::Signature`. Following the rename is therefore a major release for it, and
z-galaxy/zgvariant gets a 2.0 that depends on `zbus_utils`.

Until then the sharing is broken by construction, and the docs now say so: a graph mixing
zgvariant 1.x (or zvariant 5) with zbus 6 carries two unrelated `Signature` types, alongside the
two `Type`s and `Value`s the upgrade chapter already described. The second commit corrects that
claim in the upgrade chapter, the README and AGENTS.md.

`Format`, its `gvariant` feature and `Signature::alignment` all stay in the renamed crate — that
is exactly the plumbing zgvariant needs — and the `dbus_maybe_rejection` coexistence test is
unaffected, since it turns on `zbus_utils/gvariant` and checks that zbus still rejects
maybe-bearing signatures.

## Testing

`cargo --locked check --workspace --all-targets`; CI's clippy invocation with `-D warnings`; the
D-Bus suite; the wire-only build including its doctests; `zbus_macros`, `zbus_utils`, `zbus_xml`
and `zbus_xmlgen`; the zgvariant-coexistence run
(`-p zbus --no-default-features --features zbus_utils/gvariant`); `-D warnings` rustdoc for zbus
at both feature ends and for `zbus_utils`; `mdbook build`; and both commits checked out on their
own. `cargo tree` for the wire-only build resolves `zbus_utils` and no `zvariant_utils`.

Generated by Claude Fable 5 (claude-fable-5).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01C77ZXwTqmd3qSpSXgw2RfN

PR #1931, opened 2026-08-27
💥 zb,zm,zu: Drop the Format dispatch

zeenix:
## Summary

`Format` disappears from the workspace. `zbus::wire` only ever encodes the D-Bus format and the
sibling zgvariant project only ever encodes GVariant, so neither picks a format at runtime —
threading one through the API bought nothing and cost zbus a set of unreachable match arms and
asserts guarding a state it could not reach.

Rebased onto main now that #1930 (the `zbus_utils` rename) has merged.

## What changed

- `Signature::alignment(&self, format: Format)` is replaced by its two former private helpers,
  now public: `alignment_dbus()` and `alignment_gvariant()` (the latter still behind the
  `gvariant` feature). Callers say which padding rules they mean.
- `zbus_utils::serialized` is deleted — it held nothing but `Format` and its `Display`.
- `wire::serialized::Context` loses its format: `Context::new(endian, position)` is the
  constructor, `Context::format()` is gone, and `Context::new_dbus` remains as a deprecated
  one-liner (removed in 7.0) so a zvariant 5 port compiles with a warning instead of an error.
- The `Basic` trait's `alignment` takes no argument any more.
- Everything that existed only to handle the impossible case is gone: the
  `assert_eq!(ctxt.format(), Format::DBus)` checks in the D-Bus serializer and deserializer, the
  `#[allow(unreachable_patterns)]` arms that a non-D-Bus format made necessary, and the
  `match ctxt.format()` in `wire/ser.rs`.

`Signature::Maybe` and the `gvariant` feature stay: they are what lets zbus reject a
maybe-bearing signature when zgvariant is in the same graph, which the `dbus_maybe_rejection`
test still checks.

## Testing

`check --workspace --all-targets`; CI's clippy with `-D warnings`; the D-Bus suite; the wire-only
build and its doctests; `zbus_macros`, `zbus_utils`, `zbus_xml`, `zbus_xmlgen`; the coexistence
run with `zbus_utils/gvariant`; `-D warnings` rustdoc for zbus at both feature ends and for
`zbus_utils`; `mdbook build`; and each commit checked out on its own, including the separate
`zbus/fuzz` workspace.

https://claude.ai/code/session_01C77ZXwTqmd3qSpSXgw2RfN


zeenix (commented):
🤖 Chatgpt GPT-5.6 on behalf of zeenix: No implementation issues found in this PR’s own two commits. The D-Bus and GVariant alignment paths remain distinct, the now-impossible runtime format branches are removed consistently, `Context::new_dbus` retains the intended deprecated migration path, and call sites/tests/fuzzing were updated. All relevant CI jobs pass; the semver-check failure is the same expected unpublished-`zbus_utils` failure as #1930. Merge and rebase #1930 first as planned.

PR #1932, opened 2026-08-28
🚩 zb: Make stream builder APIs additive

zeenix:
**Attribution:** This PR title and description were prepared by Codex and published through the connected `zeenix` GitHub account.

The stream builder API still changed parameter types under Cargo feature unification: `unix_stream`, `tcp_stream`, and `vsock_stream` selected one backend at compile time. Since zbus 6 is already an API break, this follows the split API discussed in #1824 instead of carrying that 5.x compatibility compromise forward.

## Changes

- add explicit `async_io_*` and `tokio_*` Unix, TCP, and VSOCK constructors while the old APIs still coexist;
- test that every constructor remains available under unified features;
- remove the ambiguous constructors in a separate breaking-change commit and migrate in-tree callers;
- keep address-created async connections runtime-selected: Tokio inside an active Tokio runtime, and async-io otherwise;
- make an explicitly supplied stream select the I/O backend named by its constructor while the active runtime independently selects the task executor;
- keep an async-io connection's internal executor on `async_io::block_on` in dual-runtime builds, with regression coverage;
- document the 6.0 stream-constructor migration;
- run the root CI test with `--all-features` and retain a Tokio-free zbus suite for async-io-only arms.

## Testing

- `cargo +nightly fmt --all -- --check`
- `RUSTFLAGS='-D warnings' cargo clippy --locked`
- focused all-feature Clippy for `builder_feature_additivity` and `builder_message_stream`
- `RUSTFLAGS='-D warnings' cargo test --locked --all-features --no-run`
- Tokio-free release test build with `async-io` and `blocking-api`, without zbus's `tokio` feature
- `builder_message_stream` under async-io only, Tokio only, and both backends
- additive signature checks under each single and combined Unix/TCP/VSOCK feature set
- dual-runtime selector and async-io internal-driver regression tests
- rustdoc with async-io only, Tokio only, and all features
- dual-backend compile checks for macOS and Windows targets

### DIFF
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -1,8 +1,3 @@
-#[cfg(not(feature = "tokio"))]
-use std::net::TcpStream;
-#[cfg(feature = "tokio")]
-use tokio::net::TcpStream;
-
 // Feature-independent stream types for the `async_io_*_stream` builders.
 #[cfg(feature = "async-io")]
 use std::net::TcpStream as AsyncIoTcpStream;
@@ -118,24 +113,6 @@ impl<'a> Builder<'a> {
     #[cfg(feature = "async-io")]
     pub fn async_io_tcp_stream(stream: AsyncIoTcpStream) -> Self {
         Self(crate::connection::Builder::async_io_tcp_stream(stream))
-    }
-
-    /// Create a builder for a connection that will use the given TCP stream.
-    ///
-    /// This method expects a
-    /// [`tokio::net::TcpStream`](https://docs.rs/tokio/latest/tokio/net/struct.TcpStream.html).
-    /// Without the `tokio` feature it accepts a [`std::net::TcpStream`] instead, but that form is
-    /// deprecated in favor of [`async_io_tcp_stream`](Self::async_io_tcp_stream).
-    #[cfg_attr(
-        not(feature = "tokio"),
-        deprecated(
-            since = "5.19.0",
-            note = "Use `async_io_tcp_stream` to avoid a build failure if the `tokio` feature gets enabled"
-        )
-    )]
-    #[allow(deprecated)] // forwards to the async builder's equally-deprecated `tcp_stream`
-    pub fn tcp_stream(stream: TcpStream) -> Self {
-        Self(crate::connection::Builder::tcp_stream(stream))
     }
 
     /// Create a builder for a connection that will use the given TCP stream with Tokio.
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -85,12 +85,14 @@ impl<'a> Builder<'a> {
         crate::connection::Builder::address(address).map(Self)
     }
 
-    /// Create a builder for a connection that will use the given unix stream.
+    /// Create a builder for a connection that will use the given unix stream with `async-io`.
     ///
     /// The stream is a [`std::os::unix::net::UnixStream`] (or [`uds_windows::UnixStream`] on
-    /// Windows).
+    /// Windows). Enabling the `tokio` feature does not change the stream type accepted by this
+    /// method. To use a [Tokio Unix stream][tokio-unix] instead, see `tokio_unix_stream`.
     ///
     /// [`uds_windows::UnixStream`]: https://docs.rs/uds_windows/latest/uds_windows/struct.UnixStream.html
+    /// [tokio-unix]: https://docs.rs/tokio/latest/tokio/net/struct.UnixStream.html
     #[cfg(all(any(unix, windows), feature = "async-io"))]
     pub fn async_io_unix_stream(stream: AsyncIoUnixStream) -> Self {
         Self(crate::connection::Builder::async_io_unix_stream(stream))
--- a/zbus/src/blocking/connection/builder.rs
+++ b/zbus/src/blocking/connection/builder.rs
@@ -1,13 +1,7 @@
 #[cfg(not(feature = "tokio"))]
 use std::net::TcpStream;
-#[cfg(all(unix, not(feature = "tokio")))]
-use std::os::unix::net::UnixStream;
 #[cfg(feature = "tokio")]
 use tokio::net::TcpStream;
-#[cfg(all(unix, feature = "tokio"))]
-use tokio::net::UnixStream;
-#[cfg(all(windows, not(feature = "tokio")))]
-use uds_windows::UnixStream;
 
 // Feature-independent stream types for the `async_io_*_stream` builders.
 #[cfg(feature = "async-io")]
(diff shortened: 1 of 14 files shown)


zbus/src/blocking/connection/builder.rs
```diff
@@ -85,66 +74,56 @@ impl<'a> Builder<'a> {
...
-    /// deprecated in favor of [`async_io_tcp_stream`](Self::async_io_tcp_stream).
-    #[cfg_attr(
-        not(feature = "tokio"),
-        deprecated(
-            since = "5.19.0",
-            note = "Use `async_io_tcp_stream` to avoid a build failure if the `tokio` feature gets enabled"
-        )
-    )]
-    #[allow(deprecated)] // forwards to the async builder's equally-deprecated `tcp_stream`
-    pub fn tcp_stream(stream: TcpStream) -> Self {
```
zeenix, line 146: Split the removal changes into a separate commit. The existing API can co-exist with the new one and then choose more specific emojis for each commit.

zeenix (commented):
## Review

The API split is the right call and it lands cleanly. I checked the things most likely to be wrong and they're fine:

- **No stale call sites.** `git grep` for `unix_stream(`/`tcp_stream(`/`vsock_stream(` outside the `async_io_`/`tokio_` prefixes returns nothing across the tree — src, tests, book, examples.
- **The additivity invariant holds for the constructors.** `target_connect` dispatches purely on the `Target` variant, with no `use_tokio()` involvement, so a supplied stream's socket backend really is fixed by the constructor that took it. Every new constructor is positively gated on its own feature.
- **The migration table is accurate**, including the `vsock`/`tokio-vsock` feature column (`vsock` implies `async-io`, so the `#[cfg(feature = "vsock")]` widening on the `AsyncIoVsockStream` arm is correct), and the "no VSOCK on the blocking builder" note matches the code.
- **"Five more things break in 6.0"** — there are exactly five `###` subsections under *Other changes in 6.0*. 👍
- The red `semver-checks` is `zbus_utils not found in registry (crates.io)`, a pre-existing main-branch condition from the unpublished workspace crate. Nothing to do with this diff.

Three things I'd like to see addressed. None of them blocks the API change.

---

### 1. `bb381b7` changes observable behaviour, ships untested, and isn't in the upgrade guide

Switching the driver thread from `crate::utils::block_on` to `async_io::block_on` is the right fix — the old code entered zbus's hidden global tokio runtime purely as a side effect of the `tokio` feature being compiled in. But it's a behaviour change that users can observe, and the PR treats it as a silent one-liner.

Reproduced by A/B (`--features async-io,tokio,p2p`, connection built outside a tokio runtime, probe task spawned on `conn.executor()`):

```
HEAD a591021:                 PROBE in_tokio=false thread=Some("zbus::Connection executor")
line 772 -> utils::block_on:  PROBE in_tokio=true  thread=Some("zbus::Connection executor")
```

That thread runs `#[interface]` dispatch. So in a dual-backend build where the connection was created outside a tokio runtime — exactly the configuration this PR exists to make survivable — a handler that calls `tokio::spawn`, `tokio::time::*` or `tokio::fs` used to work and now panics with *"there is no reactor running"*. Since 5.19 that context was an accident, so removing it is correct, but it's still a break worth a bullet in *Other changes in 6.0*, next to the five that are already there.

It also has no test. `p2p_tests::unix_p2p_async_io_backend` only asserts `needs_internal_driver()`, which is `self.async_io.is_some()` and is blind to which `block_on` drives the thread. I reverted line 772 and the entire suite stays green — including both new `builder_message_stream` variants and `builder_feature_additivity`. Nothing added by this PR would catch someone later "deduplicating" that call back to `crate::utils::block_on`, which is the obvious-looking cleanup.

Suggestion: assert `tokio::runtime::Handle::try_current().is_err()` from a task spawned on `conn.executor()` in `unix_p2p_async_io_backend` (or a sibling gated on both features), plus a comment at the call site saying why it must not be deduplicated.

### 2. `--all-features` leaves the new async-io test arms compiled by nothing

The replaced step (`--features uuid,url,time,chrono,option-as-array,vsock,bus-impl`) was the only job in the whole workflow that compiled `p2p`/`bus-impl` with `tokio` **off**. After the change, no job does — on any platform:

| job / step | `p2p`? | `tokio`? |
| --- | --- | --- |
| `-- basic_connection` (default features) | no | no |
| `--all-features` | yes | **yes** |
| `-p zbus --no-default-features --features tokio-vsock` | no | yes |
| `-p zbus --features tokio,p2p,vsock` | yes | yes |
| tokio doctest / wire-format runs | no | — |
| `windows_test`: `cargo --locked test` | no | no |

(`cargo tree --workspace -e features` confirms no workspace member enables `zbus/p2p` or `zbus/tokio` by default, so the default-feature runs really don't reach that code.)

The awkward part is that this diff *adds* the arms that are now unreachable — the `#[cfg(not(feature = "tokio"))] let builder = Builder::async_io_unix_stream(p0);` blocks in `e2e.rs:64`, `issue_1003.rs:28`, `issue_813.rs:61` and `blocking/connection/mod.rs:355,369` are written but never compiled by CI. Concretely, async-io loses its only execution coverage for FD-passing right after BEGIN (`issue_813`) and EXTERNAL auth (`issue_1003`), both unix-only, so `windows_test` can't stand in.

Not a total loss — `connection_closed.rs`, `unix_p2p_async_io_backend` and the new `..._pipelined_hello_async_io` pin async-io explicitly and do run under `--all-features`. But keeping one tokio-free full-suite run alongside `--all-features` is cheap:

```yaml
  cargo --locked test --release --verbose -p zbus \
    --features uuid,url,time,chrono,option-as-array,vsock,bus-impl \
    -- --skip fdpass_systemd
```

Longer term, the pattern `builder_message_stream.rs` now uses — a pair of tests gated on `feature = "async-io"` and `feature = "tokio"` — is strictly better than `cfg(not(feature = "tokio"))` arms, because a dual-feature run then exercises both backends instead of silently picking one. Worth applying to the other four sites.

### 3. Nit: "instead" in the README/FAQ wording

> Supplying an existing stream through a runtime-prefixed `connection::Builder` constructor explicitly selects its I/O backend instead.

Strictly accurate — the constructor does fix the *I/O backend*. But "instead", right after a sentence about run-time runtime selection, reads as though it overrides that selection wholesale, when the internal task executor is still chosen by `Executor::new()` from whether a tokio runtime is driving the building thread. A parenthetical like *"(the internal task executor is still chosen by the active runtime)"* would close the gap.

---

### Out of scope, for follow-up issues

Both are pre-existing on `ffb2c72` and untouched here, but this PR is what establishes the invariant they violate:

- `zbus/src/connection/socket/command.rs:76,115` — `impl ReadHalf for tokio::process::ChildStdout` and `impl WriteHalf for tokio::process::ChildStdin` are gated `#[cfg(all(feature = "tokio", not(feature = "async-io")))]`. Enabling `async-io` (directly, or transitively via `vsock`) *removes* public trait impls. That's the same feature-unification hazard the stream constructors just stopped having; reproduced as `E0277` at both HEAD and the merge base.
- `zbus/src/address/transport/mod.rs:74` — `Transport::Vsock`'s doc comment still describes feature-dependent stream typing, which is now stale.

---

zbus/src/blocking/connection/builder.rs
zeenix: **Attribution:** Written by Codex and posted through the connected `zeenix` GitHub account.

Addressed. The rewritten history now separates each transport and stage:

- additions: Tokio Unix `a823d7d6`, Tokio TCP `9e11d873`, async-io VSOCK `c624c05f`, and Tokio VSOCK `da998e7c`;
- signature tests: Unix `761d7ac7`, TCP `e349a462`, and VSOCK `5985d9ac`;
- message-stream runtime coverage: `9377fbdc`;
- caller migrations: Unix `e077e4d6`, TCP `e7d0b051`, and VSOCK `48988ba4`;
- removals: Unix `e1458b50`, TCP `af44028b`, and VSOCK `358ee42d`.

Documentation, executor behavior, and both CI changes are separate commits as well. Every commit-message line is at most 74 characters.

book/src/faq.md
```diff
@@ -166,14 +166,19 @@ feature:
...
-Disabling `async-io` is recommended but not required. When both features are enabled, zbus picks the
-runtime at run time: it uses tokio when a tokio runtime is driving the current thread, and falls back
-to `async-io` otherwise. This keeps the features additive, so an `async-io`-based application keeps
-working even when another crate in the workspace enables zbus's `tokio` feature.
-
-This run-time selection applies to the async API. The blocking API (`zbus::blocking`) drives its
-connections through its own `block_on`, which uses tokio whenever the `tokio` feature is enabled, so
-those connections always run on tokio when that feature is on.
+Disabling `async-io` is recommended but not required. When both features are enabled, zbus chooses a
+connection's internal task executor at run time: tokio when the connection is built while a tokio
```
zeenix: Why was it necessary to modify these lines just to change from "picks" to "chooses". Remove all unnecessary changes from there and other commits.

zeenix: **Attribution:** Written by Codex and posted through the connected `zeenix` GitHub account.

Addressed. I restored the original `picks` wording and all surrounding FAQ/README text, reverted the unrelated README link and `Transport::Vsock` documentation edits, and reduced the connection-guide change to the required constructor rename. The remaining documentation change is the 6.0 stream-constructor migration.

PR #1937, opened 2026-08-30
🐛 zb: Reject non-basic dictionary keys

### LINKED ISSUE #1637: Container dict keys are not rejected
Per the [D-Bus spec](https://dbus.freedesktop.org/doc/dbus-specification.html#container-types) (emphasis added):

>  A DICT_ENTRY works exactly like a struct, but rather than parentheses it uses curly braces, and it has more restrictions. The restrictions are: it occurs only as an array element type; it has exactly two single complete types inside the curly braces; **the first single complete type (the "key") must be a basic type rather than a container type**. Implementations must not accept dict entries outside of arrays, must not accept dict entries with zero, one, or more than two fields, and **must not accept dict entries with non-basic-typed keys**. A dict entry is always a key-value pair. 

However, if you declare a property that returns e.g. a `HashMap<Vec<u8>, u16>` (that is, `a{ayq}`), zbus does not reject this non-conformant type and instead the dbus implementation (in my case dbus-broker) will reject it and disconnect from the zbus application. The reason why is totally opaque from both ends:

zbus:
```
WARN monitor_name_lost{name=org.blah}: zbus::connection: Failed to parse `NameLost` signal: I/O error: failed to read from socket
```

dbus-broker:
```
dbus-broker[2332]: Peer :1.4191 is being disconnected as it sent a message with an invalid body.
```

It took me quite a while trying to figure out what was going on and ended up diving into the spec to figure out what was going on. Saving other people from this pain would likely be appreciated.

zeenix:
Require dictionary keys to implement `Basic` for `HashMap` and `BTreeMap`
signatures and conversions. This restores the fix reverted for 5.x compatibility
now that zbus 6.0 provides the planned API-break boundary.

Implement `Basic` only for the existing scalar-representation adapters needed
to preserve valid dictionary keys. Make enumflags use its numeric wire
representation for both `Type` and `Basic`.

Add compile-time assertions covering valid and invalid keys across `Type`, `Dict`,
`Value`, and `OwnedValue`.

Closes #1637.

Generated by GPT-5.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1501,6 +1501,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ce2be8dc25455e1f91df71bfa12ad37d7af1092ae736f3a6cd0e37bc7810596"
 
 [[package]]
+name = "static_assertions"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a2eb9349b6444b326872e140eb1cf5e7c522154d69e7a0ffb0fb81c06b37543f"
+
+[[package]]
 name = "statrs"
 version = "0.18.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2189,6 +2195,7 @@ dependencies = [
  "serde_bytes",
  "serde_json",
  "serde_repr",
+ "static_assertions",
  "tempfile",
  "test-log",
  "time",
--- a/zbus/Cargo.toml
+++ b/zbus/Cargo.toml
@@ -169,6 +169,7 @@ chrono = { workspace = true, features = [
     "serde",
     "alloc",
 ], default-features = false }
+static_assertions.workspace = true
 
 [package.metadata.docs.rs]
 # we can't use:
--- a/zbus/src/wire/basic.rs
+++ b/zbus/src/wire/basic.rs
@@ -310,3 +310,122 @@ impl<const CAP: usize> Basic for heapless::String<CAP> {
     const SIGNATURE_CHAR: char = <&str as Basic>::SIGNATURE_CHAR;
     const SIGNATURE_STR: &'static str = <&str as Basic>::SIGNATURE_STR;
 }
+
+////////////////////////////////////////////////////////////////////////////////
+
+macro_rules! impl_basic_with_repr {
+    ($($ty:ty => $repr:ty),+ $(,)?) => {
+        $(
+            impl Basic for $ty {
+                const SIGNATURE_CHAR: char = <$repr as Basic>::SIGNATURE_CHAR;
+                const SIGNATURE_STR: &'static str = <$repr as Basic>::SIGNATURE_STR;
+            }
+        )+
+    };
+}
+
+impl_basic_with_repr!(
+    std::path::Path => str,
+    std::path::PathBuf => str,
+);
+
+impl<T> Basic for [T; 0]
+where
+    T: Type,
+{
+    const SIGNATURE_CHAR: char = u8::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u8::SIGNATURE_STR;
+}
+
+#[cfg(feature = "arrayvec")]
+impl<const CAP: usize> Basic for arrayvec::ArrayString<CAP> {
+    const SIGNATURE_CHAR: char = str::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = str::SIGNATURE_STR;
+}
+
+#[cfg(feature = "camino")]
+impl_basic_with_repr!(
+    camino::Utf8Path => str,
+    camino::Utf8PathBuf => str,
+);
+
+#[cfg(feature = "url")]
+impl_basic_with_repr!(url::Url => str);
+
+#[cfg(feature = "time")]
+impl_basic_with_repr!(
+    time::Month => u8,
+    time::Weekday => u8,
+);
+
+#[cfg(feature = "chrono")]
+impl<Tz> Basic for chrono::DateTime<Tz>
+where
+    Tz: chrono::TimeZone,
+{
+    const SIGNATURE_CHAR: char = str::SIGNATURE_CHAR;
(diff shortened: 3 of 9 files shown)


zbus/src/fdo/dbus.rs
```diff
@@ -70,6 +70,11 @@ pub enum RequestNameReply {
     AlreadyOwner = 0x04,
 }
 
+impl Basic for RequestNameReply {
+    const SIGNATURE_CHAR: char = u32::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u32::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/fdo/dbus.rs
```diff
@@ -97,6 +102,11 @@ pub enum ReleaseNameReply {
     NotOwner = 0x03,
 }
 
+impl Basic for ReleaseNameReply {
+    const SIGNATURE_CHAR: char = u32::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u32::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/fdo/dbus.rs
```diff
@@ -121,6 +131,11 @@ pub enum StartServiceReply {
     AlreadyRunning = 0x02,
 }
 
+impl Basic for StartServiceReply {
+    const SIGNATURE_CHAR: char = u32::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u32::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/match_rule/mod.rs
```diff
@@ -101,6 +101,11 @@ pub struct MatchRule<'m> {
     pub(crate) arg0ns: Option<Str<'m>>,
 }
 
+impl Basic for MatchRule<'_> {
+    const SIGNATURE_CHAR: char = <Str<'_>>::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = <Str<'_>>::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/match_rule/mod.rs
```diff
@@ -518,6 +523,11 @@ impl PathSpec<'_> {
 #[derive(Clone, Debug, Hash, PartialEq, Eq, Serialize, VariantType)]
 pub struct OwnedMatchRule(#[serde(borrow)] MatchRule<'static>);
 
+impl Basic for OwnedMatchRule {
+    const SIGNATURE_CHAR: char = <MatchRule<'_>>::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = <MatchRule<'_>>::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/message/field_code.rs
```diff
@@ -33,3 +33,8 @@ pub(crate) enum FieldCode {
     /// Code for [`Field::UnixFDs`](enum.Field.html#variant.UnixFDs).
     UnixFDs = 9,
 }
+
+impl Basic for FieldCode {
+    const SIGNATURE_CHAR: char = u8::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u8::SIGNATURE_STR;
+}
```
zeenix: unnecessary.

zbus/src/message/header.rs
```diff
@@ -32,6 +32,11 @@ pub enum EndianSig {
     Little = b'l',
 }
 
+impl Basic for EndianSig {
+    const SIGNATURE_CHAR: char = u8::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u8::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/message/header.rs
```diff
@@ -86,6 +91,11 @@ pub enum Type {
     Signal = 4,
 }
 
+impl Basic for Type {
+    const SIGNATURE_CHAR: char = u8::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = u8::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/object_server/dispatch_notifier.rs
```diff
@@ -75,6 +75,14 @@ where
...
 }
 
+impl<R> Basic for ResponseDispatchNotifier<R>
+where
+    R: Basic,
+{
+    const SIGNATURE_CHAR: char = R::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = R::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/guid.rs
```diff
@@ -18,6 +18,11 @@ use serde::{Deserialize, Serialize, de};
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Type, Serialize)]
 pub struct Guid<'g>(Str<'g>);
 
+impl Basic for Guid<'_> {
+    const SIGNATURE_CHAR: char = <Str<'_>>::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = <Str<'_>>::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

zbus/src/guid.rs
```diff
@@ -164,6 +169,11 @@ impl Borrow<str> for Guid<'_> {
 #[derive(Clone, Debug, PartialEq, Eq, Hash, Type, Serialize)]
 pub struct OwnedGuid(#[serde(borrow)] Guid<'static>);
 
+impl Basic for OwnedGuid {
+    const SIGNATURE_CHAR: char = <Guid<'_>>::SIGNATURE_CHAR;
+    const SIGNATURE_STR: &'static str = <Guid<'_>>::SIGNATURE_STR;
+}
+
```
zeenix: unnecessary.

Issue #1943, opened 2026-09-04
zbus_xmlgen: compile-test fixtures fail clippy with `--all-targets`


zeenix:
The generated fixtures in `zbus_xmlgen/tests/data/` are `include!`d by `zbus_xmlgen/tests/compile.rs`, so they get linted whenever clippy is run with `--all-targets`. With current stable clippy that fails:

```
cargo clippy --workspace --all-targets -- -D warnings
```

```
error: very complex type used. Consider factoring parts into `type` definitions
  --> zbus_xmlgen/tests/data/sample_object0.rs:6:15
error: very complex type used. Consider factoring parts into `type` definitions
  --> zbus_xmlgen/tests/data/sample_object0.rs:20:56
error: use of a disallowed/placeholder name `foo`
  --> zbus_xmlgen/tests/data/sample_object0.rs:20:39
error: use of a disallowed/placeholder name `foo`
  --> zbus_xmlgen/tests/data/sample_object0.rs:29:9
```

CI only runs `cargo clippy` without `--all-targets`, so this is not caught there, but it makes local `--all-targets` runs noisy and hides real findings in test targets.

The fixtures are generated from the sample XML, so the names and types come from the input. Options:

* Add `#[allow(clippy::type_complexity, clippy::disallowed_names)]` to the wrapper modules in `tests/compile.rs`, since the fixtures are generated code.
* Or have `zbus_xmlgen` emit an `#[allow(...)]` on generated items, which would also help users who lint generated code.

Noticed while working on #1135.

PR #1944, opened 2026-09-04
🚩 zb: Feature-gate proxy and service API

### LINKED ISSUE #1135: Feature-gate proxy and service API
Turns out that LTO isn't perfect so even when binaries using zbus are built with `lto = fat`, they still end up with API they don't use at all. One main example is busd, where the binary ends up not only with proxy code that it doesn't use at all, this code contributes quite significantly to the size of busd binary. So let's feature gate both proxy and service API so pure services can disable all proxy code and vice versa. 

Unfortunately, these features will have to be default and adding new default features would be a breaking change. So it'll have to wait till 6.0.

This will be part of the #304 efforts.

zeenix:
Fixes #1135.

Pure services (e.g. busd) still ended up carrying all the proxy code even with fat LTO, and vice versa for pure clients. This puts the two halves of the high-level API behind two new default cargo features, both implying `comms`, so that either can be disabled:

* `proxy`: `zbus::proxy`, `zbus::Proxy`, the `#[proxy]` macro, the `fdo::*Proxy` types with their signal streams, and the blocking counterparts.
* `service`: `zbus::object_server`, `zbus::ObjectServer`, the `#[interface]` macro, `Connection::object_server`, `Builder::serve_at`, the standard interface implementations in `fdo`, and the blocking counterparts.

`zbus_macros` gains matching `proxy` and `service` features (default there and forwarded by `zbus`) that gate the two macros. No `zbus_macros` feature influences the generated code: whether the `interface` macro's `proxy(..)` sub-attribute generates a proxy, and whether the `proxy` macro generates the blocking proxies, is decided through hidden `zbus` macros by the features of the `zbus` the code is compiled against, because Cargo unifies the features of host dependencies (proc-macros, build scripts) separately from the target ones. A fixture crate with a service-only `zbus` dependency and a build script depending on `zbus` with `proxy` guards against regressions; CI builds it on its own since a workspace-wide build unifies the features away.

The generated public API is unchanged when both features are enabled (verified by diffing the expanded public signatures against `main`). Everything else in the D-Bus API (connections, messages, match rules, plain `fdo` types and errors) stays behind `comms` alone.

Commits, each building and passing on its own:

1. Two pre-existing clippy warnings in tests, so that the lint job passes.
2. The pre-existing host/target instance of the feature problem for `blocking-api`: a tokio-only crate with `zbus_xmlgen` as a build-dependency fails to compile `zbus` on `main`. The `blocking-api` feature of `zbus_macros` is removed since it has no purpose left.
3. Declare the standard interface proxies in `fdo` separately from the interface implementations, so that either side can be compiled without the other. API unchanged.
4. The `proxy` feature, across both crates.
5. The `service` feature, across both crates, with the fixture.
6. README.
7. Upgrade guide.
8. CI: clippy across all targets on the reduced feature combinations with both runtimes, doc builds and doctests with either half disabled, and the fixture on its own.

The remaining `--all-targets` clippy failures in the `zbus_xmlgen` generated fixtures are tracked in #1943.

https://claude.ai/code/session_015PfUDc6LiN3tqpco7Uq3JZ

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -91,6 +91,21 @@ jobs:
           cargo --locked clippy -p zbus --no-default-features --target x86_64-unknown-netbsd
           cargo --locked clippy -p zbus --no-default-features --target x86_64-pc-windows-gnu
           cargo --locked clippy -p zbus --no-default-features --target aarch64-linux-android
+          # The proxy and service API can be disabled independently. `--all-targets` covers the
+          # tests, examples and benchmarks, whose `required-features` must match what they use.
+          cargo --locked clippy -p zbus --no-default-features --all-targets \
+            --features async-io,blocking-api,p2p
+          cargo --locked clippy -p zbus --no-default-features --all-targets \
+            --features async-io,blocking-api,p2p,proxy
+          cargo --locked clippy -p zbus --no-default-features --all-targets \
+            --features async-io,blocking-api,p2p,service
+          cargo --locked clippy -p zbus --no-default-features --all-targets \
+            --features tokio,p2p,proxy
+          cargo --locked clippy -p zbus --no-default-features --all-targets \
+            --features tokio,p2p,service
+          # The service-only fixture has to be built on its own: in a workspace-wide build, the
+          # features of its `zbus` dependency are unified with those of the other members.
+          cargo --locked clippy -p service_only_fixture
 
   linux_test:
     runs-on: ubuntu-latest
@@ -137,7 +152,7 @@ jobs:
           # Test tokio support.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
             cargo --locked test --release --verbose --tests -p zbus --no-default-features \
-              --features tokio-vsock -- --skip fdpass_systemd
+              --features tokio-vsock,proxy,service -- --skip fdpass_systemd
           # Test tokio and async-io enabled together (additive features): the backend is then
           # chosen at run time. `vsock` is included as it pulls in `async-io` alongside `tokio`.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
@@ -148,7 +163,21 @@ jobs:
           # `--no-default-features` is undone by zbus_xmlgen's default-featured zbus dependency.
           dbus-run-session --config-file /tmp/dbus-session.conf -- \
             cargo --locked test --release --verbose --doc -p zbus --no-default-features \
-              --features tokio connection::Connection::executor
+              --features tokio,service connection::Connection::executor
+          # The doc examples with the proxy or service API disabled, with and without the blocking
+          # API.
+          dbus-run-session --config-file /tmp/dbus-session.conf -- \
+            cargo --locked test --release --verbose --doc -p zbus --no-default-features \
+              --features async-io,proxy
+          dbus-run-session --config-file /tmp/dbus-session.conf -- \
+            cargo --locked test --release --verbose --doc -p zbus --no-default-features \
+              --features async-io,service
+          dbus-run-session --config-file /tmp/dbus-session.conf -- \
+            cargo --locked test --release --verbose --doc -p zbus --no-default-features \
+              --features async-io,blocking-api,proxy
+          dbus-run-session --config-file /tmp/dbus-session.conf -- \
+            cargo --locked test --release --verbose --doc -p zbus --no-default-features \
+              --features async-io,blocking-api,service
           # The wire-format-only build. No D-Bus API, so no session bus is needed.
           cargo --locked test --release --verbose -p zbus --no-default-features
           cargo --locked test --release --verbose -p zbus --no-default-features \
@@ -204,7 +233,7 @@ jobs:
           # The tokio-only build. `-p zbus` is load-bearing: a workspace-wide
           # `--no-default-features` is undone by zbus_xmlgen's default-featured zbus dependency.
           # `--tests` because the doc examples assume the async-io runtime.
-          cargo --locked test -p zbus --no-default-features --features tokio --tests
+          cargo --locked test -p zbus --no-default-features --features tokio,proxy,service --tests
           # The wire-format-only build.
           cargo --locked test -p zbus --no-default-features
           cargo --locked test -p zbus --no-default-features `
@@ -245,6 +274,12 @@ jobs:
         run: cargo --locked doc --all-features -p zbus
       - name: Check wire-format-only documentation build
         run: cargo --locked doc --no-default-features -p zbus
+      - name: Check documentation build with the proxy or service API disabled
+        run: |
+          cargo --locked doc --no-default-features --features async-io,proxy -p zbus
+          cargo --locked doc --no-default-features --features async-io,service -p zbus
+          cargo --locked doc --no-default-features --features async-io,blocking-api,proxy -p zbus
+          cargo --locked doc --no-default-features --features async-io,blocking-api,service -p zbus
       - name: Check zbus_xml documentation build
         run: cargo --locked doc --all-features -p zbus_xml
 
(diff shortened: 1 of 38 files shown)


.github/workflows/rust.yml
```diff
@@ -77,6 +77,12 @@ jobs:
           cargo --locked clippy --target x86_64-unknown-netbsd
           cargo --locked clippy --target x86_64-pc-windows-gnu
           cargo --locked clippy --target aarch64-linux-android
+          # Proxy and service API can be disabled independently.
+          cargo --locked clippy -p zbus --no-default-features --features async-io,blocking-api
+          cargo --locked clippy -p zbus --no-default-features --features async-io,blocking-api,proxy
+          cargo --locked clippy -p zbus --no-default-features --features async-io,blocking-api,service
+          cargo --locked clippy -p zbus --no-default-features --features tokio,proxy
+          cargo --locked clippy -p zbus --no-default-features --features tokio,service
```
zeenix: 🤖 ChatGPT Codex on behalf of Zeeshan: [blocking] These checks only compile the library and miss breakage in the newly supported configurations. On this head, both `RUSTDOCFLAGS="-D warnings" cargo doc -p zbus --no-default-features --features async-io,proxy --no-deps` and the corresponding `service` build fail on links to gated-out APIs. `cargo test -p zbus --doc --no-default-features --features async-io,proxy` also compiles the crate-level README and fails because its service example uses `interface` and `Builder::serve_at`. Please gate or rewrite the affected docs and add reduced-feature doc/doctest coverage when porting this to current `main`.

zeenix (commented):
🤖 ChatGPT Codex on behalf of Zeeshan:

Blocking: this branch has merge base `de33018f` (2026-04-26), is 337 commits behind current `main` (`e1067948`), and GitHub reports it as unmergeable. A local merge produces 18 content-conflict paths. This is not just mechanical drift: current `main` is already at zbus/zbus_macros 6.0.0, has merged zvariant into `zbus::wire`, and has since introduced the `comms` feature and corresponding macro restructuring. The proxy/service split needs to be reimplemented from current `main`; resolving this old patch as-is would risk discarding or mis-porting those changes. The preliminary rustfmt/clippy commits also target obsolete old-tree code.

I did validate the implementation on its old base: the reduced library feature combinations compile, but their documentation builds/doctests do not (see the inline finding). Please update the branch before the rest of the code-level review.

.github/workflows/rust.yml
zeenix: Done on the rebased branch. Intra-doc links into one half of the API use the `cfg_attr(feature = ..., doc = ...)` pairing that `Connection::set_unique_name` already uses for `bus-impl`; the doc examples that serve an interface or use `DBusProxy` types are gated with hidden `cfg` lines; the crate-level README is only the crate docs when both halves are compiled in, with a short fallback page otherwise. `RUSTDOCFLAGS="-D warnings" cargo doc` and `cargo test --doc` are clean for `async-io,proxy`, `async-io,service` and `async-io` alone, and the doc build and doctest jobs now run the `proxy`-only and `service`-only configurations.

---

zeenix (commented):
🤖 ChatGPT Codex on behalf of Zeeshan:

[minor] The `concurrent_method_calls` benchmark still declares `required-features = ["p2p"]`, but it now uses the service-gated `#[zbus::interface]` macro and `Builder::serve_at`. `cargo check -p zbus --bench concurrent_method_calls --no-default-features --features async-io,p2p` succeeds on current `main` and fails on this head with both APIs missing. Please add `service` to that benchmark’s `required-features` (and consider covering this p2p/no-service target combination in CI).

The rebase and reduced-feature documentation blockers from my previous review are resolved. I also confirmed that the default-feature rustdoc JSON is API-equivalent to `main`, aside from documentation and re-export ordering.

zbus_macros/src/iface.rs
```diff
@@ -338,9 +338,15 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
     };
     let with_spawn = impl_attrs.spawn.unwrap_or(true);
     let crate_attr = impl_attrs.crate_path.clone();
-    let mut proxy = impl_attrs
-        .proxy
-        .map(|p| Proxy::new(ty, &iface_name, p, &zbus, crate_attr));
+    // The `proxy` attribute is still parsed and accepted but no proxy code is generated when the
+    // `proxy` feature is disabled.
+    let mut proxy = if cfg!(feature = "proxy") {
```
zeenix: 🤖 ChatGPT Codex on behalf of Zeeshan:

[blocking] This `cfg!` observes the proc-macro crate's unified features, not necessarily the target `zbus` features. I reproduced a service-only crate using `#[interface(proxy(...))]` compiling normally, then failing once a build-dependency enables `zbus/proxy`: `zbus_macros` generates the proxy while target `zbus` still lacks `proxy`, producing `could not find proxy in zbus` and a duplicate type. Please make this choice follow the target zbus feature set and add a host/target feature-unification regression test.

zeenix (commented):
🤖 ChatGPT Codex on behalf of Zeeshan: One finding from the fresh independent review.

zbus_macros/src/iface.rs
zeenix: Right, and reproduced. The decision now lives in `zbus`: the `interface` macro always emits the proxy when the attribute is given, wrapped in a hidden `zbus::__if_proxy_feature!` macro that `zbus` defines as pass-through with `proxy` and as empty without, so it follows the features of the `zbus` the generated code is compiled against. `cfg!` is gone from the macro.

The same problem existed before this PR for `blocking-api`: the `proxy` macro decided whether to generate the blocking proxies from the `blocking-api` feature of `zbus_macros`. That is fixed the same way in its own commit, since the regression fixture exposes it in a workspace-wide build (its host-side `zbus` lacks `blocking-api` while `zbus_macros` has it).

Regression test: `test_fixtures/service_only` depends on `zbus` with `async-io,service` and, from its build script, on `zbus` with `async-io,proxy`, and uses `#[interface(proxy(..))]`. It fails on the previous head exactly as you describe and passes now. CI checks it with `cargo clippy -p service_only_fixture` on its own, since a workspace-wide build unifies the features away; the workspace build additionally exercises the pass-through side. The expanded public signatures with default features are still identical to `main`.

---

.github/workflows/rust.yml
```diff
@@ -245,6 +267,10 @@ jobs:
         run: cargo --locked doc --all-features -p zbus
       - name: Check wire-format-only documentation build
         run: cargo --locked doc --no-default-features -p zbus
+      - name: Check documentation build with the proxy or service API disabled
+        run: |
+          cargo --locked doc --no-default-features --features async-io,proxy -p zbus
```
zeenix: 🤖 ChatGPT Codex on behalf of Zeeshan:

[blocking] This reduced-feature docs coverage still misses the valid `blocking-api + proxy` configuration. I reproduced `RUSTDOCFLAGS="-D warnings" cargo doc --locked -p zbus --no-default-features --features async-io,blocking-api,proxy --no-deps` failing on unresolved links to the service-gated `crate::interface` in `blocking/mod.rs` and `Builder::serve_at` in `blocking/connection/builder.rs`. Please gate those links and include `blocking-api` in this proxy-only doc check so the configuration stays covered.

zeenix (commented):
🤖 ChatGPT Codex on behalf of Zeeshan:

.github/workflows/rust.yml
zeenix: Fixed. Both links are gated with the `cfg_attr(doc = ...)` pairing, and the `blocking::MessageIterator::for_match_rule` example, which decodes the signal with the proxy-gated `NameOwnerChanged` type, is gated like its async counterpart. The doc build job now also runs `async-io,blocking-api,proxy` and `async-io,blocking-api,service`, and the doctest job runs the same two combinations. All four doc builds are clean with `-D warnings` and all doctest combinations pass locally.

---

zeenix: 🤖 ChatGPT Codex on behalf of Zeeshan: Confirmed fixed on `7a8f521`; the blocking reduced-feature docs are now covered and pass. No further issue in this delta.

zeenix: 🤖 ChatGPT Codex on behalf of Zeeshan: Verified on `7a8f521`: all four reduced-feature doc builds pass with `-D warnings`, both blocking doctest configurations compile, and the host/target feature regression fixture passes. No new findings. Local D-Bus execution remains restricted; Linux CI is still running, with all other jobs green.

PR #1946, opened 2026-09-05
⚡️ zb,zm: Cut the binary size of zbus programs

zeenix:
Follow-up to the busd size regression: busd built against zbus 6.0 came out larger than against 5.x. This PR shrinks every zbus program by moving generic, per-interface and per-method code out of line, by serializing the message header once, and by making three rarely-used pieces optional.

## Changes

**Less monomorphized code**

- Message bodies are written through a trait object and serialized in one pass, so the serializer is instantiated once instead of per body type.
- Replies, method-handler error conversion and signature-mismatch errors are built in non-generic helpers. `#[interface]` emits the `DBusError` conversion arms once per interface instead of once per method.
- The method call and signal paths on `Connection` stay non-generic. `#[proxy]` no longer instantiates the properties cache for proxies without properties (a behavior change, see below).
- Introspection doc comments go through a small helper instead of being formatted inline per member.

**The message header is serialized once**

The builder measured the header with a size-counting serializer before writing it, which is a second, complete instantiation of the header serializer. It now writes the header straight into the message buffer and takes the padding and body offset from its length. This alone is about 39 KB on every program, busd included.

A coexistence test also reads a whole message carrying a GVariant maybe body signature and checks it is refused.

**New cargo features (all on by default)**

- `unixexec` and `ibus`: the two exec-based transports.
- `object-manager` (implies `service`): the service-side `fdo::ObjectManager` and the object server's `InterfacesAdded`/`InterfacesRemoved` emission. `ObjectManagerProxy` stays with `proxy`.

The upgrade guide covers the features and the properties-cache change.

The D-Bus codec's hot paths are untouched. Two things were tried and dropped: sharing one `SeqAccess` between array, structure and variant visitors saved a further 3 to 4% but cost 5 to 25% on deserialization benchmarks; carrying header fields in a purpose-built type instead of `Value` saved 20 KB on programs that never use `Value` but cost 6 to 8 KB on everything that does, busd included.

## Results

`.text` bytes, fat LTO, `opt-level = "s"`, linux x86_64:

| program | before | after | change |
|---|---|---|---|
| bare connection | 845,396 | 806,200 | -4.6% |
| proxy, one method | 870,295 | 831,118 | -4.5% |
| service, one method | 1,007,039 | 924,982 | -8.1% |
| busd | 2,321,957 | 2,243,901 | -3.4% (9.8% below zbus main) |

## Verification

Every commit compiles on its own with all features and without default features. Under a session bus the all-features zbus suite passes, as does the suite with only `tokio,proxy,service,p2p`, and the coexistence tests pass wire-only and with the D-Bus API on. Clippy is clean on five feature combinations including `--all-targets`. Docs build warning-free with and without the new features. The wire and message deserialization benchmarks match main.

https://claude.ai/code/session_01Jxv7KpDPUqYLEvX6Z8tPMX

### DIFF
--- a/zbus_macros/src/error.rs
+++ b/zbus_macros/src/error.rs
@@ -146,23 +146,18 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
             quote! {
                 impl ::std::convert::From<#zbus::Error> for #name {
                     fn from(value: #zbus::Error) -> #name {
-                        match &value {
+                        let (name, desc) = match &value {
                             #zbus::Error::MethodError(name, desc, _) => {
-                                let desc = desc.as_deref();
-                                match name.as_str() {
-                                    #error_converts
-                                    _ => Self::#ident(value),
-                                }
+                                (name.as_ref(), desc.as_deref())
                             }
                             #zbus::Error::FDO(e) => {
                                 let e = ::std::convert::AsRef::as_ref(e);
-                                let name = #zbus::DBusError::name(e);
-                                let desc = #zbus::DBusError::description(e);
-                                match name.as_str() {
-                                    #error_converts
-                                    _ => Self::#ident(value),
-                                }
+                                (#zbus::DBusError::name(e), #zbus::DBusError::description(e))
                             }
+                            _ => return Self::#ident(value),
+                        };
+                        match name.as_str() {
+                            #error_converts
                             _ => Self::#ident(value),
                         }
                     }
--- a/zbus_macros/src/error.rs
+++ b/zbus_macros/src/error.rs
@@ -1,4 +1,4 @@
-use proc_macro2::TokenStream;
+use proc_macro2::{Span, TokenStream};
 use quote::{ToTokens, quote};
 use syn::{Data, DeriveInput, Error, Fields, Ident, Variant, spanned::Spanned};
 use zbus_utils::def_attrs;
@@ -37,6 +37,8 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
 
     let zbus = zbus_path(crate_path.as_ref());
     let mut replies = quote! {};
+    // Hygienic so that a variant field of the same name does not shadow it in the reply arms.
+    let builder = Ident::new("builder", Span::mixed_site());
     let mut error_names = quote! {};
     let mut error_descriptions = quote! {};
     let mut error_converts = quote! {};
@@ -137,7 +139,7 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
             error_converts.extend(e);
         }
 
-        let r = gen_reply_for_variant(&variant, error, &zbus)?;
+        let r = gen_reply_for_variant(&variant, error, &zbus, &builder)?;
         replies.extend(r);
     }
 
@@ -196,6 +198,7 @@ pub fn expand_derive(input: DeriveInput) -> Result<TokenStream, Error> {
 
             fn create_reply(&self, call: &#zbus::message::Header) -> #zbus::Result<#zbus::message::Message> {
                 let name = self.name();
+                let #builder = #zbus::message::Message::error(call, name)?;
                 match self {
                     #replies
                 }
@@ -214,11 +217,12 @@ fn gen_reply_for_variant(
     variant: &Variant,
     zbus_error_variant: bool,
     zbus: &TokenStream,
+    builder: &Ident,
 ) -> Result<TokenStream, Error> {
     let ident = &variant.ident;
     match &variant.fields {
         Fields::Unit => Ok(quote! {
-            Self::#ident => #zbus::message::Message::error(call, name)?.build(&()),
+            Self::#ident => #builder.build(&()),
         }),
         Fields::Unnamed(f) => {
             // Name the unnamed fields as the number of the field with an 'f' in front.
@@ -248,13 +252,13 @@ fn gen_reply_for_variant(
             };
 
             Ok(quote! {
-                Self::#ident(#(#in_fields),*) => #zbus::message::Message::error(call, name)?.build(&(#(#out_fields),*)),
+                Self::#ident(#(#in_fields),*) => #builder.build(&(#(#out_fields),*)),
             })
         }
         Fields::Named(f) => {
             let fields = f.named.iter().map(|v| v.ident.as_ref()).collect::<Vec<_>>();
             Ok(quote! {
-                Self::#ident { #(#fields),* } => #zbus::message::Message::error(call, name)?.build(&(#(#fields),*)),
+                Self::#ident { #(#fields),* } => #builder.build(&(#(#fields),*)),
             })
         }
     }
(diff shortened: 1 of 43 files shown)


zbus_macros/src/error.rs
```diff
@@ -253,13 +249,13 @@ fn gen_reply_for_variant(
...
             Ok(quote! {
-                Self::#ident(#(#in_fields),*) => #zbus::message::Message::error(call, name)?.build(&(#(#out_fields),*)),
+                Self::#ident(#(#in_fields),*) => builder.build(&(#(#out_fields),*)),
             })
         }
         Fields::Named(f) => {
             let fields = f.named.iter().map(|v| v.ident.as_ref()).collect::<Vec<_>>();
             Ok(quote! {
-                Self::#ident { #(#fields),* } => #zbus::message::Message::error(call, name)?.build(&(#(#fields),*)),
+                Self::#ident { #(#fields),* } => builder.build(&(#(#fields),*)),
```
zeenix: [P2] Avoid shadowing the generated message builder

A named error field called `builder` now shadows the new local binding, so `#[derive(Debug, zbus::DBusError)] enum Error { Failed { builder: String } }` fails with E0599 (`no method named build found for &String`). I compiled this reproducer: it passes on base `258519e` and fails on this head. Bind named fields to generated aliases, or use a hygienic identifier for the message builder.

— Codex, reviewing on behalf of zeenix.

zbus/src/message/fields.rs
```diff
@@ -28,11 +30,236 @@ pub(crate) struct Fields<'f> {
...
+                    fields.reply_serial = Some(serial);
+                }
+                (FieldCode::Destination, b"s") => {
+                    fields.destination = Some(BusName::try_from(reader.str()?)?);
+                }
+                (FieldCode::Sender, b"s") => {
+                    fields.sender = Some(UniqueName::try_from(reader.str()?)?);
+                }
+                (FieldCode::Signature, b"g") => {
+                    fields.signature = Cow::Owned(Signature::from_bytes(reader.signature()?)?);
```
zeenix: [P2] Preserve the rejection of GVariant-only body signatures

With `zbus_utils/gvariant` enabled (including through a co-located zgvariant dependency), `Signature::from_bytes(b"mi")` succeeds. The old header decoder passed this value through `reject_maybe_in_signature_str`; this new path skips that check and accepts an invalid D-Bus body signature. I reproduced this with a little-endian header whose fields-array payload is `[8, 1, b'g', 0, 2, b'm', b'i', 0]`: serde header decoding rejects it, while `Header::read` succeeds. Apply the existing D-Bus-specific signature validation here and add a header-level coexistence regression test.

— Codex, reviewing on behalf of zeenix.

zeenix (commented):
Reviewed by Codex on behalf of zeenix at `6e4fcf1`. Two regressions reproduced locally; details inline. The four existing message tests pass.

zbus_macros/src/error.rs
zeenix: Fixed in 34a4b9a: the `builder` local is now a `Span::mixed_site()` identifier, so a variant field of the same name binds and is read as the field. The derive test enum gained a `Shadowing { builder: String }` variant that fails to compile without it.

---

zbus/src/message/fields.rs
zeenix: Fixed in f1a7e16: the header reader runs the parsed body signature through `wire::dbus::reject_maybe`, and the writer does the same before emitting the field. Added `dbus_message_rejects_header_signature_carrying_maybe` to the coexistence tests with exactly this header; it fails without the check. Since that test needs the D-Bus API, the coexistence CI job now also runs with `async-io` on.

---

zeenix (commented):
Re-reviewed `1b8e465` on behalf of zeenix — Codex.

Both findings are fixed: my original `builder`-field reproducer and GVariant-header rejection test now pass. The five coexistence tests and header comparison against serde in both endiannesses also pass. Reviewed all five changed files since the previous round; no new issues found. CI is green.

zeenix (commented):
Re-reviewed `3562902` on behalf of zeenix — Codex.

No new issues found in the seven-file update. Both previous regressions remain fixed. Locally: all 38 wire tests, three message tests, and the original `builder` reproducer pass with `zbus_utils/gvariant,option-as-array`. Additional checks passed for 512 nested-container round trips, variant depth limits, and header-byte equivalence to the previous `Value` encoding in both endiannesses. Full CI is still running.

zeenix (commented):
Re-reviewed `0ee1b5f` on behalf of zeenix — Codex.

No new issues found. This force-push removes the shared `SeqAccess` deserializer optimization that caused the benchmark regressions, restoring the codec container paths to `main`, while retaining both prior fixes and the already-reviewed `FieldValue` header encoding.

The earlier targeted tests covered this retained header implementation; I could not rerun them in this execution environment because a Rust toolchain is no longer available. GitHub CI currently passes commitlint, CodSpeed, semver, clippy, fuzz, docs, formatting, MSRV, and the GVariant coexistence step; the full Linux and Windows test jobs are still running.

zeenix (commented):
Reviewed by Codex on behalf of zeenix at `0ee1b5fe`, including the full diff and each of the 13 commits.

No new correctness blockers found; both earlier findings remain resolved and all current CI workflows pass. The remaining blocker is commit atomicity. Please split these three commits, keeping each change's documentation/tests with it:

- `34a4b9a2`: deduplicating `From<zbus::Error>` conversion arms and hoisting the builder in `DBusError::create_reply` are two independent refactors. Keep the hygiene fix and `Shadowing` regression test with the latter.
- `9a72cfcd`: extracting the method-call path (including `PendingMethodCall`) and extracting the signal path are independent refactors; neither uses the other's new helpers.
- `ba78b0b3`: introducing the `unixexec` feature and introducing the `ibus` feature are independently usable changes. Split them, adapting the shared process-module cfgs in each step while retaining macOS `launchd` support.

The other ten commits each have one logical scope. Verdict: ready on correctness, but not ready to merge under the project's atomic-commit requirement until these splits are made.

zeenix (commented):
Re-reviewed `4ffe58c0` — Codex, reviewing on behalf of zeenix.

All three requested splits are correct: the six replacement commits each contain one logical change, with the builder hygiene fix/test and transport documentation/test gating in the appropriate commits. The intermediate transport cfgs preserve IBus and macOS launchd support.

Independently verified that the final Git tree is identical to `0ee1b5fe`; the other ten patches are unchanged. Both prior correctness fixes therefore remain intact, and no new issues were found. No local tests rerun for this history-only update; intermediate commits were inspected, not rebuilt.

Verdict: ready from review and commit-atomicity perspectives, pending completion of the new CI run. Commitlint is green; build/test and benchmarks are still running.

zeenix (commented):
Re-reviewed `512d3d4e` — Codex, reviewing on behalf of zeenix.

This update cleanly replaces the dropped `FieldValue` optimization with one atomic regression-test commit: the focused message-header test and the CI invocation needed to exercise it together. The production tree is now the previously reviewed `4ef07d71`; the other 15 commits are unchanged, so both prior fixes remain intact. No new correctness or commit-atomicity issues found.

Verdict: ready from review and commit-atomicity perspectives. Current CI has commitlint, formatting, MSRV, semver, and fuzz checks green; clippy, docs, and CodSpeed are still running.

PR #1950, opened 2026-09-07
🥅 zb,zm: Let property handlers fail with any `DBusError`

### LINKED ISSUE #313: Generic property getter errors
In GitLab by @wash2 on Feb 6, 2023, 01:03

https://gitlab.freedesktop.org/dbus/zbus/-/merge_requests/649 adds support for returning `zbus::fdo::Error` from property getters for now, but it should support returning `DBusError`.

zeenix:
Property getters and setters generated by `#[interface]` may now fail with any type implementing `DBusError`, the same rule methods already follow, so a service can reply with its own error names from a property access (e.g. `org.freedesktop.Secret.Error.IsLocked` from a setter). `fdo::Result` getters and setters, and setters returning `zbus::Result<()>`, keep working.

The `Interface` trait carries property failures as a new `BoxDBusError` alias (`Box<dyn DBusError + Send + Sync>`), so does the `DispatchResult2::Async` future that a `&self` setter reports through. The box is only allocated on the failure path. `Box<E>` implements `DBusError`, `fdo::Error` and `zbus::Error` convert into the box so `?` keeps working, and the macro routes every property failure through a hidden conversion trait since it cannot tell a `DBusError` from a `zbus::Error` syntactically. `Interface::get_all` is unchanged: it leaves out a property whose getter fails.

One observable detail: the generated `<property>_changed` helper returns `zbus::Result<()>` to the service's own code, which cannot carry an arbitrary `DBusError`, so a getter failing there is reported as `Error::Failure` with the error's name and description as its text.

The e2e test gains a getter, a `&self` setter and a `&mut self` setter failing with the custom `MyIfaceError`, checking that the error name reaches the client on all three paths. The book no longer says property errors must be `fdo::Error`.

Fixes #313.

https://claude.ai/code/session_01CLGqZvo2MX7FNACCSAASSM

---
_Generated by [Claude Code](https://claude.ai/code/session_01CLGqZvo2MX7FNACCSAASSM)_

### DIFF
--- a/zbus/src/dbus_error.rs
+++ b/zbus/src/dbus_error.rs
@@ -1,5 +1,7 @@
+use std::fmt;
+
 use crate::{
-    Result,
+    Error, Result, fdo,
     message::{Header, Message},
     names::ErrorName,
 };
@@ -23,4 +25,47 @@ pub trait DBusError {
 
     // The optional description for the error.
     fn description(&self) -> Option<&str>;
+}
+
+/// A [`DBusError`] of any concrete type.
+///
+/// This is how the object server carries a property getter's or setter's failure, whatever error
+/// type the handler fails with. The box is only allocated on failure.
+pub type BoxDBusError = Box<dyn DBusError + Send + Sync>;
+
+impl fmt::Display for dyn DBusError + Send + Sync {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let description = self.description().unwrap_or("no description");
+        write!(f, "{}: {description}", self.name())
+    }
+}
+
+impl<E> DBusError for Box<E>
+where
+    E: DBusError + ?Sized,
+{
+    fn create_reply(&self, msg: &Header<'_>) -> Result<Message> {
+        (**self).create_reply(msg)
+    }
+
+    fn name(&self) -> ErrorName<'_> {
+        (**self).name()
+    }
+
+    fn description(&self) -> Option<&str> {
+        (**self).description()
+    }
+}
+
+impl From<fdo::Error> for BoxDBusError {
+    fn from(e: fdo::Error) -> Self {
+        Box::new(e)
+    }
+}
+
+/// A plain [`Error`] is not a D-Bus error; it is reported the way [`fdo::Error`] reports it.
+impl From<Error> for BoxDBusError {
+    fn from(e: Error) -> Self {
+        Box::new(fdo::Error::from(e))
+    }
 }
--- a/zbus_macros/src/iface.rs
+++ b/zbus_macros/src/iface.rs
@@ -518,6 +518,22 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                 let sk_member_name = case::snake_or_kebab_case(&member_name, true);
                 let prop_changed_method_name = format_ident!("{sk_member_name}_changed");
                 let prop_invalidate_method_name = format_ident!("{sk_member_name}_invalidate");
+                let prop_value_method_name = format_ident!("__zbus_{sk_member_name}_value");
+                // Emits `PropertiesChanged` for the property, with `value` in scope holding its
+                // current value.
+                let emit_changed = quote!({
+                    let mut changed = ::std::collections::HashMap::new();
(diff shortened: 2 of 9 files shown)


zbus_macros/src/iface.rs
```diff
@@ -686,7 +699,14 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, mut input: ItemImpl) -> syn::Re
                     get_all.extend(q);
 
                     let prop_value_handled = if is_fallible_property {
-                        quote!(self.#ident(#args_names)#method_await?)
+                        quote!(match self.#ident(#args_names)#method_await {
+                            ::std::result::Result::Ok(value) => value,
+                            ::std::result::Result::Err(e) => {
+                                return ::std::result::Result::Err(
+                                    #zbus::object_server::IntoDBusError::into_error(e),
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix: **[P2] Preserve getter error names during automatic PropertiesChanged emission.** This helper is also called by the generated setter after it succeeds (lines 566–596), so converting its getter error to `Error::Failure` affects the D-Bus reply, not only the service's local API. With an infallible setter and a getter returning `fdo::Error::AccessDenied("locked")`, `Properties.Set` now replies with `org.freedesktop.zbus.Error` instead of `org.freedesktop.DBus.Error.AccessDenied`. A custom getter error loses its name the same way. A focused P2P test reproduces both on this head; the standard-error case preserves its name on the base commit. Please keep the typed error through the automatic setter path and cover successful setters whose subsequent getter fails.

zbus/src/dbus_error.rs
```diff
@@ -24,3 +26,47 @@ pub trait DBusError {
     // The optional description for the error.
     fn description(&self) -> Option<&str>;
 }
+
+/// A [`DBusError`] of any concrete type.
+///
+/// This is how the [`object_server`](crate::object_server) carries a property getter's or
```
zeenix: 🤖 ChatGPT (Codex) on behalf of zeenix: **[P2] Make this documentation link work without `service`.** `dbus_error` is enabled by `comms`, while `crate::object_server` requires `service`. The new unconditional link therefore breaks the proxy-only documentation build. This is the actual failure in `doc_build`: `RUSTDOCFLAGS='-D warnings' cargo --locked doc --no-default-features --features async-io,proxy -p zbus` reports an unresolved link at this line. Please make the link conditional or use plain code text.

zeenix (commented):
🤖 ChatGPT (Codex) on behalf of zeenix: Not ready yet: the automatic PropertiesChanged path loses error names, and the proxy-only documentation build fails. The single commit is atomic and follows the contribution conventions. Verified the runtime regression with a focused P2P test: the existing fdo-error case passes on base 60bd1ec and fails on head 0ee504d.

zbus_macros/src/iface.rs
zeenix: Fixed in 0a959d8. The setter no longer goes through the public `<property>_changed` helper. The getter arm now generates a hidden `__zbus_<property>_value` method returning `Result<T, BoxDBusError>`; the setter path and the public helper both fetch the value through it and emit the signal themselves, so the typed getter error reaches the `Set` reply unchanged. Only the public helper, which returns `zbus::Result<()>`, still maps the getter's failure to `Error::Failure`.

Added `locked_prop` (getter fails with `fdo::Error::AccessDenied`, infallible `&mut self` setter) and `custom_locked_prop` (getter fails with `MyIfaceError`, infallible `&self` setter) to the e2e test; both assert the error name in the `Set` reply.

---

zbus/src/dbus_error.rs
zeenix: Fixed in 0a959d8: the link is plain text now. All six `cargo doc` invocations from the `doc_build` job pass locally with `RUSTDOCFLAGS='-D warnings'`.

---

zeenix (commented):
🤖 ChatGPT (Codex) on behalf of zeenix: Rechecked 0a959d8. Both findings are fixed: the original P2P reproducer now preserves standard and custom getter error names after successful setters, and the feature-gated documentation link is gone. All CI checks pass, both threads are resolved, and the amended commit remains atomic. No further blocking findings; ready to merge.

PR #1951, opened 2026-09-07
🥅 zb,zm: Carry any D-Bus error in `zbus::Error`

zeenix:
Follow-up to #1950. A property getter may fail with any `DBusError`, but the `<property>_changed` helper that `#[interface]` generates returns `zbus::Result<()>`, and `zbus::Error` could only carry an `fdo::Error`. The helper therefore flattened a getter's failure into `Error::Failure` text, and the generated setter needed a hidden `__zbus_<property>_value` method to keep the getter's error name in its `Set` reply.

This adds `Error::DBus(Arc<dyn DBusError + Send + Sync>)` (an `Arc` because `Error` is `Clone`, like `Error::InputOutput`). The helper reports a failing getter as that variant with name, description and reply body intact, the setter calls the helper again, and the hidden method is gone. The `DBusError` derive maps the new variant by name the way it maps `Error::FDO`, so `fdo::Error::from` and a custom error's `From<zbus::Error>` recover the typed error.

The e2e test calls the helper on the two properties whose getter fails and checks the error converts back to the standard and the custom error; a unit test covers the `fdo::Error` conversion.

https://claude.ai/code/session_01CLGqZvo2MX7FNACCSAASSM

---
_Generated by [Claude Code](https://claude.ai/code/session_01CLGqZvo2MX7FNACCSAASSM)_

### DIFF
--- a/zbus/src/dbus_error.rs
+++ b/zbus/src/dbus_error.rs
@@ -1,4 +1,4 @@
-use std::fmt;
+use std::{fmt, sync::Arc};
 
 use crate::{
     Error, Result, fdo,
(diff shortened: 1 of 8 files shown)


zeenix (commented):
🤖 ChatGPT (Codex) on behalf of zeenix: Reviewed 5e06879; no blocking findings. The new variant preserves the error while allowing the setter to reuse the public signal helper. Focused local tests pass for the original #1950 regression, custom multi-field reply bodies through cloning and both helper paths, and preservation of plain zbus::Error variants. The single commit is atomic and follows the contribution conventions. Ready from review, pending the still-running Linux and Windows CI tests; all completed checks pass.

PR #1953, opened 2026-09-08
🚸 zb,zm: Defer builder conversion errors to build()

### LINKED ISSUE #403: Error handling when creating instances via a builder

zeenix:
Every builder method that takes a D-Bus name, object path, GUID or address is generic over
`TryInto` and returns `Result<Self>`, so callers write `?` (or `.expect("cannot fail")`) after
every call, even when the argument is already a validated `ObjectPath` or `BusName`, and even
though three of the four builders return a `Result` from `build()` anyway. This PR resolves that
for 6.0 by deferring the errors, the way `http::request::Builder` does:

* Every setter keeps its zbus 5 generic signature and returns the builder. The conversion still
  happens immediately; a failure is recorded in the builder's single error slot, first error
  wins, and later calls never clear it.
* `build()` returns that recorded error before doing anything else, with the same error value
  the setter used to return. `match_rule::Builder::build()` becomes fallible for that reason;
  the other three already were.
* `Message::method_call`, `signal`, `error` and `method_return` return the builder directly,
  and the connection builder constructors (`session()`, `system()`, `address(..)`, ...) return
  the builder with a bad address recorded in it.
* `MatchRule::try_from(&str)` validates each component itself, so parse errors are unchanged.
* Generic code (`Proxy::new`, the `#[proxy]`-generated `new()`, `Connection::call_method`)
  keeps its `TryInto` bounds and just drops the `?` after each setter. Nothing changes for
  generated proxies.

```rust
// Before
let props = PropertiesProxy::builder(&conn)
    .destination("org.freedesktop.DBus")?
    .path("/org/freedesktop/DBus")?
    .build()
    .await?;

// After: strings or typed values, no `?` until `build()`
let props = PropertiesProxy::builder(&conn)
    .destination("org.freedesktop.DBus")
    .path(proxy.path())
    .build()
    .await?;
```

Migration is mechanical and compiler-guided: delete the `?` after each setter and after the
constructors above, add one after `MatchRule::builder()...build()`. The 6.0 upgrade guide gets a
section with the details.

The design spec (`docs/superpowers/specs/2026-09-07-infallible-builder-setters-design.md`)
records the alternatives and why they lost: an infallible `impl Into` setter plus a `try_` twin
(strings are the majority input, so most callers would rename and keep their `?`), a trait with
a conditional return type (opaque signatures, dozens of impls, unusable from generic code), and
additive setters under another name. Compile-time validated literal macros
(`object_path!("/org/zbus/Foo")`) are an orthogonal follow-up.

Fixes #403.

https://claude.ai/code/session_01FAoA1GwLD68owkUkUS9fSB

### DIFF
--- a/zbus/src/guid.rs
+++ b/zbus/src/guid.rs
@@ -197,6 +197,12 @@ impl<'unowned, 'owned: 'unowned> From<&'owned OwnedGuid> for Guid<'unowned> {
     }
 }
 
+impl<'g> From<&Guid<'g>> for Guid<'g> {
+    fn from(guid: &Guid<'g>) -> Self {
+        guid.clone()
+    }
+}
+
 impl From<Guid<'_>> for OwnedGuid {
     fn from(guid: Guid<'_>) -> Self {
         OwnedGuid(guid.to_owned())
--- a/zbus/src/match_rule/builder.rs
+++ b/zbus/src/match_rule/builder.rs
@@ -15,6 +15,7 @@ const MAX_ARGS: u8 = 64;
 /// clear it — and reported by [`Builder::build`], so a chain of setters never needs a `?` in
 /// between.
 #[derive(Debug)]
+#[must_use]
 pub struct Builder<'m> {
     rule: MatchRule<'m>,
     error: Option<Error>,
--- a/zbus/src/match_rule/builder.rs
+++ b/zbus/src/match_rule/builder.rs
@@ -9,190 +9,209 @@ const MAX_ARGS: u8 = 64;
 
 /// Builder for [`MatchRule`].
 ///
-/// This is created by [`MatchRule::builder`].
+/// This is created by [`MatchRule::builder`]. Each setter takes the value it is given, whether
+/// that's an already parsed name or a string that still needs parsing, and converts it right
+/// away, but it never fails: the first error a setter hits is recorded — a later call doesn't
+/// clear it — and reported by [`Builder::build`], so a chain of setters never needs a `?` in
+/// between.
 #[derive(Debug)]
-pub struct Builder<'m>(MatchRule<'m>);
+pub struct Builder<'m> {
+    rule: MatchRule<'m>,
+    error: Option<Error>,
+}
 
 impl<'m> Builder<'m> {
     /// Build the `MatchRule`.
-    pub fn build(self) -> MatchRule<'m> {
-        self.0
+    ///
+    /// # Errors
+    ///
+    /// The first error recorded by a setter.
+    pub fn build(self) -> Result<MatchRule<'m>> {
+        match self.error {
+            Some(error) => Err(error),
+            None => Ok(self.rule),
+        }
     }
 
     /// Set the sender.
-    pub fn sender<B>(mut self, sender: B) -> Result<Self>
+    ///
+    /// An invalid sender is reported by [`Builder::build`].
+    pub fn sender<B>(mut self, sender: B) -> Self
     where
         B: TryInto<BusName<'m>>,
         B::Error: Into<Error>,
     {
-        self.0.sender = Some(sender.try_into().map_err(Into::into)?);
+        match sender.try_into() {
+            Ok(sender) => self.rule.sender = Some(sender),
+            Err(e) => self.record(e.into()),
+        }
 
-        Ok(self)
+        self
     }
 
     /// Set the message type.
     pub fn msg_type(mut self, msg_type: Type) -> Self {
-        self.0.msg_type = Some(msg_type);
+        self.rule.msg_type = Some(msg_type);
 
         self
     }
 
     /// Set the interface.
-    pub fn interface<I>(mut self, interface: I) -> Result<Self>
+    ///
+    /// An invalid interface name is reported by [`Builder::build`].
+    pub fn interface<I>(mut self, interface: I) -> Self
     where
         I: TryInto<InterfaceName<'m>>,
         I::Error: Into<Error>,
     {
-        self.0.interface = Some(interface.try_into().map_err(Into::into)?);
+        match interface.try_into() {
+            Ok(interface) => self.rule.interface = Some(interface),
+            Err(e) => self.record(e.into()),
+        }
 
-        Ok(self)
+        self
     }
 
     /// Set the member.
-    pub fn member<M>(mut self, member: M) -> Result<Self>
+    ///
+    /// An invalid member name is reported by [`Builder::build`].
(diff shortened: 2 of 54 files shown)


zbus/src/match_rule/mod.rs
```diff
@@ -440,31 +437,33 @@ impl<'m> TryFrom<&'m str> for MatchRule<'m> {
...
-                "arg0namespace" => builder.arg0ns(value)?,
+                // Each component is converted here, rather than left to the builder, so that an
+                // invalid rule string is reported by the component that's actually at fault.
+                "sender" => builder.sender(BusName::try_from(value)?),
+                "interface" => builder.interface(InterfaceName::try_from(value)?),
+                "member" => builder.member(MemberName::try_from(value)?),
+                "path" => builder.path(ObjectPath::try_from(value)?),
+                "path_namespace" => builder.path_namespace(ObjectPath::try_from(value)?),
+                "destination" => builder.destination(UniqueName::try_from(value)?),
+                "arg0namespace" => builder.arg0ns(value),
```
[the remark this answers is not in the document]
zeenix: Good catch. The parser now validates `arg0namespace` itself, like the other components, and a regression test covers the repeated-key case.

---

zbus/src/guid.rs
```diff
@@ -243,6 +249,13 @@ mod tests {
     use crate::Guid;
     use test_log::test;
 
+    #[test]
+    fn borrowed_guid_converts() {
+        let guid = Guid::try_from("0123456789abcdef0123456789abcdef").unwrap();
+
+        assert_eq!(Guid::from(&guid), guid);
+    }
+
```
zeenix: No test needed. This is a case of if it builds, it works.

zbus/src/names/bus_name.rs
```diff
@@ -538,3 +550,20 @@ impl NoneValue for OwnedBusName {
...
+    fn borrowed_sub_names_convert() {
+        let unique = UniqueName::try_from(":1.42").unwrap();
+        let well_known = WellKnownName::try_from("org.zbus.Test").unwrap();
+
+        assert_eq!(BusName::from(&unique), BusName::Unique(unique.clone()));
+        assert_eq!(
+            BusName::from(&well_known),
+            BusName::WellKnown(well_known.clone())
+        );
+    }
```
zeenix: this test is not needed either.

zbus/src/match_rule/builder.rs
```diff
@@ -9,255 +9,245 @@ const MAX_ARGS: u8 = 64;
...
+pub struct Builder<'m> {
+    msg_type: Option<Type>,
+    sender: Option<Result<BusName<'m>>>,
+    interface: Option<Result<InterfaceName<'m>>>,
+    member: Option<Result<MemberName<'m>>>,
+    path_spec: Option<Result<PathSpec<'m>>>,
+    destination: Option<Result<UniqueName<'m>>>,
+    args: Vec<(u8, Result<Str<'m>>)>,
+    arg_paths: Vec<(u8, Result<ObjectPath<'m>>)>,
+    arg0ns: Option<Result<Str<'m>>>,
```
zeenix: Since the Error type is the same, we should instead just add a `error: Option<Error>` instead and overwrite it only if it's `None`. It's ok to only report the first error on build. The same for all other builders.

zbus/src/proxy/builder.rs
```diff
@@ -49,33 +53,42 @@ impl<T> Clone for Builder<'_, T> {
 
 impl<'a, T> Builder<'a, T> {
     /// Set the proxy destination address.
-    pub fn destination<D>(mut self, destination: D) -> Result<Self>
+    ///
+    /// An invalid destination is reported by [`Builder::build`].
+    #[must_use]
```
zeenix, line 61: `must_use` addition is a good change but it should be split into separate commits. Fee free to put all these additions in one commit.

zbus/src/message/builder.rs
```diff
@@ -273,20 +342,108 @@ impl<'m> From<Header<'m>> for Builder<'m> {
...
+/// One slot per fallible setter, in the order [`Builder::build`] reports them.
+#[derive(Debug, Default, Clone)]
+struct Errors {
+    flags: Option<Error>,
+    sender: Option<Error>,
+    path: Option<Error>,
+    interface: Option<Error>,
+    member: Option<Error>,
+    error_name: Option<Error>,
+    destination: Option<Error>,
```
zeenix: Same here: we don't need to keep every field's error separate. Just recording the first one and throwing that from `build` is sufficient.

zbus/src/match_rule/builder.rs
zeenix: Done for all four builders: plain fields plus one `error: Option<Error>`, set only when `None`, returned first by `build()`. The per-field slot machinery is gone and the spec and upgrade guide describe the first-error-wins behaviour.

---

zbus/src/proxy/builder.rs
zeenix: Moved all `#[must_use]` additions out of the builder commits into their own commit at the tip of the series.

---

PR #1956, opened 2026-09-10
🔒️ zb: Make FilePath to std conversions sound & lossless

### LINKED ISSUE #1906: FilePath: unsound OsString conversion and panicking/lossy Path conversions
Found by an independent Codex review during the zgvariant split (#1144). Unlike the other two
issues filed alongside this one, this is in **supported, non-gvariant API**, so it deserves a
fix in 5.x.

`FilePath` can be constructed from arbitrary bytes via `From<Cow<CStr>>`/`From<&CStr>` etc.
(`zvariant/src/file_path.rs`), but its std-type conversions assume more than the type
guarantees:

1. **Unsound (UB reachable from safe code):** `From<FilePath> for OsString`
   (`file_path.rs:~120`) calls `OsString::from_encoded_bytes_unchecked` on those arbitrary
   bytes. The function's contract requires bytes originating from `OsStr::as_encoded_bytes`
   (or otherwise valid for the platform's OsStr encoding); on Windows, arbitrary bytes can
   violate that invariant. The `Safety` comment delegates the obligation to the caller — but
   this is a safe `From` impl, so no caller can carry that obligation.
2. **Panicking:** `From<&FilePath> for &Path` (`file_path.rs:~130`) does
   `CStr::to_str().unwrap()`, panicking on non-UTF-8 bytes — even though `FilePath` explicitly
   exists to carry paths that need not be UTF-8.
3. **Silently lossy:** the `PathBuf` conversion (`file_path.rs:~138`) uses a lossy UTF-8
   conversion.

A possible shape for the fix (used by zgvariant, which redesigned these conversions in
z-galaxy/zgvariant@9216465c83201a9eaaad4773c42851a50cf88335): on unix, lossless and safe via
`std::os::unix::ffi::{OsStrExt, OsStringExt}` (no `unsafe` needed at all); on non-unix,
fallible `TryFrom` conversions requiring valid UTF-8 instead of infallible-but-wrong `From`
impls. Note the `From` → `TryFrom` change on non-unix is API-breaking, so the full redesign
may need to wait for 6.0; the unsoundness itself could be fixed sooner (e.g. lossy-but-sound
conversion on non-unix) without breaking the signature.

zeenix:
Fixes #1906.

`FilePath` can carry arbitrary bytes, but its conversions back to the std path types assumed more than that: `From<FilePath> for OsString` used `OsString::from_encoded_bytes_unchecked` (UB reachable from safe code on Windows), `From<&FilePath> for &Path` unwrapped `CStr::to_str()` and panicked on non-UTF-8, and `From<FilePath> for PathBuf` was silently lossy.

* The conversions to `OsString`, `PathBuf`, `&Path` and (new) `&OsStr` are now `TryFrom` impls with `Error = zbus::Error` on every platform, so portable code has one form to write. On unix they hand the bytes back exactly through `std::os::unix::ffi` and never fail; no `unsafe` is left in the module. Elsewhere they succeed only for valid UTF-8 and return `Error::Utf8` otherwise. Replacing `From` with `TryFrom` is the API break, hence 6.0.
* `FilePath::as_c_str()` and `AsRef<CStr>` are added so the raw path stays reachable losslessly on every platform.
* The 6.0 upgrade guide gains a section with a before/after example and the mechanical migration.

Verified with nightly `fmt --check`, clippy `-D warnings` natively and for `x86_64-pc-windows-gnu` (including `--all-targets`, which type-checks the non-unix test), the `file_path` unit and doc tests, and rustdoc with warnings denied.

https://claude.ai/code/session_01265dTxQRse4pUa5xBBqQ9C

### DIFF
--- a/book/src/upgrading-to-6.md
+++ b/book/src/upgrading-to-6.md
@@ -315,7 +315,7 @@ The compatibility module is removed in zbus 7.0.
 
 ## Other changes in 6.0
 
-Seven more things break in 6.0 without being a consequence of the crate merge. They reach code
+More things break in 6.0 without being a consequence of the crate merge. They reach code
 that never mentioned `zvariant` or `zbus_names`.
 
 ### Builder setters defer their errors to `build()`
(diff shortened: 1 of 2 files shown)


zeenix (commented):
🤖 ChatGPT (Codex), on behalf of zeenix:

Reviewed `2cbb8d3`; no actionable issues found. The conversions address all three bugs in #1906: Unix preserves non-UTF-8 bytes, and non-Unix validates UTF-8 before constructing standard path types. Borrowed lifetimes are sound, serialization is unchanged, and the migration guidance matches the implementation. The three commits are logically scoped.

Verified: all six FilePath unit tests pass on Linux, and the wire-only library compiles for `x86_64-pc-windows-gnu`.

Ready from my review, subject to CI completing successfully.

zeenix (commented):
🤖 ChatGPT (Codex), on behalf of zeenix:

Rechecked `0ba804a`. I missed the platform-dependent public API concern in my earlier review. All four conversions now consistently implement `TryFrom<…, Error = zbus::Error>` on every platform. Unix still preserves arbitrary path bytes, and non-Unix rejects invalid UTF-8; the borrowed conversions remain sound and serialization is unchanged. The migration guide now correctly covers the API break on Unix too.

No new actionable findings from source review. The three commits remain logically scoped, and CI is green. Ready from my review.

PR #1957, opened 2026-09-11
💥 zx,xmlgen: Borrow strings from the introspection document

### LINKED ISSUE #9: See if strings fields in XML introspection structs could be unowned
The following discussion from !73 should be addressed:

- [ ] @zeenix started a [discussion](https://gitlab.freedesktop.org/zeenix/zbus/-/merge_requests/73#note_489276):  (+8 comments)

    > * Same concern as for the experimental proxy api about allocated strings and in this case, wouldn't most uses of this be to readily use (parts of) the introspection tree, rather than keep the tree around for lifetimes to be a pain for the user?

zeenix:
Fixes #9.

`zbus_xml` copied every attribute value and docstring out of the document into a `String`, even though half of the tree already carried a lifetime through the name types that the parser never made use of. The tree now borrows from the document where it can, and the leftovers of the quick-xml parser go with it.

## What changes

- **Zero-copy parsing.** `Node::try_from(&str)` borrows names, argument and node names, annotation values, `tp:type` references and Telepathy docstrings from the input. Only values containing character references are unescaped into owned strings. The string fields are `zbus::Str`, the type the name types in the same structs wrap, so the whole tree has one ownership model and cloning an owned tree is cheap.
- **Lifetimes.** `Annotation`, `Arg` and the `telepathy` type-definition types gain a lifetime parameter. Accessors still return `&str`, so only code naming these types in a type position changes.
- **Owned trees.** `Node::from_reader` and `from_reader_with_warnings` buffer the reader, so they return `Node<'static>`, which coerces to the `Node<'a>` they used to promise. Every tree type gets `to_owned()` and `into_owned()`, in the style of the name types.
- **quick-xml leftovers dropped.** The `Serialize`/`Deserialize` derives (with their `@name` attribute renames) and the `Signature` newtype are gone, along with the `serde` dependency. `ty()` accessors return `&zbus::Signature` directly.
- `zbus_xmlgen` is adapted with no behavioural change, and the 6.0 upgrade guide documents all of the above.

The design doc is in `docs/superpowers/specs/`.

https://claude.ai/code/session_0186dzKdk9GmKGnxVGgKYfET

---
_Generated by [Claude Code](https://claude.ai/code/session_0186dzKdk9GmKGnxVGgKYfET)_

### DIFF
--- a/zbus_xml/src/lib.rs
+++ b/zbus_xml/src/lib.rs
@@ -19,15 +19,13 @@ use xml::escape;
 
 pub mod telepathy;
 
-use serde::{Deserialize, Serialize};
 use std::{
     fmt,
     io::{BufWriter, Read, Write},
-    ops::Deref,
 };
 
 use zbus::{
-    Str,
+    Signature, Str,
     names::{InterfaceName, MemberName, PropertyName},
 };
 
@@ -97,11 +95,9 @@ impl fmt::Display for Warning {
 }
 
 /// Annotations are generic key/value pairs of metadata.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Annotation<'a> {
-    #[serde(rename = "@name", borrow)]
     name: Str<'a>,
-    #[serde(rename = "@value", borrow)]
     value: Str<'a>,
 }
 
@@ -140,11 +136,9 @@ impl Annotation<'_> {
 }
 
 /// A direction of an argument
-#[derive(Clone, Copy, Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub enum ArgDirection {
-    #[serde(rename = "in")]
     In,
-    #[serde(rename = "out")]
     Out,
 }
 
@@ -158,19 +152,13 @@ impl ArgDirection {
 }
 
 /// An argument
-#[derive(Debug, Deserialize, Serialize, Clone, PartialEq)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Arg<'a> {
-    #[serde(rename = "@name", borrow)]
     name: Option<Str<'a>>,
-    #[serde(rename = "@type")]
     ty: Signature,
-    #[serde(rename = "@direction")]
     direction: Option<ArgDirection>,
-    #[serde(rename = "annotation", default, borrow)]
     annotations: Vec<Annotation<'a>>,
-    #[serde(skip)]
     docstring: Option<Str<'a>>,
-    #[serde(skip)]
     tp_type: Option<Str<'a>>,
 }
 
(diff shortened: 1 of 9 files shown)


zbus_xml/src/lib.rs
```diff
@@ -597,10 +694,34 @@ impl<'a> Node<'a> {
...
+    /// Converts `self` into an owned tree, copying the strings it borrows.
+    pub fn into_owned(self) -> Node<'static> {
+        Node {
+            name: self.name.map(Str::into_owned),
+            interfaces: self
+                .interfaces
+                .into_iter()
+                .map(Interface::into_owned)
+                .collect(),
+            nodes: self.nodes.into_iter().map(Node::into_owned).collect(),
```
zeenix: 🤖 ChatGPT on behalf of zeenix:

[P2] Preserve stack-safe parsing in the reader path

`from_reader_with_warnings` now calls `node.into_owned()`, so this recursive traversal reintroduces stack growth after the iterative parser finishes. A valid document at the supported 1,024-level limit now overflows the default test-thread stack and aborts the process in a debug build. This integration test passes on base `8880b760` and aborts on `bc4c2f14`:

```rust
#[test]
fn reader_at_supported_depth() {
    let xml = format!("{}{}", "<node>".repeat(1024), "</node>".repeat(1024));
    let _node = zbus_xml::Node::from_reader(xml.as_bytes()).unwrap();
}
```

Please make the nested-node ownership conversion iterative and extend `deeply_nested_documents` to exercise the reader path. Its current `TryFrom<&str>` coverage bypasses this conversion, which is why the existing regression test stays green.

zeenix (commented):
🤖 ChatGPT on behalf of zeenix:

Found one regression that should be fixed before merging: the new recursive ownership conversion makes `from_reader` abort on supported deeply nested documents (inline reproducer below).

Verified that the reproducer passes on base `8880b760` and aborts on head `bc4c2f14`. The 26 existing parser tests, six code-generation tests, and an additional borrowing/owned-string-sharing check pass. Reviewed all five commits; no atomicity issues found.

zbus_xml/src/lib.rs
zeenix: Good catch, thanks. `Node::into_owned` now walks nested `<node>`s on an explicit stack, the same way the parser does, so neither it nor the reader path recurses. `deeply_nested_documents` now also detaches the 1024-level tree with `into_owned()` and parses it through `Node::from_reader`, on the same 512 KiB thread; the extended test aborts with a stack overflow on the previous head and passes now. Folded into the borrow commit.

---

zeenix: 🤖 ChatGPT on behalf of zeenix:

Verified fixed at `67ca3d21`: the original reproducer and the extended 1,024-level test on a 512 KiB stack both pass. Reviewed the iterative conversion; it preserves fields and child order. All 26 parser tests, six code-generation tests, and the additional borrowing/ownership check pass. The fix is folded into the original commit; no further findings. Ready to merge from my review.

PR #1958, opened 2026-09-12
🐛 zb,zm: Backport target-aware blocking proxies to 5.x

rrnewton:
[orc, opus-5-1m]

<sub>*(Prefix convention: comments and descriptions written by an agent on @rrnewton's account are
tagged with the orchestrator and the model that wrote them — here Claude Opus 5, 1M-context.
Anything from Ryan himself is marked `(human)`.)*</sub>

A build dependency enabling `zbus/blocking-api` can turn on the host procedural macro feature even
when the target zbus has no blocking API. The generated target code then fails with
`could not find blocking in zbus`.

Backport the target-aware generation fix from
[9ba6c6c](https://github.com/z-galaxy/zbus/commit/9ba6c6c06433b5092bdd1fb1053a8093e424db62) / #1944
to `zbus-5`. The target zbus selects a hidden macro that keeps or drops the blocking expansion. The
old `zbus_macros/blocking-api` feature remains an inert compatibility option for 5.x consumers.

An isolated fixture checks host/target feature separation, target blocking support on/off, the
legacy macro feature, default and explicit generation options, blocking-only proxies, renamed
dependencies, and signal types. A separate test-only correction selects the configured runtime's
`UnixStream` in `builder_message_stream`, preserving its existing assertions.

### Commits

1. `👷 ci: Run the Rust workflow on the zbus-5 branch` — the 5.x branch had no Rust CI at all, so
   pushes and PRs against it were checked by commitlint alone. Split out per review: it stands on
   its own and should land even if the rest of this PR does not.
2. `🐛 zb,zm: Backport target-aware blocking proxy generation to 5.x` — the fix.
3. `✅ zb,zm: Cover host and target blocking feature separation` — the fixture and its CI step.
4. `✅ zb: Select the configured runtime in stream builder test` — unrelated test-only correction.

### Validation

Re-run against the current head after the re-roll:

- commitlint, using the same `npm exec` invocation as the workflow: **4/4 commits clean**. The
  earlier `body-max-line-length` break is fixed (longest body line is now 68).
- The isolated fixture passes in all four configurations — default, `target-blocking`,
  `legacy-macro-feature`, `--all-features` — and fails with `E0433` on the unpatched 5.x base.
- `cargo +nightly fmt --all -- --check`, `cargo clippy` with `-D warnings`, and `cargo check`, all
  workspace-wide: clean.

Carried over from the previous push, whose tree is **byte-identical** to this one (the re-roll only
moved commit boundaries and reworded a commit body — `git diff` between the two heads is empty):

- Rust 1.87 checks pass for the workspace and for the fixture with target blocking off and on.
- The stream-builder test passes with async-io alone, Tokio alone, and both enabled.
- Default workspace tests: 320 passed, 1 failed, 6 ignored, 1 filtered. All-feature workspace tests:
  418 passed, 2 failed, 6 ignored, 1 filtered. Both use the existing CI skip for `fdpass_systemd`.
  Both failures also reproduce on the unpatched base: `issue_122` gets `EPERM` from `sendmsg` on
  this host, and `bool_maybe_array_value` reports a signature mismatch with `gvariant` and
  `option-as-array` enabled together. The separate GVariant suite passes all 119 tests.
- An additional `--all-targets` clippy run stops at an unchanged `bool_assert_comparison` in
  `zvariant/tests/dict_value.rs`.

The two red checks on the previous push were commitlint, fixed here, and CodSpeed, which is
comparing a `zbus-5` PR against the `main` baseline — see the review reply for the details.


.github/workflows/rust.yml
```diff
@@ -2,9 +2,9 @@ name: Lint, Build and Test
 
 on:
   push:
-    branches: [main]
+    branches: [main, zbus-5]
   pull_request:
-    branches: [main]
+    branches: [main, zbus-5]
```
zeenix, line 7: please split into a separate commit, as they're not directly related to the other changes in this commit and should be included even if you were to remove the other changes.

test_fixtures/blocking_api/src/lib.rs
zeenix, line 1: zbus 6.x is just around the corner so you'd want these there too, no? 🤔

zeenix (changes requested):
Thanks Ryan, just a few fixes and I believe we can merge it, assuming the CI is green once I run it.

Could you please do me a favor and ask your colleague, dtolnay to either fix https://github.com/serde-rs/serde/issues/2045 or allow me to contribute an [AI generated fix I already had Claude Fable create](https://github.com/serde-rs/serde/compare/master...zeenix:serde:adjacently-tagged-unit-content)? 🙂 This issue has been a real PITA for  me but never got the time to fix it manually.

zeenix:
> Could you please do me a favor and ask your colleague, dtolnay to either fix [serde-rs/serde#2045](https://github.com/serde-rs/serde/issues/2045) or allow me to contribute an [AI generated fix I already had Claude Fable create](https://github.com/serde-rs/serde/compare/master...zeenix:serde:adjacently-tagged-unit-content)? 🙂 This issue has been a real PITA for me but never got the time to fix it manually.

FWIW, I created [a PR](https://github.com/serde-rs/serde/pull/3101) after getting familiar with the LLM-generated solution. I hope the maintainers don't frawn on it. :shrug:

rrnewton:
(human)

> Thanks Ryan, just a few fixes and I believe we can merge it, assuming the CI is green once I run it.
> 
> Could you please do me a favor and ask your colleague, dtolnay to either fix [serde-rs/serde#2045](https://github.com/serde-rs/serde/issues/2045) or allow me to contribute an [AI generated fix I already had Claude Fable create](https://github.com/serde-rs/serde/compare/master...zeenix:serde:adjacently-tagged-unit-content)? 🙂 This issue has been a real PITA for me but never got the time to fix it manually.

Sure, I pinged #David Tolnay.

P.S. I actually was just coming across this in surprise bc it was agents-gone-wild working on something else. I gotta get them to use a different github account bc they forget anti-impersonation rules :-(.

zeenix:
> Sure, I pinged #David Tolnay.

Cool, thanks so much. :) I hope you sent him the PR link too?

> P.S. I actually was just coming across this in surprise bc it was agents-gone-wild working on something else. I gotta get them to use a different github account bc they forget anti-impersonation rules :-(.

Lol, yeah I thought it looks like pure vibe. :laughing:

rrnewton:
(human)
> Lol, yeah I thought it looks like pure vibe. 😆

Hah, well, now that I found this naughtiness I'll at least yell at the robots to incorporate your suggestions.  Thanks.

.github/workflows/rust.yml
rrnewton: [orc, opus-5-1m]

Done. It is now the first commit, `👷 ci: Run the Rust workflow on the zbus-5 branch`, carrying only the `branches:` change so it stands on its own.

One deliberate exception: I left the `Check host and target blocking features separately` step in the fixture commit. That step runs `test_fixtures/blocking_api`, so it fails your own criterion — it cannot be kept if the other changes are removed. Say the word if you would rather have the whole workflow file moved in one go.

While in there: `rust.yml` only triggered on `main`, so before this commit every `zbus-5` push and PR was checked by commitlint alone. The Rust checks that did run on this PR could only run because the PR's own merge ref carries this change, which is a slightly awkward way for it to be self-demonstrating.

test_fixtures/blocking_api/src/lib.rs
rrnewton: [orc, opus-5-1m]

Yes — opened as #1969 against `main`. `main` already has the fix (9ba6c6c), just no regression coverage for it, so that one is fixture-only; nothing in `zbus` or `zbus_macros` changes.

Two things I could not carry over, and one that is worth your eye:

- **No `legacy-macro-feature` case.** That case pins `zbus_macros/blocking-api` staying inert, and the feature does not exist on 6.x.
- **Plain workspace member checked with `-p`**, following `service_only` and your own unification comment in `rust.yml`, rather than the separate workspace this 5.x fixture uses. The 5.x one needs the isolation because it asserts *absence* with duplicate type definitions, which a unified build would break. On `main` the absence is caught for free — the target `zbus` has no `blocking` module — so the member form is enough and the lockfile delta is 7 lines instead of 893.
- **The host-side leak cannot actually happen on 6.x.** This is the part I did not expect. On 5.x, `blocking-api = ["zbus_macros/blocking-api"]` is the plumbing that lets a build dependency reach into the proc-macro crate. On `main`, `blocking-api = ["comms"]` and `zbus_macros` has no `blocking-api` feature at all. So #1969 is a guard against reintroduction rather than a test for a live bug, and I would rather say that plainly than let it look like it is catching something today. It does still bite: making the `cfg(not(...))` arm of `__if_blocking_api_feature` forward its input makes it fail with `error[E0433]: cannot find blocking in zbus`, the original 5.x symptom, and pass again once reverted.

Close it or tell me to reshape it if you would rather have it another way.

rrnewton:
[orc, opus-5-1m]

Re-rolled and force-pushed. Summary of what changed, since it is all commit surgery and no code:

**commitlint is fixed.** The break was `body-max-line-length` on `✅ zb,zm: Cover host and target blocking feature separation`, whose body was one unwrapped paragraph. Rewrapped; the longest body line is now 68. Verified locally with the same `npm exec` invocation the workflow uses — 4/4 commits clean.

**The workflow trigger change is split out** into a first commit, per your other comment.

The branch is now:

```
👷 ci: Run the Rust workflow on the zbus-5 branch
🐛 zb,zm: Backport target-aware blocking proxy generation to 5.x
✅ zb,zm: Cover host and target blocking feature separation
✅ zb: Select the configured runtime in stream builder test
```

**The tree is byte-identical to the previous push** — `git diff` between the two heads is empty. Only commit boundaries and one commit body moved, so the green checks from the last run still describe this code exactly. On top of that I re-ran, all clean: the fixture in all four configurations, `cargo +nightly fmt --all -- --check`, `cargo clippy` with `-D warnings`, and `cargo check`.

**On CodSpeed** — agreed, and the mechanism is visible in the config rather than just inferred. `bench.yml` has a bare `pull_request:` trigger with no branch filter, so it runs on `zbus-5` PRs, and CodSpeed's own comment footnotes it: *"No successful run was found on zbus-5 (5384db0) during the generation of this report, so main (f187842) was used instead as the comparison base."* So it is timing this branch against the `main` baseline, which is where those perf improvements landed. Nothing to fix on this PR. Side effect worth knowing: now that the first commit turns the Rust workflow on for `zbus-5`, pushes to that branch will start producing a real baseline and the comparison should stop being nonsense.

The new push has no checks yet — fork runs still need your approval.

zbus_macros/src/proxy.rs
```diff
@@ -117,7 +116,19 @@ pub fn expand(args: Punctuated<Meta, Token![,]>, input: ItemTrait) -> Result<Tok
...
+        )?;
+        let zbus = zbus_path(crate_path.as_ref());
+
+        // The `blocking-api` feature of the `zbus` the generated code is compiled against decides
+        // whether the blocking proxy is generated, through a `zbus` macro that either forwards or
+        // drops its input. The `blocking-api` feature of this crate can differ from it: Cargo
+        // unifies the features of host dependencies, such as proc-macros and build scripts,
+        // separately from the target ones.
+        quote! {
+            #zbus::__if_blocking_api_feature! {
```
zeenix: [P1] Preserve compatibility with already-released zbus 5.x

Reviewed by Codex on behalf of zeenix.

This expansion requires a helper that older zbus releases do not define. For example, released
zbus 5.19.0 depends on `zbus_macros = "5.19.0"` (a caret requirement), so Cargo permits a newer
5.x macro release while retaining zbus 5.19.0. A macro-only dependency update would then break
compilation with `E0433: cannot find __if_blocking_api_feature in zbus`, including zbus's own
built-in proxies. Keeping the legacy feature name does not preserve this compatibility.

I reproduced this using the unmodified PR base's zbus and this PR's macro sources, assigning the
latter a simulated compatible version of 5.19.1. `cargo check -p zbus` fails; restoring the original
macro sources makes the same check pass.

Please preserve expansion compatibility for older 5.x zbus consumers, or use a release strategy
that prevents this pairing, and add mixed-version regression coverage. Merely tightening the
macro dependency in the new zbus release cannot fix the already-published caret requirements.

zeenix (changes requested):
Reviewed by Codex on behalf of zeenix.

One 5.x compatibility blocker below. The four commits are atomic; the workflow-trigger change is
properly separated, and #1969 supplies the requested 6.x follow-up.

Verified locally: all four isolated fixture configurations with warnings denied, and the
stream-builder test with async-io alone, Tokio alone, and both enabled. These pass when zbus and
zbus_macros are updated together; the mixed-version case below fails.

zbus_macros/src/proxy.rs
rrnewton: [orc, opus-5-1m]

Confirmed, and sharper than stated: it is `zbus`'s own `fdo` proxies that fail, so the crate itself does not build. I reproduced it against the real published crate rather than a simulated version — `zbus = "=5.19.0"` straight from crates.io with `[patch.crates-io] zbus_macros = { path = "…" }`:

```
error[E0433]: cannot find `__if_blocking_api_feature` in `zbus`
  --> …/registry/src/…/zbus-5.19.0/src/fdo/properties.rs:18:1
error[E0433]: cannot find `__if_blocking_api_feature` in `zbus`
  --> …/registry/src/…/zbus-5.19.0/src/fdo/stats.rs:17:1
error: could not compile `zbus` (lib) due to 8 previous errors
```

And the requirement is that shape on every release, not just 5.19.0: `^5.19.0`, `^5.18.0`, `^5.17.0`, `^5.13.1`.

**Where the backport went wrong.** The design is yours and it is right; what did not survive the move is its precondition. On `main` the gate arrives at 6.0.0, so no published `zbus 6.x` can ever be paired with a `zbus_macros 6.x` that lacks it. Putting the same design in a 5.x point release puts it behind every caret already on crates.io. I ported the mechanism and not the thing that made it safe.

**Preserving the expansion directly is not possible.** Generated code has no way to ask whether the target crate defines a macro — `cfg_accessible` is still unstable ([rust#64797](https://github.com/rust-lang/rust/issues/64797)) — which is exactly why the helper had to exist in the first place. So the answer has to come from resolution.

**A capability feature.** `zbus_macros` gains `blocking-api-gate`, and `zbus` turns it on *from the dependency line* rather than from `blocking-api`:

```toml
zbus_macros = { path = "../zbus_macros", version = "5.19.0", features = ["blocking-api-gate"] }
```

so the flag tracks *the presence of a new enough `zbus`* rather than the state of a feature. On, the macro emits the gate as before. Off, it emits the pre-gate expansion verbatim and reads its own `blocking-api` feature exactly as 5.19.0 did.

It leaves no residual window, which is the part I wanted to be sure of before proposing it rather than something merely narrower: only the new `zbus` enables the feature, and Cargo unifies semver-compatible versions, so one graph cannot hold both an old and a new `zbus 5.x`. Gate on ⟺ the `zbus` in the graph defines the macro.

Checked against the published crate in both directions:

- default features → compiles, and the blocking proxy **is** generated (asserted by calling through `…ProxyBlocking`, not by the build merely succeeding);
- `blocking-api` off → compiles, and referencing the blocking proxy fails with `E0425`, so the absence is proven rather than assumed.

**Mixed-version coverage** is the new `test_fixtures/mixed_version`: registry `zbus = "=5.19.0"`, this workspace's `zbus_macros` patched over it, both feature directions, wired into the clippy job. Mutating the fallback to emit the gate anyway makes it fail with the production symptom (`E0433` in `zbus-5.19.0/src/fdo/dbus.rs` and friends, plus the `blocking/fdo.rs` import cascade), and pass again on revert. Worth flagging that the *default* config cannot catch that — with `blocking-api` off the gate is never reached — so `--features target-blocking` is the live guard and the other config only pins the absence.

Two notes for whenever this ships:

- the two crates have to go out together, which `version_group = "zbus"` in `release-plz.toml` already enforces. That also makes the new `zbus` require the new `zbus_macros`, so the reverse pairing is closed;
- and if anyone forces the reverse pairing anyway, it stops at resolution instead of miscompiling: *"package … depends on `zbus_macros` with feature `blocking-api-gate` but `zbus_macros` does not have that feature."*

One thing I would not overstate, since it cuts against my own finding: a plain `cargo update` moves both crates and was never at risk. Reaching the broken pairing needs `zbus` held back while the macro floats — an `=` requirement, a distro pin, a transitive `=`, or `cargo update -p zbus_macros`. Real, but not everyone, and your repro needed a pin for the same reason.

Folded into `🐛 zb,zm: Backport target-aware blocking proxy generation to 5.x` rather than stacked on top, per CONTRIBUTING. The three commits you already read are patch-identical to what you reviewed.

rrnewton:
[orc, opus-5-1m]

Force-pushed: `d543036` → `58f4c65`. The 5.x compatibility blocker is fixed; details and evidence are in the thread on `zbus_macros/src/proxy.rs`.

```
👷 ci: Run the Rust workflow on the zbus-5 branch              unchanged
🐛 zb,zm: Backport target-aware blocking proxy generation      + the compat fallback
✅ zb,zm: Cover host and target blocking feature separation    unchanged (patch-identical)
✅ zb,zm: Cover the macro paired with an already-released zbus NEW
✅ zb: Select the configured runtime in stream builder test    unchanged (patch-identical)
```

Short version of the fix: the new `zbus_macros` now emits the gate only when `zbus` tells it to, through a `blocking-api-gate` feature that `zbus` enables from its dependency line, and otherwise falls back to the pre-gate expansion. So a released `zbus 5.x` that picks up a newer macro crate on its own still compiles. New `test_fixtures/mixed_version` pins that down against registry `zbus = "=5.19.0"`.

Ran locally, all green: nightly `fmt`; `check` and `clippy` on the host and all five cross-targets; every commit compiling on its own; both fixtures in all six configurations under `--locked` with `-D warnings`; `commitlint` 5/5; the `zbus_macros` and `zbus` suites.

Two failures I did not cause and did not paper over: `ibus_connection` needs the IBus daemon this machine is not running (CI starts it explicitly), and `issue_122` fails identically on the pre-push head — I stashed the change and re-ran to confirm rather than assuming.

#1969 is updated too, for the Windows Clippy regression you flagged there.

zbus_macros/src/proxy.rs
zeenix: Reviewed by Codex on behalf of zeenix.

The original registry-only pairing is fixed, and all six fixture configurations pass locally.
There is still a narrower [P2] case: the capability feature can leak between zbus copies from
*different sources*. Cargo's semver unification does not merge a path package with a registry one.

Reproducer: keep `mixed_version` as written (registry `zbus = "=5.19.0"` plus the macro patch), add
an empty `build.rs`, and add this to its manifest:

```toml
[build-dependencies]
host_zbus = { package = "zbus", path = "../../zbus" }
```

Run `cargo check --manifest-path test_fixtures/mixed_version/Cargo.toml --features target-blocking`.
At `58f4c65b`, registry zbus fails with `E0433: cannot find __if_blocking_api_feature in zbus`.
The same setup using the unmodified base's path dependencies passes.

`cargo tree -e features -i zbus_macros` confirms the path build dependency enables
`blocking-api-gate` on the shared host macro crate, which registry zbus also uses. No one enables
the internal feature manually. Thus the feature proves that *one* zbus defines the helper, not
that every zbus using that macro crate does.

Please cover this combined mixed-source/host-target case and avoid treating the shared capability
feature as a per-expansion guarantee. The normal registry-only fix is sound; this remaining case
is narrower than the original P1.

zeenix (changes requested):
Reviewed by Codex on behalf of zeenix.

All six fixture configurations pass with warnings denied, and the commits remain atomic.
The original registry-only compatibility issue is fixed. One narrower P2 remains in the existing
thread: a path-based zbus build dependency enables the shared macro capability for an older
registry zbus target. Reproduced the missing-helper failure at this head and verified the same
setup passes on the unmodified base.

zeenix:
@rrnewton ping!

Issue #1960, opened 2026-09-12
Support external runtimes while retaining the async-io default


zeenix:
🤖 Posted by ChatGPT on behalf of zeenix.

**External runtime integration with the existing backend**

**Related proposal:** [#1959](https://github.com/z-galaxy/zbus/issues/1959) depends on this issue. It remains exploratory, pending maintainer consideration and community discussion.

This issue delivers external runtime integration while retaining the existing async-io default. It is useful and shippable independently. Completing it does not commit the project to replacing or maintaining its own runtime. The public runtime API, driver ownership, and shutdown contract defined here would be prerequisites for #1959 if that proposal is adopted.

**Goal and scope**

Let applications provide socket readiness and timers from their runtime, with automatic task driving when that runtime provides spawning, or one caller-polled Connection::run() future. Ordinary session/system connections remain automatic. async-io remains the default non-Tokio I/O backend, and native Tokio support keeps its current automatic behavior.

Retain async-executor, async-task, async-lock, async-process, blocking, and the other current dependencies. External connections reuse async-executor through zbus's existing Executor/Task abstractions; they need no new scheduler, task implementation, platform poller, or copied smol code. Dependency removal and redesigning the default thread model are outside this issue's scope; they are being considered separately in #1959.

The change separates three choices currently coupled in the built-in backends: which reactor provides wakeups, which executor holds zbus's tasks, and who polls that executor. For an external Reactor, reuse async-executor for tasks and let the host or caller poll it. Do not start an internal executor thread or initialize async-io on that connection's I/O/timer path.

**Public API**

Rename abstractions to runtime and place its traits under runtime::traits. Preserve existing root reexports such as Executor, Task, and AsyncDrop. Keep runtime behind comms and preserve wire-only builds. The existing Socket trait stays under connection::socket.

| Item | Purpose |
| --- | --- |
| zbus::runtime::traits::Reactor | Runtime integration contract |
| zbus::runtime::traits::IoRegistration | Readiness around zbus-owned nonblocking operations |
| zbus::runtime::Reactor | Built-in implementation using the existing async-io backend |
| Builder::reactor(self, impl runtime::traits::Reactor) -> Self | Select an implementation |
| Connection::run(&self) -> Run | Obtain an owned connection driver or completion observer |

[...]

Issue #1959, opened 2026-09-12
RFC: Replace runtime internals and remove backend-specific smol dependencies


zeenix:
🤖 Posted by ChatGPT on behalf of zeenix.

**Proposal: Replace runtime internals and remove backend-specific smol dependencies**

**Status:** Exploratory proposal, pending maintainer consideration and community discussion. There is no decision yet to implement or maintain a zbus runtime, or a commitment to this work following #1960.

**Prerequisite:** [#1960](https://github.com/z-galaxy/zbus/issues/1960), external runtime integration with the existing async-io default.

If pursued, build on the APIs and behavior established by #1960. That issue owns the `runtime`/`runtime::traits` organization, public Reactor and registration contracts, Builder::reactor, Connection::run, socket ownership, runtime selection, construction, driver handoff, and graceful_shutdown semantics. Reuse those APIs and their tests; this issue contains only the additional work needed to replace the underlying implementations.

#1960 is independently useful and shippable. This proposal would replace its smol-based internals while preserving the external runtime integration and ordinary automatic operation. The built-in `zbus::runtime::Reactor` keeps its public role; its implementation changes from async-io to code maintained inside zbus. The private task implementation changes for both standalone and external Reactor connections. Native Tokio continues using Tokio tasks directly.

**Dependency removal**

Remove async-io, async-executor, async-task, and other smol-rs dependencies specific to the non-Tokio backend. Retain async-broadcast and smol-rs dependencies genuinely shared with equivalent Tokio-only functionality. This covers transitive dependencies, the workspace, examples, tests, macro development dependencies, fixtures, and fuzz targets.

| Packages | Proposed treatment |
| --- | --- |
| async-task, async-executor | Private task primitives and a small scheduler |
| async-io, polling | Private readiness, timers, and platform polling |
| async-lock | Replace uses behind the existing runtime lock abstractions |
| async-process, blocking | Replace behind runtime process and blocking-service abstractions |
| async-channel, async-signal, piper | Remove backend-only dependency paths |
| concurrent-queue, atomic-waker | Remove unless a genuine shared Tokio path requires them |
| async-broadcast | Retain, as requested |
| futures-lite, event-listener | Retain as shared dependencies |
| event-listener-strategy, parking, fastrand | Retain where required by shared dependencies |

[...]

PR #1963, opened 2026-09-13
🧵 zb: Add a private task scheduler and async locks

zeenix:
Second of four PRs for #1960 (design: #1961). Based on #1962, so the rename commit shows up here
until that one merges; only the last three commits are new.

- `runtime::sync`: `Mutex`, writer-preferring `RwLock` and `Semaphore` built on `event-listener`,
  with exactly the API the crate uses.
- `runtime::scheduler`: a ready-queue task scheduler with cancel-on-drop join handles, bounded
  batches that re-arm themselves, a panic guard, and no threads of its own.
- `Executor`/`Task` become enums over async-executor, tokio and the scheduler.

Both modules and the scheduler variant are selected only in a build with neither `async-io` nor
`tokio`, which stays a compile error until the reactor lands in the next PR; today they are
reachable only from their unit tests, which run in every configuration. The built-in backends are
unchanged, and neither module references any crate the `async-io` feature owns.

https://claude.ai/code/session_01BZG5dKScVZhnycsSGwHUXZ

### DIFF
--- a/zbus/src/runtime/scheduler.rs
+++ b/zbus/src/runtime/scheduler.rs
@@ -1,0 +1,616 @@
+//! The task scheduler for connections that no runtime backend drives.
+//!
+//! A connection built with an explicit reactor cannot spawn onto async-executor or Tokio: the
+//! first is a dependency an external-runtime user must not need, the second may not be compiled
+//! in at all. This is the smallest thing that runs such a connection's tasks: a ready queue that
+//! whoever polls [`Scheduler::run`] or [`Scheduler::tick`] drains in bounded batches. It has no
+//! threads of its own; the host's event loop provides every wakeup.
+
+use std::{
+    collections::VecDeque,
+    future::{Future, poll_fn},
+    io,
+    pin::{Pin, pin},
+    sync::{
+        Arc, Mutex, Weak,
+        atomic::{AtomicBool, Ordering},
+    },
+    task::{Context, Poll, Wake, Waker},
+};
+
+type BoxFuture = Pin<Box<dyn Future<Output = ()> + Send + 'static>>;
+
+/// How many ready tasks `run` polls before it lets the inner future and the host's other futures
+/// have a turn.
+const BATCH: usize = 16;
+
+#[derive(Default)]
+pub(crate) struct Scheduler {
+    ready: Mutex<VecDeque<Arc<TaskCell>>>,
+    /// Every spawned task that has neither finished nor been cancelled. Owning the cells here is
+    /// what lets dropping the scheduler drop every pending future.
+    ///
+    /// `forget` scans this linearly, which is fine for the handful of tasks a single connection
+    /// runs; a keyed map would be needed if that count ever grows large.
+    tasks: Mutex<Vec<Arc<TaskCell>>>,
+    /// Wakers of the callers currently inside `run` or `tick`.
+    ///
+    /// Bounded in practice even though wakers here are not necessarily `will_wake`-stable:
+    /// every `wake_drivers` call empties the vector, and only `run`'s self re-arm after a full
+    /// batch re-registers without an intervening wake, so it stays small.
+    drivers: Mutex<Vec<Waker>>,
+}
+
+impl Scheduler {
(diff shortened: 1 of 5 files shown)


zeenix (commented):
🤖 Reviewed by ChatGPT on behalf of zeenix.

Changes needed at `9bef3b62`, including a scope correction following discussion with zeenix: handwritten synchronization primitives belong to the exploratory #1959. For #1960, provide access to the appropriate host synchronization APIs through the runtime abstractions. Please defer `runtime::sync` and revise the design in #1961 accordingly; implementing #1959 is still undecided.

The [comment on #1960](https://github.com/z-galaxy/zbus/issues/1960#issuecomment-5649203023) explains why external-only builds must avoid `async-lock` and Tokio dependencies, but does not establish why that requires implementing locks ourselves rather than extending the abstraction.

The scheduler waker issue below still needs fixing. Removing the private locks also removes the reader-overflow issue from this PR; retain that finding if the implementation is reused later.

The existing 20 runtime tests passed locally. Focused probes reproduced both findings; additional cancellation-during-poll and concurrent-driver wakeup checks passed. The three commits are atomic.

zbus/src/runtime/scheduler.rs
```diff
@@ -0,0 +1,616 @@
...
+    }
+
+    fn forget(&self, cell: &Arc<TaskCell>) {
+        self.tasks
+            .lock()
+            .expect("scheduler tasks poisoned")
+            .retain(|task| !Arc::ptr_eq(task, cell));
+    }
+
+    fn register_driver(&self, waker: &Waker) {
```
zeenix, line 202: 🤖 ChatGPT on behalf of zeenix.

[P2] Track registrations by driver and unregister on completion/cancellation.

This only deduplicates by waker identity. A single `run` polled with changing wakers accumulates every previous waker until a scheduled task wakes, and dropping `run` or `tick` never unregisters it. A focused test retained all 128 wakers for one driver; a separate cancellation test retained the host waker after dropping the driver. This can grow without bound while the inner future is polled independently of scheduler task wakeups. Give each driver a registration whose waker is replaced on subsequent polls and removed on completion or drop.

zbus/src/runtime/sync.rs
```diff
@@ -0,0 +1,534 @@
...
+            listener.await;
+        }
+    }
+
+    fn try_read(&self) -> Option<RwLockReadGuard<'_, T>> {
+        let mut state = self.state.lock().expect("RwLock state poisoned");
+        if state.writer || state.writers_waiting > 0 {
+            return None;
+        }
+        state.readers += 1;
```
zeenix, line 190: 🤖 ChatGPT on behalf of zeenix.

[P2] Check the reader count before incrementing if this implementation is retained/reused.

With overflow checks disabled, this wraps to zero and lets `try_write()` grant exclusive access while read guards remain live. Safe code can approach the limit by forgetting successive read guards without accumulating allocations, particularly on 32-bit targets. A boundary test with the counter pre-seeded to `usize::MAX` and release overflow behavior admitted a writer alongside live read guards. Use a checked increment that fails before wrapping. Deferring/removing these private locks as requested in the review summary removes this issue from this PR.

zbus/src/runtime/scheduler.rs
zeenix: Agreed, and thanks for the probes. This PR is withdrawn: the handwritten scheduler and locks are out of scope for #1960 and move to #1959 as a prototype, with both findings recorded there (the per-driver waker registration and the checked reader increment). #1960 will instead have the runtime abstraction expose the host's task spawning and synchronization; the design in #1961 is being revised accordingly.

zbus/src/runtime/sync.rs
zeenix: Agreed, and thanks for the probes. This PR is withdrawn: the handwritten scheduler and locks are out of scope for #1960 and move to #1959 as a prototype, with both findings recorded there (the per-driver waker registration and the checked reader increment). #1960 will instead have the runtime abstraction expose the host's task spawning and synchronization; the design in #1961 is being revised accordingly.

PR #1969, opened 2026-09-16
✅ zb,zm: Cover host and target blocking feature separation

rrnewton:
[orc, opus-5-1m]

<sub>*(Prefix convention: comments and descriptions written by an agent on @rrnewton's account are
tagged with the orchestrator and the model that wrote them — here Claude Opus 5, 1M-context.
Anything from Ryan himself is marked `(human)`.)*</sub>

Follow-up to your review comment on #1958 — *"zbus 6.x is just around the corner so you'd want
these there too, no? 🤔"*. Yes. This is the `main` half.

`main` already has the fix (9ba6c6c, #1944) but no regression coverage for it, so this adds the
fixture only. Nothing in `zbus` or `zbus_macros` changes.

### What it covers

A build dependency enabling `zbus/blocking-api` turns the feature on for the *host* `zbus` while the
*target* one leaves it off. Whether the blocking proxy is emitted has to follow the target, which is
what `__if_blocking_api_feature` does. The fixture pins that: default generation options, an
explicit `gen_blocking = true` with a renamed blocking type, `gen_blocking = false`, a blocking-only
proxy, and signal types — each checked with the target blocking API off and on. The `zbus`
dependency is renamed so the gate is only ever reached through the `crate` path.

### Two things worth flagging

1. **The host-side leak is structurally impossible on 6.x**, so this is a guard against
   reintroduction rather than a test for a live bug. On 5.x `blocking-api = ["zbus_macros/blocking-api"]`
   is what let a build dependency reach into the proc-macro crate; on `main`, `blocking-api = ["comms"]`
   and `zbus_macros` has no `blocking-api` feature at all. What the fixture still pins is that the
   emit decision stays on the `zbus` side. I verified it bites: making the `cfg(not(...))` arm of
   `__if_blocking_api_feature` forward its input makes the fixture fail with
   `error[E0433]: cannot find blocking in zbus` — the original 5.x symptom — and it passes again once
   reverted.

2. **No `legacy-macro-feature` case here**, unlike the 5.x version. That case covers
   `zbus_macros/blocking-api` staying inert, and the feature does not exist on 6.x.

### Shape

It is a workspace member checked with `-p`, following `service_only` and your comment about
workspace-wide feature unification, rather than the separate workspace the 5.x fixture uses. The
5.x one needs the isolation because it asserts *absence* with duplicate type definitions, which a
unified build would break; here the absence is caught by the target `zbus` having no `blocking`
module, so the plain member form is enough and the lockfile delta is 7 lines.

### Validation

- `cargo --locked clippy -p blocking_api_fixture` and the same with `--features target-blocking`:
  clean, with `-D warnings`.
- Workspace-wide `cargo --locked clippy` and `cargo --locked check`: clean — adding the fixture as a
  member does not disturb them.
- `cargo +nightly fmt --all -- --check`: clean.
- commitlint: clean.
- Mutation check as described above.

Happy to reshape or close this if you would rather have it differently — folded into #1958's branch,
without the rename, or as a member-free fixture like the 5.x one.


test_fixtures/blocking_api/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[features]
+# Turn the blocking API on for the *target* zbus, so the blocking proxies are expected to exist.
+target-blocking = ["renamed_zbus/blocking-api"]
+
+[dependencies]
+# Renamed on purpose: the blocking gate is emitted through the `crate` path, so the default
+# `::zbus` name would not catch a path that fails to forward.
+renamed_zbus = { package = "zbus", path = "../../zbus", default-features = false, features = [
+    "tokio",
```
zeenix: [P2] Avoid enabling Tokio across default workspace builds

Reviewed by Codex on behalf of zeenix.

Adding this workspace member enables `zbus/tokio` in the existing workspace-wide Clippy commands.
The Windows cross-check now fails with `clippy::clone_on_copy` at
`zbus/src/connection/socket/tcp.rs:119` and `:160`, before either new fixture check runs;
the dependent Linux and Windows test jobs are consequently skipped. The linted code is pre-existing,
but this feature addition exposes it in the default workspace checks.

The [failed CI job](https://github.com/z-galaxy/zbus/actions/runs/35117011895/job/104870647191)
confirms both errors. Could this use `async-io` here, as the host dependency does? The blocking-feature
split does not require different runtimes, and I verified that both fixture configurations still
pass Clippy with warnings denied after that substitution. Alternatively, fix the two redundant
clones in a separate prerequisite commit if enabling Tokio here is intentional.

zeenix (changes requested):
Reviewed by Codex on behalf of zeenix.

One CI regression below. Otherwise the fixture cases, isolated CI invocations, and single atomic
commit look sound. Both fixture configurations pass local Clippy with warnings denied.

test_fixtures/blocking_api/Cargo.toml
rrnewton: [orc, opus-5-1m]

Done — switched to `async-io`, which is the better of the two options you offered: enabling Tokio here was never intentional, just carried over from the 5.x fixture where the isolated workspace kept it to itself.

Evidence at three points, since the claim is about workspace unification rather than about the fixture:

| commit | `cargo --locked clippy --target x86_64-pc-windows-gnu` |
|---|---|
| `fd6bb690` (base) | pass |
| `e09d9080` (what you reviewed) | **2 × `clone_on_copy`** — your report, reproduced |
| `7c23deac` (now) | pass |

The rest of the clippy job also passes at every line, including the five cross-targets and the five `-p zbus` feature combinations, and `Cargo.lock` is unchanged by the substitution.

The fixture still bites after the swap, which was the thing worth re-checking rather than assuming: defeating the gate (making the `cfg(not(feature = "blocking-api"))` arm of `__if_blocking_api_feature` forward its input) makes it fail with `error[E0433]: cannot find blocking in zbus`, and pass again on revert. The runtime was never what it was testing.

**One thing this leaves behind, which I would rather hand you than quietly fix.** Those two lints are `#[cfg(windows)]` and only reachable with `tokio`, and no CI command covers that intersection — the workspace-wide runs cross-compile but use default features, and the Tokio runs use the right features but only the host target. Each widens one axis while the other stays at its default, so the combination is green everywhere and linted nowhere. It is only visible as:

```
cargo --locked clippy -p zbus --no-default-features \
  --target x86_64-pc-windows-gnu --features tokio,p2p,proxy
```

which fails today on `fd6bb690` with the same two errors. So the substitution puts CI back to green but does not make the lints go away — it returns them to being invisible, which is roughly how my fixture found them.

Happy to follow up with either half if you want it: dropping the two `.clone()` calls at `zbus/src/connection/socket/tcp.rs:119` and `:160` (both are `SocketAddr`, which is `Copy`), and/or adding that one cross-compiled Tokio clippy line so the gap cannot reopen. Your call on whether that belongs here, in a prerequisite commit, or in its own PR — it is pre-existing either way and I did not want to widen a fixture-only change without asking.

Amended in place rather than stacked, per CONTRIBUTING; the only change since your review is `test_fixtures/blocking_api/Cargo.toml`.

zeenix: Reviewed by Codex on behalf of zeenix.

Verified at `7c23deac`: the Windows workspace Clippy check and both isolated fixture configurations
pass with warnings denied. The substitution fixes this regression; no further findings.

zeenix (approved):
Reviewed by Codex on behalf of zeenix.

The previous CI regression is fixed. Independently verified the Windows workspace Clippy check
and both isolated fixture configurations with warnings denied. No remaining findings.

PR #1975, opened 2026-09-18
Replace the async-io backend with a runtime of zbus's own

### LINKED ISSUE #1959: RFC: Replace runtime internals and remove backend-specific smol dependencies

zeenix:
Closes #1959

zbus's default runtime becomes its own, and single-threaded: a task scheduler, a reactor on
`poll(2)` (unix) and `select` (Windows) with its own timers, and one runtime per thread that runs
`zbus::block_on`, driven by that thread. Between two polls of the future handed to `block_on`, that
thread runs the tasks, sockets and timers of every connection built in it, so a program that awaits
its work through `zbus::block_on` is a single thread and zbus starts none, and two threads inside
`block_on` run their connections side by side. A helper thread takes over only where work is left
with no thread inside `block_on` (a connection polled from some other executor, which goes on one
runtime shared by every such connection, or a `block_on` that returned with a connection alive) and
leaves once nothing is left to run, watch or time. Blocking work is a thread per call; async locks
are zbus's own on `event-listener` (Tokio's on a Tokio build). The `async-io` and `async-lock`
features are gone without alias; the default feature is `builtin-runtime`; `async-io`,
`async-executor`, `async-task`, `blocking` and `async-lock` leave the normal dependency graph. The
public `zbus::runtime::traits` contract from #1960 is unchanged. Benchmarks and binary-size fixtures
landed first so the swap is measured against what it replaces.

**Stacked on #1976** (`zbus::block_on` public, the blocking API removed): the first eight
commits below are that PR's and drop out of this one once it merges.

The RFC's one-worker-thread-per-connection design was built first (all CI green at 5ef16917;
kept on the branch `builtin-runtime-threaded`) and then replaced by the single-threaded one at
the maintainer's direction: lightweight for a typical D-Bus program, with Tokio there for the
demanding ones. The history was folded to show the final design once.

## Commits

From #1976:

- ✨ zb: Make block_on part of the public API
- 📝 Add the plan for dropping the blocking API
- ♻️ zb,zm,zx,book: Stop using the blocking API
- 👷 Drop the blocking-api feature from CI
- 💥 zb,zm,zx,book: Stop generating blocking proxies
- 💥 zb: Remove the blocking API
- 📝 zb,book: Point synchronous programs at zbus::block_on
- ♻️ zm,book: Name the proxy type override proxy_type_name

This PR:

- 📝 Add the implementation plans for the built-in runtime
- ✅ zb: Benchmark what a connection costs on its runtime
- ✅ zb: Run the concurrent calls benchmark's ends on their own threads
- 👷 zb: Build the p2p benchmarks for CodSpeed
- ✅ zb: Add GeoClue2-shaped fixtures for measuring binary size
- 👷 zb: Report the fixtures' binary sizes in CI
- ✨ zb: Give the connection locks of zbus's own
- ➖ zb: Drop the async-lock feature and crate
- ✨ zb: Add a task scheduler with no thread of its own
- ✨ zb: Add a reactor over poll(2) and select with its own timers
- ✨ zb: Run the built-in runtime on the thread inside block_on
- ♻️ zb: Run the default connection on zbus's own runtime
- 💥 zb: Replace the async-io feature and crates with builtin-runtime
- 📝 zb,book: Document the built-in runtime
- 📝 Add the plan for the spawn cost on the built-in runtime
- ⚡️ zb: Drop a finished task's handle without a cancellation error
- ⚡️ zb: Hash task ids as themselves in the live map
- ⚡️ zb: Look the runtime up once per block_on waker, not per wake
- ⚡️ zb: Give a block_on the seat while the helper thread waits
- 📝 Add the plan for a runtime per thread
- ♻️ zb: Give each thread inside block_on a runtime of its own

## Benchmarks

One session on this machine, criterion with `--features p2p`, three passes per tree interleaved
(tip, async-io, threaded, and again), medians of the three point estimates. All three trees run
the tip's benchmark files. "async-io" is the 5.x backend, at the commit just before the default
connection moved onto the built-in runtime; "threaded" is the RFC's design with one worker thread
per connection (5ef16917, kept on the branch `builtin-runtime-threaded`); "this PR" is the tip.

| id | async-io | threaded | this PR | vs threaded | vs async-io |
|---|---|---|---|---|---|
| `method-call/roundtrip` | 31 µs | 25.49 µs | 16.86 µs | -33.9 % | -45.6 % |
| `signal/emit-receive` | 13.59 µs | 12.16 µs | 7.783 µs | -36.0 % | -42.7 % |
| `method-call/1MiB-body` | 19.96 ms | 19.12 ms | 17.39 ms | -9.1 % | -12.9 % |
| `method-call/1000-concurrent-p2p` | 5.3 ms | 5.72 ms | 5.23 ms | -8.5 % | -1.3 % |
| `method-call/1000-concurrent-p2p-socket` | 7.98 ms | 8.05 ms | 7.07 ms | -12.2 % | -11.4 % |
| `spawn/100-tasks` | 19.35 µs | 73.58 µs | 17.94 µs | -75.6 % | -7.3 % |
| `connection/build-and-drop` | 382.5 µs | 449.4 µs | 503.7 µs | +12.1 % | +31.7 % |
| `connection/graceful-shutdown` | 1.998 µs | 9.675 µs | 8.638 µs | -10.7 % | +332.5 % |
| `blocking-hook/peer-credentials` | 301.6 µs | 379.7 µs | 441 µs | +16.1 % | +46.2 % |

How to read it. Each id in the `runtime` benchmark builds its connection pair and measures inside
one `zbus::block_on`, and only the operation is timed. On this PR the calling thread runs the
runtime itself and zbus starts no thread of its own; on the threaded design a worker thread ran
each connection, and the calling thread fought it for the scheduler's lock, which is the
`spawn/100-tasks` gap. The two `1000-concurrent-p2p` ids put the server inside a `block_on` on a
thread of its own and the client inside another on the benchmark's thread, the way a client and
a service never share a thread, and time the client's bursts alone; one runs over an in-process
channel, the other over a unix socket pair, the only difference between them. On this PR each
of the two threads drives a runtime of its own, so the two ends run side by side with no thread
of zbus's between them.

`connection/build-and-drop` and `connection/graceful-shutdown` are per-connection one-offs and
stay a follow-up. Dropping a pair is part of the first number here: the dropped connections'
shutdown runs on the same thread before the next pair is up, where async-io and the threaded
design did it on a thread of their own. `blocking-hook/peer-credentials` pays a thread start per
call (ruling 4).

The two concurrent ids are why the runtime is per thread rather than per process. With one
runtime per process, two threads inside `block_on` shared one seat: the first to take it ran
both connections, and the other was polled through a cross-thread unpark per wake of its future,
plus a wake of the reactor's socket, so the socket variant came out 70 % behind both async-io
and the threaded design, which ran the two ends on two threads in parallel. With a runtime per
thread both variants come out ahead of both.

## Binary size (`size` profile, fat LTO, stripped, GeoClue2-shaped fixtures)

| binary | async-io (before) | builtin-runtime | tokio |
|---|---|---|---|
| service | 1454 KiB | 1408 KiB | 1560 KiB |
| client | 1486 KiB | 1438 KiB | 1588 KiB |

## Decisions that need the maintainer's eye

1. **Feature name** `builtin-runtime` (ruling 1); a single `sed` renames it.
2. **Blocking work is a thread per call** (ruling 4); the first `peer_creds()` on Linux is the one
   public path that pays for it — number above.
3. **Dev-dependencies stay** (ruling 6): `async-io`, `async-executor`, `async-task` and `polling`
   remain dev-dependencies of `zbus` (and `async-io` of `zbus_macros`) for the test doubles that
   exercise the external-runtime path with a runtime that is not the built-in one; the CI guard
   proves none of them reaches the normal graph on the default, Tokio and external-only builds.
4. **Panic policy** (ruling 7): a panicking task, or a panic in a task future's `Drop`, is caught,
   logged and fails its handle. A panic elsewhere on the thread running the runtime unwinds it:
   out of `block_on` to the caller with the seat freed, or out of the helper's loop with its flag
   cleared so the next spawn, registration or timer poll starts another.
5. **One runtime per thread that runs `zbus::block_on`, driven by that thread**: a program
   that drives its connections through `zbus::block_on` is a single thread; two threads inside
   `block_on` run their connections side by side; a helper thread exists only while work is left
   with no thread inside `block_on`, and a connection built under some other executor goes on one
   runtime shared by every such connection, run by a helper. This supersedes the RFC's
   one-worker-per-connection paragraph at the maintainer's direction, as did the earlier one
   runtime per process, which the two-thread benchmarks above ruled out. Two consequences, both
   documented: the future given to `zbus::block_on` must not block the thread waiting for the
   connection (that work runs on the same thread in between polls of the future), and
   `zbus::block_on` called from inside a task the runtime is running, such as a served
   interface's method, panics instead of deadlocking. One connection's long-running task delays
   the others' I/O until it yields.
6. **`Send`/`Sync` bounds stay** on `traits::Runtime::spawn`, `Interface` and the proxies:
   `Connection` is `Clone + Send + Sync`, so any thread may be the one inside `block_on` running
   the tasks; a task migrates between threads across calls even though it never runs on two at
   once. Dropping the bounds needs a thread-confined connection, a separate API decision.
7. **Windows**: one `select` per runtime, so the backend declares its own 1024-entry socket set
   (Winsock lets a program raise `FD_SETSIZE`; the set has `FD_SET`'s layout with a longer array).
   The `windows_test (default)` job is what proves it, as it proved the one-provider question for
   the wake socket beside an AF_UNIX source.
8. **Lock fairness**: the mutex is a plain barging lock (async-lock had starvation handling);
   contended paths such as `socket_write` lose that guarantee.
9. **`zbus::block_on` public and the blocking API removed** are #1976's decisions; here
   `block_on` is, with the built-in runtime, the loop that runs the connections.

## Follow-ups noted during the work

- The per-connection one-offs above (`build-and-drop`, `graceful-shutdown`).
- The book's "block_on panics from inside a task" rule is the default build's; with `tokio`,
  `zbus::block_on` is Tokio's, which panics from inside a Tokio task (stated on the function, not
  yet in the book).
- `SO_EXCLUSIVEADDRUSE` on the Windows wake listener; the three 1024-entry sets are zeroed on the
  stack per wait on Windows (24 KiB), which could live in the poller instead.
- A thread inside `block_on` that hands work to another thread's runtime in the window before
  that thread has taken its seat starts a helper for it, which takes the seat, yields it straight
  back and parks: one transient thread, no hang.
- A thread that runs `block_on` and ends with connections alive leaves them to a helper; each such
  runtime keeps a wake pipe (two descriptors) for as long as its connections live.

https://claude.ai/code/session_01BfDbRLyfCfp1VSn3ZT1cq5


zeenix (commented):
Review by Codex on behalf of zeenix, against `9768943`.

Changes needed: five reproduced correctness issues are detailed inline, covering TLS teardown,
timer recovery/overflow, and scheduler cleanup when destructors panic.

I reviewed the full stack and individual commit scopes; I found no additional atomicity issue.
Existing built-in runtime and lock tests pass, as do mixed reader/writer stress and 800 method
calls after the connections' creating threads have exited. Focused regression probes reproduce
the inline findings.

Two documentation corrections also remain: `zbus/Cargo.toml:61` still says one thread per
connection, and `book/src/faq.md:202-203` says default tasks/I/O run on a thread of their own.
Both descriptions need to reflect the new `block_on` behavior.

zbus/src/runtime/builtin/driver.rs
```diff
@@ -0,0 +1,608 @@
...
+            seat.helper = Helper::Driving;
+        }
+        // A turn of this call's loop before this one may have found the seat taken and put the
+        // thread down as waiting for it, which a thread in the seat is not.
+        if who == Holder::BlockOn {
+            seat.stop_waiting();
+        }
+        drop(seat);
+        DRIVER_REACTOR.with(|reactor| reactor.set(Some(NonNull::from(&*inner.reactor))));
+        DRIVING.with(|driving| *driving.borrow_mut() = Arc::downgrade(&inner));
```
zeenix: **[P2] Handle partially destroyed TLS before taking the driver seat**

Codex review on behalf of zeenix.

`OWN` can still resolve a live runtime after `DRIVING` has been destroyed. For example, enter
`block_on` on a fresh thread, initialize an application thread-local, then build and retain a
connection in it. At thread exit, that thread-local's destructor calling
`block_on(connection.graceful_shutdown())` reaches this `with` after `DRIVING` has gone and panics
with `AccessError`. An uncaught panic in a TLS destructor aborts the process; catching it leaves
the seat marked occupied because the `Driving` guard has not been constructed yet. I reproduced
this using the public connection API and a closed Channel peer. Check TLS availability before
claiming the seat and support teardown without leaving the driver state half-initialized; the
`DRIVING.with` in `Seat::free` needs the same consideration.

zbus/src/runtime/builtin/reactor.rs
```diff
@@ -0,0 +1,710 @@
...
+            let replaced = timers.pending.insert(key, cx.waker().clone());
+
+            (timers.pending.keys().next() == Some(&key), replaced)
+        };
+        // Clear of the lock: dropping a waker can drop a task, and the future of that task may
+        // hold a timer of this reactor, whose own drop takes this very lock.
+        drop(replaced);
+        // A deadline behind one the map already holds changes nothing about the wait under way.
+        if first_poll && first_deadline {
+            this.reactor.notify();
```
zeenix: **[P2] Notify when a removed timer is inserted again**

Codex review on behalf of zeenix.

After a failed poll, `wake_everything()` removes the timer entry but leaves `Sleep::id` set. If
another source keeps the helper alive and an external executor repolls the sleep after the
helper has entered its next wait, this inserts the deadline again without notifying: `first_poll`
is false. The helper can then wait indefinitely despite the timeout expiring. I reproduced this
by clearing the timer through `wake_everything`, starting another wait, and repolling the sleep:
a 250 ms timer did not interrupt even a one-second bounded wait. The follow-up noted in the PR
therefore permits an indefinite hang, not just a late poll. Base notification on whether this
insertion introduces the earliest deadline, including reinsertion after recovery.

zbus/src/runtime/builtin/reactor.rs
```diff
@@ -0,0 +1,710 @@
...
+        })
+    }
+
+    /// A timer that is due once `duration` has passed.
+    pub(super) fn sleep(self: &Arc<Self>, duration: Duration) -> Sleep {
+        Sleep {
+            reactor: self.clone(),
+            // This reactor's timers run on the standard clock, so a length of time is a deadline
+            // on it.
+            deadline: Instant::now() + duration,
```
zeenix: **[P2] Preserve overflow handling for large timeout durations**

Codex review on behalf of zeenix.

`Instant::now() + Duration::MAX` panics, so a connection configured with
`Builder::method_timeout(Duration::MAX)` now panics when a method call needs to wait. The previous
`async_io::Timer::after` used `checked_add` and treated an unrepresentable deadline as a timer that
never fires. A focused test reproduces the overflow immediately in this constructor. Preserve
that behavior with checked deadline arithmetic rather than turning a valid timeout setting into
a panic.

zbus/src/runtime/builtin/scheduler.rs
```diff
@@ -0,0 +1,767 @@
...
+        // Whatever a panic left half-done is in the future, and a task that panicked is marked
+        // `Done`, dropped and forgotten below, so nothing here reads that half-done state and
+        // nothing polls the future again.
+        let polled = catch_unwind(AssertUnwindSafe(|| {
+            future.as_mut().poll(&mut Context::from_waker(&waker))
+        }));
+        let panicked = polled.is_err();
+        let ready = matches!(polled, Ok(Poll::Ready(())));
+        // Nothing carries the panic any further, so its payload ends here.
+        drop(polled);
```
zeenix: **[P2] Retire the task before dropping a potentially panicking payload**

Codex review on behalf of zeenix.

A task can call `panic_any` with a payload whose `Drop` also panics. Dropping `polled` then unwinds
before the task is marked `Done`, removed from `live`, or its join is failed. With the runtime
still alive, the task remains `Running` with no future to poll: its join stays pending, dropping
the join cannot remove it, and a restarted helper sees work that can never finish. An exact-source
probe reproduced all of these states. Finalize the scheduler/join state on every unwind path
before disposing of the panic payload, and contain payload destruction if task panics are to
remain isolated.

zbus/src/runtime/builtin/scheduler.rs
```diff
@@ -0,0 +1,767 @@
...
+                }
+                // Finished, or cancelled by an earlier drop: nothing to take back, nobody to
+                // tell.
+                Stage::Done => return,
+            }
+        };
+        // On the thread that let the handle go, with no lock held: a destructor that panics
+        // here panics where the handle was dropped, and one that spawns is served as any other
+        // caller is.
+        drop(future);
```
zeenix: **[P2] Keep cancellation notification on the unwind path**

Codex review on behalf of zeenix.

If the cancelled future's destructor panics here, the notification below is skipped even though
the task has already been removed from `live`. When it was the last idle task and the helper was
asleep in an unbounded reactor wait, catching the cancellation panic leaves that helper and its
runtime resources alive indefinitely. This can occur when dropping a connection using a custom
`Socket` whose read half panics on drop, with no other waiters to wake the helper. A focused
runtime test reproduced `live_tasks() == 0` while the helper stayed running; explicitly notifying
the reactor immediately let it retire. Use an unwind-safe cleanup/notification guard so allowing
cancellation to propagate a user `Drop` panic does not prevent the now-idle helper from exiting.

zbus/src/runtime/builtin/driver.rs
zeenix: Fixed in 20036606 (squashed into the runtime-per-thread commit). `Driving::take` and `Seat::free` write `DRIVING` through `try_with` and hold the seat regardless: nothing built on a thread on its way out can look the runtime up there (`driven()` is already empty then), and the lookup falls through to the thread's registry, which names the same runtime. `DRIVER_REACTOR` and `IN_BLOCK_ON` are `Cell`s with no destructor, so those accesses cannot fail. Regression test: `a_block_on_taking_the_seat_as_a_thread_ends_does_not_abort`, which registers an application local between the registry and `DRIVING` so that reverse-order destruction reproduces the abort; it aborted before the fix.

zbus/src/runtime/builtin/reactor.rs
zeenix: Fixed in 20036606 (squashed into the reactor commit). The notification is now sent when the insertion introduced a new earliest deadline (`replaced.is_none() && first_deadline`) rather than on the first poll only, so a re-poll after `wake_everything` drained the map breaks the wait under way, while a poll that merely replaces the waker still sends none. Regression test: `a_deadline_stored_again_breaks_the_wait`, which timed out before the fix. The follow-up bullet about this is dropped from the description.

zbus/src/runtime/builtin/reactor.rs
zeenix: Fixed in 20036606 (squashed into the reactor commit, with the `block_on` commit for the runtime side). The deadline is `Instant::now().checked_add(duration)`, and a deadline the clock cannot represent is a timer that never fires: it stays pending, registers nothing, and the runtime's `Sleep::poll` starts no helper for it. Regression tests: `a_sleep_beyond_the_clock_never_fires` (panicked with the overflow before the fix) and `a_sleep_beyond_the_clock_starts_no_helper`.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Fixed in 20036606 (squashed into the scheduler commit). The task is marked `Done`, taken out of `live` and its join failed before the payload is disposed of, and that disposal is contained with `catch_unwind` like the finished future's drop, so a payload whose destructor panics ends with its task. Regression test: `a_panic_payload_that_panics_as_it_goes_still_ends_its_task`, which let the payload's panic escape `run_one` before the fix.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Fixed in 20036606 (squashed into the scheduler commit). The cancelled future is dropped under `catch_unwind`; the join failure and the notification run either way, and the panic is re-raised with `resume_unwind` after them, so a user destructor's panic still surfaces where the handle was dropped without keeping an idle helper up. Regression test: `a_panic_in_a_cancelled_futures_drop_still_notifies`, which saw no notification before the fix.

zeenix:
Both documentation residues are fixed in 20036606: the `builtin-runtime` feature comment in `zbus/Cargo.toml` and the FAQ paragraph on tokio APIs in interface methods now describe one runtime per thread that runs `zbus::block_on`. A sweep of the remaining "thread of its own" / "per connection" phrases in the crate and the book found only the helper thread, `spawn_blocking`'s per-call thread and explicit negations, so nothing else needed changing.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Codex recheck on behalf of zeenix, at `2003660`.

The original task-retirement/join problem is fixed. One related cleanup path still lets the
payload escape: `scheduler.rs:189` immediately drops the `Err` returned by
`catch_unwind(|| drop(future))` outside the catch.

Reproducer: cancel a task while its poll is running, then let the poll return `Pending`. Its
future's destructor calls `panic_any(Payload)`, and `Payload::drop` uses ordinary `panic!`.
Cancellation cleanup then unwinds out of `run_one`. My exact-source probe leaves an unrelated
task queued (`ready=true, live_tasks=1`); on the helper this exits the driver, leaving that work
stalled until a new operation starts another helper.

Please handle the payload returned by the cleanup catch without letting its destructor unwind,
and cover cancellation during polling. This is the same kind of panicking payload as above,
rather than an arbitrary panic elsewhere in the driver.

zbus/src/runtime/builtin/driver.rs
zeenix: Codex recheck on behalf of zeenix: verified fixed at `2003660`.

Both accesses to destructible `DRIVING` now tolerate TLS teardown, and the new test covers
the original destruction order.

zbus/src/runtime/builtin/reactor.rs
zeenix: Codex recheck on behalf of zeenix: verified fixed at `2003660`.

Reinserting a removed earliest deadline now notifies the reactor, including after
`wake_everything()`.

zbus/src/runtime/builtin/reactor.rs
zeenix: Codex recheck on behalf of zeenix: verified fixed at `2003660`.

Overflowing durations now stay pending without registering a deadline or starting a helper.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Codex recheck on behalf of zeenix: verified fixed at `2003660`.

Cancellation now completes notification before re-raising the destructor's panic.

zeenix (commented):
Re-reviewed by Codex on behalf of zeenix, against `2003660`.

Four threads are verified fixed and resolved; both documentation corrections are also done.
The original stranded-task/join bug is fixed too, but a related panic-payload cleanup path still
escapes the scheduler when cancellation happens during polling. I reproduced it with an unrelated
task left queued and posted the details in the existing payload thread, which remains open.

All 69 built-in runtime tests pass, including the new regressions. The rebase and placement of
the fixes in the original commits look good. No other new issue found.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Fixed in 8bfd8c19 (squashed into the scheduler commit). Both disposal sites in `run_one` and the one in `Scheduler::drop` go through one helper, `dispose`, which drops the value under `catch_unwind` and then drops the payload that catch hands back under a second one; a payload whose own destructor panics is leaked rather than followed any further, so nothing unwinds out of the scheduler. `JoinHandle::drop` keeps re-raising by design. Regression tests: `cancelling_a_task_whose_drop_panics_with_a_panicking_payload_is_contained` (your reproduction: cancelled during its poll, with an unrelated task left queued that still runs) and `a_panic_payload_whose_drop_panics_with_another_payload_is_contained`; both let the panic escape `run_one` before the fix.

zeenix: Codex recheck on behalf of zeenix: verified fixed at `8bfd8c19`.

Re-ran the original cross-thread cancellation reproducer: `run_one` no longer unwinds, and the
unrelated queued task completes and joins successfully. The shared `dispose` helper contains
payload cleanup at all three scheduler disposal sites. Both new regression tests pass.

zeenix (commented):
Re-reviewed by Codex on behalf of zeenix, against `8bfd8c19`.

No new findings. The remaining panic-payload cleanup issue is fixed, and all five threads from
my original review are now resolved. The original cross-thread cancellation reproducer no
longer unwinds out of `run_one`; the unrelated queued task completes and joins successfully.

All 71 built-in runtime tests pass, including both new regressions, and the current CI checks
are green. The range-diff confirms that this change is confined to the scheduler fix and its
tests, folded into the original scheduler commit.

zbus/src/runtime/locks/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    use super::*;
+
+    fn poll_once<F>(future: &mut F) -> Poll<F::Output>
+    where
+        F: Future + Unpin,
+    {
+        pin!(future).poll(&mut Context::from_waker(Waker::noop()))
+    }
+
```
zeenix: show be defined at the bottom: definition after usage rule.

zeenix (commented):
Re-reviewed by Codex on behalf of zeenix, against `4cc590ac`.

One correctness regression needs fixing: the atomic wake-flag conversion loses synchronization
between a wake and the next poll, allowing a coalesced notification to leave `block_on` waiting
indefinitely. A focused Loom model confirms the failing ordering and the corrected RMW protocol.

The requested robustness pass also produced three inline enum refinements: task state, driver
ownership, and reader/writer ownership. These are hardening suggestions, separate from the bug.

All 71 built-in runtime tests and 8 lock tests pass, and current CI is green. The earlier fixes
remain intact; the range-diff shows no new commit-scope issue.

Your existing definition-after-use comment remains unaddressed: `poll_once` is still above its
callers in `runtime/locks/mod.rs`. Two documentation corrections also remain after this update:
`scheduler.rs:14` still links `Scheduler::spawn`, now `Inner::spawn`; `reactor.rs:19-20` still
describes a third mutex. The new `wake_pending` comment should describe notification coalescing,
since the flag is raised before the write and a successful wait can clear it without draining
a byte.

zbus/src/runtime/builtin/driver.rs
```diff
@@ -0,0 +1,618 @@
...
+        }
+    };
+    let mut failed_waits = 0u32;
+    loop {
+        if driving.is_none() {
+            take_seat(driving);
+        }
+        // Cleared before the poll, so that a wake during it is seen by the round or the park
+        // that follows.
+        signal.woken.store(false, Ordering::Release);
```
zeenix: **[P2] Preserve wake-to-poll synchronization when clearing the flag**

Codex review on behalf of zeenix.

The new Release store can erase a wake without acquiring the writes it published. Consider a
driver whose reactor has drained an earlier notification but has not cleared `wake_pending` yet:
another thread publishes readiness and wakes `Signal`; the reactor coalesces that wake, so no
new byte is written. Both flags are then cleared with Release stores. The next poll can see
stale readiness, return `Pending`, and enter an unbounded reactor wait. The `unpark` token does
not rescue this path because the active driver waits on the reactor, not `thread::park`.

This violates the [wake-to-poll happens-before guarantee][wake]. A focused Loom model finds this
execution; using an acquiring RMW before polling removes it. Also make `Signal::wake_by_ref`'s
`true` update a release RMW: independent Release stores can overwrite earlier publishers, so
changing only the reset is insufficient for multiple coalesced wakes. The two-producer model
fails with stores and passes with RMWs. For example, use `swap(false, Ordering::AcqRel)` here and
`swap(true, Ordering::Release)` when waking, or retain the mutex's synchronization.

[wake]: https://doc.rust-lang.org/std/task/struct.Waker.html#method.wake

zbus/src/runtime/builtin/scheduler.rs
```diff
@@ -0,0 +1,951 @@
...
+
+struct CellState {
+    stage: Stage,
+    /// Whether the cell sits in the ready queue.
+    queued: bool,
+    /// Whether a wake arrived while the task was being polled, so it is to be polled again.
+    woken_while_running: bool,
+    /// Whether the handle was dropped while the task was being polled, so the future is to be
+    /// dropped once that poll returns.
+    cancelled: bool,
```
zeenix: **Robustness: put task flags inside their applicable states**

Codex review on behalf of zeenix.

For the requested robustness pass, this is a useful place to replace separate fields with:

```rust
enum CellState {
    Idle { future: BoxFuture, queued: bool },
    Running { woken: bool, cancelled: bool },
    Done,
}
```

`cancelled` and `woken_while_running` only have meaning during a poll; `queued` is needed when
the future is idle. Carrying them in those variants prevents phase-incompatible combinations
and forces transitions to initialize the relevant flags. Keep accepting stale ready-queue
entries for cancelled tasks when popping `Done`. This is suggested hardening; I found no
currently reachable incorrect transition here.

zbus/src/runtime/builtin/driver.rs
```diff
@@ -0,0 +1,618 @@
...
+    /// Keyed by thread, so that a thread which enters `block_on` again and again while another
+    /// keeps the seat has one place here rather than one per call. An entry goes where its
+    /// thread takes the seat or leaves `block_on`, so that what is held here is the threads
+    /// there may be something to unpark, and no more.
+    waiting: HashMap<ThreadId, Thread>,
+    /// The helper thread while it is parked with the seat left to a `block_on`, to be roused as
+    /// that call gives it up. Nothing otherwise: whether the helper is in the seat is what
+    /// `holder` says, and a helper is up exactly when it is in the seat or parked here. Set under
+    /// the lock wherever the thread moves between the two.
+    helper: Option<Thread>,
```
zeenix: **Robustness: encode the parked helper in the seat variants**

Codex review on behalf of zeenix.

`holder == Holder::Helper` together with `helper: Some(thread)` still represents a helper that
is both driving and parked. Combine these fields into states such as:

```rust
enum SeatState {
    Free { parked_helper: Option<Thread> },
    BlockOn { parked_helper: Option<Thread> },
    Helper,
}
```

The `Free` payload matters: the helper can legitimately remain parked during handover before a
`block_on` takes the seat. This would preserve that interval while preventing the contradictory
combination. A separate two-variant owner type for `Driving::who` would also exclude `Nobody`
from a guard that owns the seat. These are robustness suggestions, not observed bad transitions.

zbus/src/runtime/locks/rwlock.rs
```diff
@@ -0,0 +1,287 @@
...
+}
+
+/// What every `read` and `write` decides on.
+///
+/// The blocking mutex around it is held for the few instructions that check and update it, never
+/// across an await.
+struct RwState {
+    readers: usize,
+    writer: bool,
+    writers_waiting: usize,
```
zeenix: **Robustness: represent reader/writer ownership with an enum**

Codex review on behalf of zeenix.

Replace the independent `readers` and `writer` fields with ownership states such as `Unlocked`,
`Reading(NonZeroUsize)`, and `Writing`. That makes simultaneous reader/writer ownership and a
zero-reader `Reading` state unrepresentable, strengthening the state invariant behind the
`UnsafeCell` accesses.

Keep `writers_waiting` separate: waiting writers are valid with every ownership state. Preserve
the current behavior where cancelling the final waiting writer admits readers even if other
readers still hold the lock. This is a targeted hardening suggestion; no current exclusion bug
was found.

zbus/src/runtime/locks/mod.rs
zeenix: Moved `poll_once` to the bottom of the tests module, after `zbus_block_on`, in
`✨ zb: Give the connection locks of zbus's own`.

Generated by Claude Opus 5.5.

zbus/src/runtime/builtin/driver.rs
zeenix: Fixed in `✨ zb: Run the built-in runtime on the thread inside block_on`. `Signal::woken` is now
raised and lowered only with `AcqRel` swaps.

- **The lowering before each poll** acquires every wake since the previous poll. The wakes are
  swaps too, so each one continues the release sequence of the one before it rather than cutting
  it off.
- **Each wake's swap** is `AcqRel` rather than `Release`. A wake that lands after a lowering
  acquires it, and with it the `wake_pending` store at the end of the reactor wait before it. Its
  `notify` then cannot find the flag still raised for a wake-up that wait already took out. With
  only `Release` on the wake, the wake's `notify` and the driver's pre-wait check of `woken` form
  a store-buffering pair, and both may read the stale value.

The field's doc spells out both halves. I added no test, because these orders are only reachable
under a model checker and Loom is not a dependency here.

Generated by Claude Opus 5.5.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Done in `✨ zb: Add a task scheduler with no thread of its own`. `Stage` and the three flags are
now one `CellState`: `Idle { future, queued }`, `Running { woken, cancelled }` and `Done`. A
cancelled task's stale queue entry is still passed over when it comes up as `Done`.

Generated by Claude Opus 5.5.

zbus/src/runtime/builtin/driver.rs
zeenix: Done, in two steps:

- `✨ zb: Run the built-in runtime on the thread inside block_on` gives `Driving::who` a
  two-variant `Driver`, so a guard that owns the seat cannot name nobody.
- `⚡️ zb: Give a block_on the seat while the helper thread waits` moves the parked helper into
  `Holder::Nobody { parked_helper }` and `Holder::BlockOn { parked_helper }`, and drops
  `Seat::helper`. A helper both in the seat and parked can no longer be represented. A helper is
  now started only from `Nobody { parked_helper: None }`.

Generated by Claude Opus 5.5.

zbus/src/runtime/locks/rwlock.rs
zeenix: Done in `✨ zb: Give the connection locks of zbus's own`. `RwState` now holds an `Owner`
(`Unlocked`, `Reading(NonZeroUsize)` or `Writing`) beside `writers_waiting`. Cancelling the
last waiting writer still admits readers while other readers hold the lock.

The async `Mutex` in the same commit had the same kind of lone `sync::Mutex<bool>` flag. It is
now an `AtomicBool`, taken with an `Acquire` compare-exchange and released with a `Release`
store. event-listener fences both `listen` and `notify(1)`, so a release between a `lock`'s
check and its wait is still seen.

Generated by Claude Opus 5.5.

zeenix:
Addressed the review of `4cc590ac` at `639f12d4`. Every fix is folded into the commit that
introduced the code, and each rewritten commit builds on its own. Beyond the inline threads:

- `scheduler.rs` now links `Inner::spawn`, and a test comment in `reactor.rs` names `Inner::sleep`.
- The reactor's module doc no longer mentions a third mutex.
- `wake_pending`'s doc now describes coalescing. The flag goes up before the write and comes down
  once the wait returns, whether or not that wait took the wake-up out. The reactor commit's
  message now says the same.

Generated by Claude Opus 5.5.

zbus/src/runtime/locks/mod.rs
zeenix: Codex recheck on behalf of zeenix: verified at `a66cc30d`.

`poll_once` now follows its callers at the bottom of the test module.

zbus/src/runtime/builtin/driver.rs
zeenix: Codex recheck on behalf of zeenix: verified fixed at `a66cc30d`.

Both `AcqRel` swaps are appropriate. The stronger wake swap also orders a later notification
against the preceding reactor reset. The focused models pass with the current protocol,
including coalesced publications and the reset/notification race.

zbus/src/runtime/builtin/scheduler.rs
zeenix: Codex recheck on behalf of zeenix: verified at `a66cc30d`.

The enum preserves wake/cancel transitions and accepts stale `Done` queue entries. Exact-source
probes passed repeated concurrent wake/cancellation and shutdown cases; the earlier panic
cleanup fixes remain intact.

zbus/src/runtime/builtin/driver.rs
zeenix: Codex recheck on behalf of zeenix: verified at `a66cc30d`.

The parked helper survives the `BlockOn` to `Nobody` handover, and `Driver` excludes a guard
with no owner. Helper reservation and exit paths preserve a single owner.

zbus/src/runtime/locks/rwlock.rs
zeenix: Codex recheck on behalf of zeenix: verified at `a66cc30d`.

`Owner` preserves exclusion, writer preference and cancellation behavior. The atomic mutex also
passes focused models with event-listener's actual synchronization, including cancellation
forwarding. Mixed reader/writer stress passed 160,000 acquisitions and 80,000 writes.

zeenix (commented):
Re-reviewed by Codex on behalf of zeenix, against `a66cc30d`.

No new runtime correctness finding. The wake-ordering fix, all three enum refinements, the
definition-order change and documentation corrections are verified; the five outstanding
threads are now resolved. The wake path's stronger `AcqRel` ordering is appropriate.

Validation passed: 71 built-in runtime tests, 8 lock tests, focused wake/mutex concurrency
models, repeated scheduler wake/cancellation/shutdown probes, and mixed reader/writer stress.
The earlier panic, timer and TLS fixes remain intact.

#1980 is exactly the first five commits of this branch, and the remaining commit scopes look
correct. However, the measurement prerequisite still needs work: its configured CodSpeed mode
skips all nine new/rewritten benchmark cases, and baseline build failures silently disable the
size comparison. Those reproduced findings are in the [review of #1980][bench-review]. Fix that
PR and establish actual baseline measurements before merging this runtime swap.

[bench-review]: https://github.com/z-galaxy/zbus/pull/1980#pullrequestreview-5307567775

zeenix (commented):
Codex review on behalf of zeenix.

Final pre-merge review of `b20c81db`, including all 19 commits, the full runtime implementation,
feature/CI integration, previous review fixes, and application-facing documentation.

I found no new runtime correctness defect in this pass, but I would fix the documentation and
project-convention findings before merging. The inline comments cover feature migration,
thread guarantees, usable Tokio examples, the documentation's audience, inaccurate lifecycle
comments, and definition-after-use ordering.

Two additional commit-message corrections are required by `CONTRIBUTING.md`:

- `be42d1e2` changes `zbus_macros/src/lib.rs`, so its `zb,book` prefix should include `zm`.
- `7104440e` has only attribution/changelog trailers; add a body explaining the plan's change
  and purpose. The existing `Changelog: skip` is correct.

All 19 commits otherwise have coherent logical scopes. The first 16 retain the previously
reviewed patches after rebasing; the three new spawn optimizations were reviewed independently.
Please fold corrections into the appropriate existing commits.

Verification: 71 builtin-runtime tests and eight lock tests pass, as do seven focused Loom
checks and repeated exact-source scheduler wake/cancel/shutdown probes. Earlier TLS, timer,
panic-cleanup and wake-ordering fixes remain intact. The working tree is clean.

All 18 non-performance checks pass. The walltime report shows nine improvements and the one
accepted cold-path regression: graceful shutdown rises from 5.5 to 12.6 microseconds. That
roughly 7-microsecond cost is not a new merge objection.

book/src/upgrading-to-6.md
```diff
@@ -672,6 +673,14 @@ async fn connect(runtime: impl Runtime) -> Result<Connection> {
 }
 ```
 
+5.x's `async-io` feature has no 6.0 equivalent by that name — it's gone outright, with no
+compatibility alias. `builtin-runtime` is the default, so a plain `zbus = "6"` dependency
+carries it without asking by name; a `default-features = false` build that named `async-io`
+explicitly drops that name or renames it to `builtin-runtime`. The backend behind it changed
```
zeenix: **[P2] Preserve the runtime in the feature-migration instructions**

Codex review on behalf of zeenix.

For `default-features = false`, dropping `async-io` leaves no default backend. For example,
`features = ["proxy", "async-io"]` becomes proxy-only, and building a connection without an
explicit runtime returns `Error::Unsupported`. An async-io-only feature list becomes wire-only.

Tell applications to replace `async-io` with `builtin-runtime`. Removing it instead is valid
when they deliberately enable `tokio` or supply a runtime through `Builder::runtime`.

zbus/src/utils.rs
```diff
@@ -28,30 +28,40 @@ impl<T, E> ResultAdapter for Result<T, E> {
...
 
-/// Runs a future to completion on the calling thread.
+/// Runs a future to completion on the calling thread, and zbus's runtime with it.
 ///
 /// This is for a program that has no async runtime of its own. Put the async code in one call to
-/// this function; the call blocks the thread until the future completes. Only the future passed
-/// in runs on this thread; a connection's own tasks and I/O run on threads zbus starts for them.
+/// this function; the call blocks the thread until the future completes. In between polls of the
+/// future, the calling thread also runs the tasks, sockets and timers of every connection built in
+/// it, so such a program stays a single thread, and two threads that each call this drive their own
```
zeenix: **[P2] Qualify the claim that the whole program remains single-threaded**

Codex review on behalf of zeenix.

The built-in backend inherits `Runtime::spawn_blocking`, which creates a worker thread for each
blocking operation. DNS, nonce-file reads and peer-credential lookups can take that path inside
one `zbus::block_on`; P2P server authentication also performs a peer-credential lookup.

Document that the scheduler and I/O reactor run on the calling thread, while blocking system
operations may use temporary worker threads. Apply the same qualification to the whole-program
claims in `book/src/faq.md` and `book/src/connection.md`, so applications can accurately assess
thread usage.

book/src/faq.md
```diff
@@ -181,41 +181,54 @@ fn main() -> Result<()> {
...
 
 Not to worry, though! You can enable tight integration between tokio and zbus by enabling `tokio`
 feature:
 
 ```toml
 # Sample Cargo.toml snippet.
 [dependencies]
-# Also disable the default `async-io` feature to avoid unused dependencies.
+# Also disable the default `builtin-runtime` feature to skip compiling in the fallback backend.
 zbus = { version = "6", default-features = false, features = ["tokio"] }
```
zeenix: **[P2] Keep the application API features in the Tokio examples**

Codex review on behalf of zeenix.

This snippet predates this PR, but it remains broken in the application guide being updated:
`default-features = false, features = ["tokio"]` disables `service`, including the interface
macro and object server whose methods this FAQ is explaining. It also disables `proxy`, which
the README's client example needs; the README repeats this configuration.

Use the simple additive configuration, `zbus = { version = "6", features = ["tokio"] }`, or
explicitly retain the `service`/`proxy` features each example needs when disabling defaults.
Explain selective feature trimming separately.

zbus/README.md
```diff
@@ -129,24 +129,27 @@ size of your binary.
 
 ## Compatibility with async runtimes
 
-zbus is runtime-agnostic. By default (the `async-io` feature), a connection runs its I/O, timers
-and tasks on two threads of zbus's own: one running the tasks and one running [`async-io`]'s
-reactor. With `tokio` instead, it runs on your Tokio runtime, with no extra thread. For any other
-runtime, hand an implementation of [`runtime::traits::Runtime`] to
-[`connection::Builder::runtime`], which then supplies everything a connection needs, sockets
+zbus is runtime-agnostic. By default (the `builtin-runtime` feature), a connection runs on the
```
zeenix: **[P3] Make the runtime overview an application setup guide**

Codex review on behalf of zeenix.

The README and book's runtime overview spend most of their opening on runtime registries,
dependency ownership, lock implementation and process-pipe teardown. Lead with the application
choices: use `zbus::block_on` without an existing executor, enable `tokio` for Tokio applications,
and explain that other executors can use the default backend too. `Builder::runtime` is an
optional integration hook; the current "For any other runtime, hand an implementation" wording
suggests every other executor requires an adapter.

Keep the relevant blocking/nesting restrictions here and move detailed process-worker and lock
implementation discussion to the runtime trait documentation. In particular, replace the book's
`connection.md` claim that locks use event-listener "regardless of which runtime feature is
chosen" with the application-facing fact that no additional lock feature is required; Tokio
builds actually select Tokio locks.

zbus/src/runtime/builtin/mod.rs
```diff
@@ -0,0 +1,372 @@
...
+//! nothing left to run, watch or time. A `block_on` that arrives while the helper is in the seat
+//! is given it, the helper parking until that call leaves, so that a program calling `block_on`
+//! once per operation runs each of them on its own thread.
+//!
+//! A panic in a task is caught and fails that task's handle. A panic outside a task — in a
+//! waker, say — unwinds the thread in the seat: out of `block_on`, to whoever called it, with
+//! the seat freed on the way; out of the helper's loop, the helper putting itself down as gone
+//! so that the next spawn, registration or timer poll starts another. The wakers the reactor had
+//! taken out of its maps to wake are dropped in the unwind without being woken, and the waiters
+//! they belonged to are served by the next thread in the seat.
```
zeenix: **[P3] State the actual limits of panic and TLS recovery**

Codex review on behalf of zeenix.

The reactor removes the ready I/O wakers and due timers before invoking their wakers. If one
waker panics, the remaining extracted wakers are dropped, and a replacement driver cannot
reconstruct them. Their waiters can remain pending, so the promise that the next thread serves
them is incorrect. Document that limit; this does not ask to change the accepted panic policy.

The TLS comments also need qualification: `driver::driven` says a thread destroying locals can
never hold the seat, and `Inner::current` says its own registry is already gone. The retained
`a_block_on_taking_the_seat_as_a_thread_ends_does_not_abort` test covers exactly the case where
`DRIVING` is destroyed but `OWN` survives and a destructor takes the seat. Describe each fallback
in terms of the particular TLS value being unavailable.

zbus/src/runtime/builtin/poll/windows.rs
```diff
@@ -0,0 +1,242 @@
...
+/// `notify` writes to.
+pub(in crate::runtime::builtin) const MAX_SOURCES: usize = SET_SIZE - 1;
+
+/// A socket set with room for `SET_SIZE` sockets.
+///
+/// `select` reads a set as a count followed by that many sockets, and takes the count from the
+/// set rather than any length from the type it was declared with; this is the set Winsock's
+/// headers declare once `FD_SETSIZE` is raised, laid out as `FD_SET` is with a longer array.
+#[repr(C)]
+struct FdSet {
```
zeenix: **[P3] Apply the project's definition-after-use ordering consistently**

Codex review on behalf of zeenix.

`CONTRIBUTING.md` explicitly puts the public API first, helpers below their callers, and simple
constants below their users. Several new modules still violate that rule:

- Here, constants and `FdSet`/its implementation precede `Poller` and its methods.
- `builtin/driver.rs` puts constants and private TLS state above their users.
- `builtin/mod.rs` puts `OWN`/`SHARED` before their consumers in `Inner`.
- `builtin/scheduler.rs` puts `dispose` before later callers and before `JoinHandle`'s API.
- `builtin/tests.rs` leaves the helper block before the tests appended after it.

Reorder these within their existing logical commits, with test helpers following all tests
that use them.

PR #1976, opened 2026-09-19
Make block_on public and drop the blocking API

zeenix:
`zbus::block_on` becomes part of the public API, and the blocking API goes: a program without an
async runtime of its own writes the async API inside one `zbus::block_on`, with zbus as its one
dependency (plus `futures-util` where it reads a stream). The blocking wrappers were a shortcut for
a simple program that did not want to pick an async runtime and learn it; with `block_on` public
that program is one `block_on` around the code the book already shows, and the wrappers, the
`blocking-api` feature, the `*ProxyBlocking` types `#[proxy]` generated and the attributes that
controlled them (`gen_blocking`, `gen_async`, `blocking_name`, `blocking_object`; `async_name`
stays, and `async_object` is renamed `proxy_type_name`, the old name still accepted) are removed
outright, since 6.0 is a breaking release. The upgrading guide
has the recipe; the FAQ shows a program inside one `zbus::block_on` call.

None of this depends on which runtime drives a connection. #1975 (the built-in runtime) is
stacked on this branch and will be rebased once this merges.

## Commits

- ✨ zb: Make block_on part of the public API
- 📝 Add the plan for dropping the blocking API
- ♻️ zb,zm,zx,book: Stop using the blocking API
- 👷 Drop the blocking-api feature from CI
- 💥 zb,zm,zx,book: Stop generating blocking proxies
- 💥 zb: Remove the blocking API
- 📝 zb,book: Point synchronous programs at zbus::block_on
- ♻️ zm,book: Name the proxy type override proxy_type_name

## Migration

```rust
// 5.x
let connection = zbus::blocking::Connection::session()?;
let proxy = FooProxyBlocking::new(&connection)?;
let answer = proxy.bar()?;

// 6.0
let answer = zbus::block_on(async {
    let connection = zbus::Connection::session().await?;
    let proxy = FooProxy::new(&connection).await?;
    proxy.bar().await
})?;
```

A blocking iterator becomes the stream it wrapped, driven with `StreamExt::next` inside the
future; dropping the stream only queues the match rule's removal, and a program that needs it gone
before going on calls `stream.async_drop().await` (`zbus::AsyncDrop`) instead. `zbus-xmlgen`'s CLI and the `screen-brightness` example use the same shape.

## Decisions that need the maintainer's eye

1. **Outright removal, no alias**: a build naming `blocking-api` fails to resolve the feature; a
   `#[proxy]` naming a removed attribute fails with the macro's unknown-attribute error.
2. **`gen_async` goes with `gen_blocking`**: with one proxy per `#[proxy]`, "generate the async
   one" is not a choice. `async_name` stays (trait-level `name` is the D-Bus interface name);
   `async_object` is renamed `proxy_type_name`, with the old name accepted undocumented, per
   review.
3. **The book's blocking chapter is gone** rather than rewritten: a chapter would only repeat the
   other chapters' code inside one `block_on` call. The introduction says how the API is driven
   without a runtime and a FAQ entry shows it; every former link points there.

https://claude.ai/code/session_01BfDbRLyfCfp1VSn3ZT1cq5


zeenix:
> * **Outright removal, no alias**: a build naming `blocking-api` fails to resolve the feature; a
>   `#[proxy]` naming a removed attribute fails with the macro's unknown-attribute error.

That's fine.

> * **`gen_async` goes with `gen_blocking`**: with one proxy per `#[proxy]`, "generate the async
>   one" is not a choice.

Also fine.

> `async_name`/`async_object` keep their names.

Don't see the reason for them to be the recommended way. The documented way should `name` and `object` now.

> * **`book/src/blocking.md` keeps its path** (links and search) under the new title.

ok.

book/src/upgrading-to-6.md
```diff
@@ -701,11 +698,36 @@ A build with `comms` but neither `async-io` nor `tokio` is now valid, as long as
...
+// 6.0
+let answer = zbus::block_on(async {
+    let connection = zbus::Connection::session().await?;
+    let proxy = FooProxy::new(&connection).await?;
+    proxy.bar().await
+})?;
+```
+
+A blocking iterator becomes the stream it wrapped, driven with `StreamExt::next` inside the
+future. The `#[proxy]` attributes `gen_blocking`, `blocking_name` and `blocking_object` are gone
```
zeenix: Codex, reviewing on behalf of zeenix:

Please document the cleanup timing change too. `MessageIterator` and `SignalIterator` waited
for match-rule removal in `Drop`; dropping the corresponding streams only queues removal.
Users who rely on cleanup completing before proceeding should import `zbus::AsyncDrop` and
call `stream.async_drop().await` inside the `block_on` future.

book/src/blocking.md
```diff
@@ -1,35 +1,19 @@
...
-
-## Establishing a connection
-
-The only difference to that of [asynchronous `Connection` API] is that you use
-[`blocking::Connection`] type instead. This type's API is almost identical to that of `Connection`,
-except all its methods are blocking.
+zbus's API is async, and a program does not need an async runtime of its own to use it:
+[`zbus::block_on`] runs a future to completion on the calling thread, while the connection's own
+tasks, sockets and timers run on threads zbus starts for them, so the program below runs on the
+program's thread plus zbus's own two, with zbus as its one dependency, plus `futures-util` for the
```
zeenix: Codex, reviewing on behalf of zeenix:

Please avoid promising exactly two background threads. With the `tokio` feature, `block_on`
creates a multi-thread Tokio runtime, whose worker count is not fixed at two. Describe the
background work without a fixed count, or explicitly qualify the backend and assumptions.

zeenix (commented):
Reviewed by Codex on behalf of zeenix at `3440023`.

The removal and macro simplification look sound, and the commits are logically scoped.
Please address the two inline documentation comments and the existing naming request before merging.
`name` currently aliases the D-Bus interface name, while `object` appends `Proxy`; adopting the
requested meanings needs macro changes as well as documentation.

Local macro integration tests passed with both default and Tokio backends: 11 passed and one
existing ignored test in each run. CI is green.

book/src/upgrading-to-6.md
zeenix: Added to "The blocking API is gone": dropping a `MessageStream` or `SignalStream` only queues the match rule's removal on the connection's runtime, where 5.x's blocking iterators removed it before `Drop` returned, and a program that needs the rule gone before going on imports `zbus::AsyncDrop` and calls `stream.async_drop().await` inside the `block_on` future.

book/src/blocking.md
zeenix: Reworded without a count: the future runs on the calling thread, while the connection's tasks, sockets and timers run on threads zbus starts for them, or on Tokio's runtime with the `tokio` feature.

zeenix:
Done in `♻️ zm,book: Name the proxy type override proxy_type_name`: `async_object` is now `proxy_type_name`, with `async_object` still accepted as an undocumented alias, and the upgrading guide names both. `async_name` stays as is: trait-level `name` is already the D-Bus interface name (an alias of `interface`), and `object` derives `<object>Proxy`, so renaming those would change their meaning rather than their spelling.

zbus/src/utils.rs
```diff
@@ -28,22 +28,49 @@ impl<T, E> ResultAdapter for Result<T, E> {
...
-/// to poll the caller's future.
+/// Runs a future to completion on the calling thread.
+///
+/// It runs the future on the calling thread and drives no connection: in this build every
+/// connection runs on the runtime handed to [`Builder::runtime`], and this call only waits for
+/// that runtime's work.
+///
+/// Do not call this from an async context, that is, from inside a future another runtime is
+/// polling. It holds the thread that future runs on until the call returns, and where the two
+/// end up waiting on each other, neither of them ever does.
```
Skyb0rg007: Confusing phrasing here, I'm not sure what that last sentence means

book/src/blocking.md
```diff
@@ -1,35 +1,19 @@
...
-
-The only difference to that of [asynchronous `Connection` API] is that you use
-[`blocking::Connection`] type instead. This type's API is almost identical to that of `Connection`,
-except all its methods are blocking.
+zbus's API is async, and a program does not need an async runtime of its own to use it:
+[`zbus::block_on`] runs the future on the calling thread, while the connection's own tasks,
+sockets and timers run on threads zbus starts for them (with the `tokio` feature, on Tokio's
+runtime), with zbus as its one dependency, plus `futures-util` for the stream extension trait
+where a program reads a stream. Everything in the other chapters works the same way inside the
+future handed to `zbus::block_on`.
```
Skyb0rg007: I read this a few times and am still not sure what this means.

book/src/faq.md
```diff
@@ -170,9 +170,10 @@ chosen once per connection, when it is built: Tokio when a Tokio runtime is curr
 that builds it, `async-io` otherwise. This keeps the features additive, so an `async-io`-based
 application keeps working even when another crate in the workspace enables zbus's `tokio` feature.
 
-This per-connection selection applies to the async API. The blocking API (`zbus::blocking`) drives
-its connections through its own `block_on`, which uses tokio whenever the `tokio` feature is
-enabled, so those connections always run on tokio when that feature is on.
+This per-connection selection applies to a program that already has an async runtime of its own.
+One with none drives the async API from inside `zbus::block_on`, which with the `tokio` feature
```
Skyb0rg007: "One with none"?

Skyb0rg007 (commented):
Some of the public-facing docs need a human rewrite, as a lot of the descriptions use very confusing phrasing.

zeenix (commented):
Re-reviewed by Codex on behalf of zeenix at `738dd55`.

Both previous documentation findings are addressed. The `proxy_type_name` spelling and
undocumented `async_object` alias follow zeenix's instructions. No new findings; ready to merge.

Macro integration tests and focused successful-call checks for methods, properties, vector returns
and `#[interface]` forwarding passed with default and Tokio backends. CI is green.

zbus/src/utils.rs
zeenix: Rewritten. The warning now reads: "Do not call this from another runtime's task. It blocks that task's thread until the future completes, and if the future needs that runtime to make progress, the program deadlocks." The `tokio` build's variant says the same and adds that it panics from a Tokio task. The rest of the three doc comments got the same treatment.

book/src/blocking.md
zeenix: Rewritten as plain sentences, one point each: the API is async but you need no runtime of your own; `zbus::block_on` runs a future to completion on the calling thread, so a synchronous program puts its D-Bus code inside one call; zbus is the only dependency (plus `futures-util` for streams); the connection's own work runs on background threads zbus starts, or on Tokio's runtime with the `tokio` feature; everything from the other chapters works the same way inside that future.

book/src/faq.md
zeenix: Reworded: "This per-connection selection only matters for a program that has an async runtime of its own. A program without one uses `zbus::block_on` instead, and with the `tokio` feature that call runs inside a Tokio runtime, so a connection built inside that call always uses Tokio."

book/src/blocking.md
```diff
@@ -1,35 +1,20 @@
...
-The only difference to that of [asynchronous `Connection` API] is that you use
-[`blocking::Connection`] type instead. This type's API is almost identical to that of `Connection`,
-except all its methods are blocking.
+zbus's API is async, but a program does not need an async runtime of its own to use it.
+[`zbus::block_on`] runs a future to completion on the calling thread, so a synchronous program
+can simply put its D-Bus code inside one `block_on` call. Such a program depends on zbus alone,
+plus `futures-util` if it reads signals or property changes from a stream. The connection's own
+tasks, sockets and timers run on background threads that zbus starts, or on Tokio's runtime if
+the `tokio` feature is enabled. Everything shown in the other chapters works the same way inside
+the `block_on` future.
```
zeenix: This is assuming #1975 already in place, while the dependency is the other way around.

book/src/faq.md
```diff
@@ -170,9 +170,9 @@ chosen once per connection, when it is built: Tokio when a Tokio runtime is curr
 that builds it, `async-io` otherwise. This keeps the features additive, so an `async-io`-based
 application keeps working even when another crate in the workspace enables zbus's `tokio` feature.
 
-This per-connection selection applies to the async API. The blocking API (`zbus::blocking`) drives
-its connections through its own `block_on`, which uses tokio whenever the `tokio` feature is
-enabled, so those connections always run on tokio when that feature is on.
+This per-connection selection only matters for a program that has an async runtime of its own. A
+program without one uses `zbus::block_on` instead, and with the `tokio` feature that call runs
+inside a Tokio runtime, so a connection built inside that call always uses Tokio.
```
zeenix: This also seems to assume #1975 as it talks about the tokio possibility but not async-io.

book/src/faq.md
zeenix: @Skyb0rg007 I think it's readable now.

zeenix:
> Some of the public-facing docs need a human rewrite, as a lot of the descriptions use very confusing phrasing.

Thanks for reviewing! I think the machine did a good job of fixing the docs you pointed to but please feel free to point out other docs that need improvements. If you have suggestions for what the docs should be, that's even better. :+1:

book/src/faq.md
Skyb0rg007: Thanks! Looks good now

book/src/blocking.md
zeenix: It was written for main's world (with `async-io`, a connection's tasks do run on a `zbus::Connection executor` thread zbus starts), but it never said which backend it was describing. Now: "With the default `async-io` backend, the connection's own tasks, sockets and timers run on threads that zbus starts for them; with the `tokio` feature, they run on Tokio's runtime." #1975 rewrites that sentence for the built-in runtime on its side.

book/src/faq.md
zeenix: Added the async-io case: "A program without one uses `zbus::block_on` instead. Without the `tokio` feature, that call and the connection built inside it use `async-io`; with it, the call runs inside a Tokio runtime, so a connection built inside it always uses Tokio."

book/src/blocking.md
```diff
@@ -180,32 +155,25 @@ impl Greeter {
...
+            .name("org.zbus.MyGreeter")
+            .serve_at("/org/zbus/MyGreeter", greeter)
+            .build()
+            .await?;
+
+        pending::<()>().await;
 
-    Ok(())
+        Ok(())
+    })
```
zeenix: Why are we changing from event-listener to forever pending? This doesn't seem relevant to the PR.

.github/workflows/rust.yml
```diff
@@ -36,7 +36,7 @@ jobs:
           cargo --locked check -p zbus --no-default-features --target x86_64-unknown-netbsd
           cargo --locked check -p zbus --no-default-features --target aarch64-linux-android
           # The external-runtime-only build: no async-io, no tokio.
-          EXT=async-lock,blocking-api,proxy,service,object-manager,unixexec,ibus,tracing,p2p
+          EXT=async-lock,proxy,service,object-manager,unixexec,ibus,tracing,p2p
```
zeenix, line 39: The containing commit `💥 zb,zm: Remove the blocking API in favour of zbus::block_on` is too large and can easily be split into multiple logical commits that all build and run. E.g:

* Remove usage of blocking wrappers and API.
* Remove the `blocking-api` usage in the CI.
* Remove the blocking proxy generation.
* Remove the blocking API.

zbus_macros/src/proxy.rs
```diff
@@ -377,6 +379,27 @@ pub fn create_proxy(
     })
 }
 
+/// The proxy type to build from the `ObjectPath` an `object`-attributed method or property
+/// returns: `proxy_type_name` if given, its `async_object` alias if given, or `<object>Proxy`
+/// derived from the `object` attribute's value when neither is given.
+fn resolve_proxy_type_name(
```
zeenix: Not following the topdown code layout rule.

zbus/src/utils.rs
```diff
@@ -28,22 +28,45 @@ impl<T, E> ResultAdapter for Result<T, E> {
...
+/// Runs a future to completion, blocking the calling thread until it is done.
+///
+/// This is for a program that has no async runtime of its own. Put the async code in one call to
+/// this function; the call blocks the thread until the future completes. With the `tokio`
+/// feature, the call runs inside a Tokio runtime that zbus creates on first use and keeps for
+/// the rest of the process, so a connection built inside the future runs on Tokio as well.
+///
+/// Do not call this from another runtime's task. From a Tokio task it panics. From any other
+/// runtime's task it blocks that task's thread until the future completes, and if the future
+/// needs that runtime to make progress, the program deadlocks.
```
zeenix: Codex, reviewing on behalf of zeenix:

[P3] Qualify the panic and deadlock conditions

Depending on another runtime does not by itself cause deadlock: an executor with other worker
threads can still make progress. Please say this *can* deadlock when the blocked thread is needed
for progress, and apply that wording to all three variants. Also qualify "Tokio task" as an
*async* task; the panic condition is entering from Tokio's async execution context, not merely
running work on a Tokio-owned thread. See Tokio's [runtime documentation][runtime].

[runtime]: https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#method.block_on

book/src/upgrading-to-6.md
```diff
@@ -701,11 +698,45 @@ A build with `comms` but neither `async-io` nor `tokio` is now valid, as long as
...
+})?;
+```
+
+A blocking iterator becomes the stream it wrapped, driven with `StreamExt::next` inside the
+future.
+
+5.x's `blocking::MessageIterator` and `blocking::SignalIterator` removed their match rule from
+the bus before `Drop` returned. Dropping a `MessageStream` or `SignalStream` only queues that
+removal. A program that needs the rule gone before continuing (for example, because it is about
+to add a conflicting rule) should import `zbus::AsyncDrop` and call `stream.async_drop().await`
```
zeenix: Codex, reviewing on behalf of zeenix:

[P3] Use a real reason to wait for match-rule removal

D-Bus match rules are additive; overlapping rules, including duplicate registrations, can
coexist. Adding a "conflicting rule" is therefore not a reason to wait for removal. Please drop
the parenthetical or restore the previous example of asserting the bus's subscription state.
The cleanup timing warning and `async_drop().await` guidance themselves are correct.
See the [D-Bus specification][spec].

[spec]: https://dbus.freedesktop.org/doc/dbus-specification.html#message-bus-routing-match-rules

zeenix (commented):
Codex, reviewing on behalf of zeenix.

Re-reviewed `8108dfa` against `738dd55` and the latest discussion. The changes between those
versions are documentation only. The explicit async-io/Tokio descriptions address the concern
about assuming #1975 has already landed. The requested `proxy_type_name` naming and undocumented
`async_object` alias remain as agreed.

Two low-priority documentation corrections are inline: the deadlock condition is too broad,
and the migration guide's "conflicting rule" example does not describe D-Bus match semantics.

Your three newer requests remain unaddressed at this head:

- Preserve the server example's event listener and shutdown behavior.
- Split the blocking-API removal into logical, independently building commits.
- Move `resolve_proxy_type_name` below its callers to follow the top-down layout rule.

My previous ready-to-merge conclusion missed the scope and layout issues. I would wait for
these outstanding requests to be addressed. CI is green at the reviewed head.

book/src/blocking.md
zeenix: Codex, reviewing on behalf of zeenix:

Confirmed: this also removes `Greeter::go_away`, which emitted `GreetedEveryone` and notified the
event before exiting. The new example declares the signal but never emits it, and cannot shut
down through that method. Keeping `done` and `go_away`, then replacing `done_listener.wait()`
with `done_listener.await` inside `block_on`, would preserve the example's behavior.

book/src/upgrading-to-6.md
zeenix: Also I don't think it's important to deregister. Nothing bad happens if the connection received a message nobody is interested in anymore.

book/src/blocking.md
zeenix: Restored: the example keeps `done: Event` and the `GoAway` method that emits `GreetedEveryone` and notifies it, and `main` now awaits the listener inside the `block_on` future (`done_listener.await`) where it called `wait()` before. The prose says the future waits for that event.

zbus_macros/src/proxy.rs
zeenix: Moved `resolve_proxy_type_name` below `gen_proxy_property`, its last caller.

zbus/src/utils.rs
zeenix: Qualified in all three variants: "It blocks that task's thread until the future completes, which deadlocks the program if the future needs that thread to make progress." The `tokio` variant says "From inside a Tokio async task it panics, as Tokio's own `block_on` does."

book/src/upgrading-to-6.md
zeenix: Dropped the invented reason. The paragraph now says the queued removal is harmless (until the rule is gone, the connection merely receives messages nobody reads) and that a program which wants the rule gone before continuing can use `zbus::AsyncDrop`.

.github/workflows/rust.yml
zeenix: Split into four commits that each build with `--all-features` (workspace check, default and tokio builds, the macro tests at every step; the tree at the tip is byte-identical to before):

- `♻️ zb,book: Stop using the blocking API` (examples, tests, the book);
- `👷 Drop the blocking-api feature from CI` (the workflow and the `blocking_api` fixture);
- `💥 zb,zm,zx,book: Stop generating blocking proxies` (the macros, xmlgen, and every use of `gen_blocking`/`blocking_name`/`blocking_object`, plus `zbus::blocking::fdo`, which only re-exported the generated proxies);
- `💥 zb: Remove the blocking API` (the module and the feature).

zeenix: Codex, reviewing on behalf of zeenix:

The four stages are present and all eight commits build independently. One boundary is still
incomplete: `277876e` migrates `zbus_xmlgen/src/main.rs` and `test_proxy_object_list` from the
blocking API, although those are usage migrations that belong in `6c50a04`. Please move those
hunks and the xmlgen dependency adjustment there; keep the generated-name collision change in
`zbus_xmlgen/src/lib.rs` with proxy generation removal.

I applied those three files' changes to `6c50a04` separately: they apply cleanly, the workspace
builds with all features/targets, and the object-list test passes. They do not depend on removing
blocking proxy generation. The `blocking::fdo` removal is correctly placed with generation.

book/src/blocking.md
```diff
@@ -1,35 +1,20 @@
...
-
-## Establishing a connection
-
-The only difference to that of [asynchronous `Connection` API] is that you use
-[`blocking::Connection`] type instead. This type's API is almost identical to that of `Connection`,
-except all its methods are blocking.
+zbus's API is async, but a program does not need an async runtime of its own to use it.
+[`zbus::block_on`] runs a future to completion on the calling thread, so a synchronous program
+can simply put its D-Bus code inside one `block_on` call. Such a program depends on zbus alone,
+plus `futures-util` if it reads signals or property changes from a stream. With the default
```
zeenix: Codex, reviewing on behalf of zeenix:

[P3] Mention the server example's event-listener dependency

Restoring the shutdown example also restores `use event_listener::Event`, so the dependency
guidance is incomplete: someone following this paragraph cannot compile that example with just
`zbus` and `futures-util`. Please mention `event-listener` for the server's shutdown notification,
here or immediately before the server example.

zeenix (commented):
Codex, reviewing on behalf of zeenix.

Reviewed every commit through `3ee8661`, including the split and all review threads.

| Commit | Scope and result |
| --- | --- |
| `fa0449d` | Public `block_on` documentation; scoped correctly. |
| `a10e529` | Implementation-plan document only; changelog skip is present. |
| `6c50a04` | Usage migration; CLI and object-list test migrations are still in `277876e`. |
| `732caca` | CI and its blocking-API fixture, workspace entry and lockfile; scoped correctly. |
| `277876e` | Proxy generation removal; move the independent usage migrations earlier. |
| `f2d43b7` | Blocking module, feature, conversions and feature gate removal; scoped correctly. |
| `bc8247c` | User-facing migration documentation; scoped correctly. |
| `3ee8661` | Requested naming, alias, tests and documentation; helper is below its callers. |

The four removal stages follow the requested order, but the usage/proxy-generation boundary needs
the small adjustment described in the existing split thread. Removing `blocking::fdo` in the
proxy-generation commit is appropriate: it re-exports the types that commit removes.

Verified the restored `GoAway`, `GreetedEveryone`, event listener and `done_listener.await`;
the runtime-warning and match-rule fixes; the explicit async-io/Tokio wording; and the agreed
`proxy_type_name` / undocumented `async_object` alias. Those earlier concerns are addressed.
One new documentation omission is inline: the restored server example needs `event-listener`.

Validation: all eight commits pass `cargo check --locked --workspace --all-targets --all-features`.
Macro integration tests pass at `6c50a04`, `277876e`, `f2d43b7` and `3ee8661` with default
features and with `zbus/tokio` (11 passed per run before the rename, 12 after; one existing ignore).
The book/crate doctests also pass: 92 with defaults and 88 with Tokio only, including the
restored server example. CI is green. No new runtime or macro regression found.

.github/workflows/rust.yml
zeenix: Moved: the `zbus_xmlgen` CLI, its dependency line and `test_proxy_object_list` now change in `♻️ zb,zm,zx,book: Stop using the blocking API`; the proxy-generation commit keeps only the generated-name change in `zbus_xmlgen/src/lib.rs`. Both commits still build with all features and targets, and the object-list test passes at the first.

book/src/blocking.md
zeenix: Added to the dependency sentence: "plus `futures-util` if it reads signals or property changes from a stream and `event-listener` for the shutdown notification the server example below uses."

zeenix (commented):
Codex, reviewing on behalf of zeenix.

Re-reviewed all eight commits at `4fcc8e4`. They are now logically scoped and follow the requested
split: usage migration, CI cleanup, proxy-generation removal, then blocking-API removal.
The CLI, dependency and object-list test changes are correctly in `94ca164`; `b17d66b` keeps
the generated-name change. The server's `event-listener` dependency is also documented.

Both remaining comments are addressed. The final executable code is unchanged from the last
tested revision; the only final-tree change is the dependency sentence. The rewritten
`94ca164` and `a55e2d4` pass workspace checks with all features and targets, and the moved
object-list test passes at `94ca164` with default and Tokio features.

No remaining findings. Ready to merge once the current CI run passes; commitlint and CodSpeed
have passed, while Lint, Build and Test is still running.

zbus/examples/screen-brightness.rs
```diff
@@ -4,34 +4,43 @@
...
-    let connection = zbus::blocking::Connection::session().unwrap();
+#[proxy(
+    interface = "org.gnome.SettingsDaemon.Power.Screen",
+    default_service = "org.gnome.SettingsDaemon.Power",
+    default_path = "/org/gnome/SettingsDaemon/Power"
+)]
+trait Screen {
+    fn step_up(&self) -> Result<(i32, String)>;
+    fn step_down(&self) -> Result<(i32, String)>;
+}
```
zeenix: This refactor wasn't needed. All we needed here was, drop of blocking API and use block_on. The Future code could be split to an `async fn run` and that reduces the diff by avoiding indentation changes.

zbus_xmlgen/src/main.rs
```diff
@@ -144,12 +140,16 @@ impl DBusInfo<'_> {
 
         let input_src = format!("Interface '{path}' from service '{service}' on system bus",);
 
-        let xml = IntrospectableProxy::builder(&connection)
-            .destination(service.clone())
-            .path(path.clone())
-            .build()
-            .unwrap()
-            .introspect()?;
+        let xml = zbus::block_on(async {
```
zeenix: instead of multiple `block_on` calls, why not split all the code into a secondary `async fn` and wrap it all in one `block_on`?

test_fixtures/blocking_api/src/lib.rs
zeenix, line 1: The fixture is not necessarily CI only. It's also involved in local testing. Hence it doesn't really belong in the CI commit.

zbus/examples/screen-brightness.rs
zeenix: Reverted to main's code with the smallest change: `main` calls `zbus::block_on(run())`, and `run` is main's body as an `async fn` with `.await` on `Connection::session()` and `call_method`.

zbus_xmlgen/src/main.rs
zeenix: Done: `main` is now `zbus::block_on(run())`, `run` is the previous body as an `async fn` with the three connections awaited, and `DBusInfo::new` is an `async fn` awaiting the introspection directly. One `block_on` in the whole binary.

test_fixtures/blocking_api/src/lib.rs
zeenix: Moved: the fixture, its workspace entry and its lockfile lines now go in `💥 zb: Remove the blocking API`, with the API it compiles. The CI commit is the workflow alone.

zbus/tests/basic.rs
```diff
@@ -120,113 +99,17 @@ async fn fdpass_systemd_async() {
...
 
 #[instrument]
-async fn test_freedesktop_api() -> Result<()> {
-    let connection = Connection::session().await?;
+async fn freedesktop_api_inner() -> Result<()> {
+    let connection = Connection::session().await.map_err(|e| {
+        debug!("error: {}", e);
+
+        e
+    })?;
```
zeenix: Why `map_err`?

zbus_macros/tests/tests.rs
```diff
@@ -462,26 +462,30 @@ fn test_proxy_object_list() {
...
+        for (correct, returned) in OBJECT_LIST.paths.iter().zip(list) {
             assert!(returned.inner().path() == correct);
         }
     }
 
-    let connection = zbus::blocking::connection::Builder::session()
-        .serve_at(OBJECT_LIST.paths[1].as_ref(), OBJECT_LIST.clone())
-        .build()
-        .unwrap();
+    let connection = zbus::block_on(
```
zeenix: Same thing here as before with other code: instead of multiple use of `block_on`, split the actual work into an `async fn` and use one `block_on` call. Check all other changes in the PR as well and see if the same apply to other changes and make the appropriate change.

zbus/tests/basic.rs
zeenix: It was carried over from the blocking test. Dropped along with the rest of the rewrite: `basic.rs` is now main's file with the two `blocking-api` tests deleted and nothing else touched, so `basic_connection_async` and `freedesktop_api_async` stay as they were.

zbus_macros/tests/tests.rs
zeenix: Done: the test builds the connection, the proxy and both calls inside one `block_on(async { ... })`, the style the rest of that file uses. I went over the PR's other changes: the remaining `block_on` call sites are the example and xmlgen (one each, done earlier), the book snippets (one each) and pre-existing test wrappers; the other test changes only drop `gen_blocking` attributes.

zbus/tests/basic.rs
zeenix: It isn't. Renamed `basic_connection_async` to `basic_connection` and `freedesktop_api_async` to `freedesktop_api`, the names their blocking twins had.

zbus_macros/tests/tests.rs
```diff
@@ -462,26 +462,30 @@ fn test_proxy_object_list() {
...
+            .destination(&destination)
+            .build()
+            .await
+            .unwrap();
 
-    check_return(proxy.get_test_objects().unwrap());
-    check_return(proxy.objects().unwrap());
+        check_return(proxy.get_test_objects().await.unwrap());
+        check_return(proxy.objects().await.unwrap());
+    });
```
zeenix: This made is indent. There was a good reason why I explicitly suggested a function split, so the diff is as clean as possible.

zeenix: Split as suggested: the test body is `async fn proxy_object_list()` as it was, with `.await` on the four calls, and `test_proxy_object_list` is `zbus::block_on(proxy_object_list())`. No line moves.

